PR #53, opened 2017-10-31
Add a default condition to copts

brianretford:


earhart:
These options are compiler-specific; are you sure we don't just want an empty option list?

brianretford:
Actually, this is already done internally, so I'll hold off until our next public push (today or tomorrow).

brianretford:
@earhart yes, I think we should go with the default being things that work with *nix configurations and have exceptions for platforms with different compilers (like windows)

PR #56, opened 2017-11-02
vertexai-updates-20171102

brianretford:


plaidml/plaidml_test.py
```diff
@@ -13,9 +13,7 @@
 class TestPlaidML(unittest.TestCase):
 
     def setUp(self):
-        plaidml.settings.config = None
-        plaidml.settings.setup = True 
-        plaidml.settings.experimental = True 
+      testing.plaidml_config.unittest_config()
```
flaub: Indentation doesn't match

brianretford: fixed

PR #82, opened 2018-01-25
Implement ELU

KuzMenachem:


plaidml/keras/backend.py
```diff
@@ -1845,7 +1845,15 @@ def dtype(x):
...
-    _report_unimplemented('elu')
+	if alpha == 1:
+		f = """function (X, Alpha) -> (R) {
+				   A = exp(X)-1;
+				   R = (X < 0 ? A : X); }"""
+	else:
+		f = """function (X, Alpha) -> (R) {
+				   A = Alpha*exp(X)-Alpha;
+				   R = (X < 0 ? A : X); }"""
+	return _Op('elu', x.dtype, x.shape, f, {'X': x, 'Alpha': alpha}, ['R'])
```
tzerrell: The inputs to a Tile function must be tensors; `alpha` is a python float. The easiest change is to replace `'Alpha': alpha` with `'Alpha': variable(alpha)` in the input dictionary. However, since the value of alpha is known when `elu` is called, you will get (marginally) better performance if you use string manipulation to directly write the value of alpha into `f`. E.g. (for the second case)
```
f = """function (X) -> (R) {{
        A = {alpha}*exp(X) - {alpha};
        R = X < 0 ? A : X; }}""".format(alpha=alpha)
```
You can then remove alpha from the input dictionary.

KuzMenachem:
I'm not sure whether the testElu function is working as intended, I tried my best. Also sorry for the commit mess, it's my first time on github.

PR #93, opened 2018-03-20
0.3.0.dev1

brianretford:
* Add preview support for LSTM
  * Add preview LLVM HAL
  * Add more ops to the standard op library
  * Port more of the keras frontend to use the op library.
  * Supports ops used by onnx-plaidml
  * Fixes several issues relating to broadcasting


brianretford:
Yeah, it seems to be constant for now, but I agree we need to change it

PR #162, opened 2018-08-27
0.3.5

brianretford:


README.md
```diff
@@ -50,6 +52,13 @@ For examples and benchmarks, see our [blog](http://vertex.ai/blog).
 - [Reporting Issues](#reporting-issues)
 
 ### Recent Release Notes
+* PlaidML 0.3.3 - 0.3.5
+  * Support Keras 2.2.0
```
tzerrell: As far as I can tell we support Keras 2.2.0-2.2.2 (and indeed I put a bit of effort into ensuring we are compatible with both, and the version we install (e.g. with pip) if the user doesn't do anything special is 2.2.2). I think it would be reasonable to claim compatibility with that whole range?

README.md
```diff
@@ -2,6 +2,8 @@
 
 *A platform for making deep learning work everywhere.*
 
+** Vertex.AI (the creators of PlaidML) is excited to be join Intel's Artificial Intelligence Products Group. PlaidML will soon be re-licensed under Apache 2. Read the announcement [here!](https://vertex.ai) **
```
flaub: Should be: 

> is excited to join

brianretford:
Done.

brianretford:
Fixed

Issue #227, opened 2019-01-28
Plaidml not detecting Mali-T628 on ARM


nitescuc:
Hi,

I've build plaidml 0.3.5 to use on Odroid XU4 with Mali-T628 GPU with debian stretch. I manage to install the wheel, when I run plaidml-setup, I get:

"No supported devices found. Run 'clinfo' and file an issue containing the full output."

However, with plaidml 0.3.0rc1 latest available with pip install plaidml, my devices can be configured and I have 2 mali-t628 reported. "experimental.json" appears quite similar in both cases.

Any clue with what I may have done wrong building plaidml ? (used basel 0.18.1 with --config linux_arm_32v7) or what change might explain 0.3.5 not recognizing my devices where 0.3.0rc1 did ?

Thanks

Here's my clinfo report:

Number of platforms                               1
  Platform Name                                   ARM Platform
  Platform Vendor                                 ARM
  Platform Version                                OpenCL 1.2 v1.r12p0-04rel0.03af15950392f3702b248717f4938b82
  Platform Profile                                FULL_PROFILE
  Platform Extensions                             cl_khr_global_int32_base_atomics cl_khr_global_int32_extended_atomics cl_khr_local_int32_base_atomics cl_khr_local_int32_extended_atomics cl_khr_byte_addressable_store cl_khr_3d_image_writes cl_khr_fp64 cl_khr_int64_base_atomics cl_khr_int64_extended_atomics cl_khr_fp16 cl_khr_gl_sharing cl_khr_icd cl_khr_egl_event cl_khr_egl_image cl_arm_core_id cl_arm_printf cl_arm_thread_limit_hint cl_arm_non_uniform_work_group_size cl_arm_import_memory
  Platform Extensions function suffix             ARM

  Platform Name                                   ARM Platform
Number of devices                                 2
  Device Name                                     Mali-T628
  Device Vendor                                   ARM
  Device Vendor ID                                0x6200010
  Device Version                                  OpenCL 1.2 v1.r12p0-04rel0.03af15950392f3702b248717f4938b82
  Driver Version                                  1.2
  Device OpenCL C Version                         OpenCL C 1.2 v1.r12p0-04rel0.03af15950392f3702b248717f4938b82
  Device Type                                     GPU
  Device Profile                                  FULL_PROFILE
  Max compute units                               4
  Max clock frequency                             600MHz
  Device Partition                                (core)
    Max number of sub-devices                     0
    Supported partition types                     None
[...]

PR #232, opened 2019-02-03
#227 Adjust experimental.json to support Odroid XU4 Mali-T628 card

nitescuc:
As suggested by @tzerrell changing the regex in experimental.json fixed "No supported devices found ..." on ODroid XU4 with Mali-T628. Also changed for Mali-T760 but I wasn't able to test on such configuration.

### DIFF
--- a/plaidml/configs/experimental.json
+++ b/plaidml/configs/experimental.json
@@ -144,10 +144,10 @@
                     "and": {
                         "sel": [
                             {
-                                "vendor_regex": "ARM"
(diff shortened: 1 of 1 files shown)


brianretford:
Thanks for this and sorry for the delay in reviewing it - we're in the process of finalizing our procedure for accepting PRs - basically we need to run things against our internal CI. I think we have the process down and I this should get closed out in the next couple days

brianretford:
integrated in #300

Issue #243, opened 2019-02-18
extract_image_patches


iperov:
how to implement differentiable extract_image_patches in plaidML same or similar to
[tf.image.extract_image_patches](https://www.tensorflow.org/api_docs/python/tf/image/extract_image_patches)?
tf version is differentiable

kilroythethird:
+1 for a plaidml implementation of extract_image_patches or at least a workaround.
Required fe. to implement DSSIM Loss Function.

iperov:
@kilroythethird I already implemented tensorflow SSIM on pure keras that works on plaidML.

```
def dssim(kernel_size=11, k1=0.01, k2=0.03, max_value=1.0):
    # port of tf.image.ssim to pure keras in order to work on plaidML backend.

    def func(y_true, y_pred):
        ch = K.shape(y_pred)[-1]

        def _fspecial_gauss(size, sigma):
            #Function to mimic the 'fspecial' gaussian MATLAB function.
            coords = np.arange(0, size, dtype=K.floatx())
            coords -= (size - 1 ) / 2.0
            g = coords**2
            g *= ( -0.5 / (sigma**2) )
            g = np.reshape (g, (1,-1)) + np.reshape(g, (-1,1) )
            g = K.constant ( np.reshape (g, (1,-1)) )
            g = K.softmax(g)
            g = K.reshape (g, (size, size, 1, 1)) 
            g = K.tile (g, (1,1,ch,1))
            return g
                      
        kernel = _fspecial_gauss(kernel_size,1.5)

        def reducer(x):
            return K.depthwise_conv2d(x, kernel, strides=(1, 1), padding='valid')

        c1 = (k1 * max_value) ** 2
        c2 = (k2 * max_value) ** 2
        
        mean0 = reducer(y_true)
        mean1 = reducer(y_pred)
        num0 = mean0 * mean1 * 2.0
        den0 = K.square(mean0) + K.square(mean1)
        luminance = (num0 + c1) / (den0 + c1)
        
        num1 = reducer(y_true * y_pred) * 2.0
        den1 = reducer(K.square(y_true) + K.square(y_pred))
        c2 *= 1.0 #compensation factor
        cs = (num1 - num0 + c2) / (den1 - den0 + c2)

        ssim_val = K.mean(luminance * cs, axis=(-3, -2) )
        return K.mean( (1.0 - ssim_val ) / 2.0 )

    return func
```
usage:
loss = dssim()( y_true, y_pred )

[...]

PR #279, opened 2019-03-28
Added extract_images_patches impl

SleepProgger:
This PR provides an operation compatible to tensorflows extract_image_patches function. #243
The only case where it provides different results as the tensorflow implementation is when no patch can be extracted, in which case tensorflow produces a shape like [batch, 0, x, depth] and this implementation provides a shape of [0].
It does not honor the data_format of the backend. If that is required that should be easy(ish) to add tho.

Tested with this crude testcase:
```
    BATCHES = 2
    DEPTH = 3
    tf_session = tf.Session()    
    # lol
    for iy in range(1, 6):
        for ix in range(1, 6):
            images = np.random.rand(*(BATCHES, iy, ix, DEPTH))
            k_images = K.variable(images, dtype=images.dtype)
            for ky in range(1, iy):
                for kx in range(1, ix):
                    for sy in range(1, iy):
                        for sx in range(1, ix):
                            for ry in range(1, iy):
                                for rx in range(1, ix):
                                    for padding in ("SAME", "VALID"):
                                        if padding == "VALID" and (ky*ry > iy or kx*rx > ix):
                                            continue
                                        ksizes = [1, ky, kx, 1]
                                        strides = [1, sy, sx, 1]
                                        rates = [1, ry, rx, 1]
                                        print("[%s] img=%s, ksizes=%s, strides=%s, rates=%s, padding=%s" % (
                                            "?",
                                            images.shape,
                                            ksizes, strides,
                                            rates, padding
                                        ), end="\t")
                                        with tf_session.as_default():
                                            tf_patches = tf.extract_image_patches(images, ksizes=ksizes, strides=strides, rates=rates, padding=padding).eval()
                                        k_patches = extract_image_patches(k_images, ksizes, strides, rates, padding).eval()
                                        if np.any(tf_patches != k_patches):
                                            print("\033[91m\033[1mFAILED\033[0m")
                                            print("tf shape: ", tf_patches.shape)
                                            print(tf_patches)
                                            print("k_patches shape: ", k_patches.shape)
                                            print(k_patches)
                                        else:
                                            print("\033[92m\033[1mSUCCESS\033[0m")
```

For the interested here is the same implementation with pure K functions:
```
def extract_image_patches_k(images, ksizes, strides, rates, padding):
    ishape = images.shape.dims
    laxis = ksizes[1]*ksizes[2]*ishape[-1]
    patch_row_eff = ksizes[1] + ((ksizes[1] - 1) * (rates[1] -1))
    patch_col_eff = ksizes[2] + ((ksizes[2] - 1) * (rates[2] -1))
    
    if padding.upper() == "VALID":
        out_rows = math.ceil((ishape[1] - patch_row_eff + 1.) / float(strides[1]))
        out_cols = math.ceil((ishape[2] - patch_col_eff + 1.) / float(strides[2]))
        pad_top = 0
        pad_left = 0
    else:
        out_rows = math.ceil( ishape[1] / float(strides[1]) )
        out_cols = math.ceil( ishape[2] / float(strides[2]) )
        pad_top = max(0, ( (out_rows - 1) * strides[1] + patch_row_eff - ishape[1] ) // 2)
        pad_left = max(0, ( (out_cols - 1) * strides[2] + patch_col_eff - ishape[2] ) // 2)
        # we simply assume padding right + 1 == padding left (same for top/down).
        # This might lead to us padding more as we would need but that won't matter.
        # TF tries to split padding to both sides so left_pad +1 should keep us on the safe side.
        images = K.spatial_2d_padding(images, ((pad_top, pad_top+1), (pad_left, pad_left+1)))

    if out_rows * out_cols == 0:
        return K.constant(np.ndarray(0, dtype=images.shape.dtype.name.lower()))

    if rates[1] != ksizes:
        ksizes[1] = ksizes[1] + (ksizes[1] -1) * (rates[1] - 1)
    if rates[2] != ksizes:
        ksizes[2] = ksizes[2] + (ksizes[2] -1) * (rates[2] - 1)
    ishape = images.shape.dims
    patches = []
    for y in range(out_rows):
        for x in range(out_cols):
            patch = images[:, y*strides[1]:y*strides[1]+ksizes[1]:rates[1], x*strides[2]:x*strides[2]+ksizes[2]:rates[2], :]
            patch = K.reshape(patch, (ishape[0], 1, 1, laxis))
            patches.append(patch)
    patches = K.concatenate(patches, -1)
    nshape = (ishape[0], out_rows, out_cols, laxis)
    patches = K.reshape(patches, nshape)
    return patches
```

### DIFF
--- a/plaidml/keras/backend.py
+++ b/plaidml/keras/backend.py
@@ -1693,57 +1693,6 @@ def zeros_like(x, dtype=floatx(), name=None):
                 .sole_output()
 
 
-class ImagePatches(ptile.Operation):
-    def __init__(self, images, ksizes, strides, rates=(1,1,1,1), padding="VALID", name=None):
-        """
-        Compatible to tensorflow.extract_image_patches. 
-        Extract patches from images and put them in the "depth" output dimension.
-        Args:
-            images: A tensor with a shape of [batch, rows, cols, depth]
-            ksizes: The size of the patches with a shape of [1, patch_rows, patch_cols, 1]
-            strides: How far the center of two patches are in the image with a shape of [1, stride_rows, stride_cols, 1]
-            rates: How far two consecutive pixel are in the input. Equivalent to dilation. Expect shape of [1, rate_rows, rate_cols, 1]
-            padding: A string of "VALID" or "SAME" defining padding.
-        """
-        i_shape = images.shape.dims
-        patch_row_eff = ksizes[1] + ((ksizes[1] - 1) * (rates[1] -1))
-        patch_col_eff = ksizes[2] + ((ksizes[2] - 1) * (rates[2] -1))
-
-        if padding.upper() == "VALID":
-            out_rows = math.ceil((i_shape[1] - patch_row_eff + 1.) / float(strides[1]))
-            out_cols = math.ceil((i_shape[2] - patch_col_eff + 1.) / float(strides[2]))
-        else:
-            out_rows = math.ceil( i_shape[1] / float(strides[1]) )
-            out_cols = math.ceil( i_shape[2] / float(strides[2]) )
-            pad_top = max(0, ( (out_rows - 1) * strides[1] + patch_row_eff - i_shape[1] ) // 2)
-            pad_left = max(0, ( (out_cols - 1) * strides[2] + patch_col_eff - i_shape[2] ) // 2)
-            # we simply assume padding right == padding left + 1 (same for top/down).
-            # This might lead to us padding more as we would need but that won't matter.
-            # TF splits padding between both sides so left_pad +1 should keep us on the safe side.
-            images = spatial_2d_padding(images, ((pad_top, pad_top+1), (pad_left, pad_left+1)))
-
-        o_shape = (i_shape[0], out_rows, out_cols, ksizes[1]*ksizes[2]*i_shape[-1])
-        code = """function (I[B,Y,X,D]) -> (O) {{
-                    TMP[b, ny, nx, y, x, d: B, {NY}, {NX}, {KY}, {KX}, D] =
-                        =(I[b, ny * {SY} + y * {RY}, nx * {SX} + x * {RX}, d]);
-                    O = reshape(TMP, B, {NY}, {NX}, {KY} * {KX} * D);
-                }}
-        """.format(
-            NY=out_rows, NX=out_cols,
-            KY=ksizes[1], KX=ksizes[2],
-            SY=strides[1], SX=strides[2],
-            RY=rates[1], RX=rates[2]
-        )
-        super(ImagePatches, self).__init__(code,
-                [('I', images),],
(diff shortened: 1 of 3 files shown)


SleepProgger:
Oops, i just realized i forgot to remove `pad_top` and `pad_left` in the padding=="VALID" case.
I'll leave that in as i assume you guys need to merge it manually to your backend version anyway, so you cam simply remove them.

brianretford:
Awesome thank you! BTW we are talking about designing a new version of Tile, would love to include you in those discussions is you're interested

SleepProgger:
Sure, although i have to admit math is not my strongest side.
I also started to love tile to be honest. Needs a bit to get going, but after that point it really is a cool language.

*I am idling in the unofficial plaidml discord in case quicker communication is needed*

tzerrell:
LGTM, I'll point the people working on our CI at this noting it's good to go so it can be merged.

I notice there aren't unit tests -- I checked how it currently works and it seems fine, and for this sort of extension we don't strictly _need_ unit tests so it's ok to merge, but if you want us to notice if we break this you can add some unit tests.

SleepProgger:
Nice to hear that. I am going to add some light unit tests.

tzerrell:
A nice way to add unit tests would be to move this to `op.py` as you suggested and then add a test file for tests of TF-like ops (perhaps `tf_test.py`) to that directory that has a unit test comparing the results of doing this to doing it in pure TF (since this is supposed to reproduce tensorflow behavior).

Adding a test against a fixed expected value to `backend_test.py` would be easier and ok as a basic test (but agreed that it doesn't really match there as you're not comparing the behavior of fixed Keras functions across different backends).

SleepProgger:
Ok, i will move it to op.py and add tf_test.py using a similar structure to backend_test.py.

But i ran into some troubles:
My version doesn't work when width/height are symbolic which is problematic as `y_true` in loss functions is provided as fully symbolic (None all the way).
So i started to move the output shape calculation (NY, NX) into the TILE code, but that did lead to `plaidml.exceptions.Unknown: Variable _T6 used in a context which requires it to be a constant integer` for both the `TMP` sizes and the reshape call.
I assume that makes sense as the size need to be known at bind time.
Is there any way around this ?

A cheap workaround for the loss case is:
```
y_true = K.reshape(y_true, y_pred.shape.dims)
```
so it wouldn't be that bad to have it only work with known sizes, but if there is a way to allow it to work with symbolic sizes i'd highly favorr that.

SleepProgger:
I added unit tests in `tf_test.py` and copied the opTest decorator from `backend_test`.

The `extract_image_patches` function still can't handle symbolic width/height axis and i didn't found a way to do this. If there is a way to do this i'd be happy to change that, otherwise, i'll say this is good enough as is.

brianretford:
integrated in #303

PR #284, opened 2019-03-31
Enable concat to work with symbolic tensors

SleepProgger:
Currently concatenate doesn't work when the "merging" axis is symbolic.
This PR should fix that.
This is not super carefully tested, so please do so yourself before merging.
I don't see a reason it shouldn't work though.

Example to demonstrate the problem:
```
    DIM_FIX = False
    def dummy_loss(y_true, y_pred): # I am not sure why y_true has a symbolic shape for all axis at all, but that one might be on keras ?
        print("y_true shape:", K.int_shape(y_true)) # -> (None, None, None, None)
        print("y_pred shape:", K.int_shape(y_pred)) # -> (None, 64, 64, 3)
        mask = 1 # in reality this has a shape of (B, Y, X, 1)
        if DIM_FIX:
            y_true = K.reshape(y_true, K.int_shape(y_pred))       # works
            #y_true = K.reshape(y_true, (None, None, 64, None))   # works
            #y_true = K.reshape(y_true, (None, None, None, None)) # Same as unchanged. Doesn't work
            #y_true = K.reshape(y_true, (None, None, None, 3))    # Doesn't work
            #y_true = K.reshape(y_true, (None, 64, None, None))   # Doesn't work
        n_true = K.concatenate([y_true[:, :, :, i] * mask for i in range(3)])
        n_pred = K.concatenate([y_pred[:, :, :, i] * mask for i in range(3)])
        return keras.losses.mean_squared_error(n_true, n_pred)
    
    in_ = x = keras.Input((64, 64, 3))
    x = keras.layers.Conv2D(3, 3, strides=1, padding="same")(x)    
    model = keras.Model(in_, x)
    model.compile("Adam", dummy_loss)
    
    model.train_on_batch(np.zeros((32, 64, 64, 3)), np.ones((32, 64, 64, 3)))
```

With `DIM_FIX` set to false (and thus y_true being fully symbolic) i get a kernel like:
```
function (I0[N0, N1, A0], I1[N0, N1, A1], I2[N0, N1, A2]) -> (O) {
  T0[n0, n1, a+0: N0, N1, Add UINT64()] = =(I0[n0, n1, a]);
  T1[n0, n1, a+Ceiling UINT64(): N0, N1, Add UINT64()] = =(I1[n0, n1, a]);
  T2[n0, n1, a+Add UINT64(): N0, N1, Add UINT64()] = =(I2[n0, n1, a]);
  O = T0 + T1 + T2;
}
```

which of course doesn't work.

### DIFF
--- a/plaidml/keras/backend.py
+++ b/plaidml/keras/backend.py
@@ -708,7 +708,8 @@ def get_value(x):
     tensor = plaidml.Tensor(_device(), shape)
     invoker.set_output('out', tensor)
     invoker.invoke()
-    array = np.ndarray(x.shape.dims, dtype=ptile.convert_pml_dtype_to_np(x.shape.dtype))
+    out_shape = tuple(x.size for x in shape.dimensions)
+    array = np.ndarray(out_shape, dtype=ptile.convert_pml_dtype_to_np(x.shape.dtype))
     with tensor.mmap_current() as view:
         view.copy_to_ndarray(array)
     return array
--- a/plaidml/keras/backend_test.py
+++ b/plaidml/keras/backend_test.py
@@ -169,15 +169,22 @@ def opTest(in_data,
            atol=DEFAULT_ATOL,
            skip_theano=True,
            skip_tensorflow=False,
-           verbose=False):
+           verbose=False,
+           input_shapes=None):
     # If using with non-tensor parameters, all tensor params must appear before
     # all non-tensor params
-    def run_one_backend(self, data, test_func, b, *run_args):
+    # input_shapes should be None or an iterable containing tuples
+    # defining the shape for each input.
+    # Use this to define symbolic shapes (None).
+    def run_one_backend(self, data, test_func, b, shapes=None, *run_args):
         tf_session = tensorflow.Session()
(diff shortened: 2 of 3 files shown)


SleepProgger:
While implementing unit tests for this i stumbled upon some problems/unimplemented required functionality.

1. The `opTest` function in `backend_test` doesn't provided a way to use specify symbolic shapes.
I added this functionality by adding  an optional `input_shapes` parameter which can contain an iterable filled with tuples specifying the shape for each input. The shape can contain None values.

2. The `get_value` function in `op` didn't support symbolic shapes neither and used the given `shape.dims` to allocate a `np.ndarray` which would crash if one or more of the dims where symbolic.
I fixed this by getting the real shape from `invoker.get_output_shape`.

Additionally i changed my Concatenate changes to only use symbolic shapes when they are required.

Please let me know if i should split these into separate PRs.

tzerrell:
Nice! LGTM. I like these as a single PR -- you have multiple additions but they support each other.

brianretford:
closed in #305

PR #285, opened 2019-04-01
Add SpatialDropout support in dropout

SleepProgger:
This PR adds support for noise_shape in the dropout function and thus adds support for SpatialDropout*d.
I am not absolutely sure the math is correct but it should ?

On question about this (and the previous) implementation though:
`rng_step` uses `x` (the input) to get the correct shapes even with symbolic tensors.
Does that mean x is always copied onto the device even when it only is used to get the correct shape or is that optimized away ?
And if it is always copied to the device would there be a way to add support to tile to provide correct shape only without really copying the data ?

### DIFF
--- a/plaidml/keras/backend.py
+++ b/plaidml/keras/backend.py
@@ -584,13 +584,18 @@ dot = op.dot
 
 
 def dropout(x, level, noise_shape=None, seed=None):
-    if noise_shape is not None:
-        raise PlaidMLKerasException('Unimplemented noise shape in dropout')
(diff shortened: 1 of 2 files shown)


tzerrell:
I'm not certain whether we optimize away the copy onto a device, but I definitely think it would be nice to be able to pass in just the shape from a code clarity perspective if nothing else. The performance and clarity costs have been pretty minor so we haven't prioritized it.

Looks good, I'm adding a quick unit test.

brianretford:
closed in #305

PR #291, opened 2019-04-12
Added K.arange, K.eye, K.random_normal_variable backend functions

SleepProgger:
This PR adds some keras backend function implemented by wrapping numpy functions in tensors.

I also added a decorator to the `backend_test` to allow using `compareForward*` decorators with multiple inputs similar to `opTest`.

Usage like:
```
    @compareMultiple([
        [10],
        [-2, 5, 2, 'float32']
    ])
    @compareForwardExact()
    def testArange(self, b, *args):
        return b.arange(*args)
```

I hope that is ok, but imho it makes writing testcases a bit easier.


SleepProgger:
I just realized i missed to use the seed and name parameter in `random_normal_variable`.
This is now handled properly.
Sorry for that, somehow missed to copy that over.

Issue #273, opened 2019-03-23
trying to implement ReflectionPadding2D


iperov:
finally I implemented it in one op for B,H,W,C
```
class ReflectionPadding2D(PMLTile.Operation):
    def __init__(self, input, h_pad, w_pad):
        if K.image_data_format() == 'channels_last':
            if input.shape.ndims == 4:
                H, W = input.shape.dims[1:3]
                if (type(H) == int and h_pad >= H) or \
                   (type(W) == int and w_pad >= W):
                    raise ValueError("Paddings must be less than dimensions.")
                c = """ function (I[B, H, W, C] ) -> (O) {{
                        WE = W + {w_pad}*2;
                        HE = H + {h_pad}*2;
                    """.format(h_pad=h_pad, w_pad=w_pad)
                if w_pad > 0:
                    c += """
                        LEFT_PAD [b, h, w , c : B, H, WE, C ] = =(I[b, h, {w_pad}-w,            c]), w < {w_pad} ;
                        HCENTER  [b, h, w , c : B, H, WE, C ] = =(I[b, h, w-{w_pad},            c]), w < W+{w_pad}-1 ;
                        RIGHT_PAD[b, h, w , c : B, H, WE, C ] = =(I[b, h, 2*W - (w-{w_pad}) -2, c]);
                        LCR = LEFT_PAD+HCENTER+RIGHT_PAD;
                    """.format(h_pad=h_pad, w_pad=w_pad)
                else:
                    c += "LCR = I;"
                if h_pad > 0:
                    c += """
                        TOP_PAD   [b, h, w , c : B, HE, WE, C ] = =(LCR[b, {h_pad}-h,            w, c]), h < {h_pad};
                        VCENTER   [b, h, w , c : B, HE, WE, C ] = =(LCR[b, h-{h_pad},            w, c]), h < H+{h_pad}-1 ;
                        BOTTOM_PAD[b, h, w , c : B, HE, WE, C ] = =(LCR[b, 2*H - (h-{h_pad}) -2, w, c]);
                        TVB = TOP_PAD+VCENTER+BOTTOM_PAD;
                    """.format(h_pad=h_pad, w_pad=w_pad)
                else:
                    c += "TVB = LCR;"
                c += "O = TVB; }"
                inp_dims = input.shape.dims
                out_dims = (inp_dims[0], inp_dims[1]+h_pad*2, inp_dims[2]+w_pad*2, inp_dims[3])
            else:
                raise NotImplemented
        else:
[...]

PR #294, opened 2019-04-18
Added ReflectivePadding

SleepProgger:
This PR adds reflective padding equivalent to tensorflows [pad](https://www.tensorflow.org/api_docs/python/tf/pad) function with mode set to "REFLECT".
It works with symbolic and non symbolic axis.

This PRs testcase (in `tf_test.py`) needs the patch in `backend.get_value` like added in #284
If #279 is going to be merged the `opTest` function in that PR should replaced by the one in this PR.

Solves #273


SleepProgger:
I don't want to be annoying, but any ETA on this ?
Also what happened with all the commits ? github hiccup ?

brianretford:
You’re not being annoying, sorry for the delay. We had to rewrite history because there were some bad merges in it that made rebasing have conflicts, as well as some history messages that needed revision. We’ll get your patches in today

Sent from my iPhone

On Jun 17, 2019, at 3:41 PM, SleepProgger <notifications@github.com<mailto:notifications@github.com>> wrote:


I don't want to be annoying, but any ETA on this ?
Also what happened with all the commits ? github hiccup ?

Issue #256, opened 2019-03-03
Error building on Manjaro (Arch): histedit.h not found


janblumenkamp:
Hi everyone,
I am trying to build plaidml on Manjaro Linux, but I receive the following error:

```
Starting local Bazel server and connecting to it...
INFO: Analysed 2 targets (43 packages loaded, 18996 targets configured).
INFO: Found 2 targets...
ERROR: /home/jan/.cache/bazel/_bazel_jan/229fe12941da0d9c6cc2c9e386ea5376/external/llvm/BUILD.bazel:286:1: C++ compilation of rule '@llvm//:lib' failed (Exit 1)
external/llvm/lib/LineEditor/LineEditor.cpp:17:22: fatal error: histedit.h: No such file or directory
 #include <histedit.h>
                      ^
compilation terminated.
INFO: Elapsed time: 178.596s, Critical Path: 48.62s, Remote (0.00% of the time): [queue: 0.00%, setup: 0.00%, process: 0.00%]
INFO: 340 processes: 340 local.
```

Output of `bazel version`:
```
Build label: 0.19.2- (@non-git)
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Nov 20 16:00:05 2018 (1542729605)
Build timestamp: 1542729605
Build timestamp as int: 1542729605
```

The header file is definitely existing in my file system:
```
locate histedit.h
/usr/include/histedit.h
```

What am I missing?

SleepProgger:
I had the same issue on Manjaro.
"Solved" it by adding -DHAVE_LIBEDIT=0 in vendor/llvm/llvm.BUILD 
Libedit support is only needed for some GUI/CLI tool in LLVM as far as i know anyway.

PR #348, opened 2019-06-19
Removed histedit dependency.

thyrgle:
I (in addition to some others: see #256) have had trouble using the `histedit.h`. This small commit removes the `histedit.h` by adding `-DHAVE_LIBEDIT=0` to the build flags for all the LLVM vendor targets. Of course, you may feel this is not ideal, please feel free to close if you do.


flaub:
We may have a totally new way of dealing with LLVM soon. I'll update here when I know more.

flaub:
Resolved by #404

PR #360, opened 2019-07-02
Implementation of K.all

Antetokounpo:
This is a hacky implementation of K.all, basically it does a K.prod and then check if each value of the tensor is equal to 0, if so it stays the same, otherwise it changes to 1. Finally it returns a UINT8 tensor of 1s and 0s as specified in the keras documentation.

### DIFF
--- a/plaidml/keras/backend.py
+++ b/plaidml/keras/backend.py
@@ -197,8 +197,7 @@ def abs(x):
     return builtins.abs(x)
 
 
-def all(x, axis=None, keepdims=False):
-    _report_unimplemented('all')
+all = op.all
 
 
 def any(x, axis=None, keepdims=False):
--- a/plaidml/keras/backend_test.py
+++ b/plaidml/keras/backend_test.py
@@ -556,6 +556,21 @@ class TestBackendOps(unittest.TestCase):
             x / c,
             c / x,
         ]
+    @compareForwardExact()
+    def testAll(self, b):
+        return b.all(b.variable(m(3, 3)))
+
+    @compareForwardExact()
+    def testAllKeepdims(self, b):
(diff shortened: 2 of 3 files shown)


Antetokounpo:
I made a tiny rewrite to the function, because I realized that making a K.prod before converting the tensor to 1s and 0s caused problems with NaN values.

tzerrell:
Looks good! Same comments as `K.any`: This needs unit tests in plaidml/keras/backend_test.py. `testSum` might be a good example of how to write such a test. Let us know if you run into any problems.

Antetokounpo:
I added the tests, they are identical to those for K.any.

Let me know if anything is wrong or could be improved and thanks again for the help.

SleepProgger:
Nothing critical but there is a `compareMultiple` decorator in backend_test.
Use it like 
```
    @compareMultiple([
        [10],
        [-2, 5, 2, 'float32']
    ])
    @compareForwardExact()
    def testArange(self, b, *args):
        return b.arange(*args)
```
That makes the tests a bit more compact.

flaub:
Thanks for the update!

We recently enabled our new public visible CI infrastructure here:
https://buildkite.com/vertex-dot-ai

However, this PR is failing, could you try rebasing with the latest master and try again?

https://buildkite.com/vertex-dot-ai/plaidml-plaidml/builds/6#7ad5b6c5-59dd-4484-b2ab-ee9a1a4e1b65

Antetokounpo:
Okay I rebased with master.

Also @SleepProgger I read your comment, but since I'm calling a function from the backend `b` passed as an argument, I think using `@compareMultiple` will lead to a code that's less readable, because I can't just pass `*args` directly. 

Thank you for the suggestion anyway.

PR #369, opened 2019-07-15
Implementation of K.any

Antetokounpo:
This is an implementation of K.any, like my previous pull request it's a hacky solution, let me know if anything is wrong or could be better.

### DIFF
--- a/plaidml/keras/backend.py
+++ b/plaidml/keras/backend.py
@@ -199,7 +199,6 @@ def abs(x):
 
 all = op.all
 
-
 any = op.any
 
 
--- a/plaidml/keras/backend.py
+++ b/plaidml/keras/backend.py
@@ -200,8 +200,7 @@ def abs(x):
 all = op.all
 
 
-def any(x, axis=None, keepdims=False):
-    _report_unimplemented('any')
+any = op.any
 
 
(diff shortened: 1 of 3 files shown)


jinz2014:
The Keras backend function 'any' is not yet implemented in Plaid.  Sorry, is there an implementation available ? 

Thanks

tzerrell:
Looks good! This needs unit tests though, in `plaidml/keras/backend_test.py`. I think `testSum` might be a good example of how to write such a test. Let us know if you run into any problems doing this.

Antetokounpo:
Thank you for the reply and the help. I added the unit tests. I used `@compareForwardExact` for the tests instead of `@opTest`, because it seemed like TensorFlow didn't support derivatives for this operation, at least that's what it seemed like. 

Let me know if anything is wrong or could be improved.

Antetokounpo:
Okay I rebased with master. Again let me know if anything is wrong.

Thanks!

tzerrell:
The `all` PR created merge conflicts -- @Antetokounpo can you confirm that my fixes of those conflicts & the style tweaks are ok?

Antetokounpo:
Yes they're fine. Thank you!

PR #387, opened 2019-08-05
Add EDSL shared operations

tzerrell:
Adds (or in the case of var moves) operations to the shared EDSL op library. Currently adds abs, concatenate, expand_dims, squeeze, tile, transpose, and variance. Also adds or adjusts related Keras ops and the additional Keras backend ops softplus, softsign, stack, std, switch, and truncated_normal.

Note that concatenate is not working in most cases yet.

### DIFF
--- a/plaidml2/bridge/keras/__init__.py
+++ b/plaidml2/bridge/keras/__init__.py
@@ -8,6 +8,7 @@ from contextlib import contextmanager
 import six
 
 import numpy as np
+import scipy.stats
 import plaidml2 as plaidml
 import plaidml2.edsl as edsl
 import plaidml2.exec as plaidml_exec
@@ -222,7 +223,8 @@ class PlaidMLKerasException(Exception):
 
 
 def abs(x):
-    _report_unimplemented('abs')
+    logger.debug('abs(x: {})'.format(x))
+    return _KerasNode('abs', tensor=plaidml_op.abs(x.tensor))
 
 
 def all(x, axis=None, keepdims=False):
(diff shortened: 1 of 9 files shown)


plaidml2/bridge/keras/__init__.py
```diff
@@ -1085,11 +1080,13 @@ def softmax(x):
 
 
 def softplus(x):
-    _report_unimplemented('softplus')
+    logger.debug('softplus(x: {})'.format(x))
+    return log(1. + exp(x))
```
flaub: This might need to return a _KerasNode

plaidml2/bridge/keras/__init__.py
```diff
@@ -1085,11 +1080,13 @@ def softmax(x):
...
 
 
 def softsign(x):
-    _report_unimplemented('softsign')
+    logger.debug('softsign(x: {})'.format(x))
+    return x / (1. + abs(x))
```
flaub: This is safe to return like this, but it's sometimes nice to wrap as a KerasNode so you can provide a new name (e.g. 'softsign')

tzerrell:
@flaub can you re-review when you have a chance? I rebased onto updated `master`

plaidml2/bridge/keras/__init__.py
```diff
@@ -8,6 +8,7 @@
 import six
 
 import numpy as np
+import scipy.stats
```
flaub: Does this come with numpy or will we need to eventually update our `not_setup.cfg` to include scipy? Not a problem yet because we don't generate wheels for plaidml2, but eventually...

tzerrell: This has been in `plaidml/keras` for a long time. Probably it's coming from Keras.

PR #388, opened 2019-08-05
Add tests that demonstrate bad perf for the CPU jit

LLITCHEV:
Also, added ability to measure from the backend tests.

Also removed the generation of CPU sencil for XSMM since it was causing a major perf problems. All the permutations are causing the stencil selection code to perform really poorly. Having handcoded set of values is much better here.


LLITCHEV:
Merging PR.

PR #390, opened 2019-08-05
Added softplus implementation

SleepProgger:
Like mentioned on the slack, here the softplus impl.
~It somehow fails the tests on my system, but that might be due to my current setup.~ (Works fine with plaidml 0.6.4rc1)


tzerrell:
LGTM but I'll wait to see what our CI thinks of this

PR #391, opened 2019-08-06
Use vector type casts for ceiling and float

earhart:


flaub:
Why doesn't our CI catch these problems? Or rather, could we add tests to this PR that exposes these issues?

earhart:
I don’t know a good way in the current codebase to force vectorization. We could possibly build a semtree-level test framework, but that probably needs some long-term planning, not a hack for this particular case.

flaub:
Where do these issues crop up? Wouldn't they show up in correctness or performance issues on certain targets?

earhart:
This came up in nGraph testing. It’s a compilation error.

PR #394, opened 2019-08-06
pull gv100gl xception test from pr builds

drgone:


shiyao00:
xception works on Nvidia, doesn't works on Metal/iGPU.

flaub:
Yep, we have coverage of this network on gp100, so it's redundant on gv100, and since we only have 1 gv100 but we have 2 gp100, it seems like not much is lost if we just drop this from gv100.

PR #396, opened 2019-08-06
add soft_fail per workload to buildkite.py

drgone:


ci/pipeline.yml
```diff
@@ -27,6 +27,10 @@ steps:
       automatic:
         limit: {{.}}
   {{/retry}}
+  {{#softfail}}
+    soft_fail: true 
+        exit status: {{.}}
```
flaub: drop this

ci/plan.yml
```diff
@@ -198,6 +198,7 @@ SUITES:
         conda_env: conda/plaidml.yml
         precision: untested
         retry: 2
+        softfail: 1
```
flaub: you could say, `soft_fail: yes`

drgone:
@shiyao00 this adds mtl tests and allows soft fail on backend_test and mnist

flaub:
I think we should constrain soft fail to a combination of platform + workflow. Allowing soft fail to happen for an entire platform means we won't notice when tests fail for platforms we know shouldn't be failing at all.

drgone:
OK, this should work for any given workload/platform, use platform_overrides to soft_fail

platform_overrides:
  keras-stripe-mtl-uhd630:
    softfail: true

PR #400, opened 2019-08-06
WIP - make getting dev wheels an easier/better experience 

drgone:


brianretford:
Is it possible to get it to just produce the wheels, or one tarball? I'm lazy...

flaub:
How about we make a release pipeline that automatically publishes to pypi using twine?

ci/plan.yml
```diff
@@ -205,6 +207,14 @@ SUITES:
...
           keras-plaid-mtl-uhd630:
             softfail: true
+        args: 
+          - --shard-count
+          - {shards}
+          - --shard
+          - {shard}
+        params:
+          shards: [3]
+          run_shards: [1,2,3]
```
tzerrell: I think this needs to be `[0, 1, 2]` instead

ci/plan.yml
```diff
@@ -38,7 +38,15 @@ PLATFORMS:
...
-  keras-plaid-ocl-r9nano:
+  keras-plaid-ocl-1080ti:
+    variant: linux_x86_64
+  keras-stripe-ocl-1080:
+    variant: linux_x86_64
+  keras-plaid-ocl-1080ti:
+    variant: linux_x86_64
+  keras-stripe-ocl-1080:
+    variant: linux_x86_64
+    keras-plaid-ocl-r9nano:
```
flaub: Broken indentation here.

ci/plan.yml
```diff
@@ -281,24 +322,28 @@ SUITES:
...
           keras-stripe-mtl-uhd630:
-            prepend_args: [--examples, 256]
-          
+            prepend_args: [-examples, 256]
+        skip_platforms: [keras-stripe-mtl-uhd630, keras-stripe-ocl-neo] # whyfail
+
       mobilenet:
         precision: high
+        skip_platforms: [keras-plaid-mtl-uhd630] # whyfail
+        skip_platforms: [keras-plaid-mtl-uhd630, keras-stripe-mtl-uhd630, keras-stripe-ocl-neo] # whyfail
```
flaub: Multiple `skip_platforms`, not sure what behavior we get with this, what's the intent?

ci/pipeline.yml
```diff
@@ -20,6 +20,7 @@ steps:
   - label: '{{emoji}}{{engine}} {{suite}}/{{workload}}/{{platform}} BATCH_SIZE={{batch_size}} '
     command: |
       {{python}} -m pip install pyyaml --user
+      {{python}} -m pip install numpy --user
```
flaub: You can put multiple deps on a single line. But when I tried to do this I got horrible errors, perhaps because I needed to pin numpy to an earlier version.

```
{{python}} -m pip install numpy pyyaml --user
```

ci/plan.yml
drgone: just a bad paste.  this plan.yml is going to get pulled soon

ci/buildkite.py
```diff
@@ -139,10 +143,35 @@ def cmd_pipeline(args, remainder):
         yml = pystache.render(load_template('pipeline.yml'), ctx)
         print(yml)
 
+
+def wheel_path(arg):
+    if platform.system() == 'Windows':
```
flaub: Instead of using the `bazel-bin` symlink, I think we should be using `bazel info bazel-bin` to get the absolute path.

ci/report.py
```diff
@@ -24,7 +24,6 @@
 
 
 def printf(*args, **kwargs):
-    print(*args, **kwargs)
```
flaub: This got dropped again somehow.

flaub:
Why are there any changes to `backend_test`? Looks like all style changes. Did you use `yapf`? Which version are you using? You should be using the `.cenv` conda environment, it has a pinned version of `yapf`.

Issue #132, opened 2018-05-30
keras.backend.clear_session()


JimmyDuong97:
This has not been implemented, yet it's important in order to avoid memory leaks for cases when the user trains lots of models in the same session.

tzerrell:
We're putting together a release that implements this; we'll update this issue when it's out

tzerrell:
This is implemented in 0.3.5, let us know if it works for you

PR #414, opened 2019-08-10
Add operations to EDSL common library

tzerrell:


plaidml2/bridge/keras/__init__.py
```diff
@@ -40,6 +44,24 @@ def _prepend_name_scope(name, default):
...
 
+def _normalize_axis(axis, ndims, name=""):
+    negative_axis_given = False
+    if axis < 0:
+        axis += ndims
+        negative_axis_given = True
+    if axis < 0 or ndims <= axis:
+        if negative_axis_given:
+            axis -= ndims
+        if name:
```
earhart: This might actually be more clear as a one-liner -- `name = 'for {} op '.format(name) if name else ''`
Or in py3, `name = f'for {name} op' if name else ''`

plaidml2/bridge/keras/__init__.py
```diff
@@ -694,15 +792,41 @@ def identity(x):
...
 def in_train_phase(x, alt, training=None):
-    _report_unimplemented('in_train_phase')
+    logger.debug('in_train_phase(x: {}, alt: {}, training: {})'.format(x, alt, training))
+    if training is None:
+        training = learning_phase()
+        uses_learning_phase = True
+    else:
+        uses_learning_phase = False
+
+    if callable(x):
```
earhart: `cx = x() if callable(x) else x`

plaidml2/bridge/keras/__init__.py
```diff
@@ -929,7 +1131,10 @@ def random_uniform(shape, minval=0.0, maxval=1.0, dtype=None, seed=None):
 
 
 def random_uniform_variable(shape, low, high, dtype=None, name=None, seed=None):
-    _report_unimplemented('random_uniform_variable')
+    if seed:
+        np.random.seed(seed)
```
earhart: It'd be really nice to avoid these legacy global-state-polluting methods, but I'd worry about back-compat.  :-/

plaidml2/bridge/keras/__init__.py
```diff
@@ -40,6 +44,24 @@ def _prepend_name_scope(name, default):
     return r
 
 
+def _normalize_axis(axis, ndims, name=""):
```
earhart: There's some inconsistency in this file between double-quotes and single-quotes.  Please use single-quotes for all strings except doc comments.

plaidml2/edsl/__init__.py
```diff
@@ -560,6 +571,10 @@ def cast(x, dtype):
     return call("as_{}".format(dtype.info.base), x, dtype.info.bitwidth)
 
 
+def ceil(x):
+    return call("ceil", x)
```
earhart: Please use single-quote strings in here, too.

plaidml2/bridge/keras/__init__.py
```diff
@@ -295,7 +356,13 @@ def bias_add(x, bias, data_format=None):
 
 
 def binary_crossentropy(target, output, from_logits=False):
-    _report_unimplemented('binary_crossentropy')
+    logger.debug('binary_crossentropy(target: {}, output: {}, from_logits: {})'.format(
```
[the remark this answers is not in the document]
tzerrell: I'll take a quick look and see if I see something straightforward. I'm not sure whether there'll be a nice programmatic way to get at the parameter names, which seems core to having a decorator be simpler & more consistent

plaidml2/bridge/keras/__init__.py
```diff
@@ -1049,7 +1367,11 @@ def set_floatx(dtype):
 
 
 def set_learning_phase(value):
-    _report_unimplemented('set_learning_phase')
+    if value != 0 and value != 1:
```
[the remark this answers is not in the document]
tzerrell: Yes, it's kind of a mess. This is gets a basic implementation in but I want to take another look going forward b/c I think Keras has made some related changes to how learning_phase is managed and there might be a different thing we should be doing now.

plaidml2/bridge/keras/__init__.py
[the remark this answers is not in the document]
earhart: Use the `inspect` module.  `inspect.signature(some_callable).parameters` has the info.

plaidml2/bridge/keras/__init__.py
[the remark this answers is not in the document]
earhart: SGTM

plaidml2/bridge/keras/__init__.py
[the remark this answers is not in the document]
flaub: Decorator++
Adding another dependency--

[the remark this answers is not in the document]
tzerrell: It looks like inspect is a core part of python (unless I misunderstand how that works?): https://docs.python.org/3/library/inspect.html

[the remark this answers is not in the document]
flaub: My comment is about adding new pypi dependencies. If we can make it work with core python, that's good. Does the `inspect` module add much overhead? It'd be best if the overhead would only be paid if logging at debug mode is enabled.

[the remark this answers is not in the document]
tzerrell: I wrapped most of the decorator code in a `if logger.isEnabledFor(logging.DEBUG):` block. The only extra overhead when debug is off is in calling a wrapping decorator that calls the function rather than calling the function directly.

plaidml2/bridge/keras/__init__.py
```diff
@@ -348,27 +437,34 @@ def categorical_crossentropy(target, output, from_logits=False):
...
-    _report_unimplemented('ceil')
+    return _KerasNode('ceil', tensor=edsl.ceil(x.tensor))
 
 
+@_log_call
 def clear_session():
-    _report_unimplemented('clear_session')
+    global _in_train_phase, _device
+    _in_train_phase = None
+    _dev = None
```
flaub: I see `_device` and `_dev` here, that doesn't seem right

tzerrell: Thanks!

tzerrell: Whoops, my fix broke things, will investigate and put in a valid fix...

plaidml2/bridge/keras/__init__.py
```diff
@@ -348,27 +437,34 @@ def categorical_crossentropy(target, output, from_logits=False):
...
-    _report_unimplemented('ceil')
+    return _KerasNode('ceil', tensor=edsl.ceil(x.tensor))
 
 
+@_log_call
 def clear_session():
-    _report_unimplemented('clear_session')
+    global _in_train_phase, _device
+    _in_train_phase = None
+    _device = None
```
flaub: We don't really have any state to clear at the moment, but the MLIR-based EDSL might in the future. But setting `_device_ = None` will definitely break things.

plaidml2/bridge/keras/__init__.py
tzerrell: ✅

plaidml2/bridge/keras/__init__.py
tzerrell: Instead re-getting the device ... not sure that matters. There was some issue that inspired us to put code here with the previous Keras backend, I'll see if I can dig it up and see whether its concerns apply...

tzerrell: Ah, the issue I was thinking of wasn't super helpful: https://github.com/plaidml/plaidml/issues/132

PR #418, opened 2019-08-12
Build plaidml2 on Windows

earhart:


earhart:
The `py_runtime` path is much cleaner.

PR #421, opened 2019-08-14
Various fixed for placing new buffers into same arena

LLITCHEV:


tile/targets/cpu/jit.cc
```diff
@@ -1683,6 +1721,8 @@ llvm::JITSymbol Runtime::findSymbol(const std::string& name) {
       {"_libxsmm_smmdispatch", symInfo(libxsmm_smmdispatch)},
       {"libxsmm_dmmdispatch", symInfo(libxsmm_dmmdispatch)},
       {"_libxsmm_dmmdispatch", symInfo(libxsmm_dmmdispatch)},
+      {"RunTimeLogEntry", symInfo(rt::RunTimeLogEntry)},   // For debugging
+      {"_RunTimeLogEntry", symInfo(rt::RunTimeLogEntry)},  // For debugging
```
mars1026: Is the second entry really necessary? There is logic a few lines down which is intended to handle the platform-dependent leading "_" character.

tile/targets/cpu/jit.cc
```diff
@@ -216,7 +219,6 @@ Compiler::Compiler(llvm::LLVMContext* context, const Config& config)
 }
 
 ProgramModule Compiler::CompileProgram(const stripe::Block& program) {
-  IVLOG(4, program);
```
mars1026: seems harmless, easy to restore later if useful

tile/targets/cpu/jit.cc
```diff
@@ -202,10 +204,11 @@ class Compiler : private stripe::ConstStmtVisitor {
   std::map<std::string, Scalar> scalars_;
   std::map<std::string, Buffer> buffers_;
   std::map<std::string, Index> indexes_;
+  uint64_t arenaSize_;
```
mars1026: given that there are two different constructor methods, perhaps it'd be safer to use value-initialization here?

tile/targets/cpu/jit.cc
LLITCHEV: It was necessary. I tried without it and the linker couldn't find the method to link with. So, I added the second entry.

tile/targets/cpu/jit.cc
LLITCHEV: Added it back in.

tile/targets/cpu/jit.cc
LLITCHEV: Good catch! Thanks!

PR #425, opened 2019-08-15
Update BatchDot to switch between TensorFlow/Theano Behavior

dgkutnic:
- Updated the plaidml backend code to check for the environment variable `PLAIDML_BATCHDOT_TF_BEHAVIOR` and switch between Theano (default) and TensorFlow-like behaviors.
- Updated the plaidml backend tests to set `skip_tensorflow` and `skip_theano` based on the environment variable from above
- Updated the troubleshooting docs to add instructions for using the environment variable `PLAIDML_BATCHDOT_TF_BEHAVIOR` and which runtime error it's typically associated with
- Updated the runtime message typically associated with `BatchDot` issues to direct users to the troubleshooting docs

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -38,7 +38,7 @@ fixed).
 If you have existing Keras code that was written for the TensorFlow backend, 
 and it is running into this issue, you can enable experimental support for 
 TensorFlow-like `BatchDot` behavior by setting the environment variable 
-`PLAIDML_BATCHDOT_TF_BEHAVIOR` tto `True`.
+`PLAIDML_BATCHDOT_TF_BEHAVIOR` to `True`.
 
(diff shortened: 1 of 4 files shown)


tzerrell:
I thought the test failures you had were fixed by @drgone's PR #412; if they persist, perhaps talk to him about them.

dgkutnic:
Yeah, I've pulled in his changes and am rerunning CI now. I'll let him know if I'm still seeing failures this time around.

drgone:
This is unrelated, base/backend is failing across all platforms.

tzerrell:
Don't know how I missed seeing those the first time. In that case, I suspect the problem might be that we approximately never use Theano (it looks like it only is invoked in testGather and now this test, which may be relevant as the error mentions gemm and testGather does not involve matmuls). So if the install of Theano on the build machines is flakey but local installs work ok, we might see this behavior.

PR #433, opened 2019-08-19
CM_backend, commits squashed

yangleizou:
This is the first version back-end to support Intel C for Metal. Developed and tested on Ubuntu 16.04.
Passed all back-end tests. Passed correctness tests of  inception_v3, inception_resnet_v2, mobilenet, mobilenet_v2, resnet50, vgg16 and vgg19.


WORKSPACE
```diff
@@ -14,3 +14,9 @@ http_archive(
 load("//bzl:workspace.bzl", "plaidml_workspace")
 
 plaidml_workspace()
+
+new_local_repository(
+    name = "cmheaders",
+    build_file = "cmheaders.BUILD",
+    path = "/usr/include/",
```
flaub: This won't work on our build machines or for public contributors. Is there a public URL that we can use to download the headers for CM?

yangleizou: Yes, the cmheaders are contained in a zip file in https://01.org/c-for-metal-development-package/downloads. Is the size(>200MB) too big? Or I can include the headers in CM github repo (about 16MB) where the headers are not contained in a single folder. I have tried to add it in a http_archive in plaidml/bzl/workspace.bzl, but the cmheader files also depend on other header files in the /usr/include path. I will try to add all dependences from public URLs.

Issue #417, opened 2019-08-10
Bug in RandomUniform Bias Initializer


Kenivia:
This is a weird bug,
if I do this:

```
model = Sequential()
 b = keras.initializers.RandomUniform(minval=-1, maxval=-1, seed=rseed)
 model.add(layers.Dense(16,
                           input_shape=(1,1000),
                           activation=keras.activations.elu,
                           use_bias=True,
                           bias_initializer=b,

                           ))
model.add(layers.Dense(10,
                           activation='linear'  ,
                           ))

```

It works just fine, but when I do this,
`b = keras.initializers.RandomUniform(minval=0, maxval=0, seed=rseed)`
it raises an error, strangely from the next layer:
```
Using plaidml.keras.backend backend.
INFO:plaidml:Opening device "opencl_amd_stoney.0"
Traceback (most recent call last):
  File "C:/Users/allen/Desktop/Big Items/Downloads/tidy_up_version.py", line 48, in <module>
    model = init_model()
  File "C:/Users/allen/Desktop/Big Items/Downloads/tidy_up_version.py", line 44, in init_model
    activation='linear'  ,
  File "C:\Users\allen\AppData\Local\Programs\Python\Python36\lib\site-packages\keras\engine\sequential.py", line 181, in add
    output_tensor = layer(self.outputs[0])
  File "C:\Users\allen\AppData\Local\Programs\Python\Python36\lib\site-packages\keras\engine\base_layer.py", line 458, in __call__
    output = self.call(inputs, **kwargs)
  File "C:\Users\allen\AppData\Local\Programs\Python\Python36\lib\site-packages\keras\layers\core.py", line 879, in call
    output = K.dot(inputs, self.kernel)
  File "C:\Users\allen\AppData\Local\Programs\Python\Python36\lib\site-packages\plaidml\tile.py", line 256, in function
    operation = cls(*args, **kwargs)
  File "C:\Users\allen\AppData\Local\Programs\Python\Python36\lib\site-packages\plaidml\op.py", line 1568, in __init__
    x.shape.dtype, y.shape.dtype))
ValueError: Invalid dtype in multiplication: x.dtype='51', y.dtype='50'
```

dgkutnic:
Hi @Kenivia,
[...]

PR #434, opened 2019-08-19
Fix bug in RandomUniform Bias Initializer #417

WLeeeee:
The bug is mainly caused by calling random_uniform with minval == maxval.

rseed = 12345
model = keras.Sequential()
b = keras.initializers.RandomUniform(minval=-1.0, maxval=-1.0, seed=rseed)·
model.add(layers.Dense(16,
                            input_shape=(1,1000),
                            activation=keras.activations.elu,
                            use_bias=True,
                            bias_initializer=b,
                            )) 
print (model.layers[0].get_weights())

When minval = maxval, the tile language does the constant propagation and return a scalar.
However, the returned scalar is always zero no matter the value.
The reason is that we should return a tensor with broadcast zero when doing constant folding.

Below shows the aggressive optimization for the "mul" operation.
Here is an example of the compiled tile code with error:
The RevMul function only has one scalar operand.
 
DEBUG:plaidml:Doing a compilation of:
function (
 X_I_0[X_I_0_0, X_I_0_1]
 ) -> (
 X_T2,
 X_T3
 ) {
 X_T0 = 16;
 [[pid(PrngStep)]] X_T1 = prng_step(X_I_0, X_T0);
 [[pid(PrngState)]] X_T2 = prng_state(X_T1);
 [[pid(RevMul)]] X_T3 = -1.0;
}

DEBUG:plaidml:Compiling: function (
X_I_0[X_I_0_0, X_I_0_1]) -> (
X_T2,
X_T3
) {
X_T0 = 16;
[[pid(PrngStep)]] X_T1 = prng_step(X_I_0, X_T0);
[[pid(PrngState)]] X_T2 = prng_state(X_T1);
[[pid(Add)]] X_T3 = -1.0;
}

Here is the right tile code after the optimization:
X_I_0[X_I_0_0, X_I_0_1]
 ) -> (
X_T2,
X_T7
) {
X_T0 = 16;
[[pid(PrngStep)]] X_T1 = prng_step(X_I_0, X_T0);
[[pid(PrngState)]] X_T2 = prng_state(X_T1);
X_T3 = -1.0;
X_T4 = 0.0;
[[pid(PrngValue)]] X_T5 = prng_value(X_T1);
[[pid(RevMul)]] X_T6 = mul(X_T4, X_T5);
[[pid(Add)]] X_T7 = add(X_T3, X_T6);
}

### DIFF
--- a/tile/lang/compose.cc
+++ b/tile/lang/compose.cc
@@ -168,7 +168,6 @@ std::shared_ptr<Value> FunctionValue::make(std::string fn, std::vector<std::shar
       return inputs[0];
     }
   }
-  
   if (fn == "mul") {
     if (inputs[0] == zeroi || inputs[0] == zerof) {
       return inputs[0];
@@ -182,7 +181,7 @@ std::shared_ptr<Value> FunctionValue::make(std::string fn, std::vector<std::shar
     if (inputs[1] == onei || inputs[1] == onef) {
       return inputs[0];
     }
-	}
+  }
   if (fn == "match" && inputs[0] == inputs[1]) {
     return inputs[0];
   }
--- a/tile/lang/compose.cc
+++ b/tile/lang/compose.cc
@@ -168,20 +168,6 @@ std::shared_ptr<Value> FunctionValue::make(std::string fn, std::vector<std::shar
       return inputs[0];
     }
   }
-  if (fn == "mul") {
-    if (inputs[0] == zeroi || inputs[0] == zerof) {
-      return inputs[0];
-    }
-    if (inputs[1] == zeroi || inputs[1] == zerof) {
-      return inputs[1];
-    }
-    if (inputs[0] == onei || inputs[0] == onef) {
-      return inputs[1];
-    }
-    if (inputs[1] == onei || inputs[1] == onef) {
-      return inputs[0];
-    }
-  }
   if (fn == "match" && inputs[0] == inputs[1]) {
     return inputs[0];
   }
--- a/tile/lang/compose.cc
+++ b/tile/lang/compose.cc
@@ -2,7 +2,6 @@
 
 #include <algorithm>
 #include <sstream>
-#include <stack>
 
 #include "tile/lang/builtins.h"
 #include "tile/lang/fpconv.h"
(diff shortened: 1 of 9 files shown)


tile/lang/compose.cc
```diff
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 #include <sstream>
+#include <stack>
```
dgkutnic: note (to Tim and just general FYI): this is a change that was requested by cpplint.

WLeeeee:
A better way to fix this issue is to provide a general way to broadcast the constant to the target tensor with specified shape and dim. However, it requires non-trivial effort to do it. Thus, a simpler fix is to return the constant tensor in Keras backend when random_uniform is being called with minval == maxval.

tile/lang/compose.cc
```diff
@@ -952,7 +956,11 @@ void FunctionApplication::SetDone() {
     const std::shared_ptr<Value> val = func_->in_bound().at(pi.name);
     if (pi.tag == Input::FIXED) {
       if (val->num_dims() != pi.dims.size()) {
-        throw std::runtime_error("Applying function, tensor with mismatching dimensionality: " + pi.name);
+        throw std::runtime_error(
+            "Applying function, tensor with mismatching dimensionality: " + pi.name +
+            "\n Check the \"Common Issues\" section on our troubleshooting documentation to see if this issue has "
```
dgkutnic: Are these changes specific to this PR?

tzerrell: Similarly, are there cases where we expect a user to hit one of these errors and information in the "Common Issues" section will help them?

dgkutnic: Our user was running into a `ValueError`, which may be caused by many reasons (of which this is just one).

These changes were actually from my `BatchDot` PR, which is why I'm confused that they appear here.

WLeeeee: I just copy and paste the up-to-date compose.cc from the website, which has those changes.
Actually, it does not belong to this fix.

flaub: I think what's needed here is a rebase or merge with master.

WLeeeee:
Moved to:
fix random uniform problem #438

PR #435, opened 2019-08-19
WIP - Add initial presubmit yml for use with BazelCI

dgkutnic:
I'm not convinced that this yml will do the right things for us on the Bazel CI machines. However, as @flaub mentioned, there may be a bit of lead time for the Bazel CI team to respond to our request, so let's merge in what I've got so far and send in the request to their team ASAP.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,0 +1,25 @@
+---
+bazel: 0.28.1
+tasks:
+  ubuntu1604:
+    name: ":linux:"
+    shell_commands:
+    - yes | ./configure
+    build_targets:
+    - "//plaidml/keras:wheel"
+    test_targets:
+    - "..."
+  macos:
+    name: ":apple:"
+    shell_commands:
+    - yes | ./configure
+    build_targets:
+    - "//plaidml/keras:wheel"
+    test_targets:
+    - "..."
+  windows:
+    name: ":window:"
+    build_targets:
+    - "//plaidml/keras:wheel"
+    test_targets:
+    - "..."
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -6,46 +6,31 @@ tasks:
     shell_commands:
     - wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
     - bash ./Miniconda3-latest-Linux-x86_64.sh -b -p $HOME/miniconda
-    - source $HOME/miniconda/bin/activate
-    - wget https://raw.githubusercontent.com/plaidml/plaidml/master/environment.yml
-    - $HOME/miniconda/bin/conda env update -f ./environment.yml
-    - conda activate $HOME/miniconda
     build_targets:
     - "//plaidml/keras:wheel"
-    build_options:
+    build_flags:
     - "--config=linux_x86_64"
     test_targets:
     - "..."
-    test_options:
-    - "--cpu=linux_x86_64"
   macos:
     name: ":apple:"
     shell_commands:
     - curl -O https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh 
     - sh ./Miniconda3-latest-MacOSX-x86_64.sh -b -p $HOME/miniconda
-    - curl -O https://raw.githubusercontent.com/plaidml/plaidml/master/environment.yml
-    - $HOME/miniconda/bin/conda env update -f ./environment.yml
-    - conda activate $HOME/miniconda
     build_targets:
     - "//plaidml/keras:wheel"
-    build_options:
-    - "--config=windows_x86_64"
+    build_flags:
+    - "--config=macos_x86_64"
     test_targets:
     - "..."
-    test_options:
-    - "--cpu=windows_x86_64"
   windows:
     name: ":window:"
     batch_commands:
     - curl https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe -OutFile .\miniconda.exe
     - start /wait "" .\miniconda.exe /InstallationType=JustMe /S /D=%UserProfile%\Miniconda3
-    - curl https://raw.githubusercontent.com/plaidml/plaidml/master/environment-windows.yml -OutFile .\environment-windows.yml
-    - conda.exe env update -f .\environment-windows.yml
     build_targets:
     - "//plaidml/keras:wheel"
-    build_options:
-    - "--config=macos_x86_64"
+    build_flags:
+    - "--config=windows_x86_64"
     test_targets:
     - "..."
-    test_options:
-    - "--cpu=darwin_x86_64 --apple_platform_type=macos"
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -4,22 +4,48 @@ tasks:
   ubuntu1604:
     name: ":linux:"
     shell_commands:
-    - yes | ./configure
+    - wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
+    - bash ./Miniconda3-latest-Linux-x86_64.sh -b -p $HOME/miniconda
+    - source $HOME/miniconda/bin/activate
+    - wget https://raw.githubusercontent.com/plaidml/plaidml/master/environment.yml
+    - $HOME/miniconda/bin/conda env update -f ./environment.yml
+    - conda activate $HOME/miniconda
     build_targets:
     - "//plaidml/keras:wheel"
+    build_options:
+    - "--config=linux_x86_64"
     test_targets:
     - "..."
+    test_options:
+    - "--cpu=linux_x86_64"
   macos:
     name: ":apple:"
     shell_commands:
-    - yes | ./configure
+    - curl -O https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh 
+    - sh ./Miniconda3-latest-MacOSX-x86_64.sh -b -p $HOME/miniconda
+    - curl -O https://raw.githubusercontent.com/plaidml/plaidml/master/environment.yml
+    - $HOME/miniconda/bin/conda env update -f ./environment.yml
+    - conda activate $HOME/miniconda
     build_targets:
     - "//plaidml/keras:wheel"
+    build_options:
+    - "--config=windows_x86_64"
     test_targets:
     - "..."
+    test_options:
+    - "--cpu=windows_x86_64"
   windows:
     name: ":window:"
+    batch_commands:
+    - curl https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe -OutFile .\miniconda.exe
+    - start /wait "" .\miniconda.exe /InstallationType=JustMe /S /D=%UserProfile%\Miniconda3
+    - curl https://raw.githubusercontent.com/plaidml/plaidml/master/environment-windows.yml -OutFile .\environment-windows.yml
+    - conda.exe env update -f .\environment-windows.yml
     build_targets:
     - "//plaidml/keras:wheel"
+    build_options:
+    - "--config=macos_x86_64"
     test_targets:
     - "..."
+    test_options:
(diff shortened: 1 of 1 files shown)


flaub:
Looks like a good start!

I think we can use the `build_flags` and `test_flags` as a replacement for the configure script entirely.

The only thing remaining I think is bootstrapping the tools. So getting conda installed is probably the only required thing. On Windows, we might also need to have the `.cenv` conda environment setup (but note this isn't required for *nix).

dgkutnic:
Hey @flaub,
I've pushed some more changes trying to address the things you mentioned. I've added `build_flags` and `test_flags` but I may have missed some - let me know. For conda setup, I'm not sure exactly what `conda.exe activate` command we want to use for windows specifically (would it just be `conda.exe activate .\.cenv` or is there something I'm missing?)

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
   ubuntu1604:
     name: ":linux:"
     shell_commands:
-    - yes | ./configure
+    - wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
+    - bash ./Miniconda3-latest-Linux-x86_64.sh -b -p $HOME/miniconda
+    - source $HOME/miniconda/bin/activate
```
flaub: No need for this source

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
+    - wget https://raw.githubusercontent.com/plaidml/plaidml/master/environment.yml
```
flaub: We don't need to download because the current working dir should be the checkout of the repo, which should have `environment.yml` right at the top

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
+    - $HOME/miniconda/bin/conda env update -f ./environment.yml
```
flaub: There's no need to create a conda environment on unix, bazel will manage all of its conda environments directly.

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
+    - conda activate $HOME/miniconda
```
flaub: Ditto as above, we don't need to activate an environment to do a build (we'd only need to activate if we anticipate needing to run pre-commit hooks, but CI machines won't need to do that).

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
     build_targets:
     - "//plaidml/keras:wheel"
+    build_options:
+    - "--config=linux_x86_64"
     test_targets:
     - "..."
+    test_options:
+    - "--cpu=linux_x86_64"
```
flaub: To get the right set of options here, run `./configure` on a linux machine and see what file it produces (`.bazelrc.configure`). You can also take a peek on `buildkite-agent@lab-14:.bazelrc`.

It should only have this line: 
```
build --config=linux_x86_64
```

This would correspond to: 
```
build_options:
- --config=linux_x86_64
```

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
   macos:
     name: ":apple:"
     shell_commands:
-    - yes | ./configure
+    - curl -O https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh 
+    - sh ./Miniconda3-latest-MacOSX-x86_64.sh -b -p $HOME/miniconda
+    - curl -O https://raw.githubusercontent.com/plaidml/plaidml/master/environment.yml
```
flaub: Same as above for linux, we don't need to download the environment or activate it, bazel should take care of all this (we only need to ensure conda itself is installed).

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
+    - $HOME/miniconda/bin/conda env update -f ./environment.yml
+    - conda activate $HOME/miniconda
     build_targets:
     - "//plaidml/keras:wheel"
+    build_options:
+    - "--config=windows_x86_64"
     test_targets:
     - "..."
+    test_options:
+    - "--cpu=windows_x86_64"
```
flaub: This isn't needed (just as like with linux)

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
-    - yes | ./configure
+    - curl -O https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh 
+    - sh ./Miniconda3-latest-MacOSX-x86_64.sh -b -p $HOME/miniconda
+    - curl -O https://raw.githubusercontent.com/plaidml/plaidml/master/environment.yml
+    - $HOME/miniconda/bin/conda env update -f ./environment.yml
+    - conda activate $HOME/miniconda
     build_targets:
     - "//plaidml/keras:wheel"
+    build_options:
+    - "--config=windows_x86_64"
```
flaub: You probably meant `- --config=macos_x86_64` here

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
+    - curl https://raw.githubusercontent.com/plaidml/plaidml/master/environment-windows.yml -OutFile .\environment-windows.yml
+    - conda.exe env update -f .\environment-windows.yml
     build_targets:
     - "//plaidml/keras:wheel"
+    build_options:
+    - "--config=macos_x86_64"
     test_targets:
     - "..."
+    test_options:
+    - "--cpu=darwin_x86_64 --apple_platform_type=macos"
```
flaub: This isn't needed just like with linux and macos.

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
     name: ":window:"
+    batch_commands:
+    - curl https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe -OutFile .\miniconda.exe
+    - start /wait "" .\miniconda.exe /InstallationType=JustMe /S /D=%UserProfile%\Miniconda3
+    - curl https://raw.githubusercontent.com/plaidml/plaidml/master/environment-windows.yml -OutFile .\environment-windows.yml
+    - conda.exe env update -f .\environment-windows.yml
     build_targets:
     - "//plaidml/keras:wheel"
+    build_options:
+    - "--config=macos_x86_64"
```
flaub: You probably meant `- --config=windows_x86_64` here

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
     test_targets:
     - "..."
+    test_options:
+    - "--cpu=windows_x86_64"
   windows:
     name: ":window:"
+    batch_commands:
+    - curl https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe -OutFile .\miniconda.exe
+    - start /wait "" .\miniconda.exe /InstallationType=JustMe /S /D=%UserProfile%\Miniconda3
+    - curl https://raw.githubusercontent.com/plaidml/plaidml/master/environment-windows.yml -OutFile .\environment-windows.yml
```
flaub: No need to download the file, it should already be present when we do a checkout. (also, we want to make sure the environment file matches the commit we are trying to build)

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
+    build_options:
+    - "--config=windows_x86_64"
     test_targets:
     - "..."
+    test_options:
+    - "--cpu=windows_x86_64"
   windows:
     name: ":window:"
+    batch_commands:
+    - curl https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe -OutFile .\miniconda.exe
```
flaub: We might want to use a stable version number instead of always just using latest. I've run into bugs with certain version of conda.

.bazelci/presubmit.yml
```diff
@@ -4,22 +4,48 @@ tasks:
...
+    - start /wait "" .\miniconda.exe /InstallationType=JustMe /S /D=%UserProfile%\Miniconda3
+    - curl https://raw.githubusercontent.com/plaidml/plaidml/master/environment-windows.yml -OutFile .\environment-windows.yml
+    - conda.exe env update -f .\environment-windows.yml
```
flaub: We might have to deal with creating a conda environment vs updating one. It might be easier to have a python script that we call to handle all this.

Also, we need to somehow activate the environment before running the build on windows.

.bazelci/presubmit.yml
```diff
@@ -0,0 +1,51 @@
...
+    shell_commands:
+    - wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
+    - bash ./Miniconda3-latest-Linux-x86_64.sh -b -p $HOME/miniconda
+    - source $HOME/miniconda/bin/activate
+    - wget https://raw.githubusercontent.com/plaidml/plaidml/master/environment.yml
+    - $HOME/miniconda/bin/conda env update -f ./environment.yml
+    - conda activate $HOME/miniconda
+    build_targets:
+    - "//plaidml/keras:wheel"
+    build_options:
```
flaub: I think it's called `build_flags` not `build_options`

https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#using-specific-build--test-flags

.bazelci/presubmit.yml
dgkutnic: Thanks for this explanation! I had tried to derive the test flags from the info messages from `bazelisk test //...` because I didn't know about using `.bazelrc.configure` to get the correct set of options.

.bazelci/presubmit.yml
dgkutnic: Miniconda3 4.7.10 is the most recent pinned version that the [conda repos](https://repo.anaconda.com/miniconda/) have. It's also the same thing that latest currently points to. There's also a couple of versions from earlier this year, 4.6.14 and 4.5.12.

Do you think it's a good idea to try running on the Bazel CI machines with latest once as a proof of concept, then pin that version?

.bazelci/presubmit.yml
dgkutnic: Fixed in commit [`814cd24`](https://github.com/plaidml/plaidml/pull/435/commits/814cd246c8e9fee003c7abba4b074f7e7dc9259d)

.bazelci/presubmit.yml
dgkutnic: Fixed in commit [`814cd24`](https://github.com/plaidml/plaidml/pull/435/commits/814cd246c8e9fee003c7abba4b074f7e7dc9259d)

.bazelci/presubmit.yml
dgkutnic: Fixed in commit [`814cd24`](https://github.com/plaidml/plaidml/pull/435/commits/814cd246c8e9fee003c7abba4b074f7e7dc9259d)

.bazelci/presubmit.yml
dgkutnic: Fixed in commit [`814cd24`](https://github.com/plaidml/plaidml/pull/435/commits/814cd246c8e9fee003c7abba4b074f7e7dc9259d)

.bazelci/presubmit.yml
dgkutnic: Fixed in commit [`814cd24`](https://github.com/plaidml/plaidml/pull/435/commits/814cd246c8e9fee003c7abba4b074f7e7dc9259d)

.bazelci/presubmit.yml
dgkutnic: Yup, I've fixed this in commit [`814cd24`](https://github.com/plaidml/plaidml/pull/435/commits/814cd246c8e9fee003c7abba4b074f7e7dc9259d)

.bazelci/presubmit.yml
dgkutnic: Working on this now!

.bazelci/presubmit.yml
flaub: SGTM

PR #436, opened 2019-08-19
Add basic testing of PlaidML2 EDSL in plaidbench

tzerrell:
Rudimentary access to plaidbench testing with the PlaidML2/EDSL op library. This is just getting initial tests working (e.g. ResNet50); this PR does not make every network run correctly on the PlaidML2/EDSL op library


plaidbench/plaidbench.py
```diff
@@ -55,6 +56,9 @@ def make_parser():
     parser = argparse.ArgumentParser()
     plaidargs = parser.add_mutually_exclusive_group()
     plaidargs.add_argument('--plaid', action='store_true', help="Use PlaidML as the backend.")
+    plaidargs.add_argument('--plaid-edsl',
```
flaub: Since plaidml2 is a temporary transition, it'd be good if we could specify the `KERAS_BACKEND` from the outside and thus not require a separate `--plaid-edsl` flag.

plaidbench/plaidbench.py
```diff
@@ -180,7 +186,10 @@ def main():
             argv.append('--refresh-onnx-data is only meaningful with --onnx')
         if args.fix_learn_phase:
             argv.append('--fix-learn-phase')
-        if args.plaid or (not args.no_plaid and not args.caffe2):
+        if args.plaid_edsl:
+            argv.append('--plaid-edsl')
+            os.environ["KERAS_BACKEND"] = "plaidml2.bridge.keras.__init__"
```
flaub: BTW, I don't think the `__init__` is required

plaidml/keras/backend.py
```diff
@@ -1249,6 +1249,7 @@ def print_tensor(x, message=''):
 
 @_log_call
 def prod(value, axis=None, keepdims=False):
+    print("Sanity check on prod: value: {}, axis: {}, keepdims: {}".format(value, axis, keepdims))
```
flaub: Is this still needed?

tzerrell: No, thanks for the catch!

plaidbench/plaidbench.py
tzerrell: Should be doable

plaidbench/plaidbench.py
tzerrell: You're right, it's not

tzerrell:
Reverted to the `--plaid-edsl` version

PR #439, opened 2019-08-21
[skip ci] enable sharding in ci

drgone:


ci/harness.py
```diff
@@ -14,17 +14,29 @@
 
 DEFAULT_BUILD_URL = 'https://buildkite.com/plaidml'
 
+am_in_shard = False
```
flaub: Unused?

ci/report.py
```diff
@@ -24,7 +24,6 @@
 
 
 def printf(*args, **kwargs):
-    print(*args, **kwargs)
```
flaub: Is this intentional?

PR #444, opened 2019-08-22
WIP - Documentation Updates

dgkutnic:
Updating documentation to address the following:
* Command line options in `configure` script
* Updated troubleshooting docs and made them more accessible from the `README`

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -30,7 +30,8 @@ https://github.com/bazelbuild/bazelisk/releases
 
 ## Configure the build
 
-Use the `configure` script to configure your built. Note: the `configure` script requires Python 3.
+Use the `configure` script to configure your built. Note: the `configure` script
+requires Python 3.
 
 By default, running the `configure` script will:
 * Create and/or update your conda environment
--- a/docs/building.md
+++ b/docs/building.md
@@ -30,7 +30,12 @@ https://github.com/bazelbuild/bazelisk/releases
 
 ## Configure the build
 
-Note: the `configure` script requires Python 3.
+Use the `configure` script to configure your built. Note: the `configure` script requires Python 3.
+
+By default, running the `configure` script will:
+* Create and/or update your conda environment
+* Configure pre-commit hooks for development purposes
+* Configure bazelisk based on the device you're using
 
 ```
 ./configure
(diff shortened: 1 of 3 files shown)


docs/building.md
```diff
@@ -30,7 +30,13 @@ https://github.com/bazelbuild/bazelisk/releases
 
 ## Configure the build
 
-Note: the `configure` script requires Python 3.
+Use the `configure` script to configure your built. Note: the `configure` script
```
flaub: > to configure your built
should be
> to configure your build

docs/building.md
```diff
@@ -30,7 +30,13 @@ https://github.com/bazelbuild/bazelisk/releases
...
+requires Python 3.
+
+By default, running the `configure` script will:
+* Create and/or update your conda environment
+* Configure pre-commit hooks for development purposes
+* Configure bazelisk based on the device you're using
```
flaub: What does 'based on the device you're using mean?

Are you talking about the --config line?

I'd say:

* Configure bazel based on your host OS

docs/troubleshooting.md
```diff
@@ -19,9 +19,21 @@ provide you with better support on your issue.
...
+### Bazel Issues
+
+`Encountered error while reading extension file 'workspace.bzl': no such package
+'@toolchain//'`
+
+On MacOS devices, `toolchain` errors often indicate that the user does not have
+Xcode properly installed. Even if you have Xcode Command Line Tools installed,
+you may not have a proper installation of Xcode itself. You can verify your
+Xcode installation by running `spctl --assess --verbose
+/Applications/Xcode.app`.
```
flaub: I think we should point people to the bazel docs in this case:

https://docs.bazel.build/versions/0.28.0/install-os-x.html

Also, I think what we can do is say, run: 
```
xcode-select -p
``` 

And if it doesn't say `/Applications/Xcode.app/Contents/Developer` then go to the App Store and install Xcode. Then be sure to use:
```
bazelisk clean --expunge
```
before trying to use bazel again.

docs/building.md
dgkutnic: Thanks for catching this!

docs/building.md
dgkutnic: Yes, that's what I meant. I've changed the wording on this.

docs/troubleshooting.md
flaub: `spctl` just hangs for me, but why are we trying to asses something that might not yet exist? For users without Xcode installed, they won't have a `/Application/Xcode.app` directory.

dgkutnic: Fair point - I'm modifying the docs to address this.

PR #446, opened 2019-08-23
Allocate temp buffers on demand;

shiyao00:
Allocate temp buffers on demand;
Release useless buffers during execution;
Add global memory size in hardware options.


tile/platform/local_machine/mem_cache.cc
```diff
@@ -8,6 +8,25 @@ namespace vertexai {
...
 
+MemCache::MemCache(const std::shared_ptr<DevInfo>& dev_info, hal::Memory* source) :
+  total_{0}, source_{source} {
+  mem_limit_ = dev_info->settings.max_global_mem();
+}
+
+void MemCache::Cleanup(size_t new_size) {
+  while (total_ + new_size > mem_limit_) {
+    // Out of device memory, clean up cache
+    while (mem_.size() > 0 && mem_.begin()->second.empty()) {
```
earhart: What's the goal here?  The thing is, as soon as the current program finishes, it's going to release all of its temporary memory to this cache, which is just going to have to reallocate these table entries, right?

(If this is a good thing to do, why not do it directly at the points where memory is being allocated?)

earhart: Fun trick: if you consistently order comparisons s.t. you're always using '<', it makes it just a little easier to think of things as existing on a number line, and makes it just a little harder to get your comparisons wrong.  Not as big a deal for something this simple (you should just remove the comparison), but I think it's a good idea in general.

tile/platform/local_machine/mem_cache.cc
```diff
@@ -16,11 +35,16 @@ std::shared_ptr<hal::Buffer> MemCache::TryAlloc(std::size_t size) {
     l.pop();
     return result;
   }
-  return std::shared_ptr<hal::Buffer>{};
+  Cleanup(size);
+  total_ += size;
+  auto buffer = source_->MakeBuffer(size, hal::BufferAccessMask::DEVICE_RW);
```
earhart: Why are you moving the underlying allocation into the cache?

IMHO -- this is making the class more complicated than it needs to be; now it needs to know about memory allocation policy.

tile/platform/local_machine/mem_cache.h
```diff
@@ -17,11 +18,16 @@ namespace local_machine {
...
 class MemCache {
  public:
+  MemCache(const std::shared_ptr<DevInfo>& dev_info, hal::Memory* source);
   std::shared_ptr<hal::Buffer> TryAlloc(std::size_t size);
   void Free(std::size_t size, std::shared_ptr<hal::Buffer>);
+  void Cleanup(size_t new_size);
 
  private:
+  size_t total_;
+  size_t mem_limit_;
```
earhart: `mem_limit_` should be const.  (I'm not sure it belongs here, but if it does, it should be const.  :-)

tile/platform/local_machine/direct_mem_strategy.cc
```diff
@@ -37,13 +37,15 @@ class DirectMemChunk final : public MemChunk {
...
   std::uint64_t size() const final { return size_; }
+  std::uint64_t DecreaseUses() final { return --uses_; }
 
   // MemChunk implementation
   std::shared_ptr<MemDeps> deps() final { return deps_; }
   std::shared_ptr<hal::Buffer> hal_buffer() final { return mem_; }
 
  private:
   std::uint64_t size_;
+  std::uint64_t uses_;
```
earhart: Where is this being initialized?

tile/platform/local_machine/mem_chunk.h
```diff
@@ -19,6 +19,9 @@ class MemChunk : public Buffer {
 
   // Gets the chunk's underlying HAL buffer.
   virtual std::shared_ptr<hal::Buffer> hal_buffer() = 0;
+
+  // Decrease uses, return the uses after decrease
+  virtual std::uint64_t DecreaseUses() = 0;
```
earhart: This is a fragile pattern (it's super easy to get this kind of code wrong, especially when it's subsequently modified by someone else).  Use RAII patterns instead, reifying the notion of "use" as an object that can be managed via `std::shared_ptr`.

tile/targets/gpu/amd.jsonnet
```diff
@@ -196,6 +196,14 @@ local PARAMS = {
               }
             },
 
+            {
+              name: 'reduce_constraints',
```
earhart: What's this doing in this commit?

PR #449, opened 2019-08-23
Add vega56 macosx (plaid,stripe,ocl,mtl) to CI

drgone:


flaub:
Doing a little spring cleaning...

PR #459, opened 2019-08-29
fix stderr on cmd_build

drgone:


flaub:
I think this is all covered by the PR I just landed: https://github.com/plaidml/plaidml/commit/2a107266cf62d6d832ff20604057907d28333eb2

PR #465, opened 2019-08-30
Replace conda_* with in-build py_runtime

flaub:
This allows plaidml2 and a few other things to build and run on windows.


tools/analysis/analysis_test.py
```diff
@@ -23,7 +23,9 @@ def testRunNotebook(self):
             # due to bash's limit of 128 chars for the shebang line
             # see: https://github.com/pypa/pip/issues/1773
             # see: https://www.in-ulm.de/~mascheck/various/shebang/#length
-            cmd = ['python', os.path.join(conda_env, 'bin', 'jupyter-nbconvert')]
+            python = os.path.join(conda_env, 'bin', 'python')
+            nbconvert = os.path.join(conda_env, 'bin', 'jupyter-nbconvert')
+            cmd = [python, nbconvert]
```
earhart: Doesn't conda automatically ensure that binaries in the bin dir use the environment's python?  (I'm guessing you put this here because it doesn't, though...?  :-)

tools/conda_run/conda_unix.cc
```diff
@@ -0,0 +1,78 @@
...
+    // locate assets within runfiles
+    auto conda_env = fs::canonical(runfiles->Rlocation("com_intel_plaidml_conda_unix/env"));
+    auto bin_dir = conda_env / "bin";
+    auto python = fs::canonical(bin_dir / "python");
+
+    // Adjust environment variables to activate conda environment
+    std::stringstream ss;
+    ss << bin_dir << kPathSeparator << vertexai::env::Get("PATH");
+    vertexai::env::Set("PATH", ss.str());
+    vertexai::env::Set("CONDA_DEFAULT_ENV", conda_env.string());
```
earhart: FYI -- Conda also sets `CONDA_EXE`, `CONDA_PREFIX`, and `CONDA_PYTHON_EXE`.  Dunno how much those matter to our tests.

tools/analysis/analysis_test.py
flaub: It does that by using the shebang line, however this is broken on linux because the shebang line can only be something paltry like 128 chars. So we must invoke python directly.

earhart: SGTM

PR #466, opened 2019-08-30
EDSL convert backend_tests

1sanz:
convert backend_tests to EDSL


plaidml2/bridge/keras/backend_test.py
```diff
@@ -476,21 +477,39 @@ def testFlatten(self, b, x):
...
+        I.bind_dims(N)
+        O = edsl.TensorOutput(N)
+        i = edsl.TensorIndex()
+        O[i] = I[i]
+        opId = pkb._KerasNode('tile_identity', name='opId', tensor=O)
+        pkb.eval(opId)
         return 0
 
-    @unittest.skip('TODO: convert to EDSL')
+    #@unittest.skip('TODO: convert to EDSL')
```
tzerrell: Since two outputs isn't currently supported with the EDSL, maybe leave the skip in with a note to that effect

plaidml2/bridge/keras/backend_test.py
```diff
@@ -1071,49 +1090,88 @@ def testConv2dTranspose(self, b, x, k, os, st, pd, df, dr):
         ]
 
     @opTest([[m(1, 3, 3, 1), m(1, 3, 3, 1) - 2]], skip_tensorflow=True, skip_theano=True)
-    @unittest.skip('TODO: convert to EDSL')
+    #@unittest.skip('TODO: convert to EDSL')
```
tzerrell: Remove the unit test skip decorator entirely instead of just commenting it out

plaidml2/bridge/keras/backend_test.py
```diff
@@ -1071,49 +1090,88 @@ def testConv2dTranspose(self, b, x, k, os, st, pd, df, dr):
...
+        I = x.tensor
+        K = k.tensor
+        N = edsl.TensorDim()
+        L0 = edsl.TensorDim()
+        L1 = edsl.TensorDim()
+        CI = edsl.TensorDim()
+        LK0 = edsl.TensorDim()
+        LK1 = edsl.TensorDim()
+        C0 = edsl.TensorDim()
+        C1 = edsl.TensorDim()
```
tzerrell: Since the output is a fixed size and you're therefore not calling `bind_dims`, you also don't need to create all these `TensorDim`s.

plaidml2/bridge/keras/backend_test.py
```diff
@@ -1071,49 +1090,88 @@ def testConv2dTranspose(self, b, x, k, os, st, pd, df, dr):
...
-             ' * K[2 - k0, 2 - k1, co, ci]);\n}')
-        return [
-            tile.Operation(f, [('I', x), ('K', k)], [('O', tile.Shape(x.shape.dtype,
-                                                                      (1, 5, 5, 1)))],
-                           name='DefractTest').sole_output()
-        ]
+
+        I = x.tensor
+        K = k.tensor
+        N = edsl.TensorDim()
```
flaub: You could say: `N, L0, L1, CI, LK0, LK1, C0, C1 = edsl.TensorDims(8)`

plaidml2/bridge/keras/backend_test.py
```diff
@@ -1071,49 +1090,88 @@ def testConv2dTranspose(self, b, x, k, os, st, pd, df, dr):
...
+        L0 = edsl.TensorDim()
+        L1 = edsl.TensorDim()
+        CI = edsl.TensorDim()
+        LK0 = edsl.TensorDim()
+        LK1 = edsl.TensorDim()
+        C0 = edsl.TensorDim()
+        C1 = edsl.TensorDim()
+        n = edsl.TensorIndex()
```
flaub: Likewise, you could say: `n, c0, c1, co, ci, k0, k1 = edsl.TensorIndexes(7)`

plaidml2/bridge/keras/backend_test.py
```diff
@@ -476,21 +477,39 @@ def testFlatten(self, b, x):
...
-        op = tile.Operation('function (I[N]) -> (O) { O = I; }', [('I', x)],
-                            [('O', tile.Shape(x.shape.dtype, (3,)))])
-        output = op.sole_output().eval()
+        I = pkb.variable(m(3)).tensor
+        N = edsl.TensorDim()
+        I.bind_dims(N)
+        O = edsl.TensorOutput(N)
+        i = edsl.TensorIndex()
+        O[i] = I[i]
+        opId = pkb._KerasNode('tile_identity', name='opId', tensor=O)
```
flaub: Actually, I think the equivalent expression would be to pass `I` into `tensor=` kwarg of `_KerasNode`. Basically we're testing that inputs can also be outputs. I'm not actually sure if this one will pass right now :)

plaidml2/bridge/keras/backend_test.py
```diff
@@ -1599,27 +1657,26 @@ def testConvParameterRankExceptions(self):
...
+        O = edsl.TensorOutput(L, N)
+        O[i, j] = A[i, k] * B[k, j]
+        OutputAB = pkb._KerasNode('assignment_non_exception', name='opAB', tensor=O)
+        npt.assert_allclose(pkb.eval(OutputAB), np.dot(m(5, 1), m(1, 5)))
+        with self.assertRaises(RuntimeError) as cm:
+            O = edsl.TensorOutput(L, N)
+            O[i, j] = B[i, j] * A[j, k]
+            OutputBA = pkb._KerasNode('assignment_exception', name='opBA', tensor=O)
+            OutputBA.eval()
         self.assertTrue("Multiple assignment" in str(cm.exception))
```
tzerrell: This "Multiple assignment" line probably needs to be indented, I think `cm` will be out of scope otherwise

plaidml2/bridge/keras/backend_test.py
tzerrell: Oh cool, I forgot we could do this! :)

plaidml2/bridge/keras/backend_test.py
1sanz: tried passing I into _KerasNode, see updated code. The test breaks with I[i]=I[I]. Not sure how to write this one.

tzerrell:
This looks good. @flaub and I should discuss what the out as in test should look like long term but I think this can land without waiting for that discussion

plaidml2/bridge/keras/backend_test.py
tzerrell: I re-read this and now understand what you were intending; we'll include this change in the PR.

plaidml2/bridge/keras/backend_test.py
```diff
@@ -476,15 +477,18 @@ def testFlatten(self, b, x):
...
-                            [('O', tile.Shape(x.shape.dtype, (3,)))])
-        output = op.sole_output().eval()
+        I = pkb.variable(m(3)).tensor
+        N = edsl.TensorDim()
+        #I.bind_dims(N)
+        #I = edsl.TensorOutput(N) #not sure if this is needed ?code segfaults with this line in
+        i = edsl.TensorIndex()
+        I[i] = I[i]
+        opId = pkb._KerasNode('tile_identity', name='opId', tensor=I)
+        pkb.eval(opId)
```
flaub: I think here, we want:

```
I = pkb.variable(m(3)).tensor
pkb.eval(I)
```

tzerrell:
> perhaps they could go into `plaidml2/edsl/edsl_test.py`?

Good idea

plaidml2/bridge/keras/backend_test.py
1sanz: Changed as above. Fails with above code in backend_test.py but once moved to edsl_test.py passes with:  

        I = Tensor(LogicalShape(plaidml.DType.FLOAT32, [3]), name='I')
        program = Program('tile_identity', [I])
        self.assertMultiLineEqual(str(program), '''function (
  I[I_0]
) -> (
  _X0
) {
  _X0 = ident(I);
}
''')

is this the expectation?

also should the tests be removed from backend_test.py once they are rewritten in edsl_test.py?

flaub: That looks right to me. Ideally we'd actually run the program however. That gets a little harder to validate.

@tzerrell did these tests compare to tensorflow in backend_test? I can't imagine they do. Maybe just running them once to ensure they don't crash is enough?

tzerrell: They did not compare to TF; they were only checking that they don't crash. My plan was that I'm manually verifying that the code looks correct, but if you could show us how to test running the program here that would be better

plaidml2/edsl/edsl_test.py
```diff
@@ -386,6 +391,213 @@ def test_use_default(self):
...
+}
+''')
+        outputs = run(program, [(A, np.array([[1], [2], [3], [4], [5]])),
+                                (B, np.array([1, 2, 3, 4, 5]))])
+        self.assertEquals(outputs[0].tolist(),
+                          [[1., 2., 3., 4., 5.], [2., 4., 6., 8., 10.], [3., 6., 9., 12., 15.],
+                           [4., 8., 12., 16., 20.], [5., 10., 15., 20., 25.]])
+
+        O = TensorOutput(L, N)
+        O[i, j] = B[j,] * A[j, k]
```
tzerrell: This should be `O[i, j] = B[i, k] * A[k, j]`; and the shape of O should be `N, L` (reversed order). I would expect this test to fail (via not throwing where expected) -- if so, add an appropriate skip.

plaidml2/edsl/edsl_test.py
```diff
@@ -386,6 +391,213 @@ def test_use_default(self):
...
+) {
+  _X0[x2, x0 : 5, 5] = =(B[x0] * A[x0, x1]);
+}
+''')
+        with self.assertRaises(plaidml.Error) as cm:
+            outputs = run(program, [(A, np.array([[1], [2], [3], [4], [5]])),
+                                    (B, np.array([1, 2, 3, 4, 5]))])
+        self.assertTrue("More indexes than dimensions for tensor" in str(cm.exception))
+
+    #@unittest.skip('TODO: convert to EDSL ')
```
tzerrell: Remove this commented-out skip altogether

plaidml2/edsl/edsl_test.py
1sanz: with O[i, j] = B[i, k] * A[k, j] the test ran fine. 
See code for compared output. Not sure if this is desired behavior.

1sanz: O = TensorOutput(L, N) and O = TensorOutput(N, L) produced identical behavior

tzerrell: This is not desired behavior. This test is accurately reporting that we are not throwing when an illegal assignment aggregation is requested. I recommend wrapping in a `with self.assertRaises(pml2_ffi_Error):` block and skipping the test (since it will fail).

Also: The shape I had in mind for the output was `O = TensorOutput(M, M)`, but this will also test what we want it to test with `L, N` or `N, L`.

plaidml2/edsl/edsl_test.py
```diff
@@ -386,6 +391,205 @@ def test_use_default(self):
...
+  _X0 = ident(I);
+  _X1 = ident(I);
+}
+''')
+        outputs = run(program1, [(I, np.array([(1, 2, 3)]))])
+        self.assertEqual(outputs[0].tolist(), [1, 2, 3])
+        self.assertEqual(outputs[1].tolist(), [1, 2, 3])
+
+        N = TensorDim(3)
+        O1 = TensorOutput(N)
```
flaub: There's no need to say `TensorOutput` here, since we're stomping the value of `O1` and `O2` in the next lines anyways.

plaidml2/edsl/edsl_test.py
```diff
@@ -386,6 +391,205 @@ def test_use_default(self):
...
+) -> (
+  _X0
+) {
+  _X0 = ident(I);
+}
+''')
+        outputs = run(program, [(I, np.array([(1, 2, 3)]))])
+        self.assertEqual(outputs[0].tolist(), [1, 2, 3])
+
+    @unittest.skip('TODO: convert to EDSL -  exception needs to be thrown')
```
flaub: Conversion to EDSL is already done

plaidml2/edsl/edsl_test.py
```diff
@@ -386,6 +391,205 @@ def test_use_default(self):
...
+  _X0
+) {
+  _X0[x0, x1, x3, x6 : 1, 5, 5, 1] = +(I[x0, -1/2 + 1/2*x1 + 1/2*x2, -1/2 + 1/2*x3 + 1/2*x4, x5] * K[2 - x2, 2 - x4, x6, x5]);
+}
+''')
+        #outputs = run(program, [(I, np.random.rand(1, 3, 3, 1)), (K, np.random.rand(1, 3, 3, 1))])
+        #crashes when run on linux. Does not crash on mac.
+        #if any one of the divide by twos is removed it no longer crashed
+        #valgrnd was run and could not find anything, the test was written up in edsl_test.cc and run, the crash was reproduced exactly
+        #issue created in tracking system (#)
```
flaub: Could we put these comments before the line that we commented instead of after?

plaidml2/edsl/edsl_test.py
```diff
@@ -7,6 +7,11 @@
 
 import plaidml2 as plaidml
 from plaidml2.edsl import *
+import plaidml2.settings as plaidml_settings
+import plaidml2.exec as plaidml_exec
+
+device = plaidml_settings.get('PLAIDML_DEVICE')
+target = plaidml_settings.get('PLAIDML_TARGET')
```
flaub: `PLAIDML_DEVICE` and `PLAIDML_TARGET` need to be configured either in a `~/.plaidml2` file on all build machines, or added to the `plan.yml` somehow. @drgone Can you help get this setup on all the build machines?

drgone: plan.yml seems like a better idea for build machines that have more than one device.

flaub: In this context, we only need to configure build machines (not those involved with tests) so there should only ever be one device (and probably LLVM or OpenCL CPU is appropriate).

drgone: got it,  related: added env vars to ops/buildkite, and modified harness to pick it up #493

PR #469, opened 2019-09-03
Throw on invalid shape request

tzerrell:


plaidml2/edsl/__init__.py
```diff
@@ -480,7 +480,11 @@ def use_default(self, rhs):
     # Return the tensor's shape
     @property
     def shape(self):
-        return LogicalShape(ptr=ffi_call(lib.plaidml_expr_get_shape, self.as_ptr()))
+        self_ptr = self.as_ptr()
+        if not self_ptr:
+            raise RuntimeError(
+                'Cannot compute shape of non-placeholder tensor before it is assigned')
+        return LogicalShape(ptr=ffi_call(lib.plaidml_expr_get_shape, self_ptr))
```
flaub: Although it's good to validate at this level, it doesn't obviate the need to validate at the ffi layer too. In that case, I think we should do this check in `ffi.cc`.

PR #474, opened 2019-09-04
Fix linker failed

TolyaTalamanov:


TolyaTalamanov:
Failed log:
`/usr/bin/ld.gold: error: bazel-out/k8-opt/bin/external/com_intel_plaidml/plaidml/plaidml.lds:1:9: invalid use of VERSION in input file
collect2: error: ld returned 1 exit status
Target @com_intel_plaidml//plaidml:wheel failed to build
`

plaidml/BUILD
```diff
@@ -387,7 +387,7 @@ plaidml_cc_binary(
     name = "libplaidml_pub.so",
     linkopts = [
         # Prefer internal library definitions.
-        "-Wl,-Bsymbolic-functions",
+        "-Wl,-Bsymbolic-functions,-T",
```
earhart: I think the `-T` belongs on the linker script line below.  But I'm a little unclear on why this is necessary -- are you compiling with `--config=linux_x86_64`?  If not, that would explain why it's using your system toolchain.

TolyaTalamanov: Thanks!

PR #477, opened 2019-09-04
Use gcc 8.3 for the Linux toolchain

earhart:


plaidml/BUILD
```diff
@@ -394,6 +410,12 @@ plaidml_cc_binary(
 
         # Use an explicit linker script.
         "-Wl,$(location :plaidml.lds)",
+
+        # Link static C++ dependencies.
+        "-static-libgcc",
+        "-Wl,-Bstatic",
+        "-lstdc++",
+        "-Wl,-Bdynamic",
```
flaub: This seems duplicated from the base linker flags, is it required here?

earhart: Nope, no longer required.  Removed.

PR #482, opened 2019-09-06
WIP: implement batch_dot in plaidml2

dgkutnic:
* Implemented Theano-like batch_dot behavior in EDSL
* Ported the modified `testBatchDot` from plaidml to plaidml2

What's next: Implement TensorFlow-like batch_dot behavior in the EDSL


plaidml2/bridge/keras/__init__.py
```diff
@@ -324,7 +324,44 @@ def backend():
 
 @_log_call
 def batch_dot(x, y, axes=None, name=None):
-    _report_unimplemented('batch_dot')
+    X = x.tensor
+    Y = y.tensor
+    if isinstance(axes, int):
```
flaub: Could we use `six.integer_types` here to make it work on python2 + python3?

dgkutnic: Certainly

PR #488, opened 2019-09-08
prefer latest py3.6

drgone:


flaub:
Looks like https://github.com/plaidml/plaidml/pull/487 that just landed already has this?

PR #493, opened 2019-09-09
plaidml device/target env for test

drgone:


tile/targets/BUILD
```diff
@@ -11,7 +11,6 @@ gencfg(
         "//tile/targets/gpu:nvidia.json",
         "//tile/targets/gpu:intel_gen9.json",
     ] + select({
```
flaub: The way to do this would be to remove the `select()` entirely and merge the default `"//tile/targets/cpu:cpu.json"` into the `srcs` list.

tzerrell:
BTW additional conditional building of cpu targets occurs in `tile/pmlc/BUILD` and `tile/codegen/BUILD`

tile/targets/BUILD
```diff
@@ -30,11 +26,6 @@ plaidml_cc_library(
     ],
     deps = [
         "//tile/codegen",
-    ] + select({
```
flaub: Oops, this one got dropped.

drgone: so move //tile/targets/cpu to deps[]?

PR #505, opened 2019-09-11
move reshape from keras bridge to plaidml2 ops

1sanz:


plaidml2/op/lib/ops.cc
```diff
@@ -1957,17 +1957,49 @@ Value reshape(const Value& value) {
...
+        neg_idx = dim_counter;
+        dims.emplace_back(1);
+      }
+    } else if (dim.is_none()) {
+      dims.emplace_back(I_dims[dim_counter]);
     }
+    dim_counter++;
+  }
+
+  if (neg_idx > -1) {  // there was a -1 dimension which needs to be filled
```
tzerrell: This will not work correctly if any of the relevant `I_dims` are symbolic (which doesn't get tested by `backend_test` as it currently exists). In the long run we need to think about how to test this best and implement that. But for now, see if changing num to a `TensorDim` type still works for the things we are testing, and if so do that.

tzerrell: I just remembered that `opTest` has a parameter `input_shapes` used for this purpose. See `testConcatenateSymbolic` for an example of how to use it. We could add using it to test symbolic reshaping as a separate PR or as part of this PR

plaidml2/bridge/keras/backend_test.py
```diff
@@ -1222,6 +1222,21 @@ def testPool3D(self, b, x):
...
+        output = pkb.reshape(a, (2, 0, -1))
+        self.assertEqual(str(output), "reshape_2|fp32(2, 1, 30)")
+        output = pkb.reshape(a, (0, 0, 0))
+        self.assertEqual(str(output), "reshape_3|fp32(1, 1, 60)")
+        #throw runtime exceptions
+        with self.assertRaises(plaidml2.Error) as cm:
+            output = pkb.reshape(a, (-1, -1))
+        self.assertTrue("PlaidML reshape op - at most one dimension of size -1 may be provided" in
+                        str(cm.exception))
+        a = pkb.variable(m(1, 1, 60))
```
tzerrell: This final assignment of `a` is probably unnecessary

plaidml2/op/lib/ops.cc
```diff
@@ -1957,17 +1957,49 @@ Value reshape(const Value& value) {
...
   for (auto dim : args[1].as_tuple()) {
     if (dim.is_int()) {
       dims.emplace_back(dim.as_int());
     } else if (dim.is_dim()) {
       dims.emplace_back(dim.as_dim());
     } else if (dim.is_str()) {
       // TODO: handle special cases
-      dims.emplace_back(0);
+      // TODO: figure out why there are quotation marks around fill and match
+      if (strcmp("\"match\"", dim.str().c_str()) == 0) {
```
flaub: We should never use `strcmp`, it's inherently unsafe (and might be banned from automated static analysis tools).

PR #507, opened 2019-09-12
 cpu tests for ci

drgone:


ci/plan.yml
```diff
@@ -111,6 +112,8 @@ VARIANTS:
       PLAIDML_EXPERIMENTAL: 1
       PLAIDML_DEVICE: llvm_cpu.0
       PLAIDML_TARGET: llvm_cpu
+      STRIPE_JIT: 1
```
flaub: We only want to set `STRIPE_JIT` when targeting the `llvm_cpu.0` device

drgone: I believe it is STRIPE_JIT=1 for llvm_cpu, not opencl_cpu

PR #517, opened 2019-09-13
Add as_float intrinsic in the cpu jit.

LLITCHEV:


tile/targets/cpu/compiler.cc
```diff
@@ -1187,6 +1188,14 @@ void Compiler::BitLeft(const stripe::Intrinsic& stmt) {
   OutputType(ret, stmt);
 }
 
+void Compiler::AsFloat(const stripe::Intrinsic& stmt) {
+  assert(1 == stmt.inputs.size());
+  Scalar ret = Cast(scalars_[stmt.inputs[0]], DataType::FLOAT32);
```
flaub: The 2nd argument to `as_float` is the bitwidth (so not always FLOAT32)

PR #518, opened 2019-09-13
Prepare for lowering Tile dialect -> Stripe dialect -> Stripe

flaub:


flaub:
@jbruestle What do you think?

PR #520, opened 2019-09-16
Add all the as_* intrinsics functions needed at runtime.

LLITCHEV:


tile/targets/cpu/compiler.cc
```diff
@@ -1189,8 +1191,76 @@ void Compiler::BitLeft(const stripe::Intrinsic& stmt) {
 }
 
 void Compiler::AsFloat(const stripe::Intrinsic& stmt) {
-  assert(1 == stmt.inputs.size());
-  Scalar ret = Cast(scalars_[stmt.inputs[0]], DataType::FLOAT32);
+  assert(2 == stmt.inputs.size());
+  Scalar inStmt2 = scalars_[stmt.inputs[1]];
+  int bits = ((llvm::ConstantInt*)inStmt2.value)->getValue().getLimitedValue();
```
mars1026: This cast makes me a little nervous - perhaps we could use something like `llvm::cast<llvm::ConstantInt>(...)` here?

flaub: Or even `llvm::dyn_cast<>` and check for null result if you want to be even more paranoid (I don't know if it's possible to have something not be a `ConstantInt` in this context)

PR #522, opened 2019-09-16
add implementation for stripe special "shape" function

mars1026:


tile/targets/cpu/compiler.cc
```diff
@@ -1343,6 +1341,26 @@ void Compiler::PrngStep(const stripe::Special& prng_step) {
...
+  Buffer data = buffers_[shape.inputs[0]];
+  size_t data_ndims = data.refinement->interior_shape.dims.size();
+  assert(1 == shape.outputs.size());
+  Buffer out = buffers_[shape.outputs[0]];
+  assert(1 == out.refinement->interior_shape.dims.size());
+  assert(data_ndims == out.refinement->interior_shape.elem_size());
+  uint64_t elem_bits = bit_width(out.refinement->interior_shape.type);
+  for (size_t i = 0; i < data_ndims; ++i) {
+    uint64_t dim_size = data.refinement->interior_shape.dims[i].size;
+    llvm::Value* val = builder_.getIntN(elem_bits, dim_size);
```
flaub: I don't think `elem_bits` is what we want here. Consider a `1000x1000` tensor, using an `i8` data type for the elements. In that case the returned datatype for the shape tensor would be `i8` but an `i8` isn't big enough to represent `1000`.

How about we always use an `i32`? Or perhaps there's a way to use a machine-word sized type?

flaub: Oh, I just realized `elem_bits` comes from the `out` not `data`.

LGTM

PR #523, opened 2019-09-17
Enable K.reshape using K.shape for dims

tzerrell:


tzerrell:
> I think this is fine for now, but perhaps we can revisit later once MLIR work is done.

Sounds good

PR #526, opened 2019-09-17
Extend plaidbench batchnorm folding

tzerrell:
Expand the plaidbench network-level batchnorm folding so as to produce relevant perf metrics for EDSL-based PlaidML while waiting for the relevant constant folding to be fully implemented.

### DIFF
--- a/plaidbench/plaidbench/frontend_keras.py
+++ b/plaidbench/plaidbench/frontend_keras.py
@@ -79,12 +79,7 @@ class Model(core.Model):
 
     def fold_batch_norm(self, model, backend):
         import json
-        if backend == 'plaid_edsl':
-            import plaidml2.bridge.keras as K
-        elif backend == 'plaid':
-            import plaidml.keras.backend as K
-        else:
-            import keras.backend as K
+        import keras.backend as K
         from keras.models import model_from_json
 
         def make_mults(weights):
--- a/plaidbench/plaidbench/frontend_keras.py
+++ b/plaidbench/plaidbench/frontend_keras.py
@@ -68,17 +68,23 @@ def setup_imdb(train, epoch_size):
 
 class Model(core.Model):
 
-    def __init__(self, frontend, params):
+    def __init__(self, frontend, params, fold_bn_for_backend=False):
         learn_phase = params.learn_phase  # e.g. 0 for infer with fixed learning phase (but should come from param)
         if learn_phase is not None:
             from keras.backend import set_learning_phase
             set_learning_phase(learn_phase)
         self.frontend = frontend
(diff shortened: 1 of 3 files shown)


plaidbench/plaidbench/frontend_keras.py
```diff
@@ -68,17 +68,23 @@ def setup_imdb(train, epoch_size):
...
             from keras.backend import set_learning_phase
             set_learning_phase(learn_phase)
         self.frontend = frontend
         self.params = params
+        self.fold_bn_for_backend = fold_bn_for_backend
 
-    def fold_batch_norm(self, model):
+    def fold_batch_norm(self, model, backend):
         import json
-        import plaidml.keras.backend as K
```
flaub: AFAICT, we only ever use `K.get_value()` and `K.set_value()`, so why can't we always just use `import keras.backend as K`? The assumption is that `import keras.backend` will route to the appropriate backend.

plaidbench/plaidbench/frontend_keras.py
```diff
@@ -343,7 +349,11 @@ def __init__(self, backend, fp16, train):
     def model(self, params):
         if self.train:
             return TrainingModel(self, params)
-        return InferenceModel(self, params)
+        if os.getenv('USE_STRIPE', '0') == '1' or self.backend == 'plaid_edsl':
```
flaub: It seems like this should really be: 
```
enable_bn_folding = (os.getenv('USE_STRIPE', '0') == '1') or (self.backend == 'plaid_edsl')
return InferenceModel(self, params, enable_bn_folding=enable_bn_folding)
```

plaidbench/plaidbench/frontend_keras.py
tzerrell: `import keras.backend` does not actually route to the appropriate backend in the context of plaidbench in many circumstances due to technical debt lingering from when we had to monkey patch PlaidML into Keras. That said, it might be worthwhile to unwind that now rather than committing more hacky code like this

flaub: At least in the context that this matters in (i.e. only the `plaid` and `plaid-edsl` backends), it appears to work properly (at least in the testing I did locally).

tzerrell: hrm, it's been breaking my tests. I'm drilling into why though and it seems mostly unrelated, so I'm hoping I can clean this up.

tzerrell: A major part of the broken code was adjusted in #527, so if you're not super up-to-date that might explain the discrepancy in our results

PR #527, opened 2019-09-17
training fix for nightly

drgone:


ci/plan.yml
```diff
@@ -402,7 +405,6 @@ SUITES:
       # NOTE: this list must be in least to most dependent order.
       - plaidml-{version}-py2.py3-none-{arch}.whl
       - plaidml_keras-{version}-py2.py3-none-any.whl
-      - plaidbench-{version}-py2.py3-none-any.whl
```
flaub: I think we should use the wheel instead of trying to run out of the `plaidbench` directory (so that we get test coverage of the `plaidbench` wheel)

PR #532, opened 2019-09-18
Fix normalize_batch_in_training

dgkutnic:
For people who are creating `BatchNormalization` layers in Keras with an explicit axis defined, they are seeing output shape mismatches caused by `normalize_batch_in_training`.

This PR replicates the TF methodology of broadcasting/squeezing in those cases.


plaidml/keras/backend.py
```diff
@@ -1073,26 +1073,48 @@ def ndim(x):
...
 def normalize_batch_in_training(x, gamma, beta, reduction_axes, epsilon=1e-3):
     rank = x.shape.ndims
-    if reduction_axes == None:
-        axes = [rank - 1]
+    if rank == 4 and reduction_axes in [[0, 1, 2], [0, 2, 3]]:
+        # NOTE: Tensorflow's code is explicitly checking for reduction axes in
+        # the case where there are 4 dims in x. If the reduction axes are passed
+        # in explicitly when a layer is created, the broadcasting behavior
+        # differs.
+        target_shape = [1 if i in reduction_axes else i for i in range(4)]
```
tzerrell: The `else i` needs to be replaced with something that grabs the shape at axis `i`; similarly for the corresponding code int he PlaidML2 bridge

tzerrell:
Did you test the plaidml2 version of this code? I wouldn't expect your target shape formula to work there but perhaps I'm mistaken

dgkutnic:
No, I had tested the plaidml version and thought the plaidml2 version was syntactically similar.

I'll make that fix.

dgkutnic:
@flaub do you mean I should get rid of `_keras_shape` then? The lines where I'm getting `shape.dims` are in the original plaidml code, which wouldn't be using `bind_dims`. Should I be doing something else?

plaidml2/bridge/keras/__init__.py
```diff
@@ -1547,8 +1557,21 @@ def square(x):
...
+        axis = []
+        for s in range(len(x._keras_shape)):
+            if x._keras_shape[s] == 1:
+                axis.append(s)
+    if type(axis) is list:
+        x_squeezed = x
+        for i in range(len(axis)):
+            ax = axis[i] - i
+            x_squeezed = _KerasNode('squeeze', tensor=plaidml_op.squeeze(x_squeezed.tensor, ax))
+        return x_squeezed
```
tzerrell: For many cases this won't matter, but it would be nice to get multi-axis squeeze moved into the plaidml_op version of squeeze; applying it multiple times could lead to a performance hit. That could be a separate PR though.

plaidml2/bridge/keras/__init__.py
```diff
@@ -1062,14 +1062,23 @@ def not_equal(lhs, rhs):
 def normalize_batch_in_training(x, gamma, beta, reduction_axes, epsilon=1e-3):
     I = x.tensor
     ndims = I.shape.ndims
```
tzerrell: Since we're at the Keras level, I recommend replacing these two line with `ndims = ndim(x)`

plaidml2/bridge/keras/__init__.py
```diff
@@ -1547,8 +1557,21 @@ def square(x):
...
 
 @_log_call
-def squeeze(x, axis):
-    return _KerasNode('squeeze', tensor=plaidml_op.squeeze(x.tensor, axis))
+def squeeze(x, axis=None):
+    if axis is None:
+        # define axis
+        axis = []
+        for s in range(len(x._keras_shape)):
+            if x._keras_shape[s] == 1:
```
tzerrell: Replace with `int_shape(x)[s]` here. Since we're directly comparing to the python integer 1 this won't catch symbolic shapes that are 1 anyway, so just use `int_shape` to be clear about that.

plaidml2/bridge/keras/__init__.py
```diff
@@ -1062,14 +1062,23 @@ def not_equal(lhs, rhs):
 def normalize_batch_in_training(x, gamma, beta, reduction_axes, epsilon=1e-3):
     I = x.tensor
     ndims = I.shape.ndims
-    if reduction_axes == None:
-        raw_axes = [ndims - 1]
+    if ndims == 4 and reduction_axes in [[0, 1, 2], [0, 2, 3]]:
+        target_shape = [1 if i in reduction_axes else x._keras_shape[i] for i in range(4)]
```
tzerrell: I _think_ the `x._keras_shape` here should be replaced by `shape(x)`, but perhaps it should be replaced by `int_shape(x)` (and it's possible I'm overlooking something and we need a different approach from either of those suggestions). Do we test this code on tensors with symbolic dimensions? That would help clarify what we should do here.

plaidml2/bridge/keras/__init__.py
dgkutnic: I was planning on making that a separate PR -- this one's been gnarly enough as-is.

plaidml2/bridge/keras/__init__.py
```diff
@@ -1060,16 +1060,24 @@ def not_equal(lhs, rhs):
 
 @_log_call
 def normalize_batch_in_training(x, gamma, beta, reduction_axes, epsilon=1e-3):
-    I = x.tensor
-    ndims = I.shape.ndims
-    if reduction_axes == None:
-        raw_axes = [ndims - 1]
+    ndims = ndims(x)
+    if ndims == 4 and reduction_axes in [[0, 1, 2], [0, 2, 3]]:
+        target_shape = [1 if i in reduction_axes else shape(x) for i in range(4)]
```
tzerrell: This should probably be `shape(x)[i]`. Also, I think this calls `shape` for each iteration of the list comprehension loop. I think it would be better to precompute shape(x) before the list comprehension and then access different elements of the precomputed value during the list comprehension.

plaidml2/bridge/keras/__init__.py
```diff
@@ -1060,16 +1060,24 @@ def not_equal(lhs, rhs):
 
 @_log_call
 def normalize_batch_in_training(x, gamma, beta, reduction_axes, epsilon=1e-3):
-    I = x.tensor
-    ndims = I.shape.ndims
-    if reduction_axes == None:
-        raw_axes = [ndims - 1]
+    ndims = ndims(x)
```
tzerrell: Personally I slightly prefer simplicity of code to gaining a bit of perf here, but this could also be
```
shape = shape(x)
ndims = len(shape)
```
to save us one call through the FFI. @flaub thoughts?
(also, if we leave this as-is, I think the Keras function name is `ndim` not `ndims`)

plaidml2/bridge/keras/__init__.py
```diff
@@ -1547,8 +1556,22 @@ def square(x):
 
 
 @_log_call
-def squeeze(x, axis):
-    return _KerasNode('squeeze', tensor=plaidml_op.squeeze(x.tensor, axis))
+def squeeze(x, axis=None):
+    if axis is None:
+        # define axis
+        axis = []
+        x_shape = shape(x)
```
tzerrell: This one should be `int_shape` since a symbolic dimension won't `== 1` in python even if it evaluates to 1; using `int_shape` makes it clear that we're only squeezing along non-symbolic dimensions.

tile/lang/gen_stripe.cc
```diff
@@ -427,13 +427,15 @@ class StripeGenerator {
...
           for (int i = 0; i < shape.dims.size(); i++) {
+            // TODO: Confirm whether i < start case ever exists
             if (i < start) {
               access.emplace_back(Affine{});
             } else {
               const auto& dim = shape.dims[i];
               if (dim.size > 1) {
                 access.emplace_back(Affine{kernel->idxs[i + idx_offset].name});
               } else {
+                // TODO: confirm that this should be empty Affine rather than affine{0}
```
tzerrell: I just looked at this `TODO`, and `Affine{}` and `Affine{0}` should be effectively the same (they work slightly differently under the hood but both represent a constant `0` Affine). Sorry for the confusion in suggesting this might create a new monomial index.

PR #540, opened 2019-09-23
removing slow passes

1sanz:
- loc_program, loc_main and reorder_commented to check correctness


tile/targets/gpu/amd.jsonnet
```diff
@@ -31,24 +31,25 @@ local PARAMS = {
...
-              name: 'loc_main',
-              pass: {
-                '@type': 'type.vertex.ai/vertexai.tile.codegen.proto.LocateMemoryPass',
-                reqs: ['main'],
-                loc: { devs: [{ name: 'GLOBAL', units: [{ offset: 0 }] }] },
-              },
-            },
-
+            
+            // {
```
flaub: Let's remove these lines if we don't need them anymore (we can always recover them via git history)

tile/targets/gpu/cm.jsonnet
```diff
@@ -19,23 +19,24 @@ local PARAMS = {
           // Define the stripe passes
```
flaub: I'm thinking we should not touch the `cm` configuration until we have a way of running tests on the `cm` backend.

tile/targets/gpu/amd.jsonnet
1sanz: removed.

tile/targets/gpu/cm.jsonnet
1sanz: passes put back in

PR #544, opened 2019-09-24
Add the incorrect bit count to the error message.

LLITCHEV:
For bad bit count of as_XXXXX functions add the number of bits that is incorrect to the error message.

### DIFF
--- a/tile/targets/cpu/compiler.cc
+++ b/tile/targets/cpu/compiler.cc
@@ -1219,7 +1219,9 @@ void Compiler::AsFloat(const stripe::Intrinsic& stmt) {
       type = DataType::FLOAT64;
     default:
       // TODO: Add bfloat16 when added to Stripe.
-      throw std::runtime_error("Invalid bit count for as_float for CPU jit.");
+      std::ostringstream oss;
+      oss << "Invalid bit count for as_float for CPU jit - " << bits;
+      throw std::runtime_error(oss.str());
   }
 
   Scalar ret = Cast(scalars_[stmt.inputs[0]], type);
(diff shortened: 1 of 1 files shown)


tile/targets/cpu/compiler.cc
```diff
@@ -1220,7 +1220,10 @@ void Compiler::AsFloat(const stripe::Intrinsic& stmt) {
       break;
     default:
       // TODO: Add bfloat16 when added to Stripe.
-      throw std::runtime_error("Invalid bit count for as_float for CPU jit.");
+      char buff[100];
```
flaub: I think we should try to use `std::stringstream` or `boost::format` instead of `snprintf` due to security issues.

You can say: 
```
throw std::runtime_error(str(boost::format("Invalid bit count for as_float for CPU jit - %1%") % bits));
```

LLITCHEV: I thought about this and noticed the target/ cpu doesn't have dependency on the Boost framework. In fact, after a brief look it looks like the whole tile/targets backends don't have such dependency. I thought it is an overkill to introduce it just for error messaging.
Using stringstream in the past has uncovered that the that functionality is not the most performing, and the compiler throughput matters. In this case, though we are printing message before failing, so I changed it.
In general, the plain C snprintf is the fastest way to format non-complicated strings (per my experiments over the years).

flaub: `snprintf` is also hard to use correctly. Almost everyone that uses it forgets to say `snprintf(buf, sizeof(buf)-1, ...)` to leave enough room for the NULL at the end of the buffer (which leads to a nasty security vulnerability, and because this touches user controlled inputs (well, maybe not exactly this case, but it could lead to that), it's even worse).

In general, I think we should ban unsafe functions of this nature.
That said, iostreams is slow as you point out, and so is boost for that matter. My hope is to either use the `llvm::formatv`, or switch to having all of our code to use https://github.com/fmtlib/fmt (which has made it into C++20).

PR #547, opened 2019-09-24
Stripe fix aggregates

jbruestle:


pmlc/dialect/stripe/into_mlir.cc
```diff
@@ -279,14 +337,37 @@ static void BlockIntoMLIR(OpBuilder* builder, const SymbolTable& outer, const st
...
+        Value* from = AffineIntoMLIR(builder, locals, load_idx->from);
+        Type idx_base = eltwise::ScalarType::get(builder->getContext(), DataType::INT32);
+        Type idx_type = eltwise::GetTensorType(idx_base);
+        auto op = builder->create<LoadIndexOp>(unknownLoc, idx_type, from);
+        locals.scalars.emplace(load_idx->into, op);
+      } break;
       case stripe::StmtKind::Store: {
         const auto& store = stripe::Store::Downcast(stmt);
+        std::string agg_str = block.ref_by_into(store->into)->agg_op;
+        IVLOG(1, "STORE: agg_op = '" << agg_str << "'");
```
earhart: This should probably be a little less noisy.

tile/targets/gpu/amd.jsonnet
```diff
@@ -94,6 +94,17 @@ local PARAMS = {
               }
             },
 
+            // No-op MLIR pass
```
earhart: Maybe note that this is work-in-progress?

PR #556, opened 2019-09-27
Rename USE_STRIPE to PLAIDML_USE_STRIPE and enable it by default for Intel GPU targets

flaub:


plaidml/keras/backend_test.py
```diff
@@ -388,7 +388,8 @@ def a_testLearningPhase(self):
         ],
         atol=1.e-4,  # RNNs have limited precision on some devices
     )
-    @unittest.skipIf(os.environ.get("USE_STRIPE", "0") == "1", "Stripe does not work for RNNs")
+    @unittest.skipIf(
+        os.environ.get("PLAIDML_USE_STRIPE", "0") == "1", "Stripe does not work for RNNs")
```
tzerrell: Not a problem for this PR, but I thought this skip had been removed

flaub: This is the test that we had weird issues related to removing the slow locate passes. IIUC, we were going to punt on this and deal with it later, perhaps once the MLIR path is clear.

PR #557, opened 2019-09-27
move squeeze implementation to op lib

dgkutnic:
The `squeeze` operation as defined in numpy and tensorflow has two arguments: the first is a mandatory argument specifying which tensor to squeeze, and the second is an optional argument specifying which axis/axes to squeeze (defaulting to all axes with size=1 if none are specified).

The op lib originally had squeeze defined with the second argument being mandatory and only supporting one axis at a time. I had modified the keras version of squeeze to call squeeze multiple times whenever multiple axes existed.

This PR exists to squeeze multiple axes all at once, and check within the plaidml2 op itself for squeezable axes if the axes aren't specified.

Note: I did have to change one instance of `shape(x)` to `int_shape(x)` in `normalize_batch_in_training` to get plaidml2 backend tests to pass.


plaidml2/op/lib/ops.cc
```diff
@@ -2469,21 +2469,36 @@ Value squeeze(const Value& value) {
...
   std::vector<TensorIndex> O_idxs;
   I.bind_dims(I_dims);
+
+  // argument 1: axes to squeeze upon
+  std::vector<int64_t> raw_axes;
+  if (args[1].is_int()) {
+    raw_axes.push_back(args[1].as_int());
+  } else if (args[1].is_none()) {
+    for (uint64_t i = 0; i < ndims; ++i) {
+      if (I_dims[i].as_int() == 1) {
```
tzerrell: This will throw if any of the dims are symbolic; IMO we should instead skip symbolic dimensions. See edsl.h around line 100

plaidml2/bridge/keras/__init__.py
```diff
@@ -1060,7 +1060,7 @@ def not_equal(lhs, rhs):
 
 @_log_call
 def normalize_batch_in_training(x, gamma, beta, reduction_axes, epsilon=1e-3):
-    x_shape = shape(x)
+    x_shape = int_shape(x)
```
tzerrell: This will cause problems with symbolic dimensions; I'm also not sure if this PR was the source of the problem you saw, I'll investigate more

plaidml2/op/lib/ops.cc
tzerrell: Or alternatively don't have an "automatically find the size 1 dims" option; it's a bit weird in this context

dgkutnic: The "automatically find size 1 dims" feature exists everywhere, I think it's probably good to have.

plaidml2/op/lib/ops.cc
```diff
@@ -2469,21 +2469,37 @@ Value squeeze(const Value& value) {
...
   I.bind_dims(I_dims);
+
+  // argument 1: axes to squeeze upon
+  std::vector<int64_t> raw_axes;
+  if (args[1].is_int()) {
+    raw_axes.push_back(args[1].as_int());
+  } else if (args[1].is_none()) {
+    for (uint64_t i = 0; i < ndims; ++i) {
+      // this doesn't work for symbolic dimensions that are later binded to 1
+      if (I_dims[i].as_ptr() && I_dims[i].as_int() == 1) {
```
flaub: You should use `is_int()` before trying to do `as_int()`, if it's not an int, it will throw during `as_int()`. Also, shouldn't ever be a need to use `as_ptr()`, this is an internal method to get the raw pointer (and should only be used by ffi).

plaidml2/bridge/keras/__init__.py
```diff
@@ -1061,7 +1061,7 @@ def not_equal(lhs, rhs):
 @_log_call
 def normalize_batch_in_training(x, gamma, beta, reduction_axes, epsilon=1e-3):
     x_shape = shape(x)
-    ndims = len(x_shape)
+    ndims = x.tensor.shape.ndims
```
flaub: Just wondering what's going on here. It's unfortunate that this will cause another ffi round trip (on top of the one inside `shape(x)`).

plaidml2/op/lib/ops.cc
flaub: Also, probably nicer to say `are later bound` rather than `binded`

tzerrell: thanks @flaub that was my bad advice, I'd forgotten `is_int` was a thing

tzerrell:
@dgkutnic @flaub when it's convenient can you both review my revisions here? This retains the option to auto-squeeze non-symbolic 1s at the Keras level (only), and fixes the normalize batch in training tests. I also unwound some of the shape-specific logic in batch_normalize_in_training, which looks like it should be fine to do and the related tests still pass, but @dgkutnic I know you've been thinking about that part of the code more than me.

plaidml2/bridge/keras/__init__.py
```diff
@@ -1061,24 +1061,14 @@ def not_equal(lhs, rhs):
...
-        target_shape = [1 if i in reduction_axes else x_shape[i] for i in range(4)]
-        m = mean(x, axis=reduction_axes, keepdims=True)
-        m = reshape(m, target_shape)
-        v = var(x, axis=reduction_axes, keepdims=True)
-        v = reshape(v, target_shape)
-        beta = reshape(beta, target_shape)
-        gamma = reshape(gamma, target_shape)
+    ndims = ndim(x)
+    if reduction_axes == None:
+        raw_axes = [ndim - 1]
```
flaub: Should this be `[ndims - 1]`?

dgkutnic:
Thanks @tzerrell for taking over this!

From my end, the changes look good, but I'll let @flaub do the formal approval of the PR.

Issue #171, opened 2018-09-10
PlaidML extremely slow on bidirectional LSTM


wongcr:
plaidML is extremely slow on bidirectional LSTM: see Keras sample code https://raw.githubusercontent.com/keras-team/keras/master/examples/imdb_bidirectional_lstm.py

On a i7-7700HQ + Nvidia GTX1060 6Gb + CUDA8 + python 3.6

plaidML: 850s per epoch, 100% GPU load
vs
Tensorflow-gpu 1.7:  273s, 25% GPU load
vs
CNTK 2.2: 97s per epoch , 35% GPU load

(NOTE: I've observed that CNTK is faster than Tensorflow for RNN's)

gh-kb:
I also tried to train a simple LSTM network and the process was very slow on an AMD GPU and relatively more robust on the built in Intel GPU (considering its capabilities), but still slower than on the CPU. Win 10, Keras, i7 6600 (with HD graphics 530), AMD RX 570

plaidml - Intel HD graphics 530: 101 s / epoch 
plaidml - opencl_ellesmere.0 (AMD RX570): 73 s /epoch 
plaidml - opencl_cpu.0: 132 s / epoch
straight CPU: 14 s /epoch

I tried different batch sizes, but that did not change the speed much

xxshazxx:
Hi, I'm having the same issues on LSTM. Did you discover something more?
Straight CPU is performing much faster than plaidml GPU.
I've also tried Tensorflow for opencl 1.8, but the results are also very bad.

brianretford:
Currently we invoke a clumsy escape hatch to implement scatter, which causes bad LSTM training performance. I am curious what minibatch size you're using and how performance varies as you increase batch size

brianretford:
duplicate of #70

PR #558, opened 2019-09-27
implementation of "scatter" and "gather" special operations

mars1026:


Gerzer:
Does this pull request address #171 (specifically the poor performance of RNNs)? Someone mentioned in a reply to that issue that the slowness was being caused by inefficient implementations of the “scatter” and “gather” operations.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mars1026:
This is a CPU-only implementation of "scatter" and "gather" for our new, Stripe-based architecture. It is part of a broader effort toward better performance, but will not provide any immediate improvement for the issue you mention.

PR #561, opened 2019-09-27
Add initial implementation of AggregationBlockOutputInitializationPass

LLITCHEV:
This takes care of proper initialization of the agg_op outputs and new locals.

### DIFF
--- a/tile/codegen/aggrinit.cc
+++ b/tile/codegen/aggrinit.cc
@@ -1,0 +1,103 @@
+// Copyright 2018, Intel Corporation
+
+#include <vector>
+
+#include "tile/codegen/aggrinit.h"
+#include "tile/codegen/deps.h"
+#include "tile/stripe/stripe.h"
+
+namespace vertexai {
+namespace tile {
+namespace codegen {
+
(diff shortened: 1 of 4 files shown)


tile/codegen/aggrinit.cc
```diff
@@ -0,0 +1,155 @@
...
+namespace vertexai {
+namespace tile {
+namespace codegen {
+
+using namespace stripe;  // NOLINT
+using namespace math;    // NOLINT
+
+void AggregationBlockOutputInitialization(const stripe::Block* const block,
+                                          const AggregationBlockOutputInitializationPass* aggregationPass) {
+  if (block == nullptr) {
```
mars1026: this will obviously work, but is there a reason it is not an assert? is this an expected failure mode?

LLITCHEV: I always prefer to have a fail-fast code. Assert is something that runs on debug builds only and we didn't have them till recently in nightly builds (not sure they got added). It is easy to get a bad state without noticing with assert. I prefer hard failure - forces investigation and fixing it right away, irrelevant whether a debug or release builds are used.

PR #566, opened 2019-09-30
split AggInit into multiple op-specific specials

mars1026:


tile/codegen/aggrinit.cc
```diff
@@ -1,5 +1,7 @@
 // Copyright 2018, Intel Corporation
 
+#include <memory>
+#include <set>
 #include <vector>
```
LLITCHEV: Did we need to include these?

tile/codegen/aggrinit.cc
```diff
@@ -78,7 +80,7 @@ void AggregationBlockOutputInitializationPass::Apply(CompilerState* state) const
   for (const auto& toInit : this->state.blocksWithInits) {
     assert(toInit.blockToAddTo != nullptr);
     assert(toInit.refToInitialize != nullptr);
-    assert(toInit.initType == AggregationInitType::ADD || toInit.initType == AggregationInitType::MULL ||
+    assert(toInit.initType == AggregationInitType::ADD || toInit.initType == AggregationInitType::MUL ||
```
LLITCHEV: I think this change needs to change the definition of the AggregationInitType::MUL and I seem to miss that change.

tile/codegen/aggrinit.cc
mars1026: cpplint complained about their absence, due to use of set<> and make_shared<>.

tile/codegen/aggrinit.cc
LLITCHEV: Oh... that was the in assert and it is a typo. NM

PR #567, opened 2019-09-30
Enable transcoding

jbruestle:


flaub:
We should get this is once we drop the regression test that caused a crash?

PR #574, opened 2019-10-01
no need to use calloc for temp refinements as we now have explicit agg_init

mars1026:


flaub:
This seems to have failing jit tests. I'll approve once these seem to be worked out.

PR #575, opened 2019-10-01
don't explicitly close the tf_session as we are using a context manager

mars1026:
the backend_test currently produces lots of annoying, superfluous tensorflow exception warnings, since we are explicitly calling tf_session.close(), which is redundant after the context manager wrapping each test run has already closed the session


Antetokounpo:
Should we also get rid of the tensorflow and numpy deprecation warnings? They're quite annoying.

flaub:
BTW, we almost never need to use the 'Update Branch' button manually anymore thanks to mergify. But, for it to work, you have to get approval from all reviewers. Another trick is to remove reviewers if you're unsure if they will approve your PR in a timely manner.

PR #584, opened 2019-10-03
WIP: Contractions can now execute via MLIR path

flaub:


pmlc/dialect/stripe/from_mlir.cc
```diff
@@ -474,6 +474,16 @@ std::shared_ptr<stripe::Program> FromMLIR(mlir::ModuleOp module) {
   StripeBuilder builder(func);
   auto ret = std::make_shared<stripe::Program>();
   ret->entry = builder.getResult();
+  for (const auto& ref : ret->entry->SubBlock(0)->refs) {
+    if (IsReadDir(ref.dir)) {
+      IVLOG(1, "input_shape: " << ref.from << " = " << ref.interior_shape);
```
earhart: Maybe turn these down a bit?  (Do we have a guide somewhere for when to use which verbosity?  I've been kind of making it up as I go.)

pmlc/dialect/tile/program.h
```diff
@@ -2,16 +2,21 @@
...
+#include "tile/base/buffer.h"
+
 namespace pmlc {
 namespace dialect {
 namespace tile {
 
 struct TileProgram {
   mlir::OwningModuleRef module;
   mlir::BlockAndValueMapping mapper;
+  std::map<mlir::Value*, vertexai::tile::BufferPtr> ioMap;
```
earhart: Maybe use an `unordered_map` here?

tile/base/buffer.h
```diff
@@ -73,19 +76,19 @@ class Buffer {
...
   virtual ~Allocator() {}
-  virtual std::shared_ptr<Buffer> allocate(size_t size) = 0;
+  virtual BufferPtr allocate(size_t size) = 0;
 };
 
 // A mechanism used to modify / optimize constant buffers during compilation
 struct ConstBufferManager {
   std::shared_ptr<Allocator> allocator;
-  std::map<std::string, std::shared_ptr<Buffer>> buffers;
+  std::map<std::string, BufferPtr> buffers;
```
earhart: Could be an `unordered_map`

tools/conda_run/conda_unix.cc
```diff
@@ -40,6 +40,11 @@ int main(int argc, char* argv[]) {
     // Adjust environment variables to activate conda environment
     vertexai::env::Set("CONDA_DEFAULT_ENV", conda_env);
 
+    auto dyld_insert_libs = vertexai::env::Get("_DYLD_INSERT_LIBRARIES");
```
earhart: This could use a comment.

tile/lang/ast/ast.cc
```diff
@@ -601,11 +601,14 @@ class ExprOptimizer : public AstPass {
 };
 
 ProgramEvaluation Evaluate(const std::string& name, ProgramMutations mutations) {
+  std::set<Expr*> dups;
```
earhart: `unordered_set`?

pmlc/dialect/eltwise/ops.td
```diff
@@ -191,47 +191,47 @@ def AsBoolOp : EltwiseOp<"as_bool", [NoSideEffect]> {
...
+def CmpGtOp : CompareOp<"cmp_gt", [NoSideEffect]>;
+def CmpLeOp : CompareOp<"cmp_le", [NoSideEffect]>;
+def CmpLtOp : CompareOp<"cmp_lt", [NoSideEffect]>;
+def CmpNeOp : CompareOp<"cmp_ne", [Commutative, NoSideEffect]>;
+def CosHOp : UnaryFloatOp<"cosh", [NoSideEffect]>;
+def CosOp : UnaryFloatOp<"cos", [NoSideEffect]>;
+def DivOp : BinaryOp<"div", [NoSideEffect]>;
+def ExpOp : UnaryFloatOp<"exp", [NoSideEffect]>;
+def FloorOp : UnaryOp<"floor", [NoSideEffect]>;
 def IdentOp : UnaryOp<"ident">;
```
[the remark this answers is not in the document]
flaub: `ident` appears to be necessary to deal with cases where otherwise we could wind up with invalid stripe code. There are 2 cases I know about:

1) A user specifies an output multiple times, but each output gets bound to a different buffer. This means we must copy the result from the original output to any ancillary ones.
2) A user specifies an output which refers directly to an input. This could cause an empty body in the pathological case (where the only output specified is also an input). The `ident` prevents the entire body from collapsing.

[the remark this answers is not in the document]
flaub: We could rename this to: 
```
def CopyOp : UnaryOp<"copy">;
```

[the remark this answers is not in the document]
earhart: 👍

PR #585, opened 2019-10-03
Add block and alloc hardware identifiers to Stripe MLIR

earhart:


pmlc/dialect/stripe/dialect.cc
```diff
@@ -39,18 +41,19 @@ mlir::Type Dialect::parseTensor(llvm::StringRef tyData, mlir::Location loc) cons
     emitError(loc, "invalid tensor type, no ()'s on size spec");
     return Type();
   }
-  llvm::SmallVector<StringRef, 8> dims;
-  llvm::SmallVector<TensorDim, 8> odims;
+  auto dims = llvm::SmallVector<StringRef, 8>();
+  auto odims = llvm::SmallVector<TensorDim, 8>();
+  auto matches = llvm::SmallVector<StringRef, 4>();
```
flaub: Odd style, any particular reason for incurring an extra assignment?

pmlc/dialect/stripe/into_mlir.cc
```diff
@@ -32,18 +32,29 @@ static ScalarType DataTypeIntoMLIR(mlir::MLIRContext* ctx, DataType dtype) {  //
...
 
-static Type ShapeIntoTensorType(MLIRContext* ctx, const TensorShape& shape) {
+static Type ShapeIntoTensorType(MLIRContext* ctx, const TensorShape& shape, const stripe::Location& loc) {
   ScalarType dtype = DataTypeIntoMLIR(ctx, shape.type);
   llvm::SmallVector<TensorDim, 4> dims;
+  for (std::size_t dev_idx = 0; dev_idx < loc.devs.size(); ++dev_idx) {
+    const auto& dev = loc.devs.at(dev_idx);
+    for (std::size_t unit_idx = 0; unit_idx < dev.units.size(); ++unit_idx) {
+      dims.emplace_back(TensorDim{
+          0, 0, mlir::Identifier::get((boost::format("%s_%zu_%zu") % dev.name % dev_idx % unit_idx).str(), ctx)});
```
flaub: 2 thoughts:
1) move to use `llvm::formatv` in favor of `boost::format`
2) if staying with `boost::format`, it might be nicer to use `%1%_%2%_%3%` syntax (hmm that looks pretty ugly too).

pmlc/dialect/stripe/ops.td
```diff
@@ -158,3 +157,11 @@ def TerminateOp : StripeOp<"terminate", [Terminator]> {
   let printer = [{ *p << getOperationName(); }];
 }
 
+def ReturnOp : StripeOp<"return", [Terminator, HasParent<"ParallelForOp">]> {
+  let summary = "Indicates where a ParallelForOp should execute";
+
+  // 'from' represents the executor value being returned.  We're using a
+  // Variadic here in lieu of Optional; a Stripe Return op may only have zero
+  // or one argument/s.
+  let arguments = (ins Variadic<TensorRefType>:$from);
```
flaub: Can we pass in an `llvm::None` in the case of no executor being set? Variadic seems too loose.

pmlc/dialect/stripe/stripe_test.cc
```diff
@@ -89,13 +86,13 @@ TEST(Stripe, Transcode) {
...
   lbp_dev->add_units()->set_offset(1);
   codegen::LocateBlockPass{lbp}.Apply(&cstate);
-#endif
 
   IVLOG(2, "Original version:");
   IVLOG(2, *prog->entry);
 
   IVLOG(1, "Converting to MLIR");
   auto module = IntoMLIR(&context, *prog);
+  module->dump();
```
flaub: Please use IVLOG, there's a helper for this in `pmlc/dialect/eltwise/util.h`

pmlc/dialect/stripe/types.h
```diff
@@ -55,7 +55,7 @@ struct TensorDim {
   int64_t stride;
 
   // The hardware class identified by this dimension.
-  llvm::StringRef cls;
+  mlir::Identifier cls;
```
flaub: I can't default constructor TensorDim when this is an identifier. You can add a default ctor that sets this value to `"address"` or some other reasonable value.

pmlc/dialect/stripe/types.h
```diff
@@ -135,7 +135,7 @@ class TensorType : public Type::TypeBase<TensorType, Type, TensorTypeStorage> {
   Type getElementType() const { return getImpl()->elementType; }
 
   /// Return the rank.
-  int64_t getRank() const { return getImpl()->shape.size(); }
+  uint64_t getRank() const { return getImpl()->shape.size(); }
```
flaub: These should remain `int64_t` so that it's easier to convert to/from `RankedTensorType` (which is what the tile dialect uses).

pmlc/dialect/stripe/into_mlir.cc
```diff
@@ -280,22 +308,63 @@ static void BlockIntoMLIR(OpBuilder* builder, const SymbolTable& outer, const st
...
   }
 
+  // Process the block's execution location.
+  std::vector<mlir::Value*> executor;
+  if (!block.location.empty()) {
+    std::vector<TensorDim> dims;
+    for (std::size_t dev_idx = 0; dev_idx < block.location.devs.size(); ++dev_idx) {
+      const auto& dev = block.location.devs.at(dev_idx);
+      for (std::size_t unit_idx = 0; unit_idx < dev.units.size(); ++unit_idx) {
+        auto cls = (boost::format("%s_%zu_%zu") % dev.name % dev_idx % unit_idx).str();
```
flaub: This seems to be defined in multiple places, is there a way to have the format of a 'cls' be defined in one spot, like a shared function or something?

pmlc/dialect/stripe/ops.td
flaub: Also, does this replace `TerminateOp`?

pmlc/dialect/stripe/from_mlir.cc
```diff
@@ -65,6 +66,61 @@ class StripeBuilder {
...
+
+  std::vector<AffinePolynomial> access;
+
+  auto combIt = ret.access.begin();
+  auto matches = llvm::SmallVector<StringRef, 4>();
+  for (const auto& dim : ret.base_type.getShape()) {
+    if (combIt == ret.access.end()) {
+      break;  // Should never happen; this is just to be careful.
+    }
+    if (dim.cls == "address") {
```
flaub: `"address"` seems special, can we make a `const char* kDefaultClass = "address";` someplace?

pmlc/dialect/stripe/types.h
flaub: Oh, I see now why this isn't possible (or good), we need an MLIR context to create an identifier, and without it being an identifier, we have tons of copies. Oh, this is why I was thinking a `StringRef` would be safe, as long as the StringRef always came from an identifier.

pmlc/dialect/stripe/dialect.cc
earhart: Compilers are allowed to omit it (and IIRC the ones we use do); in C++17, they'll be required to.  https://en.cppreference.com/w/cpp/language/copy_elision

pmlc/dialect/stripe/stripe_test.cc
earhart: Fixed (this was just debugging code that I missed removing); thanks!

pmlc/dialect/stripe/types.h
earhart: I can't -- `Identifier::get()` requires a context.  :-/  Thoughts?  (Maybe this should go back to being an uninterned string?)

pmlc/dialect/stripe/ops.td
earhart: `llvm::None` doesn't work (`NoneType` can't be used for a `Value*`), and `nullptr` causes a segfault.  I got this style from another MLIR dialect (SPIR-V, I think?); I couldn't find a useful `Optional`.

I think we still want `TerminateOp`, at a minimum for constraints.

flaub: We could have 2 different ops? What if `TerminateOp` was the no-operand version, and `ReturnOp` was renamed `ExecuteOnOp` that takes one operand? Initially we start with `TerminateOp` from initial tile -> stripe lowering.

pmlc/dialect/stripe/types.h
flaub: It's OK as is I think (where the caller must construct a TensorDim with a valid identifier).

pmlc/dialect/stripe/into_mlir.cc
```diff
@@ -426,9 +500,20 @@ static mlir::FuncOp ProgramIntoMLIR(MLIRContext* ctx, const stripe::Block& block
...
+    // refinement locations shouldn't be dependent on any external indicies.
+    auto offsets = LocationIntoTensorOffsets(&builder, initial.idxs, ref.location, &any_non_zero_offsets);
+    if (any_non_zero_offsets) {
+      auto zero = builder.create<AffineConstOp>(loc, builder.getType<AffineType>(), builder.getI64IntegerAttr(0));
+      offsets.resize(offsets.size() + ref.interior_shape.dims.size(), zero);
+      auto refOp = builder.create<RefineOp>(loc, tensorRefOp.getType(), tensorRefOp, offsets);
+      initial.refs.emplace(ref.into(), refOp);
+    } else {
+      initial.refs.emplace(ref.into(), tensorRefOp);
+    }
```
flaub: Can we have `LocationIntoTensorOffsets` actually be `LocationIntoTensorOps` so that it can return a `tensor_ref` or a `refine` depending on any non zero offsets. I'm worried about splitting the logic like this.

earhart: That turns out to be tricky, since two of the places where it's used also need to add in the non-location offsets when they're creating the `RefineOp`.

pmlc/dialect/stripe/from_mlir.cc
earhart: Done

pmlc/dialect/stripe/into_mlir.cc
earhart: Done

pmlc/dialect/stripe/types.h
earhart: Done

pmlc/dialect/stripe/ops.td
earhart: 👍 Done

pmlc/dialect/stripe/into_mlir.cc
earhart: Done

PR #588, opened 2019-10-03
Fix type of contraction operations and assignments

LLITCHEV:
Fix type of contraction operations and assignments to be based on the inputs.

Changed the type of one_hot from BOOLEAN to FLOAT32 (after talking to Tim). The result of this operation is always used in float calculations and TF's type for the operation is float32 as well.

### DIFF
--- a/plaidml/keras/backend.py
+++ b/plaidml/keras/backend.py
@@ -1141,7 +1141,7 @@ class OneHot(ptile.Operation):
             }}""".format(idim=', '.join(['I{}'.format(k) for k in range(indices.shape.ndims)]),
                          iidx=', '.join(['i{}'.format(k) for k in range(indices.shape.ndims)]))
 
-        outshape = ptile.Shape(plaidml.DType.BOOLEAN,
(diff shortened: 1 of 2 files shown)


plaidml/keras/backend.py
```diff
@@ -1141,7 +1141,7 @@ def __init__(self, indices, num_classes):
             }}""".format(idim=', '.join(['I{}'.format(k) for k in range(indices.shape.ndims)]),
                          iidx=', '.join(['i{}'.format(k) for k in range(indices.shape.ndims)]))
 
-        outshape = ptile.Shape(plaidml.DType.BOOLEAN,
+        outshape = ptile.Shape(plaidml.DType.FLOAT32,
```
tzerrell: As a part of Keras, this might be better as
```
outshape = ptile.convert_np_dtype_to_pml(floatx())
```
This would make onehot return the user-specified default floating point type rather than fp32 regardless of what floating point type the user asked to be default.

LLITCHEV: Sure... makes perfect sense.

PR #589, opened 2019-10-04
Add plaidml2 backend_tests to CI

tzerrell:


ci/plan.yml
```diff
@@ -238,6 +238,34 @@ SUITES:
...
+      ci:
+        batch_sizes: [0]
+      plaidml:
+        batch_sizes: [0]
+      nightly:
+        batch_sizes: [0]
+    wheels:
+      # NOTE: this list must be in least to most dependent order.
+      - plaidml-{version}-py2.py3-none-{arch}.whl
+      - plaidml_keras-{version}-py2.py3-none-any.whl
```
flaub: I'm not sure we need this because 

1. this is only the wheel for legacy plaidml
2. not sure backend tests actually run from wheel installs...

If you remove this, does it work?

PR #595, opened 2019-10-08
Transfer learning demo overhaul

mwyi:


demos/TransferLearning/README.md
```diff
@@ -15,6 +15,12 @@ source .venv/bin/activate
 - matplotlib
 - jupyter
 
+For PlaidML devices, run `plaidml-setup` before starting this demo to select the Plaidml backend device.
+
+Once the interface is presented, setup the model and network along with epoch and batch size values then hit the Train button to start. The demo will initialize test images and then start with a prediction with the model selected using pretrained `imagenet` weights. The demo will then automaticall proceed into the training phase followed by another predition phase based on the trained model.
```
dgkutnic: A couple of typos here:
> The demo will then automaticall proceed into the training phase followed by another predition phase based on the trained model.

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -411,86 +220,86 @@ def compile_model(self, modelName, fine=0):
...
     nw = 'ResNet50'
     if args.network_type == 'ResNet50' or args.network_type == 'MobileNet V2':
         nw = args.network_type
-    be = "PLAIDML"
-    if args.backend != "" and args.backend in ngraph_bridge.list_backends():
+    be = "CPU"
+    if args.backend == 'CPU' or args.backend == 'TF' or args.backend == 'PLAIDML':
         be = args.backend
-    Demo(args.gui, args.training, nw, be, args.quiet)
+    if args.epochs:
```
dgkutnic: You probably want to check that this value is a positive number

demos/TransferLearning/README.md
```diff
@@ -15,6 +15,12 @@ source .venv/bin/activate
...
+
+The demo will present 9 randomly selected test images where the initial prediction incorrectly guessed and then the same 9 image predictions after the model has been trained.
```
dgkutnic: How does this sound?
>Initially, the demo will select 9 random images from the test set and perform inference on them using the pre-trained model. These initial guesses are often incorrect. After training, the demo will perform inference on the same  9 images using the fully trained model. This is shown to prove that the training resulted in improved prediction accuracy.

demos/TransferLearning/TransferLearningDemo.py
dgkutnic: While you're at it, you may want to check that `args.batch_size` and `args.workers` are positive numbers as well. I don't think any of this code will play nice if someone passes in a negative value to these args.

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -411,86 +220,86 @@ def compile_model(self, modelName, fine=0):
...
         '--network_type',
         help='selects the network used for training/classification [ResNet50]/MobileNet V2')
     parser.add_argument(
-        '--backend',
-        help=
-        'selects the backend used for training/classification (run ngraph_bridge.list_backends() for full list)'
-    )
+        '--backend', help='selects the backend used for training/classification [CPU]/PLAIDML/TF]')
     parser.add_argument('--quiet', help='disables most logging', action='store_false')
+    parser.add_argument('--epochs', help='number of epochs to train')
```
flaub: If you use `type=int` then this argument will automatically be an `int` by the time you go to use it. Plus, if a user enters something that can't be parsed as an `int`, you'll get an error.

flaub: You can also set a default value with `default=5` here.

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -411,86 +220,86 @@ def compile_model(self, modelName, fine=0):
...
-    be = "PLAIDML"
-    if args.backend != "" and args.backend in ngraph_bridge.list_backends():
+    be = "CPU"
+    if args.backend == 'CPU' or args.backend == 'TF' or args.backend == 'PLAIDML':
         be = args.backend
-    Demo(args.gui, args.training, nw, be, args.quiet)
+    if args.epochs:
+        epochs = int(args.epochs)
+    else:
+        epochs = 5
```
flaub: With the suggested additional kwargs on the `parser.add_argument` line above, this code becomes redundant.

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -411,86 +220,86 @@ def compile_model(self, modelName, fine=0):
...
+    if args.batch_size:
+        batch_size = int(args.batch_size)
+    else:
+        batch_size = 16
+    if args.workers:
+        workers = int(args.workers)
+    else:
+        workers = 1
```
flaub: Ditto if you used the same technique here.

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -411,86 +220,86 @@ def compile_model(self, modelName, fine=0):
...
+                        action='store_true')
     parser.add_argument(
         '--network_type',
         help='selects the network used for training/classification [ResNet50]/MobileNet V2')
     parser.add_argument(
-        '--backend',
-        help=
-        'selects the backend used for training/classification (run ngraph_bridge.list_backends() for full list)'
-    )
+        '--backend', help='selects the backend used for training/classification [CPU]/PLAIDML/TF]')
```
flaub: The `choices` kwarg on `add_argument` is useful for restricting which values are allowed for this argument.

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -1,11 +1,10 @@
...
-
 import ngraph_bridge
-ngraph_bridge.enable()
 
+import argparse
+import sys
+import importlib
 import numpy as np
 import matplotlib.pyplot as plt
 import matplotlib.image as mpimg
```
flaub: VSCode's python plugin has a nice feature that sorts imports. It will put python modules first, then third party next.

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -101,89 +41,64 @@ class Demo:
...
+
+    IMAGE_SIZE = 224
     IMAGE_SHAPE = (IMAGE_SIZE, IMAGE_SIZE, 3)
 
-    batch_size = 32
+    ngraph_bridge = None
+
+    batch_size = 16
     epochs = 2
     fine_tune_at = 100
```
flaub: If these values are being set by the constructor, then no need to list them here. It's a bit confusing because it's hard to tell what value `self.epochs` will take on when I find lots of places in the code where `epochs = x` where `x` is different in almost every case.

demos/TransferLearning/TransferLearningGui.py
```diff
@@ -0,0 +1,219 @@
...
+import matplotlib.pyplot as plt
+import matplotlib.image as mpimg
+
+from IPython.display import display
+from IPython.display import clear_output
+
+ngraph_backends = ['PLAIDML', 'CPU', 'DISABLED']
+batch_size = 16
+epochs = 5
+backend = ngraph_backends[0]
```
flaub: These global vars make me nervous, are they in use? Do they get overriden?

demos/TransferLearning/TransferLearningGui.py
```diff
@@ -0,0 +1,219 @@
...
+progress_box = widgets.HBox([progress_bar, progress_text, epoch_text])
+
+gui = widgets.VBox(children=[tab, progress_box])
+
+display(gui)
+
+model_dropdown.observe(on_model_change)
+ngraph_dropdown.observe(on_ngraph_change)
+train_button.on_click(on_train_clicked)
+#self.shuffle_button.on_click(self.init_model)
```
flaub: All of this code gets run when? It's hard to tell because it will happen whenever it gets imported. It's better to put this all in a function that will be explicitly called when you want it to be called.

demos/TransferLearning/TransferLearningGui.py
```diff
@@ -0,0 +1,219 @@
...
+        progress_bar.value = batch
+        pass
+
+    def on_predict_end(self, batch, logs=None):
+        progress_text.value = "Classification Complete"
+        pass
+
+
+demo = TransferLearningDemo.Demo()
+progress_info = ProgressBar(demo)
```
flaub: Global vars are hard to maintain because they are hard to track the state of them. Are they really necessary? 

Also I don't see any `def main()` in this file, so I can only assume that these objects will be initialized whenever the module is imported. This might be confusing and lead to hard to track down initialization order bugs.

demos/TransferLearning/README.md
```diff
@@ -25,9 +31,9 @@ jupyter notebook TransferLearningDemo.ipynb
 
 ## Running the demo through the command line
 ```
-usage: TransferLearningDemo [-h] [--gui] [--training]
+usage: TransferLearningDemo [-h] [--gui] [--training]
```
flaub: Not sure I understand the new architecture, but wasn't the `--gui` argument dropped?

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -1,90 +1,26 @@
...
+import importlib
 import os
 import random
-from IPython.display import display
+import sys
+import warnings
 
+import ipywidgets as widgets
+import matplotlib.image as mpimg
+import matplotlib.pyplot as plt
```
flaub: IIUC, these can be dropped because they only exist in the gui module.

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -1,90 +1,26 @@
...
+import numpy as np
 import tensorflow as tf
+from IPython.display import clear_output, display
```
flaub: Not sure we need to import this here?

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -101,89 +37,53 @@ class Demo:
...
+
+    IMAGE_SIZE = 224
     IMAGE_SHAPE = (IMAGE_SIZE, IMAGE_SIZE, 3)
 
-    batch_size = 32
+    ngraph_bridge = None
+
+    batch_size = 16
     epochs = 2
     fine_tune_at = 100
```
flaub: I think these vars should be dropped because the `__init__` is setting these values.

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -263,143 +151,46 @@ def init_images(self):
...
-            self.epochs = self.epoch_slider.value
-            self.batch_size = self.batch_slider.value
-            model = self.model_dropdown.value
-            self.progress_text.value = ''
-        else:
-            self.epochs = epochs
-            self.batch_size = batch_size
+    def setup_ngraph_bridge(self, backend):
+        # Enviornment variables
+        os.environ['PLAIDML_USE_STRIPE'] = '1'
```
flaub: This probably won't have any effect unless it happens before `import ngraph_bridge` occurs.

demos/TransferLearning/TransferLearningGui.py
```diff
@@ -0,0 +1,193 @@
+import os
+import random
+import subprocess
```
flaub: This doesn't seem to be used?

demos/TransferLearning/TransferLearningDemo.py
```diff
@@ -101,89 +37,53 @@ class Demo:
...
+    batch_size = 16
     epochs = 2
     fine_tune_at = 100
 
-    ngraph_backends = ngraph_bridge.list_backends()
-    ngraph_backends.append('DISABLED')
-    ngraph_bridge.set_backend(ngraph_backends[0])
-
     base_model = None
     model = None
```
flaub: Likewise these should be dropped.

demos/TransferLearning/TransferLearningGui.py
mwyi: os used when pulling test images from directory, random used when gathering random 9 images to display. subprocess removed.

PR #601, opened 2019-10-09
[MLIR] Introduce stripe to affine boilerplate

dcaballe:
Boilerplate code, including pass, build changes and testing.


vendor/mlir/mlir.BUILD
```diff
@@ -665,3 +665,24 @@ cc_library(
...
+    deps = [
+        ":Analysis",
+        ":IR",
+        ":Parser",
+        ":Pass",
+        ":Support",
+        ":Transforms",
+        "@llvm//:support",
+    ],
+)
```
flaub: There's already an `OptMain` cc_library defined, is this different somehow?

pmlc/tools/pmlc-opt/BUILD
```diff
@@ -9,13 +9,7 @@ plaidml_cc_binary(
...
-        "//pmlc/dialect/stripe",
-        "//pmlc/dialect/tile",
-        "@mlir//:EDSC",
-        "@mlir//:OptMain",
-        "@mlir//:Parser",
-        "@mlir//:TestTransforms",
-        "@mlir//:Transforms",
+        "//pmlc/conversion/stripe_to_affine",
+        "@mlir//:MlirOptLib",
     ],
```
flaub: Even though `//pmlc/conversion/stripe_to_affine` also depends on other things (e.g. `//pmlc/dialect/eltwise`) it's still better to explicitly list them here because `pmlc-opt` independently depends on these things.

vendor/mlir/mlir.BUILD
flaub: If we switch to using `MlirOptLib` that's fine, but we'll need to remove the old `OptMain` if that's the case. Or we just stick with `OptMain`, doesn't matter much to me.

pmlc/conversion/stripe_to_affine/convert_stripe_to_affine.h
```diff
@@ -0,0 +1,24 @@
...
+template <typename T>
+class OpPassBase;
+using FunctionPassBase = OpPassBase<FuncOp>;
+class OwningRewritePatternList;
+
+/// Creates a pass to convert Stripe dialect to affine dialect.
+std::unique_ptr<FunctionPassBase> createConvertStripeToAffinePass();
+}  // namespace mlir
+
+namespace pmlc {
```
flaub: Let's try to keep the namespace structure in-sync with the filesystem structure.

so: 
```
namespace pmlc {
namespace conversion {
namespace stripe_to_affine {
```

dcaballe: Ok! For my understanding, do you strictly create a namespace per directory, even for directories with small files like this one?

flaub: That's been the standard so far. We might not need a whole new directory for `stripe_to_affine` but we can always change it in the future.

PR #605, opened 2019-10-10
Update to bazel 1.0.0

earhart:


plaidml2/exec/BUILD
```diff
@@ -21,6 +21,7 @@ plaidml_cc_library(
         "ffi.h",
     ],
     defines = ["PLAIDML_EXEC_DLL"],
+    alwayslink = 1,
```
flaub: Not sure this is needed; there aren't any static initializers in this lib.

earhart: Without it, `plaidml_exec_init()` isn't exported by the library.

earhart: (The plaidml2 tests were failing because they couldn't find `plaidml_init()`, and then because they couldn't find `plaidml_exec_init()`.  I'm not sure what changed in the way Bazel 1.0.0 is building libraries to cause this, but `alwayslink=1` fixed it.)

flaub: Huh, weird, ok.

PR #607, opened 2019-10-10
Fix testPow on CPU

LLITCHEV:
The issue was that calling intrinsic functions was assuming always one parameter. The pow C RT function takes 2 parameters. Added support to allow to calling intrinsic with two parameters.

Also, added "cpu_thread" tag to the XSMM blocks, so we can start making progress on the MC (multi-core) jit.


tile/targets/cpu/compiler.cc
```diff
@@ -1846,16 +1846,47 @@ void Compiler::OutputBool(llvm::Value* ret, const stripe::Intrinsic& intrinsic)
...
+      }
+      break;
+    case 2:
+      if (numParams != 2) {
+        throw std::runtime_error("CallIntrinsicFunction expects 2 parameters");
+      }
+      break;
+    default:
+      throw std::runtime_error("CallIntrinsicFunction expects 1 or 2 parameters");
+  }
```
flaub: Seems like this whole thing could be something like: 
```
if (stmt.inputs.size() != numParamsIn) {
  throw std::runtime_error(llvm::formatv("CallIntrinsicFunction expects {0} parameter(s), actual: {1}", numParamsIn, stmt.inputs.size());
}

File: tile/targets/cpu/compiler.cc [hunk 2]
```diff
@@ -1846,16 +1846,47 @@ void Compiler::OutputBool(llvm::Value* ret, const stripe::Intrinsic& intrinsic)
...
+
+  Scalar op1 = Cast(scalars_[stmt.inputs[0]], stmt.type);
+  Scalar op2;
```
flaub: It makes me nervous that there's an uninitialized var, seems like we could generically just cast all the params? like:

```
std::vector<llvm::Value*> argvals;
for (auto operand : stmt.inputs) {
  argvals.emplace_back(Cast(scalars_[operand], stmt.type);
}
```

tile/targets/cpu/compiler.cc
LLITCHEV: There is handling of 2 params only. In case numParamsIn and stmt.inputs.size() are 3 for example, the code will still not work properly.

tile/targets/cpu/compiler.cc
LLITCHEV: The types will work nicely in this case (always the same) - that is the nature of pow(). But there might be cases where special attention is needed for the types. I'd rather force a manual examination instead of automatic buggeneration, where figuring out what happens would be a mistery.

flaub:
FYI: here's a more complete list of ops that we'll need to be able to support:

https://github.com/plaidml/plaidml/blob/master/pmlc/dialect/eltwise/ops.td

LLITCHEV:
@flaub Thanks for the list! We do generate most of the operations with LLVM instructions and for some we define intrinsics where we call into the C RT. All of these, but pow are 1 param functions, For now pow is the only 2 param one. Thanks again!

PR #613, opened 2019-10-11
can we use the version of tensorflow built with dnnl?

mars1026:


flaub:
Does the `tensorflow` package from conda-forge include the GPU backend? The CPU backend? Both? I'm guessing the conda-forge one has MKLDNN builtin?

flaub:
I think we should pin `tensorflow` to a specific version.

PR #614, opened 2019-10-11
Add gradient override to EDSL and use for softmax

tzerrell:
This adds a GradientOverride to the EDSL that enables the use of a custom formula for the gradient of a subgraph. This PR also uses this for Softmax to provide a better precision & performance gradient. It could alternatively be used to change the gradient, e.g. for stopping or reversing the gradient (or more complex things).


plaidml2/edsl/ffi.cc
```diff
@@ -257,16 +257,20 @@ void plaidml_expr_bind_dims(  //
     plaidml_dim_expr** dims) {
   return ffi_wrap_void(err, [&] {
     IVLOG(3, "plaidml_expr_bind_dims> " << expr->expr->str());
+    char* use_mlir_env_var = getenv("PLAIDML_MLIR");
```
flaub: `if (vertexai::env::Get("PLAIDML_MLIR", "0") == "1") {}`

PR #619, opened 2019-10-14
Add keras-tf-dnnl-cpu

flaub:


ci/plan.yml
```diff
@@ -471,19 +490,24 @@ SUITES:
         expected: "CL_MEM_OBJECT_ALLOCATION_FAILURE"
         precision: low
         args: [plaidbench.py, inception_v3, --train, --print-stacktraces]
+        skip_platforms: [keras-tf-dnnl-cpu]
```
flaub: Why do we have skips for train? I don't think `keras-tf-dnnl-cpu` is even setup to run in the `train` pipeline

PR #626, opened 2019-10-15
Pretty print

jbruestle:


pmlc/dialect/eltwise/dialect.cc
```diff
@@ -37,6 +42,16 @@ struct OpAsmInterface : public mlir::OpAsmDialectInterface {
       }
     }
   }
+  void getTypeAliases(mlir::SmallVectorImpl<std::pair<Type, StringRef>>& aliases) const final {  // NOLINT
```
flaub: I think this should take a `mlir::ArrayRef` rather than passing around a `mlir::SmallVectorImpl`?

jbruestle: Actually, this is just me implementing a virtual that MLIR defines, so not my call, but the array is basically an output array (you push all the type aliases you would like to have), so an ArrayRef wouldn't work here.

PR #629, opened 2019-10-16
Add TBB library to the PlaidML to use in CPU Jit.

LLITCHEV:


flaub:
Hmm, I'm worried about how this will work on windows?

LLITCHEV:
@flaub Pretty much all the cases online that I found for building TBB with Bazel are using a Makefile genrule and for Windows a genrule that invokes MsBuild with the TBB provided SLN or VCPROJ file.
I do agree that using a native Bazel build is better (the Makefile and the MSBuild genrules do utilize the cache, but there could be better granularity if you depend on the sources and not all the files in the TBB package).
Made a change to build the CPP files only as a native Bazel cc_library.

LLITCHEV:
@flaub Yeah, I expect tweaks will be needed to link it (and then to maintain it when new versions are pulled in), but we'll cross that bridge when we come to it. I do run tests locally that build the TBB for now. Once the Jit starts using it, it will build properly.

PR #630, opened 2019-10-16
Override gradients of several functions

tzerrell:
Namely binary crossentropy and sigmoid (nicer, equivalent formulas) and also a scale_gradient (changing the gradient by a rescaling factor)


plaidml2/op/lib/ops.cc
```diff
@@ -2034,15 +2049,41 @@ Value reshape(const Value& value) {
...
   IVLOG(1, "sigmoid");
   auto args = value.as_tuple();
   if (args.size() != 1) {
     throw std::runtime_error("sigmoid expects 1 argument");
   }
-  auto I = args[0].as_tensor();
+  auto I = ident(args[0].as_tensor());  // Copy for safe gradient override
   auto O = 1.0 / (1.0 + exp(-I));
-  return Value{O};
+  TensorDeriv deriv = [](const Tensor& Y, const Tensor& DY, const std::vector<Tensor>& X) {  //
```
flaub: So since we implemented `sigmoid` in the op lib, we can probably remove this: https://github.com/plaidml/plaidml/blob/master/plaidml2/edsl/derivs.cc#L149

PR #635, opened 2019-10-19
Integrate lit testing

flaub:
Build rule is based on one from tensorflow.

The stripe_affine_roundtrip tests are failing and so are marked with `default_tags = ["manual"]`. Remove this once tests are passing again.


pmlc/pmlc-opt.cc
```diff
@@ -11,6 +11,9 @@
 #include "llvm/Support/SourceMgr.h"
 #include "llvm/Support/ToolOutputFile.h"
```
dcaballe: We may need to add more tools, like a cpu runner so maybe we should keep pmlc-opt at least under tools dir. Was the  tools/pmcl-opt/ problematic?

flaub:
You can run that test with:
```
bazelisk run //pmlc/conversion/tests/stripe_affine_roundtrip:core_ops.mlir.test
```

pmlc/pmlc-opt.cc
flaub: I was following tensorflow's lead, that's basically how their directory structure looks.

There's actually a CPU runner built into MLIR that we could possibly use, provided we lower into LLVM dialect.

dcaballe: Yeah, maybe there is something we can reuse but we would need our own if we want the input to start from Stripe and apply transformations that are not in MLIR core.

PR #638, opened 2019-10-21
Enable bazel builtin platform configs

earhart:


earhart:
+1 -- we need the build machines and everyone's local build configs to stop using the build configs before we can remove them, but it'd definitely be nice to do so.

PR #640, opened 2019-10-21
Add pmlc-translate and more lit tests

flaub:


pmlc/dialect/tile/tests/legalize-to-stripe.mlir
```diff
@@ -0,0 +1,145 @@
...
+  %arg0: tensor<10x20x!eltwise.fp32>, 
+  %arg1: tensor<10x20x!eltwise.fp32>
+) -> tensor<10x20x!eltwise.fp32> {
+  %0 = "eltwise.add"(%arg1, %arg0) {type = !eltwise.fp32} : (
+    tensor<10x20x!eltwise.fp32>, 
+    tensor<10x20x!eltwise.fp32>
+  ) -> tensor<10x20x!eltwise.fp32>
+  return %0 : tensor<10x20x!eltwise.fp32>
+}
+
```
dcaballe: Same comment as before. Please, try to make these tests more resilient to changes not related with the goal of this tests.

You can also use the `--split-input-file` flag with the separator `// -----` to separate multiple tests within the same file: https://github.com/tensorflow/mlir/blob/master/test/Transforms/loop-fusion.mlir#L11

pmlc/pmlc-translate.cc
```diff
@@ -0,0 +1,79 @@
...
+  auto input = openInputFile(inputFilename, &error_message);
+  if (!input) {
+    errs() << error_message << "\n";
+    return 1;
+  }
+
+  // MLIRContext context;
+  // if (failed((*requested_translation)(std::move(input), output->os(), &context))) {
+  //   return 1;
+  // }
```
dcaballe: leftover?

pmlc/dialect/stripe/tests/translate/double_dot.pbtxt
```diff
@@ -0,0 +1,355 @@
...
+      attrs { key: "main" value {} }
+    }
+    loc {}
+  }
+  attrs { key: "program" value {} }
+  attrs { key: "total_macs" value { ival: 18000 } }
+}
+
+# CHECK-LABEL: func @double_dot
+# CHECK-SAME: %arg0: !fp32_2 {stripe.layout = !stripe<"tensor !eltwise.fp32(addr[10:20], addr[20:1])">, stripe.name = "_X0"}
```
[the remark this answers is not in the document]
flaub: Hmm, the goal of the test is to cover all the nooks and crannies. This particular test I control the input. I agree in general we should try to make tests resilient. Over time I don't mind changing these tests to be 'looser' but I'm not sure which parts to make loose yet.

pmlc/dialect/tile/tests/legalize-to-stripe.mlir
flaub: Yep, I know about the `--split-input-file`, thanks. However, this use case doesn't need it.

pmlc/pmlc-translate.cc
flaub: Will remove in followup.

pmlc/dialect/stripe/tests/translate/double_dot.pbtxt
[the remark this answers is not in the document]
dcaballe: We can revisit this in a follow-up PR. In general, using regex for variable names is highly recommended and matching only output variables that are captured in a FileCheck variable for the later check of users. If we want to check a lot of small details, I would recommend having multiple simpler tests. In general, we should be careful with CHECK-NEXT. Output IR + CHECK-NEXT all the way would lead to a very flaky test. CHECK-DAG or simply CHECK would be recommended unless we actually want to check that one instruction has to strictly be right after another one. If we want to match a def-use relationship of two ops, capturing the output variable name and their users is usually enough and more flexible. You can also ask for feedback in the LLVM/MLIR mailing list. I'm not sure if these guidelines are written somewhere.

PR #641, opened 2019-10-21
Fix the Metal compilation error on Catalina

shiyao00:
1) Walkaround about the limit (31) of the kernel buffer arguments
2) Set the ZERO value as float type, consistent with the usual case

### DIFF
--- a/tile/codegen/fuse.cc
+++ b/tile/codegen/fuse.cc
@@ -533,6 +533,10 @@ void FusionInner(const AliasMap& scope, Block* block, TagFusionStrategy* strateg
       if (!block2) {
         break;
       }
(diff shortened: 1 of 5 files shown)


tile/codegen/fuse.cc
```diff
@@ -533,6 +533,10 @@ void FusionInner(const AliasMap& scope, Block* block, TagFusionStrategy* strateg
       if (!block2) {
         break;
       }
+      if (block1->refs.size() + block2->refs.size() - 1 > strategy->Options().max_refs()) {
+        // Too many refinemnts in a block for the particular platform
```
tzerrell: typo: "refinemnts" should be "refinements"

PR #644, opened 2019-10-22
Add affine canonicalizers

earhart:


earhart:
@flaub SGTM -- got any good docs for writing lit tests?

flaub:
https://llvm.org/docs/TestingGuide.html
https://llvm.org/docs/CommandGuide/lit.html
https://llvm.org/docs/CommandGuide/FileCheck.html

flaub:
https://github.com/tensorflow/mlir/blob/master/g3doc/TestingGuide.md

pmlc/dialect/stripe/canon.cc
```diff
@@ -0,0 +1,164 @@
...
+
+void AffineMulOp::getCanonicalizationPatterns(OwningRewritePatternList& results, MLIRContext* context) {
+  results.insert<SimplifyMulConstOne>(context);
+  results.insert<SimplifyMulConstValue>(context);
+  results.insert<SimplifyMulConstZero>(context);
+}
+
+void RefineOp::getCanonicalizationPatterns(OwningRewritePatternList& results, MLIRContext* context) {
+  results.insert<SimplifyNopRefines>(context);
+}
```
flaub: Actually, I think all of these are really folders instead of canonicalizers (folding happens inside the canonicalize pass tho).

flaub: See:
https://github.com/plaidml/plaidml/blob/master/pmlc/dialect/eltwise/ops.td#L220
https://github.com/plaidml/plaidml/blob/master/pmlc/dialect/eltwise/ops.cc#L237

earhart: Most of them are, definitely (although not refines, I think?).  I'd missed `hasFolder` in the mlir dialects -- rewriting now... :-)

flaub: And then for tests:
https://github.com/plaidml/plaidml/blob/master/pmlc/dialect/eltwise/tests/constant-fold.mlir

PR #645, opened 2019-10-22
Print MLIR aggregate with human-readable agg_op

tzerrell:
New format looks like
```
stripe.aggregate "max" %O_1 %s_O : !fp32_4
```


tzerrell:
Not sure why mergify isn't grabbing this, but it looks good to go so I'm landing manually.

PR #650, opened 2019-10-23
Add tests to stripe_test

dgkutnic:
This PR adds various tests to `stripe_test.cc`:
  * Softmax
  * Elementwise Ops
  * Layer Normalization
  * Batch Normalization


pmlc/dialect/stripe/stripe_test.cc
```diff
@@ -156,6 +156,68 @@ TEST_P(TranscodeTest, MaxPool2d) {
...
+TEST_P(TranscodeTest, MatMul) {
+  using plaidml::edsl::LogicalShape;
+  LogicalShape A(PLAIDML_DATA_FLOAT32, {16, 16});
+  LogicalShape B(PLAIDML_DATA_FLOAT32, {16, 16});
+  using vertexai::tile::lib::LoadMatMul;
+  auto ri = LoadMatMul("matmul", A, B);
+  RunTest(ri, GetParam());
+}
+
+// These two tests fail with invalid tensor dimensions if the location parameter for RunTest is set to 0.
```
tzerrell: We'll need to consult with @flaub but there's a good chance that's a real bug

pmlc/dialect/stripe/stripe_test.cc
```diff
@@ -156,6 +156,68 @@ TEST_P(TranscodeTest, MaxPool2d) {
...
+  using plaidml::edsl::LogicalShape;
+  LogicalShape A(PLAIDML_DATA_FLOAT32, {1, 64, 64, 32});
+  using vertexai::tile::lib::LoadLayerNorm4dAx2;
+  auto ri = LoadLayerNorm4dAx2("layer_norm", A);
+  RunTest(ri, 1);
+}
+
+TEST_P(TranscodeTest, BatchNormalization) {
+  using plaidml::edsl::LogicalShape;
+  LogicalShape A(PLAIDML_DATA_FLOAT32, {1, 64, 64, 32});
```
tzerrell: Batch normalization with batch size 1 is meaningless in a practical sense. The calculations this will demonstrate are probably perfectly reasonable though.

tile/lib/lib.cc
```diff
@@ -493,6 +493,22 @@ lang::RunInfo LoadSoftmax(const std::string& name,      //
...
+  return ((I - M) * G / sqrt(V + E)) + B;
+}
+
+lang::RunInfo LoadBatchNormalization(const std::string& name,      //
+                                     const LogicalShape& input) {  //
+  auto I = Placeholder(input);
+  auto M = Placeholder(input);
+  auto V = Placeholder(input);
+  auto G = Placeholder(input);
+  auto B = Placeholder(input);
```
tzerrell: I would expect these to have the batch dimension changed to 1 rather than left the same as on the input. In the place you're using this, it already is size 1, but that is itself a bit weird. It would probably also be good to have broadcast tested and this would be a natural place to do it. So perhaps drop the first dimension from `input` to create the shape for these Tensors and see if things still work?

pmlc/dialect/stripe/stripe_test.cc
dgkutnic: Fixed this -- batch size is now 16.

flaub:
Well I'm OK with landing as is. But @dgkutnic if you want to also try to fix the bug, that's fine with me too.

PR #657, opened 2019-10-24
Add constraints to tile dialect

flaub:


pmlc/dialect/tile/contraction.cc
```diff
@@ -604,14 +613,17 @@ bool Contraction::NeedReduce() const {
...
   // Reduce if needed
   if (NeedReduce()) {
     ReduceOutputPolynomials(constraints);
     constraints = GatherConstraints(shapes);
   }
   constraints.MergeParallelConstraints();
+  IVLOG(3, "Merged Parallel Constraints:" << to_string(constraints.constraints));
   // Defract if needed (defract does early return if not required)
   Defractionalize(constraints);
+  // IVLOG(3, "Defracted:\n" << to_string(defracted));
```
earhart: Optional: remove?  (Or uncomment?)

pmlc/dialect/tile/contraction.h
```diff
@@ -48,7 +48,7 @@ struct Constraints {
 };
 
 struct Contraction {
-  explicit Contraction(ContractionOp op);
+  explicit Contraction(ContractionOp op, llvm::ArrayRef<ConstraintOp> constraintOps);
```
earhart: We usually only use `explicit` with single-parameter constructors -- are you trying to disallow copy-list initialization, or is this left over from previously?

pmlc/dialect/tile/lowering.cc
```diff
@@ -625,7 +646,7 @@ OwningModuleRef LowerIntoStripe(ModuleOp workspace) {
     IVLOG(1, "LowerIntoStripe failed: " << mlir::debugString(*module->getOperation()));
     throw std::runtime_error("Lowering to stripe dialect failure");
   }
-  // IVLOG(1, "after:\n" << mlir::debugString(*module->getOperation()));
+  IVLOG(1, "after:\n" << mlir::debugString(*module->getOperation()));
```
earhart: This seems like a slightly verbose logging level.

pmlc/dialect/tile/contraction.h
flaub: Oops just leftover, thanks for catching this.

pmlc/dialect/tile/lowering.cc
flaub: Oops, meant to re-comment or remove it

PR #658, opened 2019-10-24
some more op lib wrappers

dgkutnic:
I added wrappers for the following operations:
- Prod
- ~Relu~ (to be included in a different PR)
- Repeat
- Slice
- Softmax
- SpatialPadding

### DIFF
--- a/plaidml2/op/op.h
+++ b/plaidml2/op/op.h
@@ -80,8 +80,9 @@ inline edsl::Tensor prod(const edsl::Tensor& I, const edsl::Value& axes = edsl::
   return details::op("prod", args).as_tensor();
 }
 
-inline edsl::Tensor relu(const edsl::Tensor& I, const edsl::Tensor& alpha, const edsl::Tensor& max_value,
-                         float threshold) {
+inline edsl::Tensor relu(const edsl::Tensor& I, const edsl::Tensor& alpha = edsl::Placeholder(PLAIDML_DATA_FLOAT32, {}),
+                         const edsl::Tensor& max_value = edsl::Placeholder(PLAIDML_DATA_FLOAT32, {}),
+                         float threshold = 0.0) {
   auto args = edsl::make_tuple(I, alpha, max_value, threshold);
   return details::op("relu", args).as_tensor();
 }
--- a/plaidml2/op/op.h
+++ b/plaidml2/op/op.h
@@ -75,6 +75,26 @@ inline edsl::Tensor mean(const edsl::Tensor& I, const edsl::Value& axes = edsl::
   return details::op("mean", args).as_tensor();
 }
 
+inline edsl::Tensor prod(const edsl::Tensor& I, const edsl::Value& axes = edsl::None(), bool keepdims = false) {
+  auto args = edsl::make_tuple(I, axes, keepdims);
+  return details::op("prod", args).as_tensor();
+}
+
+inline edsl::Tensor relu(
+    const edsl::Tensor& I, 
+    const edsl::Tensor& alpha, 
+    const edsl::Tensor& max_value, 
+    const float& threshold
+) {
+  auto args = edsl::make_tuple(I, alpha, max_value, threshold);
+  return details::op("relu", args).as_tensor();
+}
+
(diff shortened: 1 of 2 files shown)


flaub:
Hmm, do you have `clang-format` configured with your git pre-commit hooks? It looks like some of this code wasn't formatted with it.

plaidml2/op/op.h
```diff
@@ -75,6 +75,26 @@ inline edsl::Tensor mean(const edsl::Tensor& I, const edsl::Value& axes = edsl::
...
+inline edsl::Tensor prod(const edsl::Tensor& I, const edsl::Value& axes = edsl::None(), bool keepdims = false) {
+  auto args = edsl::make_tuple(I, axes, keepdims);
+  return details::op("prod", args).as_tensor();
+}
+
+inline edsl::Tensor relu(
+    const edsl::Tensor& I, 
+    const edsl::Tensor& alpha, 
+    const edsl::Tensor& max_value, 
+    const float& threshold
```
flaub: No need to have PODs be `const` references.

plaidml2/op/op.h
```diff
@@ -75,6 +75,26 @@ inline edsl::Tensor mean(const edsl::Tensor& I, const edsl::Value& axes = edsl::
...
+) {
+  auto args = edsl::make_tuple(I, alpha, max_value, threshold);
+  return details::op("relu", args).as_tensor();
+}
+
+inline edsl::Tensor repeat(const edsl::Tensor& I, const int& repeats, const int& raw_axis) {
```
flaub: Same here, use `int` instead of `const int&`

plaidml2/op/op.h
flaub: These arguments should all have default values (like `None` or `0.0`)

plaidml2/op/op.h
```diff
@@ -75,6 +75,23 @@ inline edsl::Tensor mean(const edsl::Tensor& I, const edsl::Value& axes = edsl::
   return details::op("mean", args).as_tensor();
 }
 
+inline edsl::Tensor prod(const edsl::Tensor& I, const edsl::Value& axes = edsl::None(), bool keepdims = false) {
+  auto args = edsl::make_tuple(I, axes, keepdims);
+  return details::op("prod", args).as_tensor();
+}
+
+inline edsl::Tensor relu(const edsl::Tensor& I, const edsl::Tensor& alpha = edsl::Placeholder(PLAIDML_DATA_FLOAT32, {}),
```
flaub: I think we want the default values for `alpha` and `max_value` to be `edsl::None()`

dgkutnic:
Thanks for the feedback! How does it look now?

plaidml2/op/op.h
dgkutnic: Since the op itself expects a tensor for `alpha` and `max_value`, we get these c++ errors when we try to set the default values to `edsl::None`:
```
error: no viable conversion from 'plaidml::edsl::Value' to 'const edsl::Tensor'
```

flaub: Change the datatype of the argument to be an `edsl::Value`. A `Value` is sort of like a variant, a way to allow multiple data types to be packed into a single type.

flaub: Actually.
This is a case where in C++ we probably should have multiple versions of this function. One which takes all parameters for `alpha` and `max_value`. Another that only takes `alpha`, another that only takes `max_value` and yet another that doesn't take any.

Then, each impl would pass `edsl::None` to the underlying `args` Value.

flaub: To be clear, here are the functions:

```
edsl::Tensor relu(const edsl::Tensor& I, float threshold = 0.0);
edsl::Tensor relu(const edsl::Tensor& I, const edsl::Tensor& alpha, float threshold = 0.0);
edsl::Tensor relu(const edsl::Tensor& I, const edsl::Tensor& alpha, const edsl::Tensor& max_value, , float threshold = 0.0);
```

dgkutnic: Didn't see your last two comments before making my last commit. I'll do this instead.

PR #661, opened 2019-10-24
Support unit-less MLIR devices

earhart:


pmlc/dialect/stripe/into_mlir.cc
```diff
@@ -275,12 +287,15 @@ static void BlockIntoMLIR(OpBuilder* builder, const SymbolTable& outer, const st
...
-        dims.emplace_back(TensorDim{0, 0, GetDevClass(builder->getContext(), dev, dev_idx, unit_idx)});
+      if (!dev.units.size()) {
+        dims.emplace_back(TensorDim{0, 0, GetDevClass(builder->getContext(), dev, dev_idx)});
+      } else {
+        for (size_t unit_idx = 0; unit_idx < dev.units.size(); ++unit_idx) {
+          // N.B. Locations in Stripe Classic logically reference an abstract executor space, in which size and
+          // stride are not well-specified, so we leave them undefined after translation to MLIR, and ignore
+          // them on translation back to Stripe Classic.
+          dims.emplace_back(TensorDim{0, 0, GetDevClass(builder->getContext(), dev, dev_idx, unit_idx)});
+        }
```
flaub: This whole thing seems very similar to `ShapeIntoTensorType`, is there any way to reuse that function or write a common function that both this and it uses?

earhart: Done

PR #666, opened 2019-10-25
oplib wrappers

1sanz:
binary_crossentropy
clip
cumprod
cumsum
dot

### DIFF
--- a/plaidml2/op/op.h
+++ b/plaidml2/op/op.h
@@ -90,11 +90,6 @@ inline edsl::Tensor flip(const edsl::Tensor& I, int axis) {
   return details::op("flip", args).as_tensor();
 }
 
-inline edsl::Tensor hard_sigmoid(const edsl::Tensor& I, float slope) {
-  auto args = edsl::make_tuple(I, slope);
-  return details::op("hard_sigmoid", args).as_tensor();
-}
-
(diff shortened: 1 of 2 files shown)


plaidml2/op/op.h
```diff
@@ -51,6 +50,31 @@ inline edsl::Tensor argmax(const edsl::Tensor& I, const edsl::Value& axes = edsl
   return details::op("argmax", args).as_tensor();
 }
 
+inline edsl::Tensor binary_crossentropy(const edsl::Tensor& I, const edsl::Tensor& O, const float epsilon) {
```
flaub: no need for `const` on the `float epsilon`

plaidml2/op/op.h
```diff
@@ -51,6 +50,31 @@ inline edsl::Tensor argmax(const edsl::Tensor& I, const edsl::Value& axes = edsl
...
+  auto args = edsl::make_tuple(I, O, epsilon);
+  return details::op("binary_crossentropy", args).as_tensor();
+}
+
+inline edsl::Tensor clip(const edsl::Tensor& I, const edsl::Tensor raw_min, const edsl::Tensor raw_max) {
```
flaub: should probably pass by ref instead of by value

PR #667, opened 2019-10-25
Fluid Interface PoC with relu

dgkutnic:


plaidml2/op/op.h
```diff
@@ -107,6 +107,47 @@ inline edsl::Tensor prod(const edsl::Tensor& I, const edsl::Value& axes = edsl::
...
+  edsl::Tensor I_;
+  edsl::Tensor alpha_ = edsl::None().as_tensor();
+  edsl::Tensor max_value_ = edsl::None().as_tensor();
+  float threshold_ = 0.0;
+
+ public:
+  explicit Relu(const edsl::Tensor& I) { I_ = I; }
+  void setAlpha(const edsl::Tensor& alpha) { alpha_ = alpha; }
+  void setMaxValue(const edsl::Tensor& max_value) { max_value_ = max_value; }
+  void setThreshold(float threshold) { threshold_ = threshold; }
```
flaub: let's leave the `set` prefix off, since this a builder API, seems natural that calling a method on a builder means you're setting a property of the builder.

plaidml2/op/op.h
```diff
@@ -107,6 +107,47 @@ inline edsl::Tensor prod(const edsl::Tensor& I, const edsl::Value& axes = edsl::
...
+  edsl::Tensor create() {
+    // actually make the relu from the op lib
+    auto args = edsl::make_tuple(I_, alpha_, max_value_, threshold_);
+    return details::op("relu", args).as_tensor();
+  }
+};
+
+class relu : protected Relu {
```
flaub: Curious why this is 2 classes?

plaidml2/op/op.h
```diff
@@ -107,6 +107,47 @@ inline edsl::Tensor prod(const edsl::Tensor& I, const edsl::Value& axes = edsl::
...
+    setMaxValue(max_value_);
+    return *this;
+  }
+
+  relu& threshold(float threshold_) {
+    setThreshold(threshold_);
+    return *this;
+  }
+
+  edsl::Tensor create() { return Relu::create(); }
```
flaub: Hmm, I wonder how we can make this implicit so that users don't have to explicitly call 'create'. Maybe we can call it `as_tensor()`?

I'll give this more thought...

plaidml2/op/op.h
```diff
@@ -107,6 +107,47 @@ inline edsl::Tensor prod(const edsl::Tensor& I, const edsl::Value& axes = edsl::
   return details::op("prod", args).as_tensor();
 }
 
+class Relu {
+ protected:
+  edsl::Tensor I_;
+  edsl::Tensor alpha_ = edsl::None().as_tensor();
```
flaub: I'm not sure this is actually a valid thing to do? Probably need to use a default `Tensor::Tensor()` constructor.

flaub: Actually, since it has a default constructor, no need to have an `= rhs`

flaub:
Oh, what do you think about having an `operator Tensor()` on all our builder objects?

dgkutnic:
Hi @flaub,

I've overhauled the builder API with your suggestions:

* It's now a single class that contains setters for each optional parameter
* Setters have been renamed
* The function call operator `()` is overloaded to call the op lib function with all the parameters

plaidml2/op/op.h
```diff
@@ -172,6 +172,38 @@ inline edsl::Tensor prod(const edsl::Tensor& I, const edsl::Value& axes = edsl::
...
+    max_value_ = max_value;
+    return *this;
+  }
+
+  relu threshold(float threshold) {
+    threshold_ = threshold;
+    return *this;
+  }
+
+  edsl::Tensor operator()() {
```
flaub: Actually I was thinking of a cast operator, not a call operator :)

plaidml2/op/op.h
```diff
@@ -172,6 +172,38 @@ inline edsl::Tensor prod(const edsl::Tensor& I, const edsl::Value& axes = edsl::
...
 
+class relu {
+ protected:
+  edsl::Tensor I_;
+  edsl::Tensor alpha_;
+  edsl::Tensor max_value_;
+  float threshold_ = 0.0;
+
+ public:
+  explicit relu(const edsl::Tensor& I) { I_ = I; }
```
flaub: We can avoid an extra copy by using an initializer like:
```
explicit relu(const edsl::Tensor& I) : I_(I) {}
```

plaidml2/op/op.h
dgkutnic: Fixed this

PR #668, opened 2019-10-25
Refactor affine instruction into one op, AffinePolyOp.

jbruestle:


jbruestle:
Actually that snuck in in the last commit, this diff is actually removing it...

PR #671, opened 2019-10-26
Add support for use_default in tile dialect

flaub:


pmlc/dialect/tile/lowering.cc
```diff
@@ -355,12 +425,64 @@ struct AffineDomainOpConversion : public LoweringBase {
...
+      auto range = ranges[idx->getArgNumber()];
+      if (range != outShape[i].size) {
+        // The index range does not cover the entire size of the tensor dimension
+        return true;
+      }
+    }
+
+    // Now we check if we have any constraints that are 'output only'
+    // Output only indexes actually reduce the range we write to, whereas constraints
+    // that use both input + output make writes but only process some of the input
```
jbruestle: Technically this isn't true (as I recently realized when looking into the old version of this code while trying to figure out how to do padding).  But apparently, it's a sufficient approximation is every case we ever hit, so good enough for now!

PR #673, opened 2019-10-26
Make plaidml2 op library init optional

earhart:
Thoughts on this approach?

### DIFF
--- a/plaidml2/op/ffi.cc
+++ b/plaidml2/op/ffi.cc
@@ -18,13 +18,7 @@ extern "C" {
 
 void plaidml_op_init(  //
     plaidml_error* err) {
-  static std::once_flag is_initialized;
-  ffi_wrap_void(err, [&] {
-    std::call_once(is_initialized, []() {
-      IVLOG(1, "plaidml_op_init");
-      lib::RegisterOps();
-    });
-  });
+  ffi_wrap_void(err, [&] { lib::OperationRegistry::Instance(); });
 }
 
 plaidml_expr* plaidml_op_make(  //
--- a/plaidml2/op/ffi.h
+++ b/plaidml2/op/ffi.h
@@ -20,6 +20,13 @@
 extern "C" {
 #endif  // __cplusplus
 
+// Ensures that the PlaidML operation library globals have been initialized.
+//
+// Using this API is optional: if it is not called before the first use of the operation library, the
+// operation library will be automatically initialized at that time.  This API is provided for callers that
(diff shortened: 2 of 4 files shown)


plaidml2/op/ffi.cc
```diff
@@ -18,13 +18,7 @@ extern "C" {
...
 void plaidml_op_init(  //
     plaidml_error* err) {
-  static std::once_flag is_initialized;
-  ffi_wrap_void(err, [&] {
-    std::call_once(is_initialized, []() {
-      IVLOG(1, "plaidml_op_init");
-      lib::RegisterOps();
-    });
-  });
+  ffi_wrap_void(err, [&] { lib::OperationRegistry::Instance(); });
```
flaub: This needs to be safe to call multiple times. I don't see any reason not to:
1. print that `plaidml_op_init` is being called
2. wrap this all in a `std::call_once`

I want to make this interface future proof by teaching users that in order to use the library they must call the init function before using any other functions. This allows us to throw more initialization behind this interface and ensures that the order of initialization is consistent. I don't mind if users want to wrap this initialization but I don't think we should expose on-demand initialization because it makes it harder to make changes to initialization in the future.

earhart: This is safe to call multiple times... it's using a Meyers singleton rather than a call_once, but that's guaranteed to be safe.

But the other issue is more complicated -- so it's not just that you want to enable users to call `plaidml_op_init` to ensure things are initialized, you want to require them to call it?

flaub: If we don't require it, then how can we make the interface future-proof?

earhart: What kinds of future-things are you thinking we might need to handle here, that this pattern wouldn’t support?

The only one I’ve got is something like a shared library load, where the caller might want to modify the search path. But then the caller can’t ensure that nothing else in the process hasn’t already initialized the globals; if we’re worried about that, maybe we should add a context object of some sort?

(Isn’t the Python code doing initialization-on-load?)

flaub: I think we should have an init function even if today it's a no-op.

earhart: (FWIW - please feel free to end the conversation by just saying you want it to work the way it currently works; I’d like to understand your model better to help guide future changes, but we can always chat about this more offline if this PR’s not going to go in.)

earhart: Okay, sounds like a plan. Thanks!

PR #679, opened 2019-10-29
Use MLIR path for plaidml2

flaub:
This is finally ready to land!


tile/codegen/test/tile.cc
```diff
@@ -346,19 +354,6 @@ TEST(Stencil, Pass) {
   EXPECT_THAT(data["C"], ContainerEq(kMatMulExpected));
 }
 
-TEST(Codegen, TilePassBroadcast) {
```
tzerrell: Are we testing broadcast elsewhere? Is broadcast temporarily broken?

tile/lib/lib.cc
```diff
@@ -279,80 +276,90 @@ lang::RunInfo LoadConstCalc(const std::string& name) {
...
   auto K = Placeholder(kernel, "K");
-  auto runinfo = Evaluate(name, {Convolution(I, K, output)});
-  runinfo.const_inputs = {"K"};
-  runinfo.input_buffers = {{"K", MakeBuffer(kernel)}};
-  return runinfo;
+  return Program(name, {Convolution(I, K, output)});
+  // auto runinfo = Evaluate(name, {Convolution(I, K, output)});
+  // runinfo.const_inputs = {"K"};
+  // runinfo.input_buffers = {{"K", MakeBuffer(kernel)}};
+  // return runinfo;
```
tzerrell: Are the various runinfo comments in this file reminders to add `const_inputs` and `input_buffers` functionality/testing?

tile/codegen/test/tile.cc
flaub: This was an odd test, it didn't seem to have any assertions.

Also, I do have some broadcast tests elsewhere.

tile/lib/lib.cc
flaub: I will remove these comments. The` runinfo` was mixing concerns; I have a new `Binder` object on the Execution API that should provide similar functionality. Marking constant buffers is still an open item however.

PR #681, opened 2019-10-29
WIP: Round-trip Stripe func ref attrs through transcode

earhart:


pmlc/dialect/stripe/into_mlir.cc
```diff
@@ -474,6 +474,10 @@ static mlir::FuncOp ProgramIntoMLIR(MLIRContext* ctx, const stripe::Block& block
     func.setArgAttr(argIndex, attrName, builder.getStringAttr(ref.into()));
     auto attrLayout = Dialect::getDialectAttrName("layout");
     func.setArgAttr(argIndex, attrLayout, builder.getTypeAttr(tensorTypes[argIndex]));
+    auto attrs = TagsToDict(&builder, ref);
+    if (attrs.size()) {
+      func.setArgAttr(argIndex, Dialect::getDialectAttrName(Dialect::getStripeAttrsName()), attrs);
```
flaub: If we're going to prefix the attr name with the dialect name, then I don't think we need to duplicate the prefix in the name. So instead of `stripe.stripe_attrs`, how about we go with just `stripe.attrs`

earhart: SGTM.  (FYI, I'm also fixing the lit test -- forgot to actually test //pmlc :-)

pmlc/dialect/tile/tests/translate-to-stripe.mlir
```diff
@@ -13,9 +13,9 @@ func @eltwise_add(
 
 // CHECK: 0:   #program
 // CHECK-NEXT: block []:1 ( // eltwise_add
-// CHECK-NEXT:     #user none new@0x00000000 _X0[0, 0] fp32:I(10, 20):(20, 1):800 B
-// CHECK-NEXT:     #user none new@0x00000000 _X1[0, 0] fp32:I(10, 20):(20, 1):800 B
-// CHECK-NEXT:     #user none new@0x00000000 _X2[0, 0] fp32:I(10, 20):(20, 1):800 B
+// CHECK-NEXT:     none new@0x00000000 _X0[0, 0] fp32:I(10, 20):(20, 1):800 B
+// CHECK-NEXT:     none new@0x00000000 _X1[0, 0] fp32:I(10, 20):(20, 1):800 B
+// CHECK-NEXT:     none new@0x00000000 _X2[0, 0] fp32:I(10, 20):(20, 1):800 B
```
flaub: This seems wrong, don't we want to require that the `#user` tag is set somehow?

earhart: Yes, but that's not what the current Tile lowerer does; IMHO, MLIR transcoding should respect what's there, and not unconditionally set `#user`.  (In Stripe Classic, this is being done at `gen_stripe.cc:141` and `gen_stripe.cc:80`.)

flaub: I'm saying can we change the test so that the input has something that causes the `#user` tag to be set. This will be a bug without more code...

earhart:
Holding off on this for a bit while other changes go in.

PR #682, opened 2019-10-29
Port event logging context to PlaidML2

dgkutnic:
I've plumbed event logging through `plaidml_buffer_alloc`, `plaidml_buffer_mmap_current`, `plaidml_buffer_mmap_discard`, `plaidml_view_writeback`, and `plaidml_run_executable`.

You can generate the eventlog for PlaidML2 using a bazelisk commmand like this:
```
PLAIDML_EVENTLOG_FILENAME=/Users/dgkutnic/plaidml/eventlog_edsl.gz bazelisk run //plaidbench:bin -- resnet50 -n64 --plaid-edsl --print-stacktraces
```


plaidml2/core/ffi.cc
```diff
@@ -51,6 +58,20 @@ void plaidml_init(plaidml_error* err) {
...
+        auto pconfig = vertexai::ParseConfig<google::protobuf::Any>(config);
+        auto eventlog = vertexai::AnyFactoryMap<vertexai::context::EventLog>::Instance()->MakeInstance(*ctx, pconfig);
+        ctx->set_eventlog(std::move(eventlog));
+        ctx->set_is_logging_events(true);
+      } else {
+        ctx->set_is_logging_events(false);
+        ctx->set_eventlog(std::shared_ptr<vertexai::context::EventLog>());
+      }
       IVLOG(1, "plaidml_init");
       Settings::Instance()->load();
```
flaub: Let's log the entry and then also initialize settings before dealing with the eventlog.

plaidml2/core/ffi.cc
```diff
@@ -222,8 +243,8 @@ plaidml_buffer* plaidml_buffer_alloc(  //
     const char* device_id,             //
     size_t size) {
   return ffi_wrap<plaidml_buffer*>(err, nullptr, [&] {
-    Context ctx;
-    auto buffer = GetPlatform()->MakeBuffer(ctx, device_id, size);
+    Context* ctx = GlobalContext::getContext();
```
flaub: Prefer to say `auto ctx = GlobalContext::getContext()`

plaidml2/core/internal.h
```diff
@@ -44,6 +44,17 @@ struct plaidml_view {
 namespace plaidml {
 namespace core {
 
+struct GlobalContext {
+  static pmlc::dialect::tile::TileBuilder* get() {
+    static thread_local pmlc::dialect::tile::TileBuilder builder;
+    return &builder;
+  }
+  static vertexai::context::Context* getContext() {
+    static thread_local vertexai::context::Context context;
```
flaub: As @earhart mentioned in our chat, we don't want to make this `thread_local`

plaidml2/core/internal.h
```diff
@@ -44,6 +44,17 @@ struct plaidml_view {
 namespace plaidml {
 namespace core {
 
+struct GlobalContext {
+  static pmlc::dialect::tile::TileBuilder* get() {
+    static thread_local pmlc::dialect::tile::TileBuilder builder;
+    return &builder;
+  }
+  static vertexai::context::Context* getContext() {
```
flaub: Add a newline before this one.

plaidml2/core/ffi.cc
```diff
@@ -51,6 +58,20 @@ void plaidml_init(plaidml_error* err) {
           el::Loggers::setVerboseLevel(level);
         }
       }
+      auto ctx = GlobalContext::getContext();
+      auto eventlog_str = vertexai::env::Get("PLAIDML_EVENTLOG_FILENAME");
+      if (ctx && eventlog_str.size()) {
```
flaub: Why do we check for `ctx` to be non-null?

plaidml2/core/ffi.cc
```diff
@@ -51,6 +58,20 @@ void plaidml_init(plaidml_error* err) {
...
+        IVLOG(1, "Logging events to " << eventlog_str);
+        auto config = "{\"@type\": \"type.vertex.ai/vertexai.eventing.file.proto.EventLog\", \"filename\": \"" +
+                      eventlog_str + "\"}";
```
flaub: Can we use `llvm::formatv` for this instead?

Also, it'd be nice to use a raw C++ string literal so we don't have to escape the quotes.

plaidml2/core/ffi.cc
```diff
@@ -51,6 +58,20 @@ void plaidml_init(plaidml_error* err) {
...
+        auto pconfig = vertexai::ParseConfig<google::protobuf::Any>(config);
+        auto eventlog = vertexai::AnyFactoryMap<vertexai::context::EventLog>::Instance()->MakeInstance(*ctx, pconfig);
+        ctx->set_eventlog(std::move(eventlog));
+        ctx->set_is_logging_events(true);
+      } else {
+        ctx->set_is_logging_events(false);
+        ctx->set_eventlog(std::shared_ptr<vertexai::context::EventLog>());
+      }
```
flaub: Actually, can we just construct a protobuf object directly without having to parse some config string?

plaidml2/core/ffi.cc
dgkutnic: That was a relic of the past code. I'll remove it.

flaub:
Also, we need a way to ensure that the eventlog is flushed when the process completes.

dgkutnic:
Where would we want to free the context? `plaidml_program_free`?

flaub:
I don't think we ever want to free it. But we do want to flush the eventlog on shutdown. We could use `atexit`: http://www.cplusplus.com/reference/cstdlib/atexit/

plaidml2/core/ffi.cc
```diff
@@ -58,22 +61,21 @@ void plaidml_init(plaidml_error* err) {
...
-        auto eventlog = vertexai::AnyFactoryMap<vertexai::context::EventLog>::Instance()->MakeInstance(*ctx, pconfig);
+        EventLog e;
+        e.set_filename(eventlog_str);
+        auto eventlog = EventLogFactory().MakeTypedInstance(*ctx, e);
         ctx->set_eventlog(std::move(eventlog));
         ctx->set_is_logging_events(true);
       } else {
         ctx->set_is_logging_events(false);
         ctx->set_eventlog(std::shared_ptr<vertexai::context::EventLog>());
       }
```
flaub: Don't think we need this...

plaidml2/core/ffi.cc
```diff
@@ -58,22 +61,21 @@ void plaidml_init(plaidml_error* err) {
...
       auto ctx = GlobalContext::getContext();
       auto eventlog_str = vertexai::env::Get("PLAIDML_EVENTLOG_FILENAME");
-      if (ctx && eventlog_str.size()) {
+      if (eventlog_str.size()) {
         IVLOG(1, "Logging events to " << eventlog_str);
-        auto config = "{\"@type\": \"type.vertex.ai/vertexai.eventing.file.proto.EventLog\", \"filename\": \"" +
-                      eventlog_str + "\"}";
-        auto pconfig = vertexai::ParseConfig<google::protobuf::Any>(config);
-        auto eventlog = vertexai::AnyFactoryMap<vertexai::context::EventLog>::Instance()->MakeInstance(*ctx, pconfig);
+        EventLog e;
```
flaub: Would prefer to limit single letter variables to `i`, `j`, and `k` for loops.

plaidml2/core/ffi.cc
```diff
@@ -58,22 +61,21 @@ void plaidml_init(plaidml_error* err) {
...
+        e.set_filename(eventlog_str);
+        auto eventlog = EventLogFactory().MakeTypedInstance(*ctx, e);
```
flaub: I think we can directly create an instance of the `vertexai::eventing::file::EventLog` without having to go thru the `EventLogFactory` business.

plaidml2/core/ffi.cc
dgkutnic: We can default to using `eventlog.gz` if `PLAIDML_EVENTLOG_FILENAME` isn't set.

flaub: No, we don't want to enable the eventlog if it's not set...

plaidml2/core/ffi.cc
```diff
@@ -62,6 +85,8 @@ void plaidml_shutdown(plaidml_error* err) {
   ffi_wrap_void(err, [&] {
     IVLOG(1, "plaidml_shutdown");
     GetPlatform().platform.reset(nullptr);
+    auto ctx = GlobalContext::getContext();
+    ctx->eventlog()->FlushAndClose();
```
flaub: Let's reset the eventlog to `nullptr` here so that internally it gets deleted (I think this is causing CI tests to crash)

flaub: Also thinking that we should explicitly use `atexit` in the C++ code so that C++ users will also benefit from the eventlog being flushed.

plaidml2/core/ffi.cc
```diff
@@ -67,26 +67,27 @@ void plaidml_init(plaidml_error* err) {
...
-        auto eventlog = EventLogFactory().MakeTypedInstance(*ctx, e);
+        vertexai::eventing::file::proto::EventLog e_config;
+        e_config.set_filename(eventlog_str);
+        auto eventlog = std::make_shared<vertexai::eventing::file::EventLog>(e_config);
         ctx->set_eventlog(std::move(eventlog));
         ctx->set_is_logging_events(true);
-      } else {
-        ctx->set_is_logging_events(false);
-        ctx->set_eventlog(std::shared_ptr<vertexai::context::EventLog>());
+        GetPlatform();
```
flaub: I think this should go outside the `if`

plaidml2/core/ffi.cc
```diff
@@ -67,26 +67,27 @@ void plaidml_init(plaidml_error* err) {
...
+  ctx->set_eventlog(nullptr);
+}
+
 void plaidml_shutdown(plaidml_error* err) {
   ffi_wrap_void(err, [&] {
     IVLOG(1, "plaidml_shutdown");
     GetPlatform().platform.reset(nullptr);
-    auto ctx = GlobalContext::getContext();
-    ctx->eventlog()->FlushAndClose();
+    std::atexit(plaidml_reset_eventlog);
```
flaub: Actually, we'll want to register the atexit call in `init`, so that it gets called on exit. We wouldn't want to register this on shutdown...

plaidml2/core/ffi.cc
```diff
@@ -2,11 +2,20 @@
 
 #include "plaidml2/core/ffi.h"
 
+#include <stdlib.h>
```
flaub: Let's use `<cstdlib>` instead (it's the same, but at least it's consistent with the other includes.

plaidml2/core/ffi.cc
```diff
@@ -53,15 +63,31 @@ void plaidml_init(plaidml_error* err) {
...
+        e_config.set_filename(eventlog_str);
+        auto eventlog = std::make_shared<vertexai::eventing::file::EventLog>(e_config);
+        ctx->set_eventlog(std::move(eventlog));
+        ctx->set_is_logging_events(true);
+      }
     });
   });
 }
 
+void plaidml_reset_eventlog(void) {
```
flaub: Also, let's throw this in the default namespace so that it's not visible to other translation units.

PR #692, opened 2019-10-31
[MLIR] PopulateTensorRefShape analysis

dcaballe:
This commit introduces a pass that populates 'tensor_ref' types with
shape information. This is a workaround to be able to convert
'tensor_ref' types to 'memref' types.


pmlc/dialect/stripe/populate_tensor_ref_shape_pass.cc
```diff
@@ -0,0 +1,90 @@
...
+namespace {
+
+/// This pass populates the optional shape information of 'tensor_ref' types. This information is found in the 'layout'
+/// attribute of 'tensor_ref' types that are used in BlockArgument values and Alloc operations. For each 'tensor_ref'
+/// Value, the algorithm proceed as follows:
+///   1. Traverse the use-def chain until a BlockArgument/Alloc is found and retrieve its 'layout' attribute.
+///   2. Replace Value's 'tensor_ref' type with a new 'tensor_ref' type that contains the shape information.
+// NOTE: This analysis is O(n^2) so we should revisit this approach/design if the layout information is needed in
+// more cases than the Stripe->Affine dialect conversion.
+class PopulateTensorRefShape : public mlir::FunctionPass<PopulateTensorRefShape> {
```
jbruestle: Wouldn't a simple pre-order walk of the program suffice?  Basically if we've 'updated' all Ops before a given op, and we've set the values on the block arguments, then each Alloc can just set it's layout directly, and each tensor op (Refine, Load, etc) can just copy it's layout (and it's whole type in fact) from it's values immediate definition (which by virtue of being earlier is certainly already updates).  This would be linear with program size.

dcaballe: Thanks, Jeremy. Yes, that's a reasonable next step for the future.

dcaballe:
Do not merge, please. I'm revisiting the Analysis/Pass design.

dcaballe:
Thanks, Frank. Ready.
The analysis pass will be enabled/tested in the next PR

pmlc/dialect/stripe/dialect.h
```diff
@@ -6,10 +6,16 @@
...
 
+namespace llvm {
+class Regex;
+}
+
 namespace pmlc {
 namespace dialect {
 namespace stripe {
 
+class TensorDim;
```
flaub: Needs to be `struct TensorDim` to fix compiler error on macOS/clang

pmlc/dialect/stripe/dialect.cc
```diff
@@ -81,46 +82,40 @@ std::string Dialect::getCanonicalOpName(llvm::StringRef name) {
 }
 
 mlir::Type Dialect::parseTensor(llvm::StringRef tyData, mlir::Location loc) const {
```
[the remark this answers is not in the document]
dcaballe: The last set of commits included a manual merge. Do you see something  that is off?

pmlc/dialect/stripe/dialect.h
```diff
@@ -27,7 +23,7 @@ class Dialect : public mlir::Dialect {
 
   mlir::Type parseTensor(llvm::StringRef tyData, mlir::Location loc) const;
   mlir::Type parseTensorRef(llvm::StringRef tyData, mlir::Location loc) const;
-  mlir::LogicalResult parseTensorSize(llvm::StringRef sizeSpec, mlir::Location loc,    // NOLINT
+  mlir::LogicalResult parseTensorSize(llvm::StringRef sizeSpec, mlir::Location loc,
                                       llvm::SmallVectorImpl<TensorDim>& odims) const;  // NOLINT
```
flaub: The `pmlc/CPPLINT.cfg` file should allow non-const runtime references so the NOLINT should be unnecessary now.

PR #693, opened 2019-10-31
Fix input/output binding issues

flaub:


flaub:
Yeah, I actually implemented the whole thing with positional arguments and got it all working. But then the more I looked at it, it seemed like it was pointless to be binding a buffer that was never going to be used because nothing did a 'read' from it. So simply ignoring it seems to be OK. If I run into more trouble later on I could always add back the positional arguments thing.

PR #696, opened 2019-11-01
Add td include guards

earhart:


earhart:
`mlir-tblgen` is smart enough to realize that it's already included a file.  Unfortunately, it emits an error when the file is included the second time, instead of just ignoring it.  (I have to imagine there's some good reason for this, but I don't know what it is.)

Really, the include guards within the file are pointless, except that they set a define to indicate that the file's been included, so that other things avoid including them.  That's why things that include `OpBase.td` are already checking `#ifndef OP_BASE` before they `include "miler/IR/OpBase.td"; this just extends that mechanism to our own `.td` files.

flaub:
Let's chat about this, I was pretty sure we could make it work without all this...

PR #697, opened 2019-11-01
Extend model-based BN folding

tzerrell:


tzerrell:
> I'm wondering if we need this same change in plaidml2?

This change affects plaidbench regardless of whether plaidml2 is being used

PR #704, opened 2019-11-07
UW example code for toroidal shell moment about z surface integral

1sanz:


plaidml2/bridge/keras/backend_test.py
```diff
@@ -525,7 +525,7 @@ def testAddElementsRepeated(self, b, x, y, z):
         return [x + y + z]
 
     @opTest([
-        [m(3, 3), 1.0],
+        [m(3, 3), 65555],
```
flaub: This probably doesn't belong in this commit

1sanz: sorry. Fixed.

flaub:
Could we add a unit test for this so that it gets validated by CI?

networks/scitile/uw_tests.py
```diff
@@ -12,7 +16,8 @@ def test_torodial_shell_integral(self):
         minval, maxval = -1.25, 1.25
         eps = 1.0e-8
         result = toroidal_shell_integral(n, minval, maxval, eps)
-        self.assertEqual(result, 3.9926786915581705)
+        npt.assert_allclose(result, 3.9926786915581705, rtol=DEFAULT_TOL, atol=DEFAULT_ATOL)
+        #self.assertEqual(result, 3.9926786915581705)
```
flaub: It's ok to drop this line I think

PR #707, opened 2019-11-08
Build TBB with native Bazel.

flaub:
Creating a PR so I can make comments

### DIFF
--- a/bzl/tbb.BUILD
+++ b/bzl/tbb.BUILD
@@ -1,67 +1,85 @@
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+
 package(default_visibility = ["//visibility:public"])
 
 exports_files(["LICENSE"])
 
+write_file(
+    name = "gen_version_file",
+    out = "version_string.ver",
+    content = [
+        "#define __TBB_VERSION_STRINGS(N) \"Empty\"",
+    ],
+)
+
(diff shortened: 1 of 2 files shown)


bzl/tbb.BUILD
```diff
@@ -1,67 +1,57 @@
...
+    content = [
+        "#define __TBB_VERSION_STRINGS(N) \"Empty\"",
+    ],
+)
+
+copy_file(
+    name = "copy_asm_file",
+    src = "src/tbb/intel64-masm/intel64_misc.asm",
+    out = "src/tbb/intel64-masm/intel64_misc.S",
+)
```
flaub: This probably won't work across platforms. That's because each compiler deals with assembly differently. Currently, we don't have a good way of compiling assembly that works for all of gcc, msvc, and clang.

bzl/tbb.BUILD
```diff
@@ -1,67 +1,57 @@
...
+    hdrs = [
+        ":gen_version_file",
+    ],
+    defines = [
+        "TBB_USE_THREADING_TOOLS",
+        "__TBB_DYNAMIC_LOAD_ENABLED=0",
+        "__TBB_SOURCE_DIRECTLY_INCLUDED=1",
+        "__TBB_x86_64=1",
+    ] + select({
+        "@bazel_tools//src/conditions:windows": [
```
flaub: This conditional is not compatible with our build system.

Use `@com_intel_plaidml//toolchain:windows_x86_64` instead

PR #715, opened 2019-11-09
Add MLIR based tests

flaub:
This is part 4 of a series of PRs

(This comes after #714)


plaidml2/op/op_test.cc
```diff
@@ -385,14 +745,16 @@ TEST(Op, ImageResize) {
   _X7[x0, -4 + j + 5*x1, -3 + i + 4*x2, x3 : 1, 1120, 896, 3] = +(I[x0, x1, x2, x3] * _X6[j, i]);
 }
 )"));
+#endif
+#ifdef PLAIDML_MLIR
+#endif
```
tzerrell: Should there be a test here?

PR #723, opened 2019-11-13
[MLIR] Stripe->Affine conversion of Function/Types/Load

dcaballe:
It introduces basic Stripe->Affine conversion for function signature,
some types and load operations, including the polynomials involved in
the address computation.


pmlc/conversion/stripe_to_affine/convert_stripe_to_affine.cc
```diff
@@ -90,24 +188,112 @@ PatternMatchResult ParallelForOpConverter::matchAndRewrite(ParallelForOp forOp,
...
+static void convertStripeAccessToAffine(const FlatTensorAccess& accessInfo, ConversionPatternRewriter& rewriter,
+                                        Value*& base, SmallVectorImpl<Value*>& indices, AffineMap& affineMap) {
+  SmallVector<AffineExpr, 4> resultExprs;
+  unsigned numTerms = 0;
+  // Process all the polynomials in the access. Each polynomial has a set of terms consisting of {index * constant
+  // coefficient}, and an independent constant term. We create an affine expression for each polynomial.
+  for (auto polynomial : accessInfo.access) {
+    // Create an affine expression with the independent constant term.
+    AffineExpr polyExpr = rewriter.getAffineConstantExpr(polynomial.constant);
+    unsigned dimPos = 0;
```
dcaballe: There is a bug here for the more general case but I'm fixing this in a later commit.

jbruestle: I don't see the bug, just curious what it is.

dcaballe: the dim position in affine map is "global", not local to the independent polynomial expressions so we don't have to initialize `dimPos` here every time. We should use `numTerms` instead.

PR #724, opened 2019-11-13
Initial pass at op lib dialect

dgkutnic:
Here's an initial op lib dialect PoC.

The main things in this PR are:
* `ops.td` -- a `td` file that defines two op lib ops: `abs` (really simple operation) and `conv` (much more complex operation).
* `OpLibWrapperGen.cpp` -- This takes the `ops.td` file and generates op lib wrappers for it. Running `bazelisk build //pmlc/dialect/op_lib:plaidml_op_lib_wrappers` will get you an output file located here: `bazel-bin/pmlc/dialect/op_lib/op_lib_wrappers.h.inc`
* various changes/additions to the bazel build files to make everything build properly

@flaub I'd like to discuss a few things before proceeding forward-- particularly as it relates to the typing system we want to use in the op dialect. You had mentioned that we wanted to represent required parameters as operands and optional parameters as attributes. However, it seems like we lose some level of control over using attributes (i.e. let's say I want an optional parameter such as dilation, but it _has_ to be a tuple of length 2. it's much easier to do this as an operand than as an attribute). I also wanted to get your input on whether we should create a custom type in MLIR that mimics `edsl::Value`. I have been trying to use standard types wherever possible. Anyway, we can discuss these and more design decisions when we're both in the office.


vendor/mlir/mlir.BUILD
```diff
@@ -39,7 +39,7 @@ cc_binary(
     srcs = glob([
         "tools/mlir-tblgen/*.cpp",
         "tools/mlir-tblgen/*.h",
-    ]),
+    ]) + ["@com_intel_plaidml//pmlc/dialect/op_lib:gen_op_lib_wrappers"],
```
flaub: We should make the `tool` attribute of the `mlir_tblgen` allow a user-specified label. We shouldn't be pulling in `com_intel_plaidml` from MLIR's perspective.

vendor/mlir/mlir.bzl
```diff
@@ -9,6 +9,7 @@ TBLGEN_ACTIONS = [
     "-gen-op-interface-defs",
     "-gen-reference-implementations",
     "-gen-rewriters",
+    "-gen-op-lib-wrappers",
```
flaub: We should make this extensible some how. Or just allow any flag to be passed.

pmlc/dialect/op_lib/ops.td
```diff
@@ -0,0 +1,79 @@
...
+  }];
+
+  let arguments = (
+      ins AnyTensor:$x,
+      I32ArrayAttr:$strides,
+      I32ArrayAttr:$dilations,
+      I32ArrayAttr:$data_dilations,
+      I32ArrayAttr:$filter_shape,
+      APIntAttr:$groups,
+      StrAttr:$autopad_mode,
```
flaub: Many of these string-based attributes can be replaced with proper enums now that we have tablegen.

pmlc/dialect/op_lib/ops.td
```diff
@@ -0,0 +1,79 @@
...
+include "pmlc/util/enums.td"
+include "pmlc/util/interfaces.td"
+
+//===----------------------------------------------------------------------===//
+// PML op dialect definition
+//===----------------------------------------------------------------------===//
+
+def PML_Op_Dialect : Dialect {
+   let name = "pml_op";
+   let cppNamespace = "pmlc::dialect::pml_op";
```
flaub: I think we should go with `pmlc::dialect::op`

dgkutnic:
Summary of where things are at:
* Improved build rules: I created a rule `op_lib_tblgen` that lives pretty much completely outside of `vendor/mlir`.  The details of the rule are in `pmlc/dialect/op_lib/op_lib.bzl`. Currently, actions have to be specified through that file, and the tool being used is a custom binary (`op-lib-tblgen`) that builds _only_ the TableGen passes that I built for the op lib dialect. Perhaps this can be improved upon further, but at least it's better than my initial approach.
* Improved typing: In `ops.td`, I've changed most of the `StrAttr` types into custom `I32EnumAttr` types. I also added a custom typing emitter that emits the enumeration types specified in the `td`.
* Refactored code: I tried (as much as I could) to separate the TableGen process into two steps:
    1. Read the dialect's info from LLVM `Records` into a `DialectInfo` struct (sub-structs `TypeInfo` and `OpInfo`). In time, we could perhaps transfer the concept of `DialectInfo` into something like protobuf or jsonnet. For now, though, it's just a bunch of structs.
    2. Emit from a `DialectInfo` struct to the targeted language (`cpp` is the one I've mainly focused on, I've built a skeleton for `python` as well). 

Known future work:
* Add lit tests for cpp generation
* Split `OpLibWrapperGen.cpp` (perhaps by language)
* Continue development of `python` emitter.

pmlc/dialect/op_lib/OpLibWrapperGen.cpp
```diff
@@ -24,28 +27,175 @@
...
+    type = namedResult.constraint.getDescription();
+  }
+  return type;
+}
+
+struct OpInfo {
+  StringRef name_;
+  StringRef returnType_;
+  std::map<StringRef, StringRef> attributes_;
+  std::map<StringRef, StringRef> operands_;
```
flaub: Use a `MapVector` so that you don't loose the order: https://llvm.org/docs/ProgrammersManual.html#llvm-adt-mapvector-h

pmlc/dialect/op_lib/OpLibWrapperGen.cpp
```diff
@@ -24,28 +27,175 @@
...
+  explicit TypeInfo(const EnumAttr& ea) {
+    name_ = ea.getEnumClassName();
+    returnType_ = ea.getUnderlyingType();
+    for (auto eacase : ea.getAllCases()) {
+      opts_.insert({eacase.getSymbol(), eacase.getValue()});
+    }
+  }
+};
+
+// TODO(dgkutnic): Make DialectInfo into a protobuf object.
```
flaub: I'd like to avoid using protobuf if we can; it's a big dependency and adds lots of additional complexity. Using C++ structs seems like a good way to represent the 'object model' since this model doesn't ever need to be exported.

pmlc/dialect/op_lib/OpLibWrapperGen.cpp
```diff
@@ -24,28 +27,175 @@
...
-static const char* TBLGEN_DEBUG = std::getenv("TBLGEN_DEBUG");
+namespace pmlc {
+namespace dialect {
+namespace op {
+
+using namespace pmlc::dialect::op;  // NOLINT [build/namespaces]
+
+namespace tblgen {
+
+static inline std::map<StringRef, StringRef> getAttributes(const Operator& op) {
```
flaub: We probably don't need to `inline` these (they aren't one liners and `inline` is optional anyways, the compiler doesn't really need it since it's just a hint).

PR #736, opened 2019-11-19
Parallelize the block execution on CPU

LLITCHEV:
Implements a TBB based parallel runtime for executing blocks on CPUs.


plaidml/BUILD
```diff
@@ -159,6 +159,8 @@ plaidml_cc_test(
         ":api",
         "//testing:matchers",
         "//testing:plaidml_config",
+        "@tbb",
+        "@xsmm",
     ],
```
flaub: Why are we adding these deps here? Seems like they should be transitively depended on. Basically, I don't see how the `uint8_test` is directly dependent on `@tbb` or `@xsmm`

plaidml/keras/backend.py
```diff
@@ -56,15 +56,15 @@ def _log_call(func):
...
-            logger.debug('{}({})'.format(func.__name__, ', '.join(arg_str_list)))
+        #if logger.isEnabledFor(logging.DEBUG):
+        #    sig = inspect.signature(func)
+        #    arg_str_list = list()
+        #    for i, arg in enumerate(args):
+        #        arg_str_list.append('{}: {}'.format(list(sig.parameters)[i][0], arg))
+        #    logger.debug(kwargs)  # TODO
+        #    for k, v in kwargs.items():
+        #        arg_str_list.append('{}: {}'.format(k, v))
+        #    logger.debug('{}({})'.format(func.__name__, ', '.join(arg_str_list)))
```
flaub: Either remove this code or modify it so that it can be toggled at runtime instead of commenting it out for a particular run. (And we shouldn't land this anyways)

plaidml/keras/backend_test.py
```diff
@@ -1802,6 +1802,217 @@ def resnetLayer5(self, b, x, k):
         o = b.relu(c)
         return [o]
 
+
```
flaub: I'm not sure this should go into backend_test.py, backend test is already pretty slow. Also, these comments don't seem generally useful.

tile/targets/cpu/llvm_cpu.jsonnet
```diff
@@ -242,6 +196,7 @@ local PARAMS = {
                 alignment: 16,
               },
             },
+            */
```
jbruestle: Remove this block (rather than commenting it).

tile/targets/cpu/executable.cc
```diff
@@ -79,23 +80,31 @@ void Executable::Run(const std::map<std::string, void*>& buffers) {
...
   void* argvec = args.data();
   uint64_t entrypoint = engine_->getFunctionAddress(invoker_name_);
+  // To get the raw execution time for generated code.
+  // struct timeval tp;
+  // gettimeofday(&tp, NULL);
+  // long long int ms = tp.tv_sec * 1000000 + tp.tv_usec; //get current timestamp in microseconds
   ((void (*)(void*))entrypoint)(argvec);
+  // gettimeofday(&tp, NULL);
+  // long long int msDone = tp.tv_sec * 1000000 + tp.tv_usec; //get current timestamp in microseconds
+  // IVLOG(1, "Raw execution time: " << msDone - ms);
```
flaub: Is this commented code still necessary? If it's still useful, let's find a way to toggle it when appropriate.

plaidml/keras/backend_test.py
```diff
@@ -1802,6 +1802,217 @@ def resnetLayer5(self, b, x, k):
...
+    @opTest([[m(1, 7, 7, 2048), m(1, 1, 2048, 512)]],
+            do_grads=False,
+            num_iterations=10,
+            measure_eval_time=True)
+    def resnetLayer28(self, b, x, k):
+        c = b.conv2d(x, k, padding='same')
+        o = b.relu(c)
+        return [o]
+
+
```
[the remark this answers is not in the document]
LLITCHEV: Yes. These are tests and setup for the layers of resnet50. It is good to have them in ready state, so we can evaluate perf when needed. Having more tests to test the correctness of the different phases (fusion/padding) is also very helpful - they did help us find, diagnose and fix few problems in these passes.l

plaidml/keras/backend.py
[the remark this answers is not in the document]
LLITCHEV: Went with #1. I think this change is not really related to this commit, but it makes the use of log debugging impossible pretty much, hence it is part of this diff.

tile/targets/cpu/llvm_cpu.jsonnet
LLITCHEV: Taken out.

conda/unix-mkl.yml
```diff
@@ -0,0 +1,33 @@
+# rename this file to unix.yml to get TF-MKL runs for perf testing.
```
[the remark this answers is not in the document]
LLITCHEV: Sure. We will chat. But in short, we need ability to measure perf (very rarely needed) on TF and TF+MKL). This requires different libraries. So, the unix_mkl.yml is pulling the TF binary that is build to use MKL. The unix.yml uses the plain TF only.

plaidml/BUILD
LLITCHEV: Yes... This is left over from testing. Taking it out.

plaidml/keras/backend.py
LLITCHEV: Taken care of. See comment #2 above.

tile/targets/cpu/executable.cc
flaub: Is there any way to retain this code without it being a comment? Perhaps something like an env var to enable it? Or is it obsolete now?

conda/unix-mkl.yml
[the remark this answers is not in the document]
flaub: I agree with the what and why, I just think the 'how' could be done in a way that doesn't require manually renaming files to run something on-demand.

plaidml/keras/backend_test.py
LLITCHEV: The tests runs are 14-16 ms each, so it should not take long at all. Removed the comments. They were useful to me to tell me what the constants are in respect to the XSMM calls and the resnet50 layers. Taken out the tests.

tile/targets/cpu/executable.cc
LLITCHEV: Discussed with Jeremy... We dcided to remove the code for now.

bzl/tbb.BUILD
```diff
@@ -1,67 +1,168 @@
...
+    hdrs = [
+        ":gen_version_file",
+    ],
+    defines = [
+        "TBB_USE_THREADING_TOOLS",
+        "__TBB_DYNAMIC_LOAD_ENABLED=0",
+        "__TBB_SOURCE_DIRECTLY_INCLUDED=1",
+        "__TBB_x86_64=1",
+    ] + select({
+        "@bazel_tools//src/conditions:windows": [
```
flaub: Could we use `@com_intel_plaidml//toolchain:windows_x86_64` here instead? I think we can eventually use the standard condition, but we don't do that everywhere yet.

bzl/tbb.BUILD
```diff
@@ -1,67 +1,168 @@
...
-        "include/tbb/**/**",
-    ]),
-    includes = ["include"],
+               "src/tbb/*.cpp",
+               "include/tbb/*.h",
+               "src/tbb/*.h",
+               "src/rml/include/*.h",
+           ]) + ["src/rml/client/rml_tbb.cpp"] +
+           select({
+               "@bazel_tools//src/conditions:windows": [
```
flaub: Use `@com_intel_plaidml//toolchain:windows_x86_64`

bzl/plaidml.bzl
```diff
@@ -45,10 +45,7 @@ def plaidml_cc_test(copts = [], deps = (), linkopts = [], **kwargs):
     native.cc_test(
         copts = PLAIDML_COPTS + copts,
         deps = deps + [str(Label("//testing:gtest_main"))],
-        linkstatic = select({
-            "@com_intel_plaidml//toolchain:linux_x86_64": 1,
-            "//conditions:default": None,
-        }),
+        linkstatic = 1,
```
flaub: Is this still required or was it just for testing/debugging?

I'm actually curious why we need this (just for my understanding)

bzl/tbb.BUILD
```diff
@@ -1,67 +1,168 @@
...
+        # uses extra_inc=big_iron.inc to specify that static libraries are
+        # built. See https://software.intel.com/en-us/forums/intel-threading-building-blocks/topic/297792
+        make tbb_build_prefix="build" \
+            extra_inc=big_iron.inc \
+            $$COMPILER_OPT; \
+
+        echo cp build/build_{release,debug}/*.a $$DEST_DIR
+        cp build/build_{release,debug}/*.a $$DEST_DIR
+        cd $$WORK_DIR
+  """,
```
flaub: If we're going to use make to do the build, we need to ensure that it's using the internal linux toolchain that is based on GLIBC 2.17 (so that we can target ancient linux distros like CentOS).

bzl/tbb.BUILD
```diff
@@ -1,67 +1,168 @@
...
+)
+
+cc_library(
+    name = "tbb",
+    srcs = select({
+        "@bazel_tools//src/conditions:windows": glob([
```
flaub: Use `@com_intel_plaidml//toolchain:windows_x86_64`

bzl/tbb.BUILD
```diff
@@ -1,67 +1,168 @@
...
+        "@bazel_tools//src/conditions:windows": [
+            ":gen_version_file",
+        ],
+        "//conditions:default": glob([
+            "include/serial/**",
+            "include/tbb/**/**",
+        ]),
+    }),
+    defines = select({
+        "@bazel_tools//src/conditions:windows": [
```
flaub: Use `@com_intel_plaidml//toolchain:windows_x86_64`

bzl/tbb.BUILD
```diff
@@ -1,67 +1,168 @@
...
+        ]) + [
+            "src/rml/client/rml_tbb.cpp",
+            ":gen_cpu_ctl_env",
+        ],
+        "//conditions:default": [
+            "libtbb.a",
+        ],
+    }),
+    hdrs = select({
+        "@bazel_tools//src/conditions:windows": [
```
flaub: Use `@com_intel_plaidml//toolchain:windows_x86_64`

bzl/tbb.BUILD
LLITCHEV: How do I specify that?

flaub: The way I know how to do it is to use bazel to do the build, not rely on `make`. But perhaps there's a way to get the right env vars passed into `make`.

I thought we had it building for linux/macos under bazel before?

bzl/plaidml.bzl
LLITCHEV: We need this to link in the tbb into the test executables. I dumped the Linux ones (nm is your friend :) ) and the note_affinity (a  tbb method) was removed by the linker since it was not called internally by anything - just the static initializer code of tbb.
Using alwayslink flag on the TBB and the xmm did not resolve the issues. Doubler checked dumping things also and the always link will keep the methods in the static library (they are always there anyway), but it will still remove them from the final executable when it links the library in. The linkstatic flags keeps the static symbols that are used by static initializers as well. (From what I can tell running dumps and analyzing things).

flaub: The downside is that it makes our test binaries very large especially when building in debug mode, and so it causes builds to take on the order of minutes to run tests in a quick cycle. I'm OK with leaving this in for now, but I'd like to revisit this to see if we can find a better solution.

bzl/tbb.BUILD
LLITCHEV: Changed

bzl/tbb.BUILD
LLITCHEV: Changed

bzl/tbb.BUILD
LLITCHEV: Changed

bzl/tbb.BUILD
LLITCHEV: Changed

bzl/tbb.BUILD
LLITCHEV: Changed

PR #737, opened 2019-11-20
Add no-framework C++ EDSL resnet

tzerrell:
This adds a ResNet50 implementation that does not use any framework, just the op library / EDSL. It is very minimal; it does not load weights or input data, and so can't be checked for correctness (especially dangerous considering the number of typos possible when writing networks without a framework). It is also currently built on the AST version of Stripe, not MLIR.

### DIFF
--- a/networks/oplib/BUILD
+++ b/networks/oplib/BUILD
@@ -1,0 +1,17 @@
+# Copyright 2018 Intel Corporation.
+
+load("//bzl:plaidml.bzl", "plaidml_cc_binary")
+
+package(default_visibility = ["//visibility:public"])
+
+plaidml_cc_binary(
+    name = "resnet50",
+    srcs = ["resnet50.cc"],
+    deps = [
+        "//base/util",
+        "//plaidml2:api_ast",
+        "//plaidml2/edsl:api",
+        "//plaidml2/exec:api",
+        "//plaidml2/op/lib:lib_ast",
+    ],
+)
--- a/networks/oplib/BUILD
+++ b/networks/oplib/BUILD
@@ -9,9 +9,11 @@ plaidml_cc_binary(
     srcs = ["resnet50.cc"],
     deps = [
         "//base/util",
-        "//plaidml2:api_ast",
         "//plaidml2/edsl:api",
+        "//plaidml2/edsl:edsl_ast",
         "//plaidml2/exec:api",
-        "//plaidml2/op/lib:lib_ast",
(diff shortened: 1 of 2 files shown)


networks/oplib/resnet50.cc
```diff
@@ -0,0 +1,424 @@
...
+  using plaidml::networks::run_resnet50;
+
+  try {
+    auto level_str = vertexai::env::Get("PLAIDML_VERBOSE");
+    if (level_str.size()) {
+      auto level = std::atoi(level_str.c_str());
+      if (level) {
+        el::Loggers::setVerboseLevel(level);
+      }
+    }
```
flaub: This should be happening already once you call `plaidml::op::init()`, see: https://github.com/plaidml/plaidml/blob/master/plaidml2/core/ffi.cc#L54

networks/oplib/BUILD
```diff
@@ -0,0 +1,17 @@
...
+
+plaidml_cc_binary(
+    name = "resnet50",
+    srcs = ["resnet50.cc"],
+    deps = [
+        "//base/util",
+        "//plaidml2:api_ast",
+        "//plaidml2/edsl:api",
+        "//plaidml2/exec:api",
+        "//plaidml2/op/lib:lib_ast",
```
flaub: I think you want `//plaidml2/op:op_ast` here

flaub: Sorry, I think:
```
"//plaidml2/op:api",
"//plaidml2/op:op_ast",
```

networks/oplib/resnet50.cc
```diff
@@ -0,0 +1,424 @@
...
+  auto B = bias_placeholders();
+  IVLOG(1, "About to build resnet50 program");
+  auto program = resnet50_program(batch_size, I, W, B);
+  IVLOG(1, "Program built");
+  auto all_inputs = W;
+  all_inputs.insert(all_inputs.end(), B.begin(), B.end());
+  all_inputs.push_back(I);
+  auto bound_program = exec::Binder(program).set_inputs(all_inputs).build();
+  IVLOG(1, "Inputs bound");
+  bound_program->run();
```
flaub: FYI: the initial run does not include any compilation. All compilation should be complete when `.build()` returns an `Executable`

flaub: You might consider doing:  ```
auto bound_program = exec::Binder(program).set_inputs(W).set_inputs(B).set_input(I).build();
```

PR #744, opened 2019-11-21
Pass tensors to Tile domains as args

earhart:


flaub:
Can we hold off on this? Do you need this soon?

earhart:
Holding off pending more thorough design discussion.

PR #745, opened 2019-11-22
Add benchmarking of resnet50

flaub:
Also fix bugs


pmlc/dialect/tile/builder.cc
```diff
@@ -61,15 +61,15 @@ struct DomainInfo {
...
 struct TileBuilder::Impl {
   MLIRContext context;
   ModuleOp module;
   OpBuilder builder;
   std::map<AffineDomainOp, DomainInfo> domains;
-  std::map<Value*, Value*> implicitUpdates;
-  Bindings implicitBindings;
+  std::unordered_map<Value*, RankedTensorType> shapeCache;
+  std::unordered_map<Value*, Value*> implicitUpdates;
+  std::unordered_map<mlir::Value*, vertexai::tile::BufferPtr> implicitBindings;
```
tzerrell: Style: probably either use `Value*` or `mlir::Value*` in all of these lines, not a mix

PR #752, opened 2019-11-28
Start work on region-less contractions

flaub:
This is not currently hooked up thru the EDSL. A later PR will do that and also add support for the 2-sided constraint form (e.g. `0 <= x < N` -> `(x >= 0), (-(x - N) - 1 >= 0)`


pmlc/dialect/tile/contraction.cc
```diff
@@ -263,6 +263,50 @@ static IndexPoly MakePoly(mlir::Value* value) {
...
+    switch (binaryExpr.getKind()) {
+      case mlir::AffineExprKind::Add:
+        return MakePoly(binaryExpr.getLHS()) + MakePoly(binaryExpr.getRHS());
+      case mlir::AffineExprKind::Mul: {
+        auto rhs = MakePoly(binaryExpr.getRHS());
+        return MakePoly(binaryExpr.getLHS()) * rhs.constant();
+      }
+      case mlir::AffineExprKind::FloorDiv: {
+        auto rhs = MakePoly(binaryExpr.getRHS());
+        return MakePoly(binaryExpr.getLHS()) / rhs.constant();
```
tzerrell: Should we be checking that the RHS poly is constant? Or is it ok to be assuming that safe polynomials were passed in? (same question on * case)

flaub: OK I added a check for this

PR #756, opened 2019-12-03
Embed heatmap into tile/targets/cpu

flaub:


flaub:
@jbruestle @shiyao00 So should we go ahead and land this?

flaub:
Oh, looks like it doesn't build on windows, it's too big for the parser. We'll need to come up with some alternative.

flaub:
@shiyao00 is this safe to close now?

shiyao00:
> @shiyao00 is this safe to close now?

Yes, please close it.

PR #761, opened 2019-12-06
The optimization pass of automatic stencil

shiyao00:
* The optimization pass of automatic stencil.
* Add some helpers for Stripe dialect.
* Fix bugs in conversion between Stripe and MLIR

### DIFF
--- a/pmlc/dialect/stripe/autostencil_pass.cc
+++ b/pmlc/dialect/stripe/autostencil_pass.cc
@@ -1,0 +1,417 @@
+// Copyright 2019, Intel Corporation
+
+#include "pmlc/dialect/stripe/transcode.h"
+#include "pmlc/dialect/stripe/autostencil_pass.h"
+
+#include <sstream>
+#include <vector>
+
+#include "mlir/IR/Matchers.h"
+#include "mlir/Translation.h"
+
+#include "pmlc/dialect/stripe/analysis.h"
+#include "pmlc/dialect/stripe/dialect.h"
+#include "pmlc/dialect/stripe/ops.h"
+#include "pmlc/dialect/stripe/transforms.h"
+#include "pmlc/dialect/stripe/util.h"
+
+#include "base/util/logging.h"
+
+namespace pmlc {
+namespace dialect {
+namespace stripe {
+
+using vertexai::tile::codegen::proto::MLIR_AutoStencilPass;
+using BlockArgumentSet = llvm::SmallPtrSet<mlir::BlockArgument*, kIndexLimit>;
+
+// Number of tensors for the matrix multiplication
+const unsigned kNumTensors = 3;
+// Number of searching index, i.e., M, N, K
+const unsigned kNumIndex = 3;
+
+class AutoStencil {
+public:
+  explicit AutoStencil(const MLIR_AutoStencilPass& opts);
+  // Main function
+  void Stencil(ParallelForOp op);
+private:
+  // The throughput of M*N*K matrix multiplication
+  double Throughput(unsigned m, unsigned n, unsigned k);
+  // Evaluate the performance of the current searching state
+  double Evaluate();
+  // Search tensors' order
+  void SearchTensorsOrder();
+  // Search the index, i.e., M, N, K, for the inner block
+  void SearchIndex(unsigned matrix_idx);
+  // For (M, N, K) in the inner block, search their tiles
+  void SearchTiles(unsigned idx);
+  // Collect the tensors in the block
+  bool CollectTensors();
+  // Transform the current ParallelForOp
+  void Transform();
+
+  // Collect the index used by value's RefineOp
+  BlockArgumentSet RefUsedIdxs(Value* value, bool with_conflict);
+  // Test if idx is conflict with any index in innerIdxs
+  bool ConflictInnerIndex(BlockArgument* idx);
+  // Test if idx in tensors[tensor_idx] is stride one index
+  bool IsStrideOne(BlockArgument* idx, unsigned tensor_idx);
+  // Test if idx in the tensors are stride one
+  bool ValidateStrideOne(BlockArgument* idx, unsigned matrix_idx);
+
+  // Number of processors
+  const MLIR_AutoStencilPass& options;
+  // The current op
+  ParallelForOp curOp;
+  // Tensors' order
+  unsigned tensorsOrder[kNumTensors];
+  // M, N, K in inner block
+  BlockArgument* innerIdxs[kNumIndex];
+  // M, N, K's tiles
+  unsigned tiles[kNumIndex];
+  // The matrix_idx for the next search
+  unsigned nextMatrixIdx[kNumIndex] = {3, 2, 0};
+  // Target tensors, the first two are load, the third is aggregate
+  llvm::SmallVector<Value*, kNumTensors> tensors;
+  // Stride one index for the tensors
+  BlockArgumentSet strideOne[kNumTensors];
+  // The index used by the output tensor
+  BlockArgumentSet outIdxs;
+  // The accumulation index
+  BlockArgumentSet accIdxs;
+  // All used index
+  BlockArgumentSet allIdxs;
+  // Each index has a set of conflict index that can't be choosed into inner block together
+  llvm::DenseMap<BlockArgument*, BlockArgumentSet> conflict;
+  // The best performance
+  double bestPerf;
+  // The best tensors' order
+  unsigned bestTensorsOrder[kNumTensors];
+  // The best index (M, N, K)
+  BlockArgument* bestIdxs[kNumIndex];
+  // The best tiles for (M, N, K)
+  unsigned bestTiles[kNumIndex];
+};
+
+AutoStencil::AutoStencil(const MLIR_AutoStencilPass& opts) :
+  options(opts) {
+}
+
+double AutoStencil::Throughput(unsigned m, unsigned n, unsigned k) {
+  if (m == 16 && n == 16 && k == 3) {
+    return 64;
+  }
+  if (m == 32 && k == 32) {
+    return 64;
+  }
+  if (m == 16 && k == 16) {
+    return 64;
+  }
+  if (m == 8 && k == 16) {
+    return 64;
+  }
+  if (m == 48 && k == 48) {
+    return 64;
+  }
+  return -1;
+}
+
+double AutoStencil::Evaluate() {
+  unsigned tot_inner_loop = tiles[0] * tiles[1] * tiles[1];
+  double inner_time = tot_inner_loop / Throughput(tiles[0], tiles[1], tiles[2]);
+  IVLOG(3, "Inner: loop = " << tot_inner_loop << " inner_time = " << inner_time);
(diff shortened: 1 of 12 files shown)


shiyao00:
It's not completely finished. Need to merge with the heatmap and tune the performance.

pmlc/dialect/stripe/autostencil_pass.cc
```diff
@@ -0,0 +1,312 @@
...
+
+  size_t outer_batches = (tot_outer_loop - 1) / n_cpus + 1;
+  double perf = outer_batches * tot_middle_loop * inner_time;
+  IVLOG(3, "Performance = " << perf);
+
+  return perf;
+}
+
+void AutoStencil::SearchTiles(size_t idx) {
+  if (idx >= numIndex) {
```
jbruestle: I like this search mechanism

pmlc/dialect/stripe/autostencil_pass.cc
```diff
@@ -0,0 +1,312 @@
...
+  auto ref_op = mlir::dyn_cast<RefineOp>(value->getDefiningOp());
+  auto tensor = value->getType().cast<TensorRefType>();
+  if (!ref_op) {
+    throw std::runtime_error("Access a tensor not defined by RefineOp.");
+  }
+  for (int i = 0; i < tensor.getRank(); i++) {
+    auto access = AffinePolynomial(ref_op.getOffset(i));
+    for (auto [arg0, scale0] : access.terms) {
+      used_idxs.insert(arg0);
+      if (with_conflict) {
```
jbruestle: If I understand correctly, the conflict sets prevent one from picking say both 'i' and 'x' for a convolution which has D[i + x] somewhere?  So basically any index conflicts with any other index that appears in the same polynomial at any point, right?  That makes a lot of sense, but some comments about that when the conflict member variables are defined would be useful, since it took me a while to understand.

pmlc/dialect/stripe/autostencil_pass.cc
```diff
@@ -0,0 +1,312 @@
...
+// Search for matrix index (0 for M, 1 for N, 2 for K)
+void AutoStencil::SearchIndex(size_t matrix_idx) {
+  if (matrix_idx >= numIndex) {
+    SearchTiles(0);
+    return;
+  }
+  auto& idxs = (matrix_idx == 2) ? all_idxs : out_idxs;
+  for (auto idx : idxs) {
+    if (!ConflictInnerIndex(idx)) {
+      inner_idxs.push_back(idx);
```
jbruestle: Need to check for stride 1 requirements, possibly here?

pmlc/dialect/stripe/autostencil_pass.cc
```diff
@@ -0,0 +1,312 @@
...
+  // Transform
+  Transform();
+}
+
+void AutoStencilPass::runOnFunction() {
+  auto reqs = vertexai::tile::stripe::FromProto(options.reqs());
+  mlir::FuncOp f = getFunction();
+  AutoStencil as(options.processors());
+  f.walk([&reqs, &as] (ParallelForOp op) {
+    if (hasAttrs(op.getOperation(), reqs)) {
```
jbruestle: I think using tag based approach to find blocks to modify is good right now for transition, however long term it might be good to just look for blocks that have the right form.

pmlc/dialect/stripe/util.cc
```diff
@@ -35,6 +37,166 @@ void createMainParallelFor(mlir::FuncOp funcOp) {
   builder.create<TerminateOp>(funcOp.getLoc());
 }
 
+bool hasAttr(Operation* op, const std::string& attr) {
```
jbruestle: Many of these utility functions are working around a design flaw I think.  Specifically, if the 'stripe attributes' were placed directly on the Op, rather than into a DictionaryAttr on the op, then things would be much easier.  For example, 'if (hasAttr(op, "name"))' would just be 'if(op->getAttr("name")' using existing MLIR support with no additional functions required.  Additionally, adding a new attribute would also be trivial, and not involve copying a while dictionary to make a change.  Let me think about what the right thing to do here is and also talk to Frank.

pmlc/dialect/stripe/autostencil_pass.cc
```diff
@@ -0,0 +1,312 @@
...
+namespace stripe {
+
+// Number of tensors for the matrix multiplication
+const size_t numTensors = 3;
+// Number of searching index, i.e., M, N, K
+const size_t numIndex = 3;
+// Maximum number of index in a block, e.g., ParallelForOp
+const size_t indexLimit = 32;
+
+typedef llvm::SmallPtrSet<BlockArgument*, indexLimit> BlockArgumentSet;
```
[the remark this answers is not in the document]
flaub: Also, let's start trying to use the new `using` syntax since it's a little easier to read:

```
using BlockArgumentSet = llvm::SmallPtrSet<BlockArgument*, indexLimit>
```

pmlc/dialect/stripe/autostencil_pass.cc
```diff
@@ -0,0 +1,312 @@
...
+#include "pmlc/dialect/stripe/util.h"
+
+#include "base/util/logging.h"
+
+namespace pmlc {
+namespace dialect {
+namespace stripe {
+
+// Number of tensors for the matrix multiplication
+const size_t numTensors = 3;
```
flaub: For constants, I think the current style is `kNumTensor`

flaub: Also, although this is different than our current codebase, I've noticed that LLVM prefers to use `unsigned` over `size_t`

pmlc/dialect/stripe/autostencil_pass.cc
```diff
@@ -0,0 +1,312 @@
...
+  // The current op
+  ParallelForOp cur_op;
+  // M, N, K in inner block
+  llvm::SmallVector<BlockArgument*, numIndex> inner_idxs;
+  // M, N, K's tiles
+  size_t tiles[numIndex];
+  // Target tensors, the first two are load, the third is aggregate
+  llvm::SmallVector<Value*, numTensors> tensors;
+  // The index used by the output tensor
+  BlockArgumentSet out_idxs;
```
flaub: @jbruestle How closely should we match LLVM style? They'd use `oldIdxs` here

pmlc/dialect/stripe/autostencil_pass.cc
```diff
@@ -0,0 +1,312 @@
...
+    double performance = Evaluate();
+    if (performance > best_perf) {
+      best_perf = performance;
+      best_tiles.clear();
+      for (size_t i = 0; i < numIndex; ++i) {
+        best_tiles[idxName(inner_idxs[i])] = tiles[i];
+      }
+    }
+    return;
+  }
```
flaub: Seems like this could just be an `else`, unless I'm missing something...

pmlc/dialect/stripe/autostencil_pass.cc
flaub: I'm a little worried about mutating state on the `AutoStencil` object to accumulate the tile as we descend. Have you considered passing in an array that gets populated as you go down? That way it's still 'functional' and doesn't have side effects on the `AutoStencil` object.

pmlc/dialect/stripe/autostencil_pass.cc
```diff
@@ -0,0 +1,312 @@
...
+            conflict[arg1].insert(arg0);
+          }
+        }
+      }
+    }
+  }
+  return used_idxs;
+}
+
+void AutoStencil::Transform() {
```
flaub: So I'm thinking that it'd be good to have a new op that is the result of tiling, like in this case perhaps we generate an `xsmm` op so that we can just use attributes to mark the stencil properties rather than using tags like we did in stripe classic.

pmlc/dialect/stripe/util.cc
flaub: The reason for the dictionary was to isolate stripe 'tags/attrs' from MLIR ones. If we're not going to bother with translating to stripe classic, then I agree, we can just put them directly on the op.

pmlc/dialect/stripe/util.cc
```diff
@@ -35,6 +37,166 @@ void createMainParallelFor(mlir::FuncOp funcOp) {
...
+      idx_name = str_attr;
+    }
+  }
+  return idx_name.getValue();
+}
+
+std::pair<StringRef, size_t> getSingleIndex(ParallelForOp op, size_t n) {
+  auto names = op.getAttrOfType<ArrayAttr>(mlir::Identifier::get("idx_names", op.getContext()));
+  auto ranges = op.ranges();
+  auto idx_name = StringAttr::get("", op.getContext());
```
flaub: I'm thinking that in the case of an index name missing, we should generate a default one like `i{{n}}` or something.

flaub: Also, it'd be nice to have a bunch of other code that does this same logic reuse this utility function. There's places in the code already that do this.

pmlc/dialect/stripe/autostencil_pass.cc
```diff
@@ -0,0 +1,312 @@
...
+  bool ConflictInnerIndex(BlockArgument* idx);
+
+  // Number of processors
+  size_t n_cpus;
+  // The current op
+  ParallelForOp cur_op;
+  // M, N, K in inner block
+  llvm::SmallVector<BlockArgument*, numIndex> inner_idxs;
+  // M, N, K's tiles
+  size_t tiles[numIndex];
```
flaub: Isn't this a 'single' tile, not a collection of tiles?

pmlc/dialect/stripe/autostencil_pass.cc
jbruestle: We only track the 'best' output, so it's really more of a return value, but we also need the 'best cost' etc, so it would end up being some sort of composite object.  The idea that the AutoStencil object represents the search state seems Ok to me, but the fact the we reuse the AutoStencil state multiple times and clear it each time is a little awkward, so perhaps we should construct a new AutoStencil each time around the loop to make it more clear (which is basically free since it's got a mostly empty constructor).

pmlc/dialect/stripe/util.cc
jbruestle: Even if we do translate to/from stripe, the dictionary is pretty awkward, so perhaps something like either: 
- A name prefix (i.e. stripe_<some name>) or
- Assume any attribute not known to MLIR specifically should be translated back

But yes, this may all be a mute point if we end up going straight to extended affine...

pmlc/dialect/stripe/autostencil_pass.cc
jbruestle: Probably we should, but if we are going to do that, we should decide where the boundary between 'style regimes' is (probably pmc/*?) and maybe see if we can get some automatic linting?  Also, we should probably make sure everyone know where the style guide is, and resolve thinks like use of references for out-vals, etc.  Seems like something to follow up on in a meeting.

PR #766, opened 2019-12-12
draft - chcking md math fix

1sanz:


dgkutnic:
A few things:

- Can we somehow preserve the LaTeX code from the original functions? Perhaps as a comment or some other annotation?
- Any way to use colors with this approach? I've tried a few methods with no luck, and the internet doesn't seem to have any guidance.
- Can we make the text a little larger, if possible? Not critical, but the equations that existed on gitlab were bigger and therefore easier to read

1sanz:
@dgkutnic: I changed the first formula to look like what you suggested. If this looks good I'll update the rest tomorrow :)

flaub:
I'll check with @brianretford but I think we should spend some time to actually have a build step that renders our docs into github pages. It seems like we'll need this generally anyways going forward.

flaub:
My biggest worry with this approach is keeping it:
1. Maintainable (I'd like to write multi-line expressions)
2. In-sync between the rendered version and the code (ideally we'd have a single source of truth)

flaub:
Agreed about colors, they're useful for describing the different parts of contractions.

PR #768, opened 2019-12-12
Refresh docs

flaub:
This is setup for jekyll, which makes it a drop in for using GitHub Pages.


flaub:
This doesn't yet deal with LaTeX math properly, but it seems like it's possible with jekyll.

tzerrell:
Cool, I'd be happy with a GitHub Pages Jekyll LaTeX approach for docs. It appears that this ultimately goes to MathJax, which has support for the AMSMath `align` environment, which gives multiline support: http://docs.mathjax.org/en/v1.0/tex.html. It also has at least basic color support, which should be sufficient.

flaub:
Sounds like KaTeX is another option. Regardless, at least we have options with this approach without having to do any server-side/build steps (that's my hope anyways!).

flaub:
FYI, a lot of work is needed to get this to land:

* Need to update all the internal links (including images)
* Need to get LaTeX math to render properly
* Update main README.md at the root of the repo. Right now the `index.md` is duplicated, is this a good idea?

@1sanz or @dgkutnic Can I hand this off to you?

tzerrell:
Awesome!

I've heard good things about KaTeX from a "how easy is it to typeset with" perspective but I don't know the relative difficulty of integrating it with the other hosting tools we need compared to Jekyll. I suspect the later concerns should dominate as both seem like they should be fine from an expressibility perspective.

dgkutnic:
I had actually started playing around with mathjax the last time we had a docs discussion, and it seems like the right choice. I second Tim's comment that mathjax can handle different colors, fonts, sizes, etc. without needing to do any additional building.

I'd be happy to finish this up and/or collaborate with @1sanz on this.

flaub:
Cool! Looking good, nice work!

docs/docs/edsl.md
```diff
@@ -323,24 +320,27 @@ We determined the Tile code for this example by starting from imperative code,
...
+\color{green} \max_{\color{magenta} 0 \leq j < 2}
+\color{cyan} I[2i + j] 
+$$
+
 
-```math
+$$
 \Large
 \texttt{
-  if (\textcolor{magenta}{j < 2}) \{{
```
flaub: This is why we need the `{% raw %}` block, to prevent Liquid from breaking on the double `{{`, which is needed to make open curly display properly in LaTeX.

docs/docs/edsl.md
```diff
@@ -455,24 +455,22 @@ Tensor csum(const Tensor& I) {
...
+$$
 \Large
-\texttt{
-  if (\textcolor{magenta}{j < N}) \{{
-    \textcolor{red}{O(n)}
-    \textcolor{green}{+=}
-    \textcolor{cyan}{I(i - j)};
-  \}}
+
+  \text{if} ( \color{magenta}j < N \color{black} ) \color{red} O(n) \color{green}{+=} \color{cyan} I(i - j) \color{black};
```
flaub: If at all possible, I'd like to avoid smashing multiple lines down into a single one, it makes it much harder to figure out what's going on (to me anyways).

docs/_includes/mathjax_support
```diff
@@ -0,0 +1 @@
+<script type="text/javascript" src="http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>
```
flaub: Can we put a reasonable suffix on this file? like `.html`?

flaub:
OK, after some more research, it looks like we can't use any plugins with github pages, so we can't do anything server-side.
However, this is interesting:
https://karas.io/blog/math-support-with-katex-on-github-pages/

> I would have used MathJax, but for the speed and layout stability issue. MathJax is almost unreadable on mobile, especially high-latency mobile like most of the world has.

dgkutnic:
Wow! I'm impressed at how far this has gotten so quickly.

@flaub you have converted me to KaTeX. This currently looks good on mobile + my Mac.

I'm okay with pushing this to production, as long as we have considered the following (mostly minor) points:

- On the home page, we reference 3rd party sources, but we don't have the legal disclaimer. Could we add that as a footer or something? That legal disclaimer is: `* Other names and brands can be claimed as the property of others`. Along these lines, we'd also need to add an asterisk wherever a 3rd party source is mentioned. It's kind of inconsistent right now and I don't want us getting into any trouble.
- Could this framework be extensible in a way such that we could add pages for each of our MLIR dialects, similar to how TensorFlow dialects convert their `td` files into webpages?
- Perhaps we should think about modifying navigation a bit. Currently, it's only encompassing top-level pages, but we should probably include headers too.

flaub:
Where is the official guide for policy on public pages? I agree about a disclaimer but I don’t know exactly where to put it. I also don’t think we should use asterisk everywhere (what does it even mean?). Do you mean (c) and (r) for copyright and trademarks?

Yes we can add more docs later. 

Navigation happens via a theme. Let’s get this landed and then if someone wants to dig into making the navigation more flexible they can. I personally think it’s fine as is (I don’t like putting anchors into the nav).

PR #771, opened 2019-12-14
Rename combo_op tag to comb_op

earhart:


earhart:
Hmm -- macOS passes... known Linux issue?

flaub:
Why do we need this change?

flaub:
Comb seems objectively worse than combo

jbruestle:
So it is weird the combo stands for combination (which doesn't in fact have an 'o' in it), so perhaps that's the reasoning for this change?  However, I still think it's better than 'comb' (which is a thing you use to adjust your hair).  I don't know that I'd say 'comb' in 'objectively' worse, although 'abbreviation for combination' on the internet does come back with 'combo' as the first hit.  In general my approach has been: everyone just picks their best guess at names for things, and then we try to keep them consistent unless there is a clear reason to change them.  Otherwise it's just churn and without a sensible rubric for what's better (which I don't have).  I'm so glad for clang-format, so at least we don't need to adjudicate that sort of thing, if only it worked for english as well.

earhart:
The reasoning for this change is that the old Tile->Stripe pipeline generated tags named like "comb_op_foo" -- check out `gen_stripe.cc:329`.  I'm wanting this change solely for consistency with older Stripe pass logic.

PR #777, opened 2019-12-17
Op Lib Dialect: wrapper fixes & added tests

dgkutnic:
### Overview

To build the op lib wrappers: `bazelisk build //pmlc/dialect/op_lib:plaidml_op_lib_wrappers`
To run op lib dialect tests: `bazelisk test //pmlc/dialect/op_lib/tests:all`

---

### Fix make_tuple calls in the Op Lib Dialect

In previous versions of the op lib dialect, `edsl::make_tuple` calls for operator arguments were empty. They are now populated with the real arguments to the operator.

---

### Static Casting Enum Types

I was trying to pass enum classes (i.e. `TensorLayout`, `GroupLayout`, `DerivMode`, etc.) through an EDSL tuple to instantiate ops within the op lib dialect. For example:

```c++
/*
 This doesn't work because edsl::make_tuple doesn't understand the
 TensorLayout enum class
*/
TensorLayout input_layout_;
auto args = edsl::make_tuple(input_layout_);
```

We decided that it would be best to statically cast those enums to ints for the EDSL call, and statically cast them back to enums for the actual op lib implementation. (Side note: MLIR by default types its enums as `uint32_t` which I am casting to `int64_t` for now. We might want to change this.)

```c++
TensorLayout input_layout_;
auto args = edsl::make_tuple(static_cast<int64_t>(input_layout_));
```

Additionally, I removed the dummy `op_lib_wrappers.cc` file and replaced it with `ops.cc` which has the actual implementations of the ops, modified to `static_cast` the enum types to ints.

---

### Added tests

I added and renamed a few more tests for the op lib dialect. I'll add even more in an upcoming PR.


pmlc/dialect/op_lib/OpModel.h
```diff
@@ -0,0 +1,52 @@
...
+using llvm::MapVector;
+using llvm::RecordKeeper;
+using mlir::StringRef;
+using mlir::tblgen::EnumAttr;
+using mlir::tblgen::Operator;
+
+namespace pmlc::dialect::op::tblgen {
+
+struct OpInfo {
+  StringRef name_;
```
flaub: nit: no need for `_` suffix if it's just a public member

pmlc/dialect/op_lib/cpp/ImplGen.cc
```diff
@@ -0,0 +1,47 @@
...
+  for (const auto& optPair : typeInfo_.opts_) {
+    os.indent(2) << optPair.first << " = " << optPair.second << ",\n";
+  }
+  os << "};\n\n";
+}
+
+Emitter::Emitter(DialectInfo info, raw_ostream& os) : info_(info) {
+  emitHeaders(os);
+  os << "namespace plaidml {\n"
+     << "namespace op {\n\n";
```
flaub: I think we should eventually make namespaces configurable from the `.td`, but this is fine for now.

pmlc/dialect/op_lib/cpp/WrapperGen.cc
```diff
@@ -0,0 +1,162 @@
...
+
+namespace wrapper {
+
+OpEmitter::OpEmitter(const OpInfo& op, raw_ostream& os) : opInfo_(OpInfo(op)) {
+  os << formatv(commentHeader, opInfo_.name_, "wrapper");
+  os << "class " << opInfo_.name_ << " {\n";
+  os.indent(2) << "protected:\n";
+  // Emit operand and attribute declarations
+  emitDeclarations(os);
+  os.indent(2) << "private:\n";
```
flaub: This probably should be `public:`

pmlc/dialect/op_lib/cpp/WrapperGen.cc
```diff
@@ -0,0 +1,162 @@
...
+using llvm::RecordKeeper;
+
+namespace pmlc::dialect::op::tblgen::cpp {
+
+namespace wrapper {
+
+OpEmitter::OpEmitter(const OpInfo& op, raw_ostream& os) : opInfo_(OpInfo(op)) {
+  os << formatv(commentHeader, opInfo_.name_, "wrapper");
+  os << "class " << opInfo_.name_ << " {\n";
+  os.indent(2) << "protected:\n";
```
flaub: This probably should be `private:` (I don't know if we need to support having subclasses)

pmlc/dialect/op_lib/cpp/WrapperGen.cc
```diff
@@ -0,0 +1,162 @@
...
+    if (visited) os << ", ";
+    visited = true;
+    os << convertType(operandPair.second) << " " << operandPair.first;
+  }
+  os << ") : ";
+  // Initialize the private variables in the op class with the parameters passed in
+  visited = false;
+  for (const auto& operandPair : opInfo_.operands_) {
+    if (visited) os << ", ";
+    visited = true;
```
flaub: Hmm, I usually do: 

```
for (unsigned i = 0; i < opInfo_.operands.size(); i++) {
  if (i) {
    os << ", ";
  }
}
```

pmlc/dialect/op_lib/cpp/utils.h
```diff
@@ -0,0 +1,98 @@
...
+static inline const char* const initFunction = R"(
+inline void init() {  //
+  plaidml::init();
+  plaidml::edsl::init();
+  ffi::call_void(plaidml_op_init);
+}
+
+)";
+
+// Order doesn't matter here, but lookup efficiency does, so keep as a std::map.
```
flaub: Actually, an `unordered_map` should be constant lookup `O(1)` while a `map` is `O(log(n))`, so if all you need is lookups, `unordered_map` is probably better (but in the grand scheme of things doesn't really matter, but the comment isn't actually accurate).

pmlc/dialect/op_lib/cpp/utils.h
```diff
@@ -0,0 +1,98 @@
...
+    {"uint32_t", "int64_t"},
+};
+
+static inline StringRef convertType(const StringRef type) {
+  StringRef edslType = "unrecognized";
+  auto found = typeLookupTable.find(type);
+  if (found != typeLookupTable.end()) {
+    edslType = found->second;
+  }
+  return edslType;
```
flaub: This could be:

```
if (auto it = typeLookupType.find(type)) {
  return it->second;
}
llvm_unreachable("unrecognized type");
```

pmlc/dialect/op_lib/BUILD
```diff
@@ -27,10 +27,55 @@ PLATFORM_COPTS = select({
...
+    ]),
+    hdrs = glob([
+        "python/WrapperGen.h",
+        "python/utils.h",
+    ]),
+    deps = [
+        ":OpModel",
+        "@mlir//:TableGen",
+    ],
+)
```
flaub: The 'bazel' way of doing this would be to put a `BUILD` file in `python` and `cpp` subdirs.

pmlc/dialect/op_lib/BUILD
```diff
@@ -27,10 +27,55 @@ PLATFORM_COPTS = select({
     ],
 })
 
+plaidml_cc_library(
+    name = "OpModel",
```
flaub: Hmm, well so far we've been using the filenaming style of `foo_bar` instead of `FooBar`. I know LLVM/MLIR uses the latter standard. I'm not sure if we should use that style or not, but I'm pretty sure we should be consistent. Maybe we should stick with the old style for now and then convert everything to the new style as one big consolidated commit.

pmlc/dialect/op_lib/cpp/utils.h
```diff
@@ -0,0 +1,98 @@
...
+static inline const char* const ffiFunction = R"(
+namespace details {
+
+inline edsl::Value op(const std::string& name, const edsl::Value& args) {
+  return edsl::Value(ffi::call<plaidml_expr*>(plaidml_op_make, name.c_str(), args.as_ptr()));
+}
+
+} // namespace details
+
+)";
```
flaub: I'm thinking we should leave this kind of thing outside of the generated code, and instead have them be part of `ops.cc` (the file that includes `ops.h.inc`)

pmlc/dialect/op_lib/cpp/utils.h
```diff
@@ -0,0 +1,98 @@
...
+static inline const char* const includeHeader = R"(
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "plaidml2/edsl/edsl.h"
+#include "plaidml2/op/ffi.h"
+
+)";
```
flaub: Ya, I think we should not really do any of this stuff from the generated file, instead if the outside file `ops.cc` did it, then it would make it much easier to create and use test dialects.

dgkutnic:
@flaub whoops, sorry for the thrash here. We had some tests fail because we changed the binary's name from `op-lib-tblgen` to `oplib-tblgen` but did not change the tests to match the binary. I've made that change in the tests and they are passing locally now.

dgkutnic:
fixed codegen per #765

PR #781, opened 2019-12-18
Various cleanups

flaub:


dgkutnic:
> I wonder whether "ml" might be a better name for the dialect than "oplib"...?

I'm open to renaming it. I like the sound of the `plaid` dialect but that's honestly not very descriptive. The `ml` dialect may be too general, but it might be what we're going for in this dialect. I also prefer just calling it the `op` dialect over the `oplib` dialect tbh.

earhart:
IMHO -- "op" is a bit too overloaded to use in MLIR code.  :-)  "ml" seems like what this dialect is going for.  (And then the pass might be something like "lower-ml-to-tile"... with passes like "lower-ml-to-mkldnn" to get the system to use handwritten kernels for a specific device.)

earhart:
(No reason to hold up this change, though -- that was just a thought I had while reading the code.)

flaub:
Yep I'm open to a rename too, let's discuss when we're all around.

PR #783, opened 2019-12-18
Add new PXA dialect and two lowerings.  Still a WIP, but want to merge to coordinate with @flaub

jbruestle:


flaub:
Need the Ubuntu bump to land to fix this

PR #785, opened 2019-12-19
Natural Docs Prototype

dgkutnic:
This is a prototype of the `TensorDim` class in C++ and Python (along with some of the functions it references). I'm looking for feedback on the general style/layout of the comments within the code as it relates to the docs that are generated (critique on the content of the comments itself is welcome too). The goal is to rapidly prototype on this class before I go and do a bunch of classes.

We'll probably have to modify the CSS too, but that can be something we do later.


dgkutnic:
Side Note: integrating natural docs into bazel seems like it might be _interesting_. I'll leave that part for later since it isn't immediately critical to getting the docs out.

flaub:
OK, I can work on the bazel build part, let's not checkin all the generated files

dgkutnic:
I agree that we wouldn't want to check them in for an actual PR, just wanted a way to host them for now for an easier review process.

flaub:
Actually, it's very hard for me to review since there's lots of irrelevant files in the PR.

dgkutnic:
OK, I'll get rid of them now then.

flaub:
@dgkutnic Are you ok if we close this PR? Seems like sphinx will work for us?

dgkutnic:
@flaub yeah, I'm fine with closing it. I was keeping it open for reference purposes, but we've gone far enough down the Sphinx path that I'm comfortable closing this.

PR #787, opened 2019-12-20
Sphinx Prototype

dgkutnic:
`make html`

Side Note: Currently, this is spitting out ~2000 warnings related to duplicate declarations (it seems to think every operator overload is a duplicate declaration)...


flaub:
I was able to tame doxygen/sphinx. Basically, we need to manually specify which groups or classes we want to include in the docs, instead of letting the 'auto' versions do everything blindly.

dgkutnic:
@flaub I'm currently happier with this than the rest of the options we've looked at. 

Separately, do you think we should beef up `docs/README.md` to include usage of `watch.py`/a more detailed specification of the doc generation process?

PR #795, opened 2019-12-23
New MLIR pass for initializing aggregate refinements

shiyao00:
* New optimization pass
* Fix Vgg19 on CPU
* Fix partial-zero in padding pass
* Fix missing zero-init while changing large prime index in padding pass
* Automatic hardware concurrency detection for autostencil pass


tile/base/shape.cc
```diff
@@ -81,5 +81,72 @@ DataType CommonSupertype(DataType lhs, DataType rhs) {
   return lhs;
 }
 
+bool IsIntegerDataType(DataType type) {
```
jbruestle: This is somewhat duplicative of is_int || is_uint in shape.h

tile/base/shape.cc
```diff
@@ -81,5 +81,72 @@ DataType CommonSupertype(DataType lhs, DataType rhs) {
...
+  return type == DataType::INT128 ||
+         type == DataType::INT64 ||
+         type == DataType::INT32 ||
+         type == DataType::INT16 ||
+         type == DataType::INT8;
+}
+
+bool IsFloatDataType(DataType type) {
```
jbruestle: This is an exact dup of is_float

pmlc/dialect/stripe/transforms.cc
```diff
@@ -180,6 +184,50 @@ void LiftConstraint(ParallelForOp pf) {
   pf.getOperation()->erase();
 }
 
+// Parallelize a ParallelForOip for eltwise only
```
jbruestle: Looks good.

pmlc/dialect/stripe/autostencil_pass.cc
```diff
@@ -295,6 +298,38 @@ bool AutoStencil::ValidateStrideOne(BlockArgument* idx, unsigned matrix_idx) {
   return false;
 }
 
+bool AutoStencil::IndexExists(BlockArgument* idx, unsigned tensor_idx) {
+  return tensorIdxs[tensor_idx].find(idx) != tensorIdxs[tensor_idx].end();
+}
+
+// Confirm if idx exists in the right place
+bool AutoStencil::ValidateIndexExistance(BlockArgument* idx, unsigned matrix_idx) {
+  switch (matrix_idx) {
```
jbruestle: I think this is a fine approach.  In the old stencil pass, we put all of this into the configuration, but I think that's probably overkill.

pmlc/dialect/stripe/agginit_pass.cc
```diff
@@ -0,0 +1,239 @@
...
+
+// Find out where to insert the initialization of tensor
+std::pair<Operation*, Value*> AggregateInitializer::InitLocation(Value* tensor) {
+  Value* prev_tensor = nullptr;
+  Operation* last_refine = nullptr;
+  Value* cur_tensor = tensor;
+  while (cur_tensor) {
+    if (auto def_op = cur_tensor->getDefiningOp()) {
+      if (auto allocate_op = mlir::dyn_cast<AllocateOp>(def_op)) {
+        if (!last_refine) {
```
jbruestle: I think this code is not correct in the more general sense, but always works when the Stipe dialect code is generated from old stripe.  Specifically, there is no requirement of refinements on every ParallelFor, or at all even.  Let's say we have a tensor that is 1x1x1.  It's totally valid to do a 'load' directly from the allocate with no refinements, and refine(x, 0, 0, 0) might be removed by canonicalization at some point.

pmlc/dialect/stripe/agginit_pass.cc
```diff
@@ -0,0 +1,239 @@
+// Copyright 2019, Intel Corporation
```
jbruestle: Overall this pass is hard to follow.  I'm not sure if it's a design flaw in Stipe IR or something else, but I feel like this should be less complex.  Nothing needs to be done about this, but it's interesting to read this code.

pmlc/dialect/stripe/agginit_pass.cc
jbruestle: For now though, I think it's not a blocker.

PR #796, opened 2019-12-25
Experimental end to end MLIR execution

flaub:


plaidml2/exec/ffi.cc
```diff
@@ -63,15 +82,288 @@ class PlatformAllocator : public Allocator {
   std::string device_id_;
 };
```
jbruestle: I feel like some of this code belongs somewhere other than ffi.cc

flaub: Agreed. I’m currently working on a refactor and a way to configure passes based on the target. Was considering making a new dialect for pass configuration but after some experiments I’m currently thinking of sticking with protobuf.

flaub:
To run, use _edsl rules along with PLAIDML_EE=1 env var

flaub:
Try:
```
PLAIDML_EE=1 bazelisk run //plaidml2/edsl:cc_mlir_test -- --gtest_filter=CppEdsl.Dot
```

PR #797, opened 2019-12-27
Bug fixes

shiyao00:
* Fix partial-zero in padding pass
* Fix missing zero-init while changing large prime index in padding pass
* Fix missing or redundant index in tensors for autostencil pass
* Automatic hardware concurrency detection for autostencil pass


tile/targets/cpu/llvm_cpu.jsonnet
```diff
@@ -172,7 +171,7 @@ local PARAMS = {
             },
 
             // Init aggregation outputs
-            // Keet this towards the end since other passes are generating intermediate blocks and the initialization 
+            // Keet this towards the end since other passes are generating intermediate blocks and the initialization
```
flaub: Typo: Keet -> Keep?

shiyao00: This is the old wording. We will anyway replace this section soon with the new agg_init pass soon.

PR #798, opened 2019-12-27
Python EDSL test fixes and snippets for docs

1sanz:
working on adding edsl python tests for the code added in the doc
working on creating tabs


docs/docs/edsl_python.md
```diff
@@ -0,0 +1,707 @@
...
+sum over axis `0` of a 2D tensor (in Keras this would be `K.sum(I, axis=0)`):
+
+Python [ | C++ ]({{ site.baseurl }}{% link docs/edsl.md%}#sum-over-axis)
+```python
+def sum_over_axis(I):
+    M, N = TensorDims(2)
+    m, n = TensorIndexes(2)
+    I.bind_dims(M, N)
+    O = TensorOutput(N)
+    O(n) += I(m, n); // contraction
```
tzerrell: This line is still written in C++.

1sanz:
@flaub  I agree. Ideally there should be language tabs I am still trying to figure out a way to make that happen inside Jekyll. I have not had much luck so far

@tzerrell no I have not tested yet but I saw that there are some similar tests in plaidml2/edsl/edsl_test.py, I am adding tests there to check (locally) discussing in person would be helpful :)

flaub:
I can think of one fairly easy way to make the tabs work, use javascript to hook the link click so that it toggles the visibility of the various `<div>` elements for each language.

But, I think that the code snippets are only part of the problem. Some of the prose references C++ and so we'd either need to re-work those parts to be language agnostic or also put these sections under a per-language tab.

plaidml2/edsl/edsl_test.py
```diff
@@ -98,10 +105,338 @@ def arg_max(I):
...
+    B.bind_dims(K, J)
+    C = TensorOutput(I, J)
+    C[i, j] += A[i, k] * B[k, j]
+    return C
+
+def global_min(I):
+    i, j, k = TensorIndexes(3)
+    Neg = -I
+    O_Neg = TensorOutput()
+    O_Neg[()] >= Neg[i, j, k]
```
tzerrell: My memory might be bad, but I thought we had worked out that it was ok to do `O_Neg[]` instead of `O_Neg[()]`. If so, switch to the version without the extra parentheses (here and the other places it comes up).

flaub:
How about we leave out the python related docs for now, since we're going to move to the new sphinx based doc system that @dgkutnic is working on. Then let's see if we can't use the tabs that sphinx has to rewrite the EDSL article to have both C++ and python code snippets. I've verified that we can write math with latex using katex inside of sphinx.

We should also look into porting all the existing jekyll based markdown files over into sphinx. Sphinx does have a markdown mode, but it might be good to just move all the existing jekyll markdown into sphinx restructured text.

flaub:
Also, we can make use of sphinx's include directives so that we can leave the code snippets in an actual python file (like edsl_test.py) so that it is tested and also so that we don't have to duplicate code snippets into the docs.

plaidml2/edsl/edsl_test.py
```diff
@@ -98,10 +105,338 @@ def arg_max(I):
...
+    return O
 
 class TestEdsl(unittest.TestCase):
     maxDiff = None
 
+    def test_sum_over_axis(self):
+      I = Tensor(LogicalShape(plaidml.DType.FLOAT32, [1, 784]))
+      O = sum_over_axis(I)
+      program = Program('sum_over_axis', [O])
+     # print(str(program))
```
tzerrell: probably remove the various commented out `print(str(program))` lines

plaidml2/edsl/edsl_test.py
1sanz: I get a syntax error with O_Neg[] version so sticking with O_Neg[()] for now.

plaidml2/edsl/edsl_test.py
1sanz: oops. thanks!

plaidml2/edsl/edsl_test.py
```diff
@@ -50,6 +67,44 @@ def conv_2d(I, K):
...
+    Y0, Y1 = TensorDims(2)
+    Y0 = (X0 + s0 - 1) // s0
+    Y1 = (X1 + s1 - 1) // s1
+
+    #Compute the effective kernel size after dilation
+    EK0, EK1 = TensorDims(2)
+    EK0 = d0 * (K0 - 1) + 1
+    EK1 = d1 * (K1 - 1) + 1
+
+   #Compute the padding offset
```
flaub: Run `yapf` to fix formatting

plaidml2/edsl/edsl_test.py
```diff
@@ -435,6 +491,106 @@ def test_skip(self):
...
+''')
+
+    def test_complex_conv_2d(self):
+      I = Tensor(LogicalShape(plaidml.DType.FLOAT32, [1, 1, 1, 1, 1]))
+      K = Tensor(LogicalShape(plaidml.DType.FLOAT32, [1, 1, 1, 1, 1]))
+      O = complex_conv_2d(I,K,1,2,1,2)
+      program = Program('complex_conv_2d', [O])
+      if USE_MLIR():
+        self.assertMultiLineEqual(  
+                str(program), '''#map0 = (d0, d1, d2, d3, d4, d5, d6, d7) -> (d0, d1, d2, d3, d4)
```
flaub: Can we add a new line here but then call `trim` so that we aren't comparing against the extra newline?

plaidml2/edsl/edsl_test.py
```diff
@@ -40,6 +40,23 @@ def softmax(X):
...
+def conv_1d(I, K) :
+    N, X, KX, CI, CO = TensorDims(5)
+    n, x, k, ci, co = TensorIndexes(5)
+    I.bind_dims(N, X, CI)
+    K.bind_dims(KX, CI, CO)
+    O = TensorOutput(N, X - KX + 1, CO)
+    O[n, x, co] += I[n, x + k, ci] * K[k, ci, co]
+    return O
+
+def conv_2d_dialated(I, K) :
```
flaub: typo: `dilated`

1sanz: fixed.

plaidml2/edsl/edsl_test.py
1sanz: for all tests?

plaidml2/edsl/edsl_test.py
1sanz: I used yapf. It didn't delete the blank lines, are those the problem? They were for better readability in the document. I can remove them manually if needed.

plaidml2/edsl/edsl_test.py
1sanz: fixed. Let me know if this is what you meant.

plaidml2/edsl/edsl_test.py
```diff
@@ -51,6 +71,58 @@ def conv_2d(I, K):
...
 
 
+def complex_conv_2d(
+        I,
+        K,
+        s0,
+        s1,  # stride coeffs
+        d0,
+        d1  # dilation coeffs
+):
```
flaub: It doesn't look like yapf is being run because this would all get re-formatted.

plaidml2/edsl/edsl_test.py
```diff
@@ -99,9 +171,465 @@ def arg_max(I):
     return as_uint(O, 32)
 
 
+def sum_over_axis(I):
+    M, N = TensorDims(2)
+    m, n = TensorIndexes(2)
+    I.bind_dims(M, N)
+    O = TensorOutput(N)
+    O[n] += I[m, n]
+    # contraction
```
flaub: either remove or put on the previous line

plaidml2/edsl/edsl_test.py
```diff
@@ -99,9 +171,465 @@ def arg_max(I):
...
     maxDiff = None
 
+    def test_sum_over_axis(self):
+        I = Tensor(LogicalShape(plaidml.DType.FLOAT32, [1, 784]))
+        O = sum_over_axis(I)
+        program = Program('sum_over_axis', [O])
+        if USE_MLIR():
+            self.assertMultiLineEqual(
+                str(program), ('''
+
```
flaub: remove this newline

plaidml2/edsl/edsl_test.py
```diff
@@ -99,9 +171,465 @@ def arg_max(I):
...
+!fp32 = type tensor<!eltwise.fp32>
+module {
+  func @max_over_axis(%arg0: tensor<1x784x!eltwise.fp32>) -> tensor<784x!eltwise.fp32> {
+    %cst = "eltwise.sconst"() {value = 0.000000e+00 : f64} : () -> !fp32
+    %0 = tile.cion max, none, %cst, %arg0 {sink = #map0, srcs = [#map1]} : !fp32, tensor<1x784x!eltwise.fp32> -> tensor<784x!eltwise.fp32>
+    return %0 : tensor<784x!eltwise.fp32>
+  }
+}
+''').lstrip())
```
flaub: Why don't we just trim both lhs and rhs? You can make a helper function to make it nicer.

plaidml2/edsl/edsl_test.py
flaub: Yes, but it'd be nice to use a helper function so we don't to put `lstrip` everywhere. Also, there's an extra newline that we don't need.

PR #804, opened 2019-12-31
Add gradient support to MLIR version

tzerrell:
Add support for derivatives in MLIR PlaidML.

This is not quite complete (notably, it is missing simple_reduce), but it is sufficiently complete that no MLIR backend_tests fail that are not already failing on HEAD. Submitting as draft because it's known to be incomplete, but I can change that if we decide we want to land anyway.


flaub:
Awesome, let's get this in before the MLIR bump

PR #805, opened 2020-01-01
Move to MLIR included with llvm-project

flaub:
The biggest change here is adapting to `Value` having value semantics (as opposed to using `Value*` everywhere).

It appears that `bool operator<(Value lhs, Value rhs)` has broken behavior, so there are some hacks to allow `Value` to be a key (or part of a key) for `std::map`.

We should look into upstreaming a proper `bool Value::operator<(Value rhs)` impl.


flaub:
@mars1026 Any ideas why this might be failing on just a single test: `TestBackendOps.testSelfMult`?

The only thing I needed to change in the LLVM JIT was related to an API change that requires `llvm::MaybeAlign` instead of directly specifying alignment. That doesn't seem related to this, so I guess I'm worried that some deeper change has happened on upstream LLVM.

flaub:
Temporarily skipping the failing test only on `llvm_cpu` for now

PR #809, opened 2020-01-03
Sphinx Prototype, continued

dgkutnic:
This PR includes the following:
* adds `watch.py` to bazel as a `py_binary`
* sets up `autodoc` for python documentation through Sphinx
* prototypes `sphinx-tabs` for seamless C++/Python documentation

I'm planning a separate PR with content once we like the way that the infrastructure looks.


BUILD
```diff
@@ -47,3 +47,21 @@ toolchain(
     toolchain = ":py_runtime_pair",
     toolchain_type = "@rules_python//python:toolchain_type",
 )
+
+py_binary(
```
flaub: Now that we have this working in bazel, let's move the `watch.py` into the `docs` dir.

docs/conf.py
```diff
@@ -10,9 +10,9 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
+import os
+import sys
+sys.path.insert(0, os.path.abspath('..'))
```
flaub: I wonder why we need this? It's possible that if we move the `watch.py` into `docs` this might go away?

docs/manual_api/edsl.rst
```diff
@@ -6,15 +6,43 @@ Embedded Domain Specific Language
 
 .. contents::
 
-.. doxygenfunction:: plaidml::edsl::init
+-------
+Initialization
+-------
+
+.. tabs::
```
flaub: cool!

docs/conf.py
dgkutnic: This is a relic of my attempt to enable `autodoc` before I switched to the Bazel approach. It's not needed anymore, thanks for catching this

PR #811, opened 2020-01-06
Tile -> PXA conversion

flaub:


plaidml2/edsl/edsl_test.cc
```diff
@@ -756,6 +755,9 @@ module {
 }
 
 TEST(CppEdsl, GlobalMin) {
+  if (vertexai::env::Get("PLAIDML_EE") == "1") {
+    FAIL() << "Assertion failed: (map.getNumInputs() == mapOperands.size() && \"inconsistent index info\")";
```
jbruestle: Huh?

pmlc/conversion/tile_to_pxa/pass.cc
```diff
@@ -0,0 +1,553 @@
...
+struct CondOp {
+  Value create(ConversionPatternRewriter& rewriter, Location loc, Type resultType, ArrayRef<Value> operands) {
+    // TODO: add cast op for each operand wherever appropriate
+    CmpOpBuilder cmpOpBuilder;
+    auto cmp = cmpOpBuilder.create(rewriter, loc, resultType, operands.take_front(2));
+    auto zero = createInit(rewriter, loc, resultType, AggregationKind::add);
+    return rewriter.create<mlir::SelectOp>(loc, cmp, operands[2], zero).getResult();
+  }
+
+  Value createInit(OpBuilder& builder, Location loc, Type type, AggregationKind agg) const {
```
jbruestle: It feels like this should be useful other places and might want to be moved?

pmlc/dialect/pxa/to_affine.cc
```diff
@@ -100,25 +100,36 @@ struct AffineParallelForOpConversion : public LoweringBase<AffineParallelForOp>
...
+  }
+
+  Value createReduction(ConversionPatternRewriter& rewriter, AffineReduceOp op, Value source) const {
     switch (op.agg()) {
+      case AggregationKind::assign:
+        return source;
       case AggregationKind::add:
-        agg = rewriter.create<mlir::AddFOp>(op.getLoc(), in, op.val());
-        break;
+        return rewriter.create<mlir::AddFOp>(op.getLoc(), source, op.val());
```
jbruestle: Clearly we need to handle the integer cases here at some point...

plaidml2/edsl/edsl_test.cc
flaub: There area a few tests that cause this assertion to fail. Let's chat about it since it relates to how we translate AffinePolynomials into AffineExprs

PR #812, opened 2020-01-06
Respect dataflow deps in dead code elimination

earhart:


flaub:
@shiyao00 Want to take a look? I think you initially wrote this one?

PR #816, opened 2020-01-08
More work on PXA lowering

flaub:


pmlc/conversion/tile_to_pxa/pass.cc
```diff
@@ -182,6 +183,42 @@ struct ScalarConstantOpConversion : public OpConversionPattern<ew::ScalarConstan
...
+    return from;
+  }
+  if (auto intoFloatType = intoType.dyn_cast<FloatType>()) {
+    if (auto fromFloatType = fromType.dyn_cast<FloatType>()) {
+      if (fromFloatType.getWidth() < intoFloatType.getWidth()) {
+        // FPExtOp: FloatType -> wider FloatType
+        return rewriter.create<mlir::FPExtOp>(loc, from, intoType).getResult();
+      }
+      // FPTruncOp: FloatType -> narrower FloatType
+      return rewriter.create<mlir::FPTruncOp>(loc, from, intoType).getResult();
```
jbruestle: Technically there is an uncovered case here of fp16 to bf16, which I think we should ignore, but just as an FYI.

pmlc/conversion/tile_to_pxa/pass.cc
```diff
@@ -234,39 +304,51 @@ struct FirstOperand {
   }
 };
 
+void promoteTypes(ArrayRef<Value> operands, llvm::SmallVectorImpl<Value>* into) {
+  // TODO: add cast op for each operand wherever appropriate
```
jbruestle: I presume this code is meant to eventually compute the common subtype + then cast...

flaub: Yep, will do in a follow up

pmlc/conversion/tile_to_pxa/pass.cc
flaub: Also there's a missing case to go from integer to floating point.

PR #817, opened 2020-01-08
NFC: source tree reorganization

flaub:
The new tree looks like:

- pmlc
  - compiler
  - conversion
    - $src_to_tgt
      - tests
  - dialect
    - $dialect
      - analysis
      - ir
      - tests
      - transforms
  - target
    - $target
  - tools
    - pmlc-opt
    - pmlc-tblgen
    - pmlc-translate
  - util


earhart:
I'd refactor `dialect` a bit -- I think that in the current proposal, a dialect is modeling "a bag of stuff related to some particular dialect".  I've found it useful to think of the dialect directory as defining a datatype and low-level utilities on that datatype, and things like transforms and tests and analysis as orthogonal components operating on the various datatypes (especially when lowering from one dialect to another, or for any pass involving operations from multiple dialects).  I think that separation would help keep the system more composable, and make it easier to understand where various bits of code should go (and where one should go to find them).

pmlc/tools/pmlc-opt/BUILD
```diff
@@ -0,0 +1,28 @@
+# Copyright 2019 Intel Corporation.
+
+package(default_visibility = ["//visibility:public"])
+
+load("//bzl:plaidml.bzl", "plaidml_cc_binary")
+
+plaidml_cc_binary(
```
earhart: I think it would be useful to enable other compilers built on this technology.  For example, I think it would be useful to split `pmlc-opt` into two targets: a library containing all of these dependencies and the current `main()` entrypoint, and a trivial cc_binary that links them all together.  If `lit.bzl` and friends were also changed to take their `pmlc-opt` target as an input, that would make it easy for other compilers using these components to leverage the same lit test infrastructure, instead of having to roll their own.

(We could also do that in a separate PR -- if the basic idea sounds okay, LMK, and I'll go ahead and put it together after the current PR lands.)

tzerrell:
I like the general structure of this proposal.

It might be nice to make the various internal structures more parallel -- e.g. in `conversion`, `tile_to_pxa/tile_to_pxa.cc` vs `tile_to_stripe/lowering.cc`, or in `dialect`, where `pxa` is broken into `ir` and `transforms` while `stripe` is mostly top level files and `tile` is `ir`, `tests`, and `transforms`. That said, making all of these symmetric is a rather substantial batch of work and I'm inclined to think that even if we do want to do it that work would be better suited to a 2nd PR.

I also think that the separation @earhart mentioned is nice; IMO it would be best represented as `dialect/[whatever]/ir` and `dialect/[whatever]/transforms` (and more as appropriate), as it already is in the `pxa` and `tile` cases. To me that feels like something that belongs in whatever PR makes the internal structures more parallel.

jbruestle:
I think many transforms are dialect specific, and for those dialect/$dialect/transforms/ seems to match with MLIR's most recent organizational structure.  I do think more generic or multi-dialect transforms probably want to go elsewhere, but we can decide that when they arrive.  Agree with Tim on fixing Stripe in theory, but in practice we will probably drop it the StripeIR in the long run based on our recent discussions with Google, so no point to fixing it.

pmlc/tools/pmlc-opt/BUILD
flaub: For lit testing, you're going to need to make your own `lit.site.py` and whatnot to use your own compiler. I'm not sure what value we'd get in having a `pmlc-opt-lib`, seems like we can just have any other custom compilers bring in the pieces they need and write their own custom main as needed.

flaub:
@tzerrell  I decided not to reorg the `stripe` dialect because it's basically going away now. I wanted to get this skeleton done so that new passes, transforms, ops, types, and analysis get binned going forward, so I focused only on the dialects that will continue forward. Also I forgot to rename `lowering.cc`, I'll do that in a future PR.

flaub:
Not included in the proposal because we don't yet have any examples but totally fits is something like top-level dirs:
- analysis
- transforms

This is where generic or cross dialect things would go.

flaub:
I think generally most of our transforms will actually be dialect specific:

- tile
  - padding
  - bounds
- pxa/affine
  - autotile
  - fusion


Part of the motivation for this reorg is to align us with the upstream structure to ease the transition of the `pxa` prototype into upstream MLIR `affine` dialect.

pmlc/tools/pmlc-opt/BUILD
earhart: Yeah, that's basically what I did in another project -- I copied the PlaidML files and tweaked them to fit the other project's situation, but it seemed like suboptimal technical debt to be cleaned up at some point, not the right way forward.

flaub: Happy to see any proposals to improve the situation, I don't have a concrete feel for what change we need to make at this time.

PR #819, opened 2020-01-09
[tile->pxa] Lowering of heterogenous operand types

flaub:


pmlc/conversion/tile_to_pxa/tests/compare.mlir
```diff
@@ -0,0 +1,201 @@
...
+!u16 = type tensor<!eltwise.u16>
+!i32 = type tensor<!eltwise.i32>
+!bool = type tensor<!eltwise.bool>
+
+// CHECK-LABEL: func @cmp_u16_i32
+func @cmp_u16_i32(%u16: !u16) -> !bool {
+  %c0 = "eltwise.sconst"() {value = 0 : i64} : () -> !i32
+  %0 = "eltwise.cmp_lt"(%u16, %c0) {type = !eltwise.fp32} : (!u16, !i32) -> !bool
+  // CHECK: pxa.parallel_for
+  // CHECK: sexti
```
jbruestle: This should be definitely be a  zexti, if I do uint16_t(0x8000) < int32_t(0), it should return false.   In general, I think the extension follows the signedness of the input type.

pmlc/conversion/tile_to_pxa/tests/compare.mlir
```diff
@@ -0,0 +1,201 @@
...
+!i16 = type tensor<!eltwise.i16>
+!u32 = type tensor<!eltwise.u32>
+!bool = type tensor<!eltwise.bool>
+
+// CHECK-LABEL: func @cmp_i16_u32
+func @cmp_i16_u32(%i16: !i16) -> !bool {
+  %c0 = "eltwise.sconst"() {value = 0 : i64} : () -> !u32
+  %0 = "eltwise.cmp_lt"(%i16, %c0) {type = !eltwise.fp32} : (!i16, !u32) -> !bool
+  // CHECK: pxa.parallel_for
+  // CHECK: zexti
```
jbruestle: I think this should be sext, but I'm not sure.  Might we worth checking with via compilation.  Specifically, we could check if int16_t(-1) > uint32_t(0x100000).  If it's sign extension then upcast, we should go 0xffff -> 0xffffffff which is bigger than 0x100000 when compared as unsigned.  Otherwise, if would be 0xffff -> 0x0000ffff which would be smaller than 0x100000 when compared as unsigned.

jbruestle: Actually I got curious on this (and the comment later), so I just wrote some c test code:

#include <cstdint>
#include <stdio.h>

int main() {
  uint16_t a = 0xffff;
  int16_t b = -1;
  int32_t z = 0;
  uint32_t big = 0x100000;
  printf("Case 1: %d\n", a < z);
  printf("Case 2: %d\n", b > big);
}

The results are:

Case 1: 0
Case 2: 1

So definitely sext here.

PR #821, opened 2020-01-10
Drop obsolete toolchains

flaub:
Any reason that we still need these?


brianretford:
Not that I can think of.

PR #822, opened 2020-01-10
Fboemer/uint64

fboemer:
Fixes `FromMLIR` conversion to include type information


pmlc/dialect/stripe/from_mlir.cc
```diff
@@ -616,6 +616,14 @@ void StripeBuilder::visit(util::GenericBuilder builder) {
   for (auto operand : op->getOperands()) {
     intr->inputs.push_back(get_scalar(operand));
   }
+
+  auto result = op->getResult(0);
+  auto tensorType = eltwise::getRankedTensorType(result->getType());
+  auto scalarType = tensorType.getElementType().cast<eltwise::ScalarType>();
+
+  if (scalarType.type() != DataType::BOOLEAN) {
```
fboemer: Resolves issues in `CppEdsl.Relu` test

flaub: Actually, I think we need this: 

```
  if (intr->name == "cmp") {
    DataType superType = DataType::INVALID;
    for (auto operandType : op->getOperandTypes()) {
      auto tensorType = eltwise::getRankedTensorType(operandType);
      auto scalarType = tensorType.getElementType().cast<eltwise::ScalarType>();
      superType = CommonSupertype(superType, scalarType.type());
    }
    intr->type = superType;
  } else {
    auto result = op->getResult(0);
    auto tensorType = eltwise::getRankedTensorType(result->getType());
    auto scalarType = tensorType.getElementType().cast<eltwise::ScalarType>();
    intr->type = scalarType.type();
  }
```

LLITCHEV: Thanks @flaub!!
@fboemer Our cmp operation names are more involved - we explicitly call out the type of the operation in the operation name - cmp_eq, cmp_ne, cmp_gt, cmp_lt, cmp_ge, cmp_le.
So, I'd just change the first line of the code that @flaub so graciously supplied above to:
if (intr->name.rfind("cmp_", 0) == 0) {
I ran the tests with that code and things seems to be working fine.
Thanks for fixing this!

fboemer: Did something similar.

pmlc/dialect/stripe/from_mlir.cc
```diff
@@ -616,6 +616,23 @@ void StripeBuilder::visit(util::GenericBuilder builder) {
   for (auto operand : op->getOperands()) {
     intr->inputs.push_back(get_scalar(operand));
   }
+
+  static std::unordered_set<std::string> boolean_op_names{"cmp_eq", "cmp_ne", "cmp_gt", "cmp_lt", "cmp_ge", "cmp_le"};
+
+  if (boolean_op_names.find(intr->name) != boolean_op_names.end()) {
```
LLITCHEV: I would probably have these two lines replaced by:
if (intr->name.rfind("cmp_", 0) == 0) {
Just in case new cmp_ operations get added we won't need to revisit this code as well. It probably runs faster too - just comparison of up to 3 characters, instead of calculating hashes and doing a HT lookup.

PR #845, opened 2020-01-16
Various Documentation Improvements

dgkutnic:
* added some placeholders for core and exec docs
* fixed `docs/README` to reflect how to run `watch.py` with Bazel
* got rid of extraneous `#include <...>` directives in the documentation
* added a bunch of docstrings
* added a couple of missing ops to the python EDSL


docs/manual_api/core.rst
```diff
@@ -1,3 +1,43 @@
...
+
+.. tabs::
+
+   .. group-tab:: C++
+
+      .. doxygenfunction:: plaidml::init
+
+   .. group-tab:: Python
+
+      .. autofunction:: plaidml.core.__init
```
flaub: I think we should avoid adding private impl details into docs. (In python, anything that starts with `__` is private, and `_` is semi-private)

flaub: There isn't any correlated initialization in python, it happens when the module gets imported. We could mention that in this section, but not actually point to `__init` which is private an may change at any time

docs/manual_api/edsl.rst
```diff
@@ -34,22 +34,61 @@ Objects
...
+      .. autofunction:: plaidml.edsl.cosh
+      .. autofunction:: plaidml.edsl.exp
+      .. autofunction:: plaidml.edsl.gather
+      .. autofunction:: plaidml.edsl.ident
+      .. autofunction:: plaidml.edsl.index
+      .. autofunction:: plaidml.edsl.log
+      .. autofunction:: plaidml.edsl.pow
+      .. autofunction:: plaidml.edsl.prng
+      .. autofunction:: plaidml.edsl.reshape
+      .. autofunction:: plaidml.edsl.reshape
```
flaub: This is a dup

docs/manual_api/exec.rst
```diff
@@ -1,2 +1,41 @@
...
+
+.. tabs::
+
+   .. group-tab:: C++
+
+      .. doxygenfunction:: plaidml::exec::init
+
+   .. group-tab:: Python
+
+      .. autofunction:: plaidml.exec.__init
```
flaub: Ditto here as with the other sections

plaidml/edsl/__init__.py
```diff
@@ -27,6 +29,15 @@ class LogicalShape(ForeignObject):
     __ffi_repr__ = lib.plaidml_logical_shape_repr
 
     def __init__(self, dtype=None, dims=[], ptr=None):
+        """Initializes the LogicalShape.
+
+        Args:
+            self (pointer): The object pointer for a LogicalShape
+            dtype: Description of dtype
```
flaub: This is of `plaidml.DType` type

plaidml/exec/__init__.py
```diff
@@ -13,6 +13,9 @@
 
 
 def __init():
+    """
+    Initializes PlaidML's execution engine
```
flaub: I think we should say `Execution API`

plaidml/exec/exec.h
```diff
@@ -14,6 +14,9 @@
 namespace plaidml {
 namespace exec {
 
+///
+/// Initializes PlaidML's execution engine
```
flaub: Execution API

docs/manual_api/edsl.rst
```diff
@@ -34,22 +34,61 @@ Objects
...
+      .. doxygengroup:: edsl_primitives
+         :content-only:
+
+   .. group-tab:: Python
+
+      .. autofunction:: plaidml.edsl.abs
+      .. autofunction:: plaidml.edsl.as_float
+      .. autofunction:: plaidml.edsl.as_int
+      .. autofunction:: plaidml.edsl.as_uint
+      .. autofunction:: plaidml.edsl.as_bool
```
flaub: I'd like to deprecate these older `as_XXX` casting methods with a `cast` method. For instance, we have no good way to cast between a `BF16` and an `FP32` using `as_float()`

docs/manual_api/core.rst
```diff
@@ -18,7 +18,11 @@ Initialization
 
    .. group-tab:: Python
 
-      .. autofunction:: plaidml.core.__init
+      .. note::
+
+         Initialization in Python follows standard conventions: the PlaidML Core
+         APIs are initialized automatically wherever the namespace
```
flaub: In python, namespace isn't a thing, but we could say `module` here instead.

flaub: I also wouldn't say there's a 'standard' convention for initialization.

plaidml/edsl/__init__.py
```diff
@@ -15,7 +15,7 @@
 
 def __init():
     """
-    Initializes PlaidML's EDSL.
+    Initializes PlaidML's EDSL APIs.
```
flaub: Singular `API`, the EDSL is one API, Execution is another one, etc...

PR #852, opened 2020-01-17
Use a TypeSwitch

flaub:


pmlc/dialect/tile/ir/ops.cc
```diff
@@ -125,39 +126,47 @@ OpFoldResult DimOp::fold(ArrayRef<Attribute> operands) {
...
+        })
+        .template Case<AffineMulOp>([this](auto op) {
+          visitOperands(op.lhs(), op.rhs());
+          return static_cast<T*>(this)->visitMulOp(op);
+        })
+        .template Case<AffineDivOp>([this](auto op) {
+          visitOperands(op.lhs(), op.rhs());
+          return static_cast<T*>(this)->visitDivOp(op);
+        })
+        .template Case<AffineSubOp>([this](auto op) {
```
earhart: Just curious -- why'd you swap sub and neg?  (Otherwise, LGTM -- and thanks for pointing out this functionality!)

flaub: Is this swapped? Looks ok to me?

earhart: I think the code is fine -- it's just that the two cases appear swapped compared to their ordering in the original code; I'm guessing it doesn't matter, but in case it does, I wanted to mention it (and maybe learn more about TypeSwitch).

flaub: Oh, I was simply following the order of the visitors listed for reference, I can swap it if you think this is a problem.

flaub: Could you approve this then?

PR #853, opened 2020-01-17
Primitive Op Docs

dgkutnic:
* added documentation for some python primitives
* standardized all `\ingroup` directives to be on top of the `\class` or `\struct` directive for cleanliness
* added some primitive ops that didn't exist in c++/python
    * open question: do we want to add `gradients`, `min`, and `max` to C++ too? Not sure if those were intended to be missing from the C++ API. I can do that in a separate PR.


flaub:
Yes, the API between python and C++ should be functionally identical.

plaidml/edsl/__init__.py
```diff
@@ -669,17 +669,42 @@ def call(fn, *args):
...
 
 def abs(x):
-    """A placeholder docstring."""
+    """Computes the elementwise absolute value of ``x``.
+
+        Args:
+            x (Tensor): The tensor used to peform the elementwise ``abs``.
+
+        Returns:
+            y (Tensor): The result of the elementwise ``abs`` operation.
```
flaub: Are we required to name the output? I'd rather not if we don't have to

plaidml/edsl/__init__.py
```diff
@@ -723,57 +796,161 @@ def gradients(loss, variables):
...
     return call('prng', state, *shape)
 
 
 def reshape(x, dims):
-    """A placeholder docstring."""
+    """Takes a tensor ``x`` and reshapes it according to ``dims``.
+
+        Args:
+            x (Tensor): The tensor to reshape.
+            dims (list): The desired shape of the tensor.
```
flaub: Is it possible to say what kind of list we support? BTW I think we also support numpy arrays here @tzerrell is that true?

plaidml/edsl/__init__.py
dgkutnic: Just checked - we are _not_ required to name outputs. I was just doing it because the template I was using had named the outputs.

I'll fix this in my next go-around. I didn't like the inconsistency of it either...

plaidml/edsl/__init__.py
dgkutnic: I believe it is possible to specify multiple types. I wasn't sure exactly which types of list we supported here, so one of us can add them if we have a full list of them.

PR #855, opened 2020-01-17
C++ Execution API Documentation

dgkutnic:


plaidml/exec/exec.h
```diff
@@ -59,6 +71,12 @@ inline std::vector<std::string> list_devices() {
   return ret;
 }
 
+///
+/// \ingroup exec_functions
+/// Lists the available targets.
+/// Use this list of targets to set the environment variable `PLAIDML_TARGET`.
```
flaub: This list isn't really related to env vars itself. This list simply represents the targets that are available. I don't think I'd mention anything related to `PLAIDML_TARGET` here, that probably belongs in a usage/configuration section, it's not really part of the API itself.

plaidml/exec/exec.h
```diff
@@ -135,6 +162,11 @@ class Binder {
   using BindingMap = std::map<edsl::TensorRef, Buffer>;
 
  public:
+  ///
+  /// Constructs a Binder.
+  /// By default, this constructor uses the environment variables `PLAIDML_DEVICE` and `PLAIDML_TARGET` to specify your
```
flaub: This constructor doesn't really depend on env vars. It's more like it depends on those settings to have been configured. The default values for those settings default to the env var values.

PR #859, opened 2020-01-17
Added right shift for signed integers

fboemer:
I'm happy to add the expected program strings as well, but I'm not sure how to generate them (printing the program locally seems doesn't have the `idxs, src, maps` info).


pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -811,8 +831,10 @@ struct LoweringPass : public mlir::ModulePass<LoweringPass> {
         EltwiseOpConversion<ew::BitOrOp, StdOp<mlir::OrOp>, OperandsAre<EltwiseInteger>>,
         EltwiseOpConversion<ew::BitXorOp, StdOp<mlir::XOrOp>, OperandsAre<EltwiseInteger>>,
         EltwiseOpConversion<ew::BitShlOp, StdOp<mlir::ShiftLeftOp>, OperandsAre<EltwiseInteger>>,
-        EltwiseOpConversion<ew::BitShrOp, StdOp<mlir::SignedShiftRightOp>, OperandsAre<EltwiseSigned>>,
-        EltwiseOpConversion<ew::BitShrOp, StdOp<mlir::UnsignedShiftRightOp>, OperandsAre<EltwiseUnsigned>>,
+        EltwiseOpConversion<ew::BitShrOp, StdOp<mlir::SignedShiftRightOp>,
+                            FirstOperandIsAndRestAre<EltwiseSigned, EltwiseInteger>>,
+        EltwiseOpConversion<ew::BitShrOp, StdOp<mlir::UnsignedShiftRightOp>,
+                            FirstOperandIsAndRestAre<EltwiseUnsigned, EltwiseInteger>>,
```
flaub: Actually, it's only the first operand that we care about, the rest should always be unsigned; we should update the eltwise `ops.td` to reflect that. Once that's in, the predicate here would only be `FirstOperandIs<xxx>`

plaidml/edsl/edsl_test.cc
```diff
@@ -143,9 +143,9 @@ TEST(CppEdsl, BitLeft) {
   }
 }
 
-TEST(CppEdsl, BitRightSignedScalarSigned) {
+TEST(CppEdsl, BitRightSignedScalar) {
   auto A = Placeholder(DType::INT64, {3, 3});
-  auto C = A >> 1;
+  auto C = A >> std::uint64_t{1};
```
flaub: I actually think we should let users use integers directly, but then at a lower level prevent non-negative integers from being specified.

pmlc/dialect/eltwise/ir/ops.td
```diff
@@ -143,6 +148,12 @@ class EW_BinaryIntOp<string mnemonic, list<OpTrait> traits = [NoSideEffect]> :
     Arguments<(ins EltwiseInteger:$lhs, EltwiseInteger:$rhs)>,
     Results<(outs EltwiseInteger:$result)>;
 
+class EWBinaryIntUnsignedOp<string mnemonic, list<OpTrait> traits = [NoSideEffect]> :
```
flaub: Could you use the `EW_` prefix?

plaidml/edsl/edsl_test.cc
flaub: @jbruestle Do you have any thoughts on this? I was hoping we could have only 2 constant types: `double` and `int64_t`, but really these ought to be arbitrary precision types.

pmlc/dialect/tile/builder.cc
```diff
@@ -145,9 +145,16 @@ struct TileBuilder::Impl {
   }
 
   Value makeScalarConstantOp(double value) {
+    IVLOG(5, "makeScalarConstantOp double " << value);
     auto type = builder.getType<ScalarType>(DataType::f32);
```
flaub: Might as well make this a `f64`

pmlc/dialect/eltwise/ir/predicates.td
```diff
@@ -31,7 +31,7 @@ def ScalarBool : Type<And<[
 
 def ScalarIndex : Type<And<[
     IsScalarPred,
-    CPred<"$_self.cast<ScalarType>().type() == DataType::i32">
+    CPred<"$_self.cast<ScalarType>().type() == DataType::i32 || $_self.cast<ScalarType>().type() == DataType::i64">
```
flaub: We have been considering adding a new `DataType` enum value to represent machine-word sized integers (e.g. indexes). I'm OK with doing it like this for now, but I expect there to be some churn on this as we start getting lowering working on different targets.

pmlc/dialect/tile/ir/ops.td
```diff
@@ -315,8 +315,8 @@ def GatherOp : SpecialOp<"gather">, HasCanonicalizer {
 
 def PrngOp : SpecialOp<"prng", []>, HasCanonicalizer {
   let summary = "pseudorandom number generator";
-  let arguments = (ins RankedTensorOf<[ScalarUINT32]>:$state, Variadic<EltwiseIndex>:$dims);
-  let results = (outs RankedTensorOf<[AnyScalarFloat]>:$result, RankedTensorOf<[ScalarUINT32]>:$new_state);
+  let arguments = (ins RankedTensorOf<[AnyScalarUnsigned]>:$state, Variadic<EltwiseIndex>:$dims);
+  let results = (outs RankedTensorOf<[AnyScalarFloat]>:$result, RankedTensorOf<[AnyScalarUnsigned]>:$new_state);
```
flaub: I think this needs to remain a 32-bit value. The lowering for PRNG hasn't been done yet, but the impl in plaidml-v0 was for the state to be UINT32, it's not really up to the user, this is like opaque data that the user shouldn't be concerned with. If there was a make it completely hidden that's what we'd probably want to do.

plaidml/core/internal.h
```diff
@@ -49,7 +49,8 @@ using Variant = std::variant<  //
     std::monostate,            // PLAIDML_VALUE_NONE
     plaidml_dim_expr,          // PLAIDML_VALUE_DIM
     plaidml_expr,              // PLAIDML_VALUE_EXPR
-    double,                    // PLAIDML_VALUE_FLOAT
+    double,                    // PLAIDML_VALUE_DOUBLE
+    float,                     // PLAIDML_VALUE_FLOAT
```
flaub: These are meant to be "arbitrary precision constant values", so I don't think we should make a version for each bitwidth.

fboemer:
> I think we should put a pause on this PR until we have a design review. I'm not comfortable making these changes as is.

Fair enough. Thanks for the help so far!

fboemer:
Closed in favor of https://github.com/plaidml/plaidml/pull/897/

PR #863, opened 2020-01-19
Add missing affine.if for constraints

flaub:


flaub:
Yes, I've been chatting with the OrcV2 maintainer (which is what the MLIR EE uses) and this appears to be a known issue: https://bugs.llvm.org/show_bug.cgi?id=40074

PR #866, opened 2020-01-20
Treat boolean as u1 instead of i1

flaub:


earhart:
@flaub FWIW -- one of the tricky bits with the nGraph integration was that they treated booleans as just another integer type, and then tested to see whether various operations produced bit-correct values, which didn't work so great with OpenCL (where a non-vector compare might produce 1 and a vector compare might produce -1).

IMHO it'd be better to have an actual bool type at the eltwise layer, and let lower dialects do what they want, with explicit translations when an upper layer wants to see a precise bit pattern.

flaub:
This is all going to be replaced with std types as soon as this lands: https://reviews.llvm.org/D72533

The motivation for this change is to align with this upcoming change upstream. I also think that lowering to SPIR-V would be the appropriate place to deal with difference in semantics at the `std` and `SPIR-V` (i.e. OpenCL) levels.

earhart:
👍

PR #868, opened 2020-01-21
walkaround msvc bug

cloudhan:
I (maybe) encountered a bug of msvc when building on Windows, which is described in
https://github.com/NervanaSystems/ngraph/issues/4125 and its corresponding stackoverflow question.

Here is a simple walkaround. By removing the templated default arguments, I can successfully build plaidml and ngraph (both with hacks to the build system).


plaidml/plaidml++.h
```diff
@@ -169,9 +169,13 @@ class base_shape {
 template <typename T>
 class shape : public base_shape {
  public:
-  explicit shape(const std::shared_ptr<ctx>& ctx, datatype dt = to_plaidml_datatype<T>::value) : base_shape(ctx, dt) {}
+  explicit shape(const std::shared_ptr<ctx>& ctx) : shape(ctx, to_plaidml_datatype<T>::value) {}
 
-  explicit shape(const base_shape& base, datatype dt = to_plaidml_datatype<T>::value) : base_shape(base) {
+  explicit shape(const std::shared_ptr<ctx>& ctx, datatype dt) : base_shape(ctx, dt) {}
```
flaub: `explicit` is not useful for constructors with multiple arguments.

plaidml/plaidml++.h
```diff
@@ -169,9 +169,13 @@ class base_shape {
...
+
+  explicit shape(const base_shape& base) : shape(base, to_plaidml_datatype<T>::value) {}
+
+  explicit shape(const base_shape& base, datatype dt) : base_shape(base) {
```
flaub: Same as above, omit `explicit`

plaidml/plaidml++.h
cloudhan: It is useful when you want to disallow user from implicitly using initializer list. say you have a function `void blah(const shape& s) { /*whatever*/ }`. Without `explicit`, `blah({ctx, dt})` will be valid.

flaub:
@earhart Could you review this one?

cloudhan:
> LGTM once the `explicit` on the multi-param constructors is removed.

I am confused now, since @flaub resolved the first review, in which I replied
> It is useful when you want to disallow user from implicitly using initializer list. say you have a function `void blah(const shape& s) { /*whatever*/ }`. Without explicit, `blah({ctx, dt})` will be valid.

Should I take it as `explicit` is necessary and keep it as is (even if in the second review), or, `explicit` is useless only in the second review and remove it ?

earhart:
Hmm.  IMHO, the important thing is to make sure `explicit` is there for single-argument constructors (unless you want the conversion).  I don't think I've seen problems with people using initializer lists with multi-argument constructors, so I tend to omit `explicit` on multi-argument constructors, which is allowed/encouraged by the Google coding standards we're loosely following in this repository -- see https://google.github.io/styleguide/cppguide.html#Implicit_Conversions.

So I'd tend to remove it, if only for consistency, but I don't have a strong feeling about it.  @flaub, thoughts?

flaub:
As long as there aren't any issues with the API itself, I don't mind much either way.

PR #869, opened 2020-01-21
Attempting to get resnet50 to run to completion

flaub:
* Use a forward (top-to-bottom) pass for initial canonicalization when creating a program (turning `tile.sym_cion` into `tile.cion` and computing static shapes).
* Rename `Tensor::shape()` to `Tensor::compute_shape()` and replace most usages of `tensor.shape()` with new APIs:
  * Tensor::dtype()
  * Tensor::rank()
* Make it easier to use the resnet50 EDSL model in different situations:
  * benchmarking
  * running from command line
* Add a new TraceOp to the EDSL and tile dialects. Still WIP, starting lowering into a `std.call` but still need to complete stub generation during LLVM dialect lowering.

### DIFF
--- a/plaidml/edsl/edsl.h
+++ b/plaidml/edsl/edsl.h
@@ -1054,6 +1054,14 @@ inline Tensor tan(const Tensor& x) { return Call("tan", x); }
 inline Tensor tanh(const Tensor& x) { return Call("tanh", x); }
 
 ///
+/// Adds a tracepoint to the graph
+///
+inline Tensor trace(const Tensor& x, const std::string& msg) {
+  auto ptr = ffi::call<plaidml_expr*>(plaidml_expr_trace, x.as_ptr(), msg.c_str());
+  return Tensor{ptr};
+}
+
+///
 /// Returns a Tensor with a value of 0.
 /// \return Tensor
 ///
--- a/plaidml/edsl/edsl.h
+++ b/plaidml/edsl/edsl.h
@@ -90,7 +90,7 @@ class Program {
   ///
   /// Return the Program as a string.
   ///
-  std::string str() const { return ffi::str(ffi::call<plaidml_string*>(plaidml_program_repr, ptr_.get())); }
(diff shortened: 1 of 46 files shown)


plaidml/edsl/edsl.h
```diff
@@ -657,10 +657,19 @@ class Tensor {
   ///
   /// Return the tensor's shape
   ///
-  LogicalShape shape() const {
+  LogicalShape compute_shape() const {
     return LogicalShape(ffi::call<plaidml_logical_shape*>(plaidml_expr_get_shape, as_ptr()));
   }
 
+  DType dtype() const {  //
```
tzerrell: I'm excited about these `dtype` and `rank` functions!

plaidml/edsl/edsl_test.cc
```diff
@@ -578,25 +578,25 @@ TEST(CppEdsl, LarsMomentum4d) {
 !f32 = type tensor<!eltwise.f32>
 module {
   func @lars_momentum4d(%arg0: tensor<4x7x3x9x!eltwise.f32>, %arg1: tensor<4x7x3x9x!eltwise.f32>, %arg2: !f32, %arg3: tensor<4x7x3x9x!eltwise.f32>) -> (tensor<4x7x3x9x!eltwise.f32>, tensor<4x7x3x9x!eltwise.f32>) {
-    %cst = "eltwise.sconst"() {value = 0.000000e+00 : f64} : () -> !f32
```
tzerrell: In the long run, should probably fix this test to not expect a specific order of constants. But I don't think that's for this PR

plaidml/edsl/edsl.h
flaub: Ya, they basically make constructing resnet50 very fast, instead of always recomputing a shape we never ended up needing.

plaidml/edsl/edsl_test.cc
flaub: Yes, I'd like to fix this at some point...

If we could use lit for these tests that'd be ideal

PR #873, opened 2020-01-22
minor doc fixes

1sanz:


docs/usage/edsl.rst
```diff
@@ -1,20 +1,20 @@
 Tile eDSL 
 #############
-The C++ Tile eDSL (Embedded Domain Specific Language) provides developers with a
+The Tile eDSL (Embedded Domain Specific Language) provides developers with a
```
dgkutnic: Not sure this is breaking, but `eDSL` should be `EDSL`

docs/usage/edsl.rst
```diff
@@ -107,7 +107,6 @@ code. Some portions of the notation do not perfectly correspond. Here's why:
   would result in a `0` as the last element of `O` if we're still assuming `N`
   is the size of the last dimension of `I`.
 
-- As is the case for all C++ statements, they must end with a semicolon.
```
dgkutnic: Perhaps replace this with something like:

> EDSL code should follow the syntactical rules for the language it is written in (e.g. for the C++ EDSL, make sure that your statements end with a semicolon, and for the Python EDSL, use the proper indentation for each statement).

docs/usage/edsl.rst
```diff
@@ -169,7 +168,7 @@ detail:
 
   C[i, j] = \sum_{k} (A[i, k] \cdot B[k, j])
 
-We can convert this to C++ Tile code using the same correspondence as the
+We can convert this to Tile code using the same correspondence as the
 previous example: The summation sign becomes plus-assignment, the summation
 index is omitted, dimensions are given for the output tensor, and the statement
```
dgkutnic: statement doesn't necessarily end in a semicolon if it is written in python

flaub: I'm sorta thinking this whole section ought to be tabbed so that there's a discussion about the differences between the mathematical representation and the form it takes in a target language.

PR #876, opened 2020-01-24
Arbitrary precision constants

fboemer:
Added `floatx, intx` arguments to Program creation API. These default to the highest available precision on the host language (currently float64 / int64) on python/C++.  A tile `ConstantTypesPass` is used to lower the precision of the constants if a lower precision is selected.


plaidml/core/__init__.py
```diff
@@ -164,7 +164,8 @@ def sizes(self):
     @property
     def strides(self):
         return [
-            ffi_call(lib.plaidml_shape_get_dim_stride, self.as_ptr(), i) for i in range(self.ndims)
+            ffi_call(lib.plaidml_shape_get_dim_stride, self.as_ptr(), i)
+            for i in range(self.ndims)
```
flaub: I'm wondering how these changes got in, `yapf` should be making this all uniform and I don't see any other change other than style here.

plaidml/edsl/edsl.h
```diff
@@ -107,13 +113,43 @@ class Program {
...
+
+  void set_intx(const DType& intx) {
+    if (intx != DType::INT32 && intx != DType::INT64) {
+      std::stringstream ss;
+      ss << "Invalid intx_ requested by Program. Expected either int32 or int64";
+      throw std::runtime_error(ss.str());
+    }
+    intx_ = intx;
+  }
+
```
flaub: I was thinking that `floatx` and `intx` would be part of the built `Program` state and that we would avoid allowing mutating these parameters after the `Program` was constructed. What use case do you have in mind where you want to change them after the `Program` was constructed? Also, wouldn't we need to re-compile if these values were changed? Currently the compilation happens in the `Program` constructor.

plaidml/edsl/edsl.h
```diff
@@ -81,10 +82,15 @@ class Program {
...
   ///
   /// Program constructor
-  ///
-  Program(                                 //
-      const std::string& name,             //
-      const std::vector<Tensor>& outputs,  //
+  /// floatx is used to specify the underlying datatype of floating-point constants
+  /// and should be either DType::FLOAT32 or DType::FLOAT64.
+  /// intx is used to specify the underlying datatype of integer constants
+  /// and should be either DType::INT32 or DType::INT64.
```
flaub: I think we should allow any floating point type for `floatx` and any integer type for `intx`.

plaidml/edsl/edsl_test.cc
```diff
@@ -46,6 +46,44 @@ Tensor Softmax(const Tensor& X) {
...
+
+!i32 = type tensor<!eltwise.i32>
+!f32 = type tensor<!eltwise.f32>
+module {
+  func @lower_precision(%arg0: tensor<3x3x!eltwise.f64>) -> tensor<3x3x!eltwise.f64> {
+    %c1 = "eltwise.sconst"() {value = 1 : i64} : () -> !i32
+    %cst = "eltwise.sconst"() {value = 2.000000e+00 : f64} : () -> !f32
+    %0 = "eltwise.add"(%arg0, %c1) : (tensor<3x3x!eltwise.f64>, !i32) -> tensor<3x3x!eltwise.f64>
+    %1 = "eltwise.add"(%0, %cst) : (tensor<3x3x!eltwise.f64>, !f32) -> tensor<3x3x!eltwise.f64>
+    return %1 : tensor<3x3x!eltwise.f64>
```
flaub: This seems like `floatx` isn't working properly here, wouldn't we expect to see a `f32` return type?

plaidml/edsl/edsl_test.cc
```diff
@@ -46,6 +46,44 @@ Tensor Softmax(const Tensor& X) {
   return E / N;
 }
 
+TEST(CppEdsl, LowerPrecision) {
+  auto A = Placeholder(DType::FLOAT64, {3, 3});
+  auto C = A + 1 + 2.0;
+
+  Program program("lower_precision", {C}, DType::FLOAT32, DType::INT32);
+
+  // consatnt_types pass doesn't lower precision to float32/int32 until program is compiled
```
flaub: typo: "constant_types"

But I think we should make the constant types pass happen during the construction of a program, not afterwards.

plaidml/edsl/edsl_test.py
```diff
@@ -104,8 +104,8 @@ def complex_conv_2d(
     O = TensorOutput(N, Y0, Y1, G, GCO)
 
     # Compute the convolution
-    O[n, x0, x1, g, gco] += I[n, s0 * x1 + d0 * k0 - P0, s1 * x1 + d1 * k1 -
-                              P1, g, gci] * K[k0, k1, g, gci, gco]
+    O[n, x0, x1, g, gco] += I[n, s0 * x1 + d0 * k0 - P0, s1 * x1 + d1 * k1 - P1, g, gci] * K[
+        k0, k1, g, gci, gco]
```
flaub: Why is `yapf` doing this?

plaidml/edsl/edsl_test.py
```diff
@@ -258,6 +258,31 @@ class TestEdsl(unittest.TestCase):
...
+
+!i32 = type tensor<!eltwise.i32>
+!f32 = type tensor<!eltwise.f32>
+module {
+  func @lower_precision(%arg0: tensor<3x3x!eltwise.f64>) -> tensor<3x3x!eltwise.f64> {
+    %c1 = "eltwise.sconst"() {value = 1 : i64} : () -> !i32
+    %cst = "eltwise.sconst"() {value = 2.000000e+00 : f64} : () -> !f32
+    %0 = "eltwise.add"(%arg0, %c1) : (tensor<3x3x!eltwise.f64>, !i32) -> tensor<3x3x!eltwise.f64>
+    %1 = "eltwise.add"(%0, %cst) : (tensor<3x3x!eltwise.f64>, !f32) -> tensor<3x3x!eltwise.f64>
+    return %1 : tensor<3x3x!eltwise.f64>
```
flaub: Again, I would expect this to return `f32`

plaidml/edsl/edsl_test.py
```diff
@@ -872,8 +897,8 @@ def test_defract_long(self):
         K = Tensor(LogicalShape(plaidml.DType.FLOAT32, shape), name='K')
         n, x0, x1, c0, c1, co, ci, k0, k1 = TensorIndexes(9)
         O = TensorOutput(1, 5, 5, 1)
-        O[n, x0, x1, co] += (I[n, (x0 + k0 - 1) // 2,
-                               (x1 + k1 - 1) // 2, ci] * K[2 - k0, 2 - k1, co, ci])
+        O[n, x0, x1, co] += (I[n, (x0 + k0 - 1) // 2, (
+            x1 + k1 - 1) // 2, ci] * K[2 - k0, 2 - k1, co, ci])
```
flaub: `yapf`

plaidml/edsl/ffi.cc
```diff
@@ -987,6 +989,17 @@ plaidml_program* plaidml_program_evaluate(  //
         args[i].buffer = nullptr;
       }
     }
+
+    auto context = ret->program->module->getContext();
+    ret->program->dtypes.floatx =
+        pmlc::dialect::eltwise::ScalarType::get(context, static_cast<pmlc::util::DataType>(floatx))
+            .toStandard()
```
flaub: We should avoid using standard types at this stage. At this stage, we should use the `ScalarType`.

It's only later when we lower from `tile` to `std` would we want to transition to using standard types.

plaidml/exec/ffi.cc
```diff
@@ -125,15 +127,19 @@ plaidml_executable* plaidml_compile(  //
...
     auto exec = std::make_unique<plaidml_executable>();
     std::vector<void*> bufptrs(args.size());
     for (unsigned i = 0; i < args.size(); i++) {
       auto view = args[i].buffer->MapCurrent();
       bufptrs[i] = view->data();
     }
-    exec->exec = std::make_unique<Executable>(program->program->entry, target, *program->program->module, bufptrs);
+
+    exec->exec = std::make_unique<Executable>(program->program->entry, target, *program->program->module, bufptrs,
+                                              program->program->dtypes.floatx, program->program->dtypes.intx);
```
flaub: I don't think the execution API should have anything to do with arbitrary precision constants. I think we should handle this all during program construction, before we go to compile down to a particular target.

pmlc/compiler/compiler.cc
```diff
@@ -122,6 +123,10 @@ Executable::Executable(StringRef entry, StringRef target, ModuleOp programModule
   manager.addNestedPass<FuncOp>(createCanonicalizerPass());
   manager.addNestedPass<FuncOp>(createCSEPass());
 
+  manager.addPass(dialect::tile::createConstantTypesPass(floatx, intx));
+  manager.addNestedPass<FuncOp>(createCanonicalizerPass());
+  manager.addNestedPass<FuncOp>(createCSEPass());
```
flaub: I think this belongs in the `builder.cc`, take a look at the `TileBuilder::MakeProgram` method, here is where takes a set of outputs to start with and traces down all the expressions that those outputs depend on. Then, it runs a pipeline to canonicalize, but just before canonicalization (which is presently where type inference is occuring), we'd want to update the types on the constant ops.

pmlc/dialect/eltwise/ir/ops.td
```diff
@@ -56,6 +56,7 @@ def ScalarConstantOp :
 
   let extraClassDeclaration = [{
     Attribute getValue() { return getAttr("value"); }
+    void setValue(Attribute attr) {setAttr("value", attr); }
```
flaub: I don't think we need to modify the value, we only need to modify the result type. The value itself can have whatever physical precision it has. But the logical type should follow `floatx` or `intx`. All constant ops basically have 2 types: physical (this is the type on the attribute) and a logical type (which type is returned from this op).

pmlc/dialect/tile/program.h
```diff
@@ -19,11 +20,17 @@ struct ProgramArgument {
   pmlc::util::BufferPtr buffer;
 };
 
+struct ProgramDType {
+  mlir::FloatType floatx;
+  mlir::IntegerType intx;
+};
+
```
flaub: If we kept the constant materialization pass local to the builder, we wouldn't need to expose this outwards to the later compiler/execution API.

pmlc/dialect/tile/transforms/constant_types.cc
```diff
@@ -0,0 +1,61 @@
...
+      signalPassFailure();
+    }
+  });
+}
+
+std::unique_ptr<mlir::OpPassBase<mlir::FuncOp>> createConstantTypesPass(mlir::FloatType floatx,
+                                                                        mlir::IntegerType intx) {
+  auto pass = std::make_unique<ConstantTypesPass>();
+  pass->floatx_ = floatx;
+  pass->intx_ = intx;
```
flaub: You could add a constructor to `ConstantTypesPass` that takes the `floatx` and `intx` parameters. Then you could say:

```
return std::make_unique<ConstantTypesPass>(floatx, intx);
```

pmlc/dialect/tile/transforms/constant_types.h
```diff
@@ -0,0 +1,19 @@
...
+#include "pmlc/dialect/eltwise/ir/ops.h"
+#include "pmlc/dialect/tile/ir/ops.h"
+
+namespace pmlc::dialect::tile {
+
+struct ConstantTypesPass : public mlir::FunctionPass<ConstantTypesPass> {
+  void runOnFunction() final;
+
+  mlir::FloatType floatx_;
+  mlir::IntegerType intx_;
```
flaub: We should avoid the use of standard types here and only use either `ScalarType` or even the `DataType` enum here instead.

pmlc/dialect/tile/transforms/registration.cc
```diff
@@ -1,13 +1,18 @@
...
-static mlir::PassRegistration<ComputeBoundsPass> pass(  //
-    "tile-compute-bounds",                              //
+static mlir::PassRegistration<ComputeBoundsPass> compute_bounds_pass(  //
+    "tile-compute-bounds",                                             //
     "Compute bounds for contractions");
 
+static mlir::PassRegistration<ConstantTypesPass> constant_types_pass(  //
+    "tile-constant-types",                                             //
+    "Set constants to specified types");
+
```
flaub: We'll need to support the ability to pass options for the `floatx` and `intx` values so that we can use this from the `pmlc-opt` command line tool.

plaidml/core/__init__.py
fboemer: resolved using correct yapf version

plaidml/edsl/edsl.h
fboemer: Removed these functions.

plaidml/edsl/edsl.h
fboemer: done

plaidml/edsl/edsl_test.cc
fboemer: done

plaidml/edsl/edsl_test.cc
fboemer: correct; fixed

plaidml/edsl/edsl_test.py
fboemer: resolved using correct yapf version

plaidml/edsl/edsl_test.py
fboemer: done

plaidml/exec/ffi.cc
fboemer: Agreed; removed these changes

pmlc/compiler/compiler.cc
fboemer: removed these changes

pmlc/dialect/tile/transforms/registration.cc
fboemer: Added `tile-constant-types-floatx` and `tile-constant-types-intx` command-line options

fboemer:
For now, I'm using float32 / int32 defaults

plaidml/core/core.h
```diff
@@ -10,9 +10,12 @@
 #include <vector>
 
 #include "plaidml/core/ffi.h"
+#include "pmlc/util/enums.h"
```
flaub: We need to keep the `pmlc` and `plaidml` APIs distinct. The idea is that end-users will pull in `plaidml/edsl/edsl.h`, and when they do they should only be including `plaidml` because `pmlc` might be in a different `.so`.

plaidml/core/core.h
```diff
@@ -128,6 +131,39 @@ inline const char* to_string(DType dtype) {
   }
 }
 
+inline DataType to_datatype(DType dtype) {
```
flaub: Unfortunately we can't expose the `pmlc::util::DataType` to end-users without forcing them to include all of `pmlc`. The deployment model is that users will pull in `pmlc` as a `.so` (a binary distribution, not a source-based one).

plaidml/edsl/edsl.h
```diff
@@ -14,6 +14,7 @@
 
 #include "plaidml/core/core.h"
 #include "plaidml/edsl/ffi.h"
+#include "pmlc/util/enums.h"
```
flaub: Same as above, we can't expose `pmlc` from the `plaidml` API.

plaidml/edsl/edsl_test.cc
```diff
@@ -46,6 +46,50 @@ Tensor Softmax(const Tensor& X) {
...
 }
 
+TEST(CppEdsl, LowerPrecisionInvalidNegative) {
+  auto A = Placeholder(DType::FLOAT32, {3, 3});
+  auto C = A * (-2);
+
+  EXPECT_ANY_THROW({ Program("lower_precision", {C}, DType::FLOAT16, DType::UINT16); });
+}
+
+TEST(CppEdsl, LowerPrecision) {
```
flaub: I'd call this `LowerPrecisionConstants` because it's only modifying the constant types, it's not causing the entire test to produce lower precision results.

plaidml/edsl/edsl_test.cc
```diff
@@ -46,6 +46,50 @@ Tensor Softmax(const Tensor& X) {
   return E / N;
 }
 
+TEST(CppEdsl, LowerPrecisionInvalidNegative) {
+  auto A = Placeholder(DType::FLOAT32, {3, 3});
+  auto C = A * (-2);
+
+  EXPECT_ANY_THROW({ Program("lower_precision", {C}, DType::FLOAT16, DType::UINT16); });
```
flaub: I'm not sure why we shouldn't allow this case. I would assume that the result would be `f16` and that truncation would be happening somewhere in the middle of the program.

plaidml/edsl/edsl_test.py
flaub: Isn't this still returning an `f64`? Shouldn't it return an `f32`?

pmlc/dialect/tile/builder.cc
```diff
@@ -693,6 +694,7 @@ std::shared_ptr<TileProgram> TileBuilder::MakeProgram(StringRef name, const Prog
   // Do some optimization passes
   mlir::PassManager pm(&impl->context);
   pm.addPass(MakeProgramPass::create());
+  pm.addPass(dialect::tile::createConstantTypesPass(mutations.floatx, mutations.intx));
```
flaub: Since this translation unit is in the `pmlc::dialect::tile` namespace, there's no need to prefix with `dialect::tile::`

pmlc/dialect/tile/transforms/BUILD
```diff
@@ -25,7 +25,7 @@ plaidml_cc_library(
 
 plaidml_cc_library(
     name = "transforms",
-    srcs = ["registration.cc"],
+    srcs = ["registration.cc", "constant_types.cc"],
```
flaub: I think we should try to keep all registration in `registration.cc`

pmlc/dialect/tile/transforms/constant_types.cc
```diff
@@ -0,0 +1,121 @@
...
+    auto elementType = ScalarType::get(type.getContext(), floatx_);
+    auto new_type = RankedTensorType::get(shape, elementType);
+
+    if (new_type == type) {
+      return matchFailure();
+    }
+    rewriter.replaceOpWithNewOp<ScalarConstantOp>(constOp, elementType, value);
+    return matchSuccess();
+
+  } else if (int_attr && intx_ != DataType::invalid) {
```
flaub: Since the previous `if` is returning, then no need to `else if`, just use `if() {...}`

pmlc/dialect/tile/builder.cc
flaub: This pass needs to run before the initial `MakeProgramPass` because we want constant folding to use the updated constant datatype.

pmlc/dialect/tile/tests/constant-types.mlir
```diff
@@ -0,0 +1,22 @@
...
+    %1 = "eltwise.add"(%0, %cst) : (tensor<3x3x!eltwise.f32>, !f64) -> tensor<3x3x!eltwise.f64>
+    return %1 : tensor<3x3x!eltwise.f64>
+  }
+}
+
+// CHECK-LABEL: func @lower_precision
+// CHECK: eltwise.sconst
+// CHECK-SAME: {value = 1 : i64} : () -> !i16
+// CHECK: eltwise.sconst
+// CHECK-SAME: {value = 2.000000e+00 : f64} : () -> !f16
```
flaub: We should be checking that the ultimate result type of the function is now `f16` instead of `f64`

pmlc/dialect/tile/transforms/constant_types.cc
```diff
@@ -0,0 +1,121 @@
...
+  DataType intx_;
+};
+
+void ConstantTypesPass::runOnOperation() {
+  OwningRewritePatternList patterns;
+  patterns.insert<ConstantTypesRewriter>(&getContext(), floatx_, intx_);
+  applyPatternsGreedily(getOperation()->getRegions(), patterns);
+}
+
+std::unique_ptr<mlir::Pass> createConstantTypesPass(const DataType& floatx, const DataType& intx) {
```
flaub: No need for `const DataType&`, just pass by value here.

pmlc/dialect/tile/transforms/constant_types.cc
```diff
@@ -0,0 +1,121 @@
...
+    rewriter.replaceOpWithNewOp<ScalarConstantOp>(constOp, elementType, value);
+    return matchSuccess();
+  } else {
+    IVLOG(1, "Warning: non-float, non-int type");
+  }
+  return matchFailure();
+}
+
+struct ConstantTypesPass : public OperationPass<ConstantTypesPass> {
+  ConstantTypesPass(const std::string& floatx = constant_types_option_floatx,
```
flaub: Use `llvm::StringRef` instead of `const std::string&`.

flaub: Actually, could we make it so that the `ConstantTypesPass` always takes in a `DataType`, and the conversion between string to int happens as part of the command line parsing code.

pmlc/dialect/tile/transforms/passes.h
```diff
@@ -2,18 +2,24 @@
...
 
 namespace pmlc::dialect::tile {
 
+using DataType = util::DataType;
+
 class ContractionOp;
 
 std::unique_ptr<mlir::OpPassBase<mlir::FuncOp>> createComputeBoundsPass();
 
+std::unique_ptr<mlir::Pass> createConstantTypesPass(const DataType& floatx, const DataType& intx);
```
flaub: Pass by value instead of const ref.

pmlc/util/enums.h
```diff
@@ -58,4 +59,27 @@ inline DataType promoteTypes(DataType lhs, DataType rhs) {  //
   return std::max(lhs, rhs);
 }
 
+inline DataType from_string(const std::string& dtype) {
```
flaub: I'm not sure why we need any of this. Tablegen generates the `symbolizeDataType` function for the purpose of converting strings into `DataType`.

pmlc/util/enums.h
```diff
@@ -58,4 +59,27 @@ inline DataType promoteTypes(DataType lhs, DataType rhs) {  //
...
+  static std::unordered_map<std::string, DataType> long_str_to_type{
+      {"bool", DataType::u1},       {"int8", DataType::i8},          {"uint8", DataType::u8},
+      {"int16", DataType::i16},     {"uint16", DataType::u16},       {"int32", DataType::i32},
+      {"uint32", DataType::u32},    {"int64", DataType::i64},        {"uint64", DataType::u64},
+      {"bfloat16", DataType::bf16}, {"float16", DataType::f16},      {"float32", DataType::f32},
+      {"float64", DataType::f64},   {"<invalid>", DataType::invalid}};
+
+  static std::unordered_map<std::string, DataType> short_str_to_type{
```
flaub: I don't think we need to have a 'long' vs 'short' string name for the internal DataType. I do think it's possible that the `plaidml::DType` might differ from the `pmlc::util::DataType` (since they live in different modules, at different layers of the system, one is user facing (the interface) and the other is part of the implementation that users would never see).

plaidml/edsl/edsl_test.cc
fboemer: `-2` is an int, which shouldn't be cast to `uint16`

plaidml/edsl/edsl_test.py
fboemer: Sorry about the incorrect previous comment. Maybe I misunderstood our previous conversation, but here, the user passes `float64` as a Placeholder, and conflicting type `floatx=float32.` Type promotion yields return `max(float32, float64) => float64`

pmlc/util/enums.h
fboemer: Currently, `to_string(plaidml::DType)` is not parseable by `pmlc::util::symbolizeDataType` (e.g. `float16` vs `float16`). Do you mean we should change `to_string(plaidml::DType)` to match that the format of `DataType`? A common string representation between `DataType` and `DType` seems to be the cleanest way of keeping the `plaidml` and `pmlc` APIs separate

pmlc/dialect/tile/transforms/BUILD
fboemer: Not sure how to do this, since `ConstantTypesPass` relies on `static llvm::cl::opt<std::string> constant_types_option_floatx`, which needs to be present in `constant_types.cc` for the default argument. Putting `constant_types_option_floatx` in a `consatnt_types.h` yields llvm errors due to multiple registration of the CL option.

pmlc/util/enums.h
flaub: I'm curious why we even need to use string values anywhere other than for printing purposes. I don't think we should have users select the dtype based on a string, they should use the enum instead.

plaidml/edsl/edsl.h
```diff
@@ -1081,7 +1084,21 @@ inline Tensor zero() { return Tensor{0}; }
 inline Program::Program(                 //
     const std::string& name,             //
     const std::vector<Tensor>& outputs,  //
+    const DType& floatx,                 //
+    const DType& intx,                   //
     const std::vector<std::tuple<Tensor, Tensor>>& updates) {
+  if (!isFloat(floatx)) {
```
flaub: Let's have all this validation occur inside the `ffi.cc` layer instead of out in the user's context.

plaidml/edsl/edsl_test.cc
```diff
@@ -46,6 +46,49 @@ Tensor Softmax(const Tensor& X) {
...
 
+TEST(CppEdsl, LowerPrecisionInvalidNegative) {
+  auto A = Placeholder(DType::FLOAT32, {3, 3});
+  auto C = A * (-2);
+
+  EXPECT_ANY_THROW({ Program("lower_precision_constants", {C}, DType::FLOAT16, DType::UINT16); });
+}
+
+TEST(CppEdsl, LowerPrecisionConstants) {
+  auto A = Placeholder(DType::FLOAT32, {3, 3});
```
flaub: Why would we want to use a FLOAT32 as an input if we are specifying `FLOAT16` for `floatx`?

plaidml/edsl/edsl_test.py
flaub: I guess I'm confused about why we'd want to have tests for situations where `floatx` and the input datatypes don't match. I think in that situation the user is requesting conflicting types but we should always choose the most promoted type between the two to resolve the conflict. However, the `floatx` API and the new `ConstantTypesPass` should allow the user to specify some other datatype (other than the default=`float32`) such that the result type is ensured to be that other datatype and that no truncation would occur.

plaidml/edsl/ffi.cc
```diff
@@ -973,6 +976,10 @@ plaidml_program* plaidml_program_evaluate(  //
       }
       mutations.updates.emplace(ProgramUpdate{src_updates[i]->value, dst_updates[i]->value});
     }
+
+    mutations.set_floatx(to_string(static_cast<plaidml::DType>(floatx)));
+    mutations.set_intx(to_string(static_cast<plaidml::DType>(intx)));
```
flaub: I don't think we should ever convert to a string unless we are: 1) printing or 2) parsing the type. But if we're just passing the type along thru different layers, we should depend on the enum (not the string value, strings just have a higher cost compared to a simple integer/aka enum value)

pmlc/dialect/tile/transforms/BUILD
flaub: Can we declare the `llvm::cl::opt` in a header file and then have the `static` initializer (i.e. telling the compiler which translation unit this declaration should be defined in) live in `registration.cc`

pmlc/dialect/tile/transforms/passes.h
```diff
@@ -2,18 +2,24 @@
 
 #pragma once
 
+#include "pmlc/util/enums.h"
 #include <memory>
```
flaub: The order of includes should be:

C++ system headers (e.g. <memory>, <string>, etc)

C system headers

Third party headers

This project's headers

pmlc/dialect/tile/builder.h
```diff
@@ -44,6 +45,21 @@ struct ProgramUpdate {
...
+    if (!isFloat(floatx)) {
+      throw std::runtime_error("Invalid floatx string " + std::string(floatx_str));
+    }
+  }
+  void set_intx(const char* intx_str) {
+    intx = util::from_string(intx_str);
+    if (!isInteger(intx)) {
+      throw std::runtime_error("Invalid intx string " + std::string(intx_str));
+    }
+  }
```
flaub: I think we should pass the `floatx` and `intx` as parameters to the `MakeProgram()` method instead of attaching them to the `ProgramMutations` (which should only concern itself with specifying the outputs and updates aka mutations of the program)

plaidml/edsl/edsl.h
```diff
@@ -81,10 +81,13 @@ class Program {
...
-  Program(                                 //
-      const std::string& name,             //
-      const std::vector<Tensor>& outputs,  //
+  /// floatx is used to specify the underlying datatype of floating-point constants
+  /// intx is used to specify the underlying datatype of integer constants
+  Program(                                   //
+      const std::string& name,               //
+      const std::vector<Tensor>& outputs,    //
+      const DType& floatx = DType::FLOAT32,  //
+      const DType& intx = DType::INT32,      //
```
flaub: Pass by value, not by const ref

pmlc/dialect/tile/transforms/BUILD
flaub: OK, this is fine the way it is, I can revisit this later perhaps. It's really an attempt to improve linker performance by minimizing the amount of `alwayslink=1` symbols to consider.

plaidml/edsl/edsl_test.cc
fboemer: You're right, this isn't really a good example. I've replaced it with a better one

plaidml/edsl/ffi.cc
```diff
@@ -973,7 +976,16 @@ plaidml_program* plaidml_program_evaluate(  //
       }
       mutations.updates.emplace(ProgramUpdate{src_updates[i]->value, dst_updates[i]->value});
     }
-    auto ret = new plaidml_program{GlobalContext::get()->MakeProgram(name, mutations)};
+
+    if (!plaidml::isFloat(static_cast<plaidml::DType>(floatx))) {
```
flaub: When we're on the `ffi` side we shouldn't be using the `plaidml` API (that's only for the user's side). Once we get across the C boundary, we should use internal types, so let's cast the `plaidml_datatype` directly into a `pmlc::util::DataType` (the enum values are intentionally identical between the 2 definitions to allow this cheap/easy conversion).

plaidml/edsl/ffi.cc
```diff
@@ -9,6 +9,7 @@
 
 #include "llvm/Support/FormatVariadic.h"
 
+#include "plaidml/core/core.h"
```
flaub: Drop this.

plaidml/edsl/ffi.cc
```diff
@@ -973,7 +976,16 @@ plaidml_program* plaidml_program_evaluate(  //
       }
       mutations.updates.emplace(ProgramUpdate{src_updates[i]->value, dst_updates[i]->value});
     }
-    auto ret = new plaidml_program{GlobalContext::get()->MakeProgram(name, mutations)};
+
+    if (!pmlc::util::isFloat(static_cast<pmlc::util::DataType>(floatx))) {
```
flaub: I think we should try to `symbolizeDataType` just before this line, so then we can call `isFloat()` and then pass it (now already converted to `pmlc::util::DataType` to the call to `MakeProgram()`

pmlc/dialect/tile/builder.h
```diff
@@ -107,7 +107,15 @@ class TileBuilder {
...
-  std::shared_ptr<TileProgram> MakeProgram(llvm::StringRef name, const ProgramMutations& mutations);
+  std::shared_ptr<TileProgram> MakeProgram(llvm::StringRef name, const ProgramMutations& mutations,
+                                           util::DataType floatx = DataType::invalid,
+                                           util::DataType intx = DataType::invalid);
+
+  std::shared_ptr<TileProgram> MakeProgram(llvm::StringRef name, const ProgramMutations& mutations,
+                                           std::uint64_t floatx, std::uint64_t intx) {
+    return MakeProgram(name, mutations, util::symbolizeDataType(floatx).getValueOr(DataType::invalid),
+                       util::symbolizeDataType(intx).getValueOr(DataType::invalid));
+  }
```
flaub: Let's isolate the conversions between raw integers to strongly typed enums in the FFI, so this method should be dropped.

plaidml/edsl/edsl_test.cc
```diff
@@ -46,6 +46,51 @@ Tensor Softmax(const Tensor& X) {
...
+  binder.input(A).copy_from(input_a.data());
+  executable->run();
+  {
+    auto view = binder.output(C).mmap_current();
+    ASSERT_THAT(view.size(), expected.size() * sizeof(expected[0]));
+    auto data = reinterpret_cast<double*>(view.data());
+    std::vector<double> actual(data, data + expected.size());
+    EXPECT_THAT(actual, ContainerEq(expected));
+  }
+}
```
flaub: Wrap this with an `#if !defined(_WIN32)`

jbruestle:
Just wanted to all say thank you for the very useful contribution and also working though the process including all the comments.

PR #882, opened 2020-01-28
MLSys 2020 Demo (initial)

dgkutnic:
I've created a python notebook to play around with MLSys Demo-related code in.

* Bazel compatibility:
`bazelisk run //demos/MLSys:demo`
* Separate dev/prod views:
Development view: This is the default jupyter view of the notebook that can be used for development purposes. `http://localhost:8888/notebooks/demos/MLSys/demo.ipynb`
Production view: This is a view of the notebook that can be used for production. The tool [voila](https://github.com/voila-dashboards/voila) renders the outputs in the notebook into something prettier: `http://localhost:8888/voila/render/demos/MLSys/demo.ipynb`

I've used some basic code for now - the function `edsl_program` currently has a hard coded `dot` just to start out. I'm using that program to autogenerate the SVG graphics that match the shape of the output tensor of the `edsl_program` as an example of what this notebook is capable of doing.


demos/MLSys/BUILD
```diff
@@ -0,0 +1,13 @@
+package(default_visibility = ["//visibility:public"])
```
flaub: Could you run buildifier on this file?

conda/unix.yml
```diff
@@ -34,5 +35,6 @@ dependencies:
     - sphinx-tabs
     - tensorflow==1.14.0
     - theano
+    - voila
```
flaub: It's unclear where this is used anywhere. Is `nbconvert` insufficient?

demos/MLSys/demo.ipynb
```diff
@@ -0,0 +1,125 @@
...
+      "text/plain": [
+       "<drawSvg.raster.Raster at 0x7f84d868e310>"
+      ]
```
flaub: Is this something that was generated? Is this something we want to commit? How do we make notebooks in a ready to commit format?

demos/MLSys/demo.ipynb
```diff
@@ -0,0 +1,125 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "binder = plaidml.exec.Binder(program)\n",
+    "executable = binder.compile()\n",
+    "executable.run()"
```
flaub: I'm wondering if we really want to run anything, it will probably go very slowly depending on what code is submitted, so generally we should probably disallow it...

Once we have the compiler API, we can use that to retrieve the output of various passes.

demos/MLSys/demo.ipynb
```diff
@@ -0,0 +1,125 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "program = Program('edsl_program', [O])\n",
+    "#print(str(program))"
```
flaub: Actually, this `str(program)` can be used for as the 'initial pass' output

conda/unix.yml
flaub: Voila is cool!

demos/MLSys/demo.ipynb
flaub: Oh, for this demo we need that, OK :)

demos/MLSys/demo.ipynb
flaub: I was thinking about the other demo, this is fine!

earhart:
Maybe include the year in the conference name?

demos/MLSys/demo.ipynb
dgkutnic: I ran `nbstripout` on the notebook and it got rid of this

demos/MLSys/BUILD
dgkutnic: Done

dgkutnic:
> Maybe include the year in the conference name?

Good idea. @earhart I've changed it within the PR and in the source code itself.

demos/MLSys2020/demo.ipynb
```diff
@@ -0,0 +1,116 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#  use placeholder, make them ints\n",
+    "A = Tensor(LogicalShape(plaidml.DType.INT32, [3, 3]))\n",
```
dgkutnic: @flaub perhaps I'm missing something extremely basic here but it seems like the `Placeholder` API only exists for the C++ EDSL? I didn't see it when I did a `git grep` and I also couldn't find any examples of Python EDSL code using `Placeholder`.

flaub: The C++ and python APIs should have parity, if it doesn't, could you even them out so that they do?

flaub:
Until we have fully decided, let's keep the filenames consistent with the existing approach (snake_case).

dgkutnic:
@flaub @cnamrata15 marking this as ready for review now. Here's what it looks in voila currently. Not perfect, but ok for now while I iterate on a better UI.

image

There's a couple of things I want to change (add the `Placeholder` API to python and fix the corresponding demo code for example), but I'll do those things in a separate PR.

demos/MLSys2020/demo.ipynb
dgkutnic: Added this to our checklist. I'll address in a separate PR.

flaub:
LGTM

PR #885, opened 2020-01-29
Add stdx dialect with initial atomic_rmw op impl

flaub:


flaub:
I've been considering the lowering, I might do a followup PR as I'm writing the RFC, but I'd like to do some level of PoC so that the RFC is a solid proposal.

PR #890, opened 2020-01-30
Updates from master

flaub:


flaub:
This should also fix the flakey build failures on windows.

PR #891, opened 2020-01-31
Add Placeholder API for Python

dgkutnic:


plaidml/edsl/__init__.py
```diff
@@ -600,6 +600,40 @@ def TensorIndexes(count):
...
+ 
+        Keyword Args:
+            name (string): Optional name to be assigned to the ``Tensor``.
+
+        Returns:
+            (Tensor): The placeholder ``Tensor``.
+
+    """
+    name = ""
+    if "name" in kwargs.keys():
```
flaub: Use single-quotes for strings for consistency.

plaidml/edsl/__init__.py
```diff
@@ -600,6 +600,40 @@ def TensorIndexes(count):
     return [TensorIndex() for i in range(count)]
 
 
+def Placeholder(*args, **kwargs):
+    """Creates a placeholder tensor, specified either by a combination of data
+       type and dimensions or by passing in a LogicalShape.
```
flaub: Can you add tests that use all the combinations possible for this function?

plaidml/edsl/__init__.py
```diff
@@ -600,6 +600,40 @@ def TensorIndexes(count):
...
+
+        Args:
+            dtype (DType): Specifies the data type of the ``Placeholder``.
+            dims (list): Specifies the dimensions of the ``Placeholder``.
+            shape (LogicalShape): Specifies the shape of the ``Placeholder``.
```
flaub: This isn't quite right, because it's not actually the case that users will do:

```
Placeholder(dtype, dims, shape)
```

There are 2 choices:

```
Placeholder(dtype, dims)
Placeholder(shape)
```

To resolve this, let's do this instead:

```
def Placeholder(dtype_or_shape, dims=[], name=''):
```

This then removes much of the bolierplate to determine how the `Placeholder` was called. The only thing to determine is whether `dtype_or_shape` is indeed a `plaidml.DType` or a `plaidml.LogicalShape`

plaidml/edsl/__init__.py
```diff
@@ -600,6 +600,40 @@ def TensorIndexes(count):
...
+        dtype = None
+        dims = []
+        for arg in args:
+            if isinstance(arg, DType):
+                dtype = arg
+            if isinstance(arg, list):
+                dims = arg
+        shape = LogicalShape(dtype=dtype, dims=dims)
+    ptr = ffi_call(lib.plaidml_expr_placeholder, shape.as_ptr(), ffi.NULL, name.encode())
+    return Tensor(expr=ptr)
```
flaub: If possible, let's reuse the `Tensor` ctor to do the work of calling down into the ffi, it just minimizes the calling surface a bit.

plaidml/edsl/__init__.py
dgkutnic: Done

plaidml/edsl/__init__.py
```diff
@@ -600,6 +600,27 @@ def TensorIndexes(count):
...
 
+def Placeholder(dtype_or_shape, dims=[], name=''):
+    """Creates a placeholder tensor, specified either by a combination of data
+       type and dimensions or by passing in a LogicalShape.
+
+        Args:
+            dtype_or_shape (DType | LogicalShape): Specifies either the data 
+            type or the logical shape of the ``Placeholder``.
+            dims (list): Specifies the dimensions of the ``Placeholder``.
+            name (string): Optional name to be assigned to the ``Tensor``.
```
flaub: These are still considered kwargs, but I'm not sure if napolean cares about that or not?

plaidml/edsl/edsl_test.py
```diff
@@ -1034,6 +1034,86 @@ def test_two_outputs(self):
...
+        expected = '''
+!i32 = type tensor<!eltwise.i32>
+module {
+  func @placeholder_dtype_only(%arg0: !i32) -> !i32 {
+    %0 = "eltwise.ident"(%arg0) : (!i32) -> !i32
+    return %0 : !i32
+  }
+}
+'''
+        self.compare_results(program, expected)
```
flaub: This is actually pretty brittle. I know we have lots of tests like this, but I don't think we should add anymore of them. So let's just create the `Program` and leave it at that.

flaub: Maybe we could do a substring or regex comparison so it's not so brittle, so that we are testing only the bits we care about from the `program.str()`

plaidml/edsl/__init__.py
```diff
@@ -600,6 +600,30 @@ def TensorIndexes(count):
...
+
+    """
+    if isinstance(dtype_or_shape, LogicalShape):
+        shape = dtype_or_shape
+    elif isinstance(dtype_or_shape, DType):
+        shape = LogicalShape(dtype=dtype_or_shape, dims=dims)
+    else:
+        raise TypeError('Unsupported type {} for dtype_or_shape={}'.format(
+            type(dtype_or_shape), dtype_or_shape))
+    return Tensor(shape=shape, name=name)
```
flaub: Good call on checking for all cases and throwing when unsupported.

plaidml/edsl/edsl_test.py
dgkutnic: This might not be perfect, but I can pair up each case with the identity `LogicalShape` placeholder and compare those, i.e.

```
    def test_placeholder_with_name(self):
        I1 = Placeholder(plaidml.DType.INT32, [1, 1], name='I')
        program1 = Program('placeholder_with_name', [I1])
        I2 = Placeholder(LogicalShape(plaidml.DType.INT32, [1, 1]), name='I')
        program2 = Program('placeholder_with_name', [I2])
        self.assertEqual(str(I1.shape), "tensor<1x1x!eltwise.i32>")
        self.assertEqual(str(I2.shape), "tensor<1x1x!eltwise.i32>")
        self.compare_results(program1, str(program2))
```

flaub: OK, comparing shapes seems good, but you can drop the `self.compare_results`

flaub: Actually, if we're comparing shapes, then there's probably no need to even make a program.

plaidml/edsl/edsl_test.py
```diff
@@ -1034,6 +1034,33 @@ def test_two_outputs(self):
...
+        self.compare_results(program1, str(program2))
+
+    def test_placeholder_with_name(self):
+        I1 = Placeholder(plaidml.DType.INT32, [1, 1], name='I')
+        program1 = Program('placeholder_with_name', [I1])
+        I2 = Placeholder(LogicalShape(plaidml.DType.INT32, [1, 1]), name='I')
+        program2 = Program('placeholder_with_name', [I2])
+        self.assertEqual(str(I1.shape), "tensor<1x1x!eltwise.i32>")
+        self.assertEqual(str(I2.shape), "tensor<1x1x!eltwise.i32>")
+        self.compare_results(program1, str(program2))
```
flaub: OK cool, I like this :)

plaidml/edsl/__init__.py
dgkutnic: It doesn't care about keyword args like this necessarily, but it has a separate way of letting you denote what goes into `*args` and `**kwargs`

PR #894, opened 2020-01-31
Fix lowering of broadcasted shapes

fboemer:


plaidml/edsl/edsl_test.py
```diff
@@ -264,6 +264,19 @@ def checkProgram(self, program, inputs, expected):
         for i in range(len(expected)):
             self.assertEqual(outputs[i].tolist(), expected[i])
 
+    def test_broadcast_cmp(self):
+        A = Tensor(LogicalShape(plaidml.DType.UINT64, [3, 4]))
+        B = Tensor(LogicalShape(plaidml.DType.UINT64, [3, 1]))
+        O = cast(A >= B, DType.UINT64)
+
+        program = Program('broadcast_cmp', [O])
+        0, 0, 0, 0, 1, 2, 3, 4, 0, 5, 0, 6
```
flaub: I'm puzzled by this line :)

flaub:
I actually think there are 2 major cases we want to cover:

* Mismatched number of dimensions between operands (that case seems to have already been handled)
* When one of the dimensions has the value of 1 (this new case you found)

pmlc/conversion/tile_to_pxa/tests/broadcast.mlir
```diff
@@ -0,0 +1,33 @@
...
+!t_2x1x3x4xu64 = type tensor<2x1x3x4x!eltwise.u64>
+!t_2x2x3x4xu1 = type tensor<2x2x3x4x!eltwise.u1>
+
+// CHECK-LABEL: func @broadcast_has_dim_one
+func @broadcast_has_dim_one(%arg0: !t_1x2x3x4xu64, %arg1: !t_2x1x3x4xu64) -> !t_2x2x3x4xu1 {
+    %0 = "eltwise.cmp_ge"(%arg0, %arg1) : (!t_1x2x3x4xu64, !t_2x1x3x4xu64) -> !t_2x2x3x4xu1
+    return %0 : !t_2x2x3x4xu1
+
+    // CHECK: affine.load
+    // CHECK-SAME:  %arg0[0, %arg4, %arg5, %arg6] : memref<1x2x3x4xi64>
```
flaub: In order to make the tests less brittle, the MLIR team recommends that we use regex patterns instead of matching against `%arg0` directly.

https://mlir.llvm.org/getting_started/TestingGuide/
https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-regex-matching-syntax

So, this would be preferred:

```
// CHECK-SAME:  %{{.*}}[0, %{{.*}}, %{{.*}}, %{{.*}}] : memref<1x2x3x4xi64>
```

PR #898, opened 2020-02-03
Add stdx.atomic_rmw & stdx.atomic_rmw.yield

flaub:


pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
```diff
@@ -0,0 +1,108 @@
...
+
+ protected:
+  LLVM::LLVMDialect& getDialect() const { return lowering.getDialect(); }
+};
+
+Optional<LLVM::AtomicBinOp> matchBinOp(stdx::AtomicRMWOp op) {
+  auto* body = op.getBody();
+  auto* terminator = body->getTerminator();
+  auto yieldOp = llvm::cast<stdx::AtomicRMWYieldOp>(terminator);
+  auto defOp = yieldOp.result().getDefiningOp();
```
jbruestle: We need to additionally should insure that the 'Add' or whatever has one of it's inputs match the argument to the region and that it is the sole instruction in the loop besides the yield (not part of something more complex, which needs the CmpXchg lowering).

pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
```diff
@@ -0,0 +1,108 @@
...
+  using LLVMLegalizationPattern<stdx::AtomicRMWOp>::LLVMLegalizationPattern;
+
+  PatternMatchResult matchAndRewrite(Operation* op, ArrayRef<Value> operands,
+                                     ConversionPatternRewriter& rewriter) const override {
+    auto atomicOp = llvm::cast<stdx::AtomicRMWOp>(op);
+    auto binOp = matchBinOp(atomicOp);
+    if (!binOp.hasValue()) {
+      return matchFailure();
+    }
+    rewriter.eraseOp(op);
```
jbruestle: I assume this would actually do a rewrite here eventually.

pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
```diff
@@ -0,0 +1,108 @@
...
+  using LLVMLegalizationPattern<stdx::AtomicRMWOp>::LLVMLegalizationPattern;
+
+  PatternMatchResult matchAndRewrite(Operation* op, ArrayRef<Value> operands,
+                                     ConversionPatternRewriter& rewriter) const override {
+    auto atomicOp = llvm::cast<stdx::AtomicRMWOp>(op);
+    auto binOp = matchBinOp(atomicOp);
+    if (binOp.hasValue()) {
+      return matchFailure();
+    }
+    rewriter.eraseOp(op);
```
jbruestle: I assume this would actually do a rewrite here eventually.

pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
flaub: Yep, still WIP, will address after I get this RFC published

pmlc/conversion/stdx_to_llvm/tests/atomic_rmw.mlir
```diff
@@ -0,0 +1,91 @@
...
+    stdx.atomic_rmw.yield %0 : f32
+  }
+  return
+}
+
+// -----
+
+// CHECK-LABEL: func @cmpxchg
+func @cmpxchg(%F : memref<10xf32>, %fval : f32, %i : index) {
+  // CHECK: llvm.br ^bb1(%{{.*}} : !llvm.float)
```
jbruestle: It might be nice to test for the initial load.  I see it being generated in the lowering code, but it's not in the test.

PR #899, opened 2020-02-03
Initial autotile pass implementation

jbruestle:


pmlc/dialect/pxa/ir/ops.cc
```diff
@@ -48,32 +48,56 @@ ParseResult parseKeywordIntoEnumAttr(OpAsmParser &parser,
...
-  SmallVector<AffineExpr, 8> ubExprs;
-  // Make range expressions for each range
-  for (int64_t range : ranges) {
-    lbExprs.push_back(builder->getAffineConstantExpr(0));
-    ubExprs.push_back(builder->getAffineConstantExpr(range));
-  }
-  // Make the maps (handling the 0-dim case carefully)
+  // Default initalize empty maps
+  auto lb_map = AffineMap::get(builder->getContext());
+  auto ub_map = AffineMap::get(builder->getContext());
```
flaub: nit: use `camelCase`

pmlc/dialect/pxa/ir/ops.cc
```diff
@@ -48,32 +48,56 @@ ParseResult parseKeywordIntoEnumAttr(OpAsmParser &parser,
...
+    ub_map = AffineMap::get(0, 0, ubExprs);
   }
-  SmallVector<int64_t, 8> steps(ranges.size(), 1);
+  // Fall through
+  build(builder, result, lb_map, {}, ub_map, {});
+}
+
+void AffineParallelOp::build(Builder *builder, OperationState &result,
+                             AffineMap lb_map, ValueRange lb_args,
+                             AffineMap ub_map, ValueRange ub_args) {
```
flaub: nit: use `camelCase`

pmlc/dialect/pxa/ir/ops.cc
```diff
@@ -89,6 +113,50 @@ AffineParallelOp::operand_range AffineParallelOp::getUpperBoundsOperands() {
...
+}
+
+AffineValueMap AffineParallelOp::getRangesValueMap() {
+  AffineValueMap out;
+  AffineValueMap::difference(getUpperBoundsValueMap(), getLowerBoundsValueMap(),
+                             &out);
+  return out;
+}
+
+bool AffineParallelOp::getConstantRanges(llvm::SmallVectorImpl<int64_t> &out) {
```
flaub: Return an `Optional<SmallVector<int64_t, N>>` here instead of `bool` + out param.

pmlc/dialect/pxa/ir/ops.td
```diff
@@ -57,11 +57,29 @@ def AffineParallelOp : PXA_Op<"parallel", [ImplicitAffineTerminator]> {
...
+    // Get various AffineValueMaps
+    AffineValueMap getLowerBoundsValueMap(); 
+    AffineValueMap getUpperBoundsValueMap(); 
+    AffineValueMap getRangesValueMap();
+
+    // Get ranges as constants, return true if it worked (non-dynamic case)
+    bool getConstantRanges(llvm::SmallVectorImpl<int64_t>& out);
     
+    // For a given dimension, check if range is constant.  If so, return, otherwise return -1
+    int64_t getRange(size_t dim);
```
flaub: Probably use `unsigned` instead of `size_t`

pmlc/dialect/pxa/ir/ops.cc
```diff
@@ -89,6 +113,50 @@ AffineParallelOp::operand_range AffineParallelOp::getUpperBoundsOperands() {
...
+    if (auto cst = expr.dyn_cast<mlir::AffineConstantExpr>()) {
+      cache.push_back(cst.getValue());
+    } else {
+      return false;
+    }
+  }
+  out = cache;
+  return true;
+}
+
```
flaub: We need a `verify` function that checks that the IR is sound. One such check is that all the number of dims in lower, upper, and steps are the same.

pmlc/dialect/pxa/ir/ops.cc
```diff
@@ -48,32 +48,56 @@ ParseResult parseKeywordIntoEnumAttr(OpAsmParser &parser,
...
+                             AffineMap lb_map, ValueRange lb_args,
+                             AffineMap ub_map, ValueRange ub_args,
+                             ArrayRef<int64_t> steps) {
+  // Verify sizes
+  size_t idxCount = lb_map.getNumResults();
+  assert(idxCount == ub_map.getNumResults());
+  assert(idxCount == steps.size());
+  // Set all of the attributes
+  result.addAttribute("lowerBoundsMap", AffineMapAttr::get(lb_map));
+  result.addAttribute("upperBoundsMap", AffineMapAttr::get(ub_map));
```
flaub: Attribute names are typically `snake_case`

pmlc/dialect/pxa/tests/autotile.mlir
```diff
@@ -0,0 +1,24 @@
...
+  // CHECK-LABEL: @dot
+  func @dot(%arg0: memref<100x100xf32>, %arg1: memref<100x100xf32>, %arg2: memref<100x100xf32>) {
+    "pxa.parallel"() ( {
+    ^bb0(%arg3: index, %arg4: index, %arg5: index):	// no predecessors
+      %0 = affine.load %arg1[%arg3, %arg5] : memref<100x100xf32>
+      %1 = affine.load %arg0[%arg5, %arg4] : memref<100x100xf32>
+      %2 = mulf %0, %1 : f32
+      pxa.reduce add %2, %arg2[%arg3, %arg4] : memref<100x100xf32>
+      "affine.terminator"() : () -> ()
+    }) {lowerBoundsMap = #map1, steps = [1, 1, 1], upperBoundsMap = #map2} : () -> ()
```
flaub: Attributes are usually `snake_case`

pmlc/dialect/pxa/transforms/BUILD
```diff
@@ -5,13 +5,25 @@ package(default_visibility = ["//visibility:public"])
 load("//bzl:plaidml.bzl", "plaidml_cc_library", "plaidml_cc_test")
 
 plaidml_cc_library(
-    name = "transforms",
+    name = "lib",
     srcs = [
+	"autotile.cc",
+	"tile.cc",
```
flaub: Whitespace: run `buildifier` on this file

pmlc/dialect/pxa/transforms/autotile.h
```diff
@@ -0,0 +1,111 @@
...
+class FixedTileSizeGenerator {
+public:
+  explicit FixedTileSizeGenerator(int64_t val) : val(val) {}
+  std::vector<int64_t> operator()(int64_t range) const { return {val}; }
+
+private:
+  int64_t val;
+};
+
+// Producs the union of multiple generators
```
flaub: typo: `Produces`

pmlc/dialect/pxa/transforms/registration.cc
```diff
@@ -0,0 +1,28 @@
+// Copyright 2020, Intel Corporation
+
+#include "pmlc/dialect/pxa/transforms/autotile.h"
+
+namespace pmlc::dialect::pxa {
+
+struct Autotile10Pass : public mlir::FunctionPass<Autotile10Pass> {
+  void runOnFunction() override {
+    auto func = this->getFunction();
```
flaub: Drop `this->`

pmlc/dialect/pxa/transforms/tile.cc
```diff
@@ -0,0 +1,50 @@
+// Copyright 2019, Intel Corporation
+
+#include <iostream>
+
+#include "pmlc/dialect/pxa/transforms/tile.h"
+
+namespace pmlc::dialect::pxa {
+
+void Tile(AffineParallelOp op, llvm::ArrayRef<int64_t> tileSizes) {
```
flaub: nit: use `camelCase`

Also, perhaps we should call it `performTiling`

pmlc/dialect/pxa/transforms/tile.cc
```diff
@@ -0,0 +1,50 @@
...
+  auto &outerLoopOps = outerBody->getOperations();
+  innerLoopOps.splice(std::prev(innerLoopOps.end()), outerLoopOps,
+                      outerLoopOps.begin(), std::prev(outerLoopOps.end(), 2));
+  // Update outer step size
+  llvm::SmallVector<int64_t, 8> newSteps;
+  auto oldSteps = op.steps().cast<ArrayAttr>().getValue();
+  for (size_t i = 0; i < dimCount; i++) {
+    newSteps.push_back(oldSteps[i].cast<IntegerAttr>().getInt() * tileSizes[i]);
+  }
+  op.setAttr("steps", builder.getI64ArrayAttr(newSteps));
```
flaub: Let's add helper methods to the ODS of the op itself so that we can shield callers from changes to the op.
So `setSteps()` would be preferable to setting attributes directly.

pmlc/dialect/pxa/transforms/tile.h
```diff
@@ -0,0 +1,11 @@
+// Copyright 2019, Intel Corporation
+
+#pragma once
+
+#include "pmlc/dialect/pxa/ir/ops.h"
+
+namespace pmlc::dialect::pxa {
+
+void Tile(AffineParallelOp op, llvm::ArrayRef<int64_t> tileSizes);
```
flaub: If tiling only works on `AffineParallelOp`, would it make sense to add a method to the `AffineParallelOp` itself?

pmlc/dialect/pxa/transforms/autotile.h
```diff
@@ -0,0 +1,111 @@
...
+template <typename Head, typename... Rest>
+class UnionGenerator {
+public:
+  UnionGenerator(const Head &head, const Rest &... rest)
```
flaub: You might want to use `&&` instead of `const &` to be able to have perfect forwarding...

pmlc/dialect/pxa/transforms/autotile.h
```diff
@@ -0,0 +1,111 @@
...
+#include "mlir/Pass/Pass.h"
+#include "pmlc/dialect/pxa/ir/ops.h"
+#include "pmlc/dialect/pxa/transforms/tile.h"
+
+namespace pmlc::dialect::pxa {
+
+// A tile size generator is a functor from range -> list of sizes
+
+// Produces all powers of 2 <= range
+std::vector<int64_t> PowerOfTwoGenerator(int64_t range);
```
flaub: nit: I'm not sure if this should be `camelCase`

pmlc/dialect/pxa/tests/autotile.mlir
```diff
@@ -0,0 +1,24 @@
...
+    return
+  }
+  // CHECK: pxa.parallel
+  // CHECK: pxa.parallel
```
flaub: It'd be good to check that these `pxa.parralel` have the correct ranges given the tiling performed. Although I could see that this might be difficult to check for without the custom printer/parser done first.

pmlc/dialect/pxa/transforms/registration.cc
```diff
@@ -0,0 +1,28 @@
...
+    FixedTileSizeGenerator always10(10);
+    func.walk([&](AffineParallelOp op) {
+      llvm::SmallVector<int64_t, 8> ranges;
+      if (!op.getConstantRanges(ranges)) {
+        return;
+      }
+      auto tileSize = FindBestTileSize(always10, DummyCostModel, ranges);
+      if (tileSize.size() == 0) {
```
flaub: nit: `tileSize.empty()`

pmlc/dialect/pxa/transforms/autotile.cc
```diff
@@ -0,0 +1,29 @@
...
+  return out;
+}
+
+std::vector<int64_t> EvenDivisionGenerator(int64_t range) {
+  std::vector<int64_t> out;
+  // TODO: Something less naive: i.e. factor with sieve and then produce
+  // divisors via that.  This is not as bad as one might imagine, since generator
+  // set is cached in autotile.
+  for (int64_t r = 1; r <= range; r++) {
+    if (range % r != 0) {
```
earhart: I think there's one trivial optimization you could do here -- just scan up to the sqrt of the range, and add the paired divisors as you go?

pmlc/dialect/pxa/transforms/autotile.cc
```diff
@@ -0,0 +1,29 @@
...
+
+std::vector<int64_t> PowerOfTwoGenerator(int64_t range) {
+  std::vector<int64_t> out;
+  for (int64_t r = 1; r <= range; r *= 2) {
+    out.push_back(r);
+  }
+  return out;
+}
+
+std::vector<int64_t> EvenDivisionGenerator(int64_t range) {
```
earhart: Could you move these generally useful sorts of things to `pmlc/util` somewhere and export them?  (I was actually about to write this for another project; I'd rather just call your version.  :-)

pmlc/dialect/pxa/transforms/autotile.h
earhart: 👍 Seems like most (all?) functions should be in the tightly-coupled-with-MLIR world

pmlc/dialect/pxa/transforms/autotile.h
earhart: Also use `std::forward` when passing `rest`

pmlc/dialect/pxa/transforms/autotile.cc
flaub: If this winds up in upstream MLIR, then it would probably belong in `mlir/Support/Generators.h` or some such.

pmlc/dialect/pxa/ir/ops.cc
jbruestle: Yup, trying to adjust my brain, will do.

pmlc/dialect/pxa/ir/ops.cc
jbruestle: Done.

pmlc/dialect/pxa/ir/ops.td
jbruestle: Actually, this has no implementation, just deleted it.

pmlc/dialect/pxa/transforms/autotile.cc
jbruestle: We don't want prime factors, we want all the divisors (including the full range).  So for 16, we want 1, 2, 4, 8, 16, and sqrt of 16 is only 4.

pmlc/dialect/pxa/tests/autotile.mlir
jbruestle: Yeah, agree, but probably a later commit.

pmlc/dialect/pxa/transforms/autotile.h
jbruestle: Make them all functors... :)

pmlc/dialect/pxa/ir/ops.cc
jbruestle: Done.

pmlc/dialect/pxa/transforms/tile.h
jbruestle: Maybe?  Not sure about this one, tiling feels too big to be in a helper function... Not quite today anyway.

pmlc/dialect/pxa/transforms/autotile.cc
jbruestle: I think we can totally move the, but I'm not sure where I'd put them right now, so leaving it.

PR #902, opened 2020-02-05
MLSys Demo: Interactive Logic (1/2)

dgkutnic:
# MLSys Demo: Interactive Logic (1/2)

This PR makes the interactive "Pick an operation" dropdown auto-populate the interactive/editable code snippet in the text box.

## Example 1:

image

## Example 2:

image

## Changes:
* Added the `edsl_snippets` dictionary to map operation names (e.g. Matrix Multiplication) to EDSL code (e.g. R[i, j] += X[i, k] * Y[k, j])
* Moved the dropdown and text box widgets into a cell that can observe/handle callback logic
* Handle the python indentation of the EDSL code with a VBox rather than &nbsp
* Rename `title`, `subdemo`, and `demo` to EDSL-specific names because these assets are not shared between the EDSL and MLIR demos.


dgkutnic:
@flaub I'm aware that the static code under `edsl_program` is inaccurate for the majority of the EDSL snippets, and will need to be dynamically changed as well. I'll make a separate PR to handle the callbacks for that logic.

dgkutnic:
I agree with you that the functional code should be separate from the UI. I've separated those components into `demo_source.py` and import that into the notebook now.

PR #905, opened 2020-02-06
Support for upcoming demos and improved plaidml API

flaub:
Adds support for `program.passes` which allows `[(pass_name, ir), ...]` to be returned from a `program` after it has been compiled.

Note that a minor breaking change is introduced in the C++ API. Use `ProgramBuilder(name, outputs).compile()` instead of `Program(name, outputs)`. The `ProgramBuilder` provides an easier way to specify which settings to use during compilation.

The `debug` argument is added to the `ProgramBuilder` (for C++) and `Program` (for python) to enable the collection of IR after each pass is performed. The `passes` property on `Program` will be empty unless `debug` is enabled.

Additionally, the `target` is now specified as part of `Program` construction rather than `Executable` construction.


plaidml/core/internal.h
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Intel Corporation.
+// Copyright 2019 Intel Corporation
```
dgkutnic: minor nit: should this date be modified to 2020 (or even 2019-2020)?

flaub: I have no idea what our policy is for source file headings... I've been updating the date for new files only thus far.

dgkutnic: Fair enough. I just saw it updated in some places and not others, so I was just wondering
<hr>

As a sidenote, I have seen other groups within Intel just do a blanket update to 2020. Not saying that needs to be addressed right now or even anytime soon.

plaidml/edsl/__init__.py
```diff
@@ -672,6 +682,16 @@ def inputs(self):
     def outputs(self):
         return [x for x in self.args if not x.is_input]
 
+    @property
+    def passes(self):
+        """Returns a list of passes in to form: [(name, ir), ...]
```
dgkutnic: Thanks for writing up a docstring about this API. We'll probably want to regenerate documentation after this lands. That reminds me that we wanted to start auto-rebuilding ReadTheDocs at some point. I've created a task in our planner for that.

flaub: I'll try to make another pass at the docs at some point before the demo and C4ML so we have something good to generate.

PR #906, opened 2020-02-06
MLSys Demo: Interactive Logic (2/2)

dgkutnic:
# MLSys Demo: Interactive Logic (2/2)

This PR makes the entire demo interactive, adding a runtime handler to run a new `Program` each time the user clicks `Run`.

It also adds `Contraction` and `Elementwise` tabs to the UI to show that tensor indexing and binding dims is not needed for eltwise ops.

Check out Teams for a video of this code in action! (Github doesn't allow videos here)

There are some known issues:
* Contraction operations with an output shape that's different from the input shape
* Operations that take a single operand as input

(... and probably more.)

@cnamrata15 when you work on error handling, it might probably be worth making sure (outside of jupyter notebooks) that the runtime handler has robust testing. I discussed this with Frank yesterday and we can discuss in more detail when you have time.


demos/mlsys_2020/demo_source.py
```diff
@@ -35,7 +53,7 @@ def draw_grid(dims, vals, grid_size=10, color='#ffffff'):
                         stroke_width=grid_size,  # outline size
                         stroke='black',  # outline color
                         fill=color))  # fill color
-                string_output = str(vals[x][y])
+                string_output = str(vals[dims[1] - 1 - y][x])
```
dgkutnic: @flaub sorry, wanted to push one more change...

Remember how the matmul results were looking a little weird yesterday? I tracked it down to the `draw_grid` function drawing the arrays weirdly (because the origin of the drawings denote the bottom left, whereas for matrices, the `array[0][0]` element is traditionally placed at the top left).

For example, the array `[[1, 2, 3], [4, 5, 6], [7, 8, 9]]` was showing up as this:
image

I fixed this now:

image

flaub: Good catch. Might be good to add a legend to the top (or bottom) and left (or right) edge. Also, might be good to say whether things are major-column or major-row layout.

PR #908, opened 2020-02-07
Enable uint64 tensors

fboemer:


plaidml/edsl/edsl.h
```diff
@@ -553,6 +553,15 @@ class Tensor {
     impl_->ptr = details::make_ptr(ffi::call<plaidml_expr*>(plaidml_expr_int, value));
   }
 
+  ///
+  /// Tensor constructor
+  /// \param value uint64_t
+  /// \return Tensor
+  ///
+  explicit Tensor(std::uint64_t value) : impl_(new Impl) {  //
```
flaub: nit: no need for `std::` prefix

pmlc/dialect/tile/builder.cc
```diff
@@ -343,6 +343,13 @@ std::vector<Value> TileBuilder::GetTupleElements(Value value) {
   throw std::runtime_error("Expected TupleOp");
 }
 
+Value TileBuilder::MakeScalarConstantOp(uint64_t value) {
+  IVLOG(5, "TileBuilder::MakeScalarConstantOp> " << value);
+  auto type = impl->builder.getType<ScalarType>(DataType::i32);
```
fboemer: Maybe use `DataType::u64` instead? For now, this is consistent with the `int64_t` case, which also uses `DataType::i32`

PR #909, opened 2020-02-07
Add initialization for aggregations

jbruestle:


pmlc/dialect/tile/builder.cc
```diff
@@ -156,10 +155,50 @@ struct TileBuilder::Impl {
...
   }
+
+  Value makeIdentity(ScalarType elemType, util::AggregationKind agg) {
+    switch (agg) {
+    case util::AggregationKind::assign:
+    case util::AggregationKind::add:
+      if (isFloat(elemType.type())) {
+        return makeScalarConstantOp(0.0);
+      } else {
+        return makeScalarConstantOp(int64_t(0));
```
flaub: use c++ cast

flaub: 0LL

PR #917, opened 2020-02-12
Added a padding pass, which computes padding and adds as attributes.

jbruestle:


pmlc/dialect/tile/transforms/padding.cc
```diff
@@ -0,0 +1,213 @@
...
+      }
+      continue;
+    }
+    // Check if it's a block argument, and if so add a 'copy'
+    if (auto arg = def.dyn_cast<mlir::BlockArgument>()) {
+      // Construct an initial identity operation
+      auto block = arg.getOwner();
+      OpBuilder inner(block, block->begin());
+      auto ident = inner.create<eltwise::IdentOp>(inner.getUnknownLoc(), def);
+      // Replace all uses with ident (expect for newly generated use)
```
flaub: typo: except?

pmlc/dialect/tile/transforms/padding.cc
```diff
@@ -0,0 +1,213 @@
...
+      // TODO: This seems like the wrong way to do things?
```
flaub: Is `replaceAllUsesWith` not sufficient here?

pmlc/dialect/tile/transforms/padding.cc
```diff
@@ -0,0 +1,213 @@
...
+      // Compute the largest region the source may access.
+      auto bounds = computeBounds(op.getSourceMap(i), *op.lowerBounds(),
+                                  *op.upperBounds());
+      // Check if there are any out-of-bounds reads.
+      auto shape =
+          op.getTensor(i).getType().cast<mlir::MemRefType>().getShape();
+      assert(bounds.size() == shape.size());
+      bool needsPadding = false;
+      for (size_t i = 0; i < bounds.size(); i++) {
+        if (shape[i] == mlir::MemRefType::kDynamicSize) {
```
flaub: I think there's a `static ShapedType::isStaticSize` that you can use here instead.

pmlc/dialect/tile/transforms/padding.cc
```diff
@@ -0,0 +1,213 @@
...
+  BoundRange visitFloorDivExpr(mlir::AffineBinaryOpExpr expr) {
+    llvm_unreachable("Invalid floor in ComputeRangeVisitor");
+  }
+
+private:
+  std::vector<BoundRange> dimRanges;
+};
+
+llvm::SmallVector<BoundRange, 4>
+computeBounds(AffineMap access, AffineMap lower, AffineMap upper) {
```
flaub: This is slightly confusing because we already have a pass called "compute bounds" and presumably this function doesn't do the same work.

pmlc/dialect/tile/transforms/padding.cc
```diff
@@ -0,0 +1,213 @@
...
+
+#include "pmlc/dialect/eltwise/ir/ops.h"
+#include "pmlc/dialect/tile/ir/ops.h"
+#include "pmlc/dialect/tile/transforms/passes.h"
+
+namespace pmlc::dialect::tile {
+
+using mlir::AffineMap;
+
+struct BoundRange {
```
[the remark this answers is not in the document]
shiyao00: Should it be a part of alias analysis?

pmlc/dialect/tile/transforms/padding.cc
```diff
@@ -0,0 +1,213 @@
...
+  for (size_t i = 0; i < size; i++) {
+    outRanges.push_back(visitor.visit(access.getResult(i)));
+  }
+  return outRanges;
+}
+
+struct PaddingInfo {
+  llvm::SmallVector<int64_t, 4> padBelow;
+  llvm::SmallVector<int64_t, 4> padAbove;
+};
```
shiyao00: Shall we put agg kinds inside PaddingInfo to avoid DenseMap<AggregationKind, PaddingInfo>? It could be  a simple "uint32_t aggKinds", which is a bitmap for all agg kinds. Then we can check if the aggKinds has multiple 1s, i.e., check if it is power of 2 using  a simple trick aggKinds^(aggKinds-1) == 0. If aggKinds has multiple 1s, padBelow and padAbove are useless and can be overwritten.

pmlc/dialect/tile/tests/padding.mlir
```diff
@@ -0,0 +1,136 @@
+// RUN: pmlc-opt -tile-compute-bounds -tile-pad %s | FileCheck %s
+
+!f32 = type !eltwise.f32
+
+#conv1dcenter = affine_map<(i, j)->(i+j-1)> 
+#conv1djustify = affine_map<(i, j)->(i+j)>
```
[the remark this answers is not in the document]
jbruestle: Made spacing match printing form...

pmlc/dialect/tile/transforms/passes.h
```diff
@@ -19,9 +19,15 @@ using DataType = util::DataType;
...
 
-std::unique_ptr<mlir::OpPassBase<mlir::FuncOp>> createComputeBoundsPass();
+std::unique_ptr<mlir::Pass> createComputeBoundsPass();
 
 std::unique_ptr<mlir::Pass> createConstantTypesPass(DataType floatx,
                                                     DataType intx);
 
+struct PadPass : public mlir::FunctionPass<PadPass> {
+  void runOnFunction() final;
+};
```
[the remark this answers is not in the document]
jbruestle: So this does need to be in the public header if we are to register it in register.cc

pmlc/dialect/tile/transforms/padding.cc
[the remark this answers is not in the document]
jbruestle: Yes I do think this will eventually move somewhere more general, but I wanted to wait until we needed a second use.  I will be building analysis structures for Affine, but for Tile I still don't have a super good idea of which parts we will use more than once, so I'd like to wait for more use case.

pmlc/dialect/tile/transforms/passes.h
```diff
@@ -19,9 +19,15 @@ using DataType = util::DataType;
...
+
+std::unique_ptr<mlir::Pass> createPadPass();
```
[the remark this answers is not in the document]
jbruestle: I called it 'PadPass' to keep the pass named in the form of a verb, which seems to be the standard.

pmlc/dialect/tile/transforms/padding.cc
```diff
@@ -0,0 +1,213 @@
...
+      for (size_t i = 0; i < bounds.size(); i++) {
+        if (shape[i] == mlir::MemRefType::kDynamicSize) {
+          op.emitRemark("padding cannot support dynamic memref sizes");
+          return;
+        }
+        if (bounds[i].min < 0 || bounds[i].max >= shape[i]) {
+          needsPadding = true;
+        }
+      }
+      // If there is no need to pad, don't bother adding an entry
```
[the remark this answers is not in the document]
jbruestle: Fixed.

pmlc/dialect/tile/transforms/padding.cc
jbruestle: Leaving for now.

pmlc/dialect/tile/transforms/padding.cc
jbruestle: I considered something along these lines, however I decided that it was premature, since we may someday soon want to pad two different ways, with one of them being a copy, and one being inline for example.  In order to do that, we would need to track more, so I just tracked it.

pmlc/dialect/tile/transforms/padding.cc
```diff
@@ -0,0 +1,215 @@
...
+      auto shape =
+          op.getTensor(i).getType().cast<mlir::MemRefType>().getShape();
+      assert(bounds.size() == shape.size());
+      bool needsPadding = false;
+      for (size_t i = 0; i < bounds.size(); i++) {
+        if (shape[i] == mlir::MemRefType::kDynamicSize) {
+          op.emitRemark("padding cannot support dynamic memref sizes");
+          return;
+        }
+        if (bounds[i].min < 0 || bounds[i].max >= shape[i]) {
```
flaub: Drop trivial braces (I have a hard time with this one myself, trying to get into the new convention)

pmlc/dialect/tile/transforms/padding.cc
```diff
@@ -0,0 +1,215 @@
...
+      }
+    }
+  });
+  Builder builder(&getContext());
+  for (const auto &kvp : toPad) {
+    // Get the value which we want to pad.
+    Value def = kvp.first;
+    // If there are two conflicting ways to pad, don't pad.
+    if (kvp.second.size() != 1) {
+      if (Operation *op = def.getDefiningOp()) {
```
flaub: Drop trivial braces

PR #921, opened 2020-02-13
Add Vulkan runner and Vulkan loader

YangleiZouIntel:


pmlc/.clang-format
```diff
@@ -1,2 +1,2 @@
 BasedOnStyle: LLVM
-AlwaysBreakTemplateDeclarations: Yes
+AlwaysBreakTemplateDeclarations: true
```
flaub: This file came directly from upstream: https://github.com/plaidml/llvm-project/blob/plaidml/plaidml-v1/mlir/.clang-format

I wonder if you're using an older version of clang-format? We use version 9.

flaub:
This appears to build correctly on our CI, so I'm happy to have this land so that we can iterate on it.

pmlc/.clang-format
YangleiZouIntel: I see. I upgraded clang-format and this issue went away.

pmlc/tools/pmlc-vulkan-runner/BUILD
```diff
@@ -0,0 +1,45 @@
...
+        "@vulkan_loader//:vulkan_loader",
+        "@llvm-project//mlir:Support",
+        "@llvm-project//mlir:SPIRVSerialization",
+        "@llvm-project//mlir:SPIRVDialect",
+    ],
+    alwayslink=1,
+)
+
+
+#plaidml_cc_test(
```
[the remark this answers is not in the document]
YangleiZouIntel: Thanks, some tests deps have been renamed/removed. I have fixed the deps and the tests passed on my local machine.

PR #923, opened 2020-02-14
Introduce xsmm.gemm op and libxsmm integration

flaub:
This introduces the `xsmm.gemm` op, which represents a standard 2-D GEMM that can be lowered into a function call to `libxsmm_gemm`.

Includes:
- The initial IR definition
- custom printer/parser
- Test integration with libxsmm (via `pmlc-jit`)

TODO:
- Pattern match `affine.parallel` to transform into `xsmm.gemm`.
- Lowering from `xsmm.gemm` to LLVM dialect.


pmlc/target/x86/tests/xsmm.mlir
```diff
@@ -0,0 +1,57 @@
...
+        %b_ref = memref_cast %b_view : memref<2x2xf32, offset: ?, strides: [2, 2]> to memref<*xf32>
+        %c_ref = memref_cast %c_view : memref<2x2xf32, offset: ?, strides: [2, 2]> to memref<*xf32>
+        call @plaidml_rt_xsmm_gemm_f32(%a_ref, %b_ref, %c_ref, %c8_i32, %c8_i32, %c8_i32, %c2_i32, %c2_i32, %c2_i32)
+          : (memref<*xf32>, memref<*xf32>, memref<*xf32>, i32, i32, i32, i32, i32, i32) -> ()
+      }
+    }
+  }
+  return
+}
+
```
jbruestle: At some point we should use something better than fill for initialization since it’s a bit too easy of a test (lots of ways to get the right answer via the wrong computing, for example always using the same input tile.

flaub: Fair enough, I'm currently working on a convolution test that is more complex. I could also look into using a `DenseElementAttr` to specify arbitrary input data.

PR #924, opened 2020-02-14
update keras backend to accommodate Tensor API changes (shape/rank)

mars1026:
The Tensor class used to have a shape attribute; this updates the keras backend to use the rank or dtype properties instead, or the compute_shape() method.

### DIFF
--- a/plaidml/bridge/keras/__init__.py
+++ b/plaidml/bridge/keras/__init__.py
@@ -124,7 +124,7 @@ class _Function(object):
 
     def _compile(self, inputs):
         for node, data in zip(self._inputs, inputs):
-            dtype = node.tensor.shape.dtype
+            dtype = node.tensor.dtype
             shape = edsl.LogicalShape(dtype, data.shape)
             node.tensor.bind(shape)
         inputs = [x.tensor for x in self._inputs]
@@ -193,7 +193,7 @@ class _KerasNode(object):
         except ValueError:
             ellipsis_idx = None
         I = self.tensor
-        ndims = I.shape.ndims
+        ndims = I.rank
         extension_length = ndims - len(key)
         if ellipsis_idx is not None:
             # The ellipsis is counted in the length of the key, but does not persist as an axis for slicing, so undo that count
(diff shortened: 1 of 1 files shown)


plaidml/bridge/keras/__init__.py
```diff
@@ -1535,11 +1537,13 @@ def softsign(x):
 
 @_log_call
 def sparse_categorical_crossentropy(target, output, from_logits=False):
-    dims = edsl.TensorDims(output.tensor.shape.ndims)
+    shape = output.tensor.compute_shape()
+    dims = edsl.TensorDims(shape.rank)
```
tzerrell: I think this should just be `dims = edsl.TensorDims(output.tensor.rank)`

plaidml/bridge/keras/__init__.py
```diff
@@ -679,7 +679,7 @@ def conv3d_transpose(x,
 @_log_call
 def count_params(x):
     result = 1
-    for dim in x.tensor.shape.int_dims:
+    for dim in x.tensor.compute_shape().into_TensorShape().sizes:
```
flaub: I don't think we need to convert to `TensorShape` here, but `compute_shape()` is probably necessary.

plaidml/bridge/keras/__init__.py
flaub: Agreed, avoiding `compute_shape()` is key because its a fairly expensive operation. (which was the motivation for the API change to begin with actually).

plaidml/bridge/keras/__init__.py
flaub: Is it possible to use `x.bind_dims()` here instead and compute the product symbolically I wonder?

plaidml/bridge/keras/__init__.py
mars1026: We can't avoid compute_shape() in this case, because one of the sizes gets passed along in the call one_hot() a couple lines further down; I thought it made more sense to grab the shape object once and reuse it.

flaub: Is it possible for `one_hot` to take the dim size symbolically?

plaidml/bridge/keras/__init__.py
tzerrell: Keras expects `count_params` to return something that's more-or-less a python integer (specifically, going by the uses in the Keras source code, if `n` is a python integer, then `n == X.count_params()` needs to behave as expected and `int(X.count_params())` needs to behave as expected). You could do something like
```
shape = edsl.TensorDims(I.rank)
I.bind_dims(*shape)
ret = 1
for dim in shape:
    ret *= dim
return ret
```
but will that cast and compare to python ints correctly?

flaub: OK, no, we have to call` compute_shape()` then.

plaidml/bridge/keras/__init__.py
tzerrell: `one_hot` needs to construct a tensor equal to `range(this_size)` where `this_size` is equal to the variable we're talking about. I don't see an immediate way to do this symbolically.

This is in `sparse_categorical_crossentropy` and we haven't yet prioritized handling the sparsity for this function in a performant way at runtime, so I'm not sure bad compile time perf matters that much in this case.

PR #932, opened 2020-02-20
NFC: Tile dialect updates

flaub:
Rename and reorganize tile dialect ops so that they are less confusing. Split the dialect into two `.td` files so its clear which ops are used by the builder vs the rest of the system.


flaub:
Non-Functional Change: there shouldn't be any changes to logic.

PR #933, opened 2020-02-21
Padding during lowering of tile to affine

flaub:
TODO: The inner `affine.if` still needs to be removed


flaub:
@jbruestle I added support for dealing with padding on contractions as well as elementwise ops. This current implementation will do a separate fill for the exterior and interior regions, even if they both use the same value, so obviously this could be improved. It'd be nice if we could defer the merging of these 2 fills in a later optimization pass because we'd want to do it after a canonicalization pass. It's not clear to me if it would be easier to just skip creating a separate fill if we know both values are the same, but that seems to be the rub, how do we know when the interior and exterior values are the same? After canonicalization we can check if they both refer to the same constant.

flaub:
Also included some folders and canonicalization patterns for AffineReduceOp because otherwise the subview doesn't canonicalize into a nice form (it's up to the users of the subview to do that apparently).

PR #935, opened 2020-02-21
Lowering Demo for MLSys

cnamrata15:
Initial version with all the IRs. 
Will make another PR for hiding the passes we talked about.
Added the demo target, although probably still need to set default conditions to use it.


demos/mlsys_2020/demo_source.py
```diff
@@ -3,10 +3,13 @@
 import drawSvg as draw
 from drawSvg.widgets import DrawingWidget, AsyncAnimation
 import ipywidgets as widgets
+from IPython.display import display
+from ipywidgets import Layout
 
 import plaidml
 import plaidml.exec
 from plaidml.edsl import *
+from ipywidgets import Textarea, VBox
```
flaub: Can you sort your imports? If you're using vscode, there's a command to do that for you.

demos/mlsys_2020/demo_source.py
```diff
@@ -83,9 +86,23 @@ def edsl_program(X, Y):
...
+                                    layout=Layout(
+                                        height='100%',
+                                        width='auto',
+                                    ))
+            box = VBox([text], layout={'height': '350px'})
+            display(box)
 
-        # Create the binder and the executable so that the program can run.
+
+# Create the binder and the executable so that the program can run.
```
flaub: Looks like whitespace differences here, use `yapf` to format the code

demos/mlsys_2020/lowering_demo.ipynb
```diff
@@ -0,0 +1,228 @@
+{
```
flaub: Is there a way to move the majority of this code into a python file? It's hard to review it in this form.

demos/mlsys_2020/demo.ipynb
```diff
@@ -184,7 +184,10 @@
     "    ]),\n",
     "    widgets.HTML(value=\"<h2>Result</h2>\"),\n",
     "    AsyncAnimation(1, output_anim, click_pause = False),\n",
-    "])"
+    "])\n",
```
flaub: Are you using the `pre-commit` hook that @dgkutnic added? It should normalize this stuff and remove it I think.

pmlc/target/demo/pipeline.cc
```diff
@@ -0,0 +1,48 @@
...
+  pm.addNestedPass<FuncOp>(createCanonicalizerPass());
+  pm.addNestedPass<FuncOp>(createCSEPass());
+
+  pm.addPass(createLowerAffinePass());
+  pm.addNestedPass<FuncOp>(createCanonicalizerPass());
+  pm.addNestedPass<FuncOp>(createCSEPass());
+}
+
+static PassPipelineRegistration<>
+    passPipelineReg("target_demo", "Target pipeline for demonstrations",
```
flaub: `target-demo`

pmlc/target/demo/pipeline.cc
```diff
@@ -0,0 +1,48 @@
+// Copyright 2019, Intel Corporation
+
+#include "mlir/Conversion/LoopToStandard/ConvertLoopToStandard.h"
```
flaub: Remove unused includes

pmlc/target/demo/pipeline.cc
```diff
@@ -0,0 +1,48 @@
+// Copyright 2019, Intel Corporation
```
flaub: Update the year

pmlc/target/demo/BUILD
```diff
@@ -0,0 +1,23 @@
...
+    name = "demo",
+    srcs = glob(["*.cc"]),
+    hdrs = glob(["*.h"]),
+    deps = [
+        "//pmlc/compiler",
+        "//pmlc/conversion/pxa_to_affine",
+        "//pmlc/conversion/tile_to_pxa",
+        "//pmlc/dialect/tile/transforms",
+        "@llvm-project//mlir:AffineToStandardTransforms",
+        "@llvm-project//mlir:GPUToSPIRVTransforms",
```
flaub: Remove unused deps

demos/mlsys_2020/lowering_demo.ipynb
cnamrata15: Thanks. I kept the exact same structure as the other demo, in case we want to merge both. 
But yes, stuff can definitely be moved around. Let me address this in another PR

demos/mlsys_2020/demo.ipynb
cnamrata15: Yes it should. Not sure how this got through. 
Thanks for pointing out.

demos/mlsys_2020/lowering_demo.ipynb
flaub: Would it be possible to do it in this PR or would that make a mess of things?

I only ask because I'm finding it hard to review the code.

cnamrata15: I agree it is hard to review.
Although, it is the same code that was in Denise's PR, minus the grids for displaying the inputs and outputs, so you might have already gone through it.
My reasoning behind keeping it the same way is this - if we want to merge the two demos(like she had asked me to) , we can restructure them together. If we keep them separate, then we can work on making them modular separately. 
Additionally, I think voila only works on the notebook. I am looking into whether it can work on a python file, if not, we probably have to keep all the interactive stuff in the notebook :(

flaub: To me it seems that whether we merge the two demos or not is orthogonal to the request to split out as much logic as possible from the notebook to make it easier to review. Would it be possible to put this split out logic into a separate python file for now and then merge the two python files later if we decide that's a good idea?

I think you're right that I think voila only works on a notebook, but I don't understand why it can't work with code that is imported from a python file? Basically it seems like the only thing we need in the notebook is something like:

```
import demo
demo.run()
```

pmlc/target/demo/pipeline.cc
```diff
@@ -0,0 +1,47 @@
...
+void addToPipeline(OpPassManager &pm) {
+  pm.addPass(pmlc::dialect::tile::createComputeBoundsPass());
+  pm.addNestedPass<FuncOp>(createCanonicalizerPass());
+  pm.addNestedPass<FuncOp>(createCSEPass());
+
+  pm.addPass(pmlc::conversion::tile_to_pxa::createLowerTileToPXAPass());
+  pm.addNestedPass<FuncOp>(createCanonicalizerPass());
+  pm.addNestedPass<FuncOp>(createCSEPass());
+
+  // TODO: do optimizations here
```
flaub: This comment doesn't make sense in this context

demos/mlsys_2020/lowering_demo.ipynb
cnamrata15: Makes sense, I will separate the logic into a python file.

conda/unix.yml
```diff
@@ -14,6 +14,7 @@ dependencies:
     - autodoc
     - bokeh
     - breathe
+    - cairosvg
```
dgkutnic: Excuse me if I'm missing something, but how exactly are we using `cairosvg`?

As far as I can know, `cairosvg` is for converting SVG images into png/pdf form (`drawSVG` has APIs that use cairo) but converting to these formats prevents any animations from rendering. Non-animated SVGs should be able to be displayed just fine alongside normal jupyter widgets with the `DrawingWidget` object, which skips the extra step of converting to another filetype.

cnamrata15: Yes, you are right, we are not using cairosvg. But drawsvg was generating a multi-line warning in red, which did not look good. Now that we are moving things to the python file I can possibly remove it.
Thanks for pointing it out.

demos/mlsys_2020/demo_handler.py
```diff
@@ -0,0 +1,137 @@
...
+
+contraction_dropdown.observe(contraction_dropdown_handler, names='value')
+
+contraction_op_widget = widgets.VBox([
+    widgets.HBox([widgets.Label(value="Pick an operation:"), contraction_dropdown]),
+    widgets.HTML(value="<code> \
+        def edsl_program(X, Y): <br>\
+        </code>"),
+    widgets.VBox([
+        widgets.HTML(value="<code> \
```
flaub: I think you can use a multi-line literal here, like:
```
"""
<code>
stuff goes here
</code>
"""

File: demos/mlsys_2020/demo_handler.py [hunk 12]
```diff
@@ -0,0 +1,137 @@
+from demo_source import *
```
flaub: It's generally frowned upon in the python world to use `from module import *`.

https://docs.python.org/2/tutorial/modules.html#more-on-modules

> Note that in general the practice of importing * from a module or package is frowned upon, since it often causes poorly readable code. However, it is okay to use it to save typing in interactive sessions.

demos/mlsys_2020/demo_handler.py
```diff
@@ -0,0 +1,137 @@
...
+
+myDemo = Demo()
```
flaub: I'm wondering if we can put all of this code behind a function, so that we limit the amount of top-level definitions?

Some reasons for this are:
1. It reduces the amount of top-level definitions and thus makes it less likely to have collisions.
2. Being able to import this module can avoid the frowned upon `import module from *` form. You can instead say
```
import demo
demo.run()
```

demos/mlsys_2020/lowering_demo_handler.py
```diff
@@ -0,0 +1,123 @@
...
+hbox_layout.width = '100%'
+hbox_layout.justify_content = 'space-around'
+
+# do this in a flexbox-friendlier way
+edsl_title = widgets.HTML(value='<div style="text-align:center"><h1>EDSL Demo</h1></div>')
+
+edsl_subdemo = widgets.HBox([left])
+edsl_subdemo.layout = hbox_layout
+
+edsl_demo = widgets.VBox([edsl_title, edsl_subdemo])
```
flaub: Missing newline

demos/mlsys_2020/demo_handler.py
```diff
@@ -0,0 +1,137 @@
...
+
+# Functional widget: Text Box
+op_textbox = lambda x: widgets.Text(value=list(myDemo.edsl_snippets[x].values())[0],
```
flaub: I think we should try to pick one naming scheme, like `snake_case` or `camelCase` but not a mix of both.

dgkutnic:
Hey @flaub can we land this as-is and push the formatting changes in a later PR? I've got some visual/UI changes I'd like to get in that depend on this to land, and I can go and make these changes as well. That'll also free up @cnamrata15's time to work on error handling.

demos/mlsys_2020/demo_handler.py
cnamrata15: Good point. I agree with you.

PR #936, opened 2020-02-22
Revive windows and get lit tests to run

flaub:
Note that `pmlc-jit` still doesn't work on Windows due to symbol resolution errors. However, almost all other lit tests now run (and pass!). There's an issue with python lit tests, but since we have coverage on other platforms, this should be OK. Also at the moment xsmm and vulkan are both disabled since we need to make some changes to get them to build on windows.

Also note that on windows, the following bazel flags are required now:

```
--enable_runfiles
--test_env=PATHEXT
```


earhart:
Let's talk about whether we should use `--enable_runfiles`.

PR #939, opened 2020-02-25
Lowering of xsmm.gemm to std.call

flaub:


pmlc/dialect/pxa/analysis/strides.cc
```diff
@@ -183,4 +221,46 @@ Optional<StrideInfo> computeStrideInfo(pmlc::dialect::pxa::AffineReduceOp op) {
...
+    if (item.index())
+      os << ", ";
+    os << item.value();
+  }
+  os << ']';
+}
+
+Optional<StrideArray> computeStrideArray(AffineMap map) {
+  std::vector<SmallVector<int64_t, 8>> flat;
+  if (failed(getFlattenedAffineExprs(map, &flat, nullptr)))
```
jbruestle: I think this will do something incorrect if there are any ceildivs, etc.  I read the related code, and it looks like the class is uses (SimpleAffineExprFlattener) adds 'locals' as needed to handle modulus, etc, but then the function call that constructs the class ignores the induced locals.  So I think it will silently produce garbage data (probably an oversight?). Might be worth a test.  The correct behavior for use is to fail on ceildiv, etc.

pmlc/target/x86/xsmm_lowering.cc
```diff
@@ -0,0 +1,385 @@
...
+      ArrayRef<NamedAttribute> attrs{};
+      builder.create<FuncOp>(loc, symbol, funcType, attrs);
+      return SymbolRefAttr::get(symbol, context);
+    }
+
+    PreparedOperand prepareOperand(Value operand, AffineMap accessMap,
+                                   ValueRange mapOperands, AffineMap tileMap,
+                                   unsigned leadingDimPos) {
+      ArrayRef<Value> empty{};
+      auto offsets = expandAffineMap(rewriter, loc, accessMap, mapOperands);
```
jbruestle: I'm think there might be an easier way to do this, not a blocker for now, although it brings in a lot of copied code.

flaub: My plan is to make this code accessible upstream so that we don't need to duplicate it. But this requires a patch upstream.

PR #942, opened 2020-02-27
Initial affine-stencil pass.

flaub:
(Very rough for now).


pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -0,0 +1,160 @@
...
+    IVLOG(3, "Not a Gemm match.");
+    return;
+  }
+}
+
+void StencilPass::runOnFunction() {
+  auto func = getFunction();
+  auto context = &getContext();
+  func.walk([&](mlir::AffineParallelOp op) {
+    Stencil as;
```
jbruestle: Since the lifetime of the Stencil object is one per op, we should probably make the op be a constructor parameter.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -0,0 +1,160 @@
...
+    as.DoStenciling(op, context);
```
jbruestle: No need to pass context, since almost everything in MLIR can access is (ie op.getContext() works here).

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -0,0 +1,160 @@
...
+};
+
+mlir::Optional<GemmOperationMatch> Stencil::getGemmOperation(mlir::AffineParallelOp op, mlir::MLIRContext* mlirContext) {
+  mlir::Optional<GemmOperationMatch> ret;
+  Builder builder(mlirContext);
+  auto *body = op.getBody();
+  // Get the instructions in the body and match for load, load, mulXXX, reduce add operations.
+  // For everything else we fail.
+  if (body->getOperations().size() != kNumValidInstrInGemmRegion) {
+    op.emitRemark("the ParallelOp region didn't have the right number of instructions for a Gemm");
```
jbruestle: Remarks like this are very helpful, but we should have a mechanism to only enable them during some level of debugging/verbosity.  We could use vlog level.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -0,0 +1,160 @@
...
+      lhs = llvm::dyn_cast_or_null<mlir::AffineLoadOp>(mulfOp.lhs().getDefiningOp());
+      rhs = llvm::dyn_cast_or_null<mlir::AffineLoadOp>(mulfOp.rhs().getDefiningOp());
+    }
+    else if (muliOp != nullptr) {
+      mulOpType = MulOperationType::IntTy;
+      lhs = llvm::dyn_cast_or_null<mlir::AffineLoadOp>(muliOp.lhs().getDefiningOp());
+      rhs = llvm::dyn_cast_or_null<mlir::AffineLoadOp>(muliOp.rhs().getDefiningOp());
+    }
+    else {
+      op.emitRemark("unhandled multiplication type in stenciling");
```
jbruestle: Isn't this code unreachable given the early check on mulfOp == nullptr && muliOp == nullptr?

pmlc/dialect/pxa/transforms/stencil.cc
flaub: I agree with @jbruestle  but would also add that we should probably do one or the other of IVLOG and remarks, probably not both.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Now the data is part of the Stencil object.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Passed to the constructor of Stencil now.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: You are right... In general, I like to have reminders in code when it is likely to add more cases (like BF6 mul op for example at some point). And have the run/code tell me what is wrong instead of spending time trying to figure out what went wrong. Took it out, though...

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -0,0 +1,703 @@
...
+    mlir::MulFOp mulfOp  = llvm::dyn_cast_or_null<mlir::MulFOp>(valDef);
+    mlir::MulIOp muliOp  = llvm::dyn_cast_or_null<mlir::MulIOp>(valDef);
+    if (mulfOp == nullptr && muliOp == nullptr) {
+      op.emitRemark("The source of the reduce is not a multiplication operation");
+      return;
+    }
+
+    mlir::AffineLoadOp lhs;
+    mlir::AffineLoadOp rhs;
+    if (mulfOp != nullptr) {
```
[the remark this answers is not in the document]
LLITCHEV: I always preferred to have "in my case" types in the checks, but sure...

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Using IVLOG only.

PR #944, opened 2020-02-28
EDSL Demo Improvements

dgkutnic:


dgkutnic:
Pulled most of @flaub's changes in.

There's still plenty of work that needs to be done in terms of adding the HTML boilerplates, naming conventions, etc.

However, I did get my existing boilerplate-EDSL (matmul and sum over axis) working in both `//demos/mlsys_2020:demo` and `//demos/mlsys_2020:voila`

flaub:
Cool, glad that's working for you.

dgkutnic:
Things left to be done at this point:
* Lowering Demo: display each of the lowering steps in Voila
* Integrate the lowering demo UI code into `app.py` and create a parameter `demo_name` within `app.run()` such that if you pass in "edsl" then it displays the EDSL demo, and if you pass in "mlir" then it displays the MLIR.
* Modify each of the jupyter notebooks so that the `app.run()` call becomes `app.run(demo_name="edsl")` and `app.run(demo_name="mlir")` for the edsl and lowering demos respectively.
* Error handling

flaub:
We could also just have two different top level run functions...

dgkutnic:
> We could also just have two different top level run functions...

We could have two separate functions, but this functionality stays the same across both demos and should be shared by both demos (perhaps by creating a separate function that initializes this, which is then called by both demos):
image

flaub:
It's useful to think of UIs as a collection of components. If we split the UI into components, we can see that the image you posted is just a component. Then, each application can just decide which components it wants to integrate together.

PR #945, opened 2020-02-28
add conversions in stdx dialect from float to signed & unsigned int 

mars1026:


flaub:
How's this coming @mars1026? It looks like there are some failing tests.

pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
```diff
@@ -345,6 +345,42 @@ struct CmpXchgLowering : public LoadStoreOpLowering<stdx::AtomicRMWOp> {
...
+  using LLVMLegalizationPattern<stdx::FPToSIOp>::LLVMLegalizationPattern;
+  using Base = LLVMLegalizationPattern<stdx::FPToSIOp>;
+
+  PatternMatchResult
+  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (auto fptosiOp = cast<stdx::FPToSIOp>(op)) {
+      auto value = fptosiOp.getOperand();
+      auto stdxType = fptosiOp.getOperation()->getResult(0).getType();
+      auto llvmType = typeConverter.convertType(stdxType);
```
flaub: Actually, we don't need to do this, this should already be done via the `operands` being passed to `matchAndRewrite`.

pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
```diff
@@ -345,6 +345,42 @@ struct CmpXchgLowering : public LoadStoreOpLowering<stdx::AtomicRMWOp> {
...
 };
 
+struct FPToSILowering : public LLVMLegalizationPattern<stdx::FPToSIOp> {
+  using LLVMLegalizationPattern<stdx::FPToSIOp>::LLVMLegalizationPattern;
+  using Base = LLVMLegalizationPattern<stdx::FPToSIOp>;
+
+  PatternMatchResult
+  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (auto fptosiOp = cast<stdx::FPToSIOp>(op)) {
```
flaub: `cast` can never fail, it only has an `assert` internally. But I don't think we need to actually downcast in this case.

pmlc/dialect/stdx/ir/ops.cc
```diff
@@ -91,6 +91,80 @@ LogicalResult verifyAtomicRMWYieldOp(AtomicRMWYieldOp op) {
...
+  Type srcType, dstType;
+  return failure(parser.parseOperand(srcInfo) ||
+                 parser.parseOptionalAttrDict(result.attributes) ||
+                 parser.parseColonType(srcType) ||
+                 parser.resolveOperand(srcInfo, srcType, result.operands) ||
+                 parser.parseKeywordType("to", dstType) ||
+                 parser.addTypeToList(dstType, result.types));
+}
+
+LogicalResult verifyFPToSIOp(FPToSIOp op) {
```
flaub: I think we can do this verification declaratively with ODS. In other words, if we tweak the acceptable types for the `arguments` and `results` in the `.td` file, then this will generate logically equivalent C++ verify code.

pmlc/dialect/stdx/ir/ops.td
```diff
@@ -42,4 +42,26 @@ def AtomicRMWYieldOp :
   let arguments = (ins AnyType:$result);
 }
 
+def FPToSIOp :
+    StdX_Op<"fptosi"> {
+  let summary = "cast from floating-point to signed integer type";
+  let arguments = (ins AnyType:$result);
+  let results = (outs AnyType);
```
flaub: Use something like `AnyFloatType` for the arguments and `AnyIntegerType` for the result.

pmlc/dialect/stdx/ir/ops.td
```diff
@@ -42,4 +42,26 @@ def AtomicRMWYieldOp :
   let arguments = (ins AnyType:$result);
 }
 
+def FPToSIOp :
+    StdX_Op<"fptosi"> {
```
flaub: We need a `NoSideEffect` trait.

pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
```diff
@@ -352,14 +352,11 @@ struct FPToSILowering : public LLVMLegalizationPattern<stdx::FPToSIOp> {
...
                   ConversionPatternRewriter &rewriter) const override {
-    if (auto fptosiOp = cast<stdx::FPToSIOp>(op)) {
-      auto value = fptosiOp.getOperand();
-      auto stdxType = fptosiOp.getOperation()->getResult(0).getType();
-      auto llvmType = typeConverter.convertType(stdxType);
-      rewriter.replaceOpWithNewOp<LLVM::FPToSIOp>(op, llvmType, value);
-      return matchSuccess();
-    }
-    return matchFailure();
+    auto value = op->getOperand(0);
```
flaub: If you want access to a particular post-converted operand, you can use the Adapter class, generated by ODS (i.e. tablegen).

flaub:
@mars1026 Do you want to land this as is? I can approve if so, or did you want to try to do more?

PR #946, opened 2020-02-28
Lowering Demo Functionality

cnamrata15:
Its in between modifications for Denise to look at.


dgkutnic:
Hey @flaub,

I got Namrata's MLIR demo working -- running the lowering demo now displays three separate textareas that show the `tile`, `convert-pxa-to-affine`, and `lower-affine` passes and update each time someone clicks Run.

PR #951, opened 2020-03-03
Stenciling collects tensors and stride info

LLITCHEV:


pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -69,6 +81,7 @@ mlir::Optional<GemmOperationMatch> Stencil::getGemmOperation() {
     // Not check the reduceOp aggregation.
     if (reduceOp.agg() != AggregationKind::add) {
       IVLOG(3, "the reduce operation is not addition");
+      ret = false;
       return;
```
jbruestle: This should be "return false", not sure how the compiler even allowed this.

jbruestle: Same for 3 cases below.

flaub: This is inside an `op.walk` lambda. I think you can return `WalkResult::interrupt()` and it looks like the return type of `walk` is templated, so you should be able to return a bool directly.

flaub:
Also, please run clang-format/cpplint on this file. This should happen automatically for you if you've installed the pre-commit hook.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -107,33 +121,76 @@ mlir::Optional<GemmOperationMatch> Stencil::getGemmOperation() {
...
   }
+
+  if (!CollectTensors())
+    return;
+
+  if (!ComputeStrideInfo())
+    return;
+
+  Builder builder(op.getContext());
+  op.setAttr("is_gemm", builder.getUnitAttr());
```
flaub: There's no reason to create a builder just to create a `UnitAttr`. You can call `UnitAttr::get()`

pmlc/dialect/pxa/transforms/stencil.cc
flaub: Here's what we should do, use the `m_Op` matcher found in `mlir/IR/Matchers.h`

Use it like:
```
auto pattern = m_Op<ReduceOp>(m_Op<AddFOp>(m_Op<MulFOp>(a, b), c));
bool ok = pattern.match(&ops.back());
```

LLITCHEV: Remove the walk all together after talking to Jeremy. The pattern we are detecting is very strict, so no need to look for the ReduceOp. We can just get the operation before last (terminator) in the ParallelOp region and check to see if it is ReduceOp.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Not getting the builder anymore and using the static factory.

LLITCHEV:
Ran the formatting.

LLITCHEV:
Resolved upstream conflicts.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -38,102 +34,155 @@ class Stencil {
...
+
+  // Found Gemm operation data
+  MulOperationType mulOpType;
+  mlir::AffineLoadOp in1Op;
+  mlir::AffineLoadOp in2Op;
+  AffineReduceOp outOp;
+
 public:
-  explicit Stencil(mlir::AffineParallelOp opIn) : op(opIn) {}
+  explicit Stencil(mlir::AffineParallelOp opIn) : op(opIn){};
```
flaub: drop the ending semicolon, cpplint catches this.

Can you please install the `pre-commit` hooks?

pmlc/dialect/pxa/transforms/stencil.cc
flaub: Do you think we could use the `m_Op` matcher? Seems like a really clean way to do complex matching.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Installed the pre-commit package and cpplint as well.
Ran them and submitted the changes.

PR #960, opened 2020-03-07
Add best-tile searching code

LLITCHEV:


pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -253,8 +295,210 @@ void Stencil::CollectStrideOneIndices() {
...
+  IVLOG(3, "Outer: loop = " << tot_outer_loop);
+
+  for (auto &kvp : outer_idxs) {
+    if (kvp.second > 1) {
+      IVLOG(3, kvp.first << ": " << kvp.second);
+    }
+  }
+
+  unsigned outer_batches =
+      (tot_outer_loop - 1) / std::thread::hardware_concurrency() + 1;
```
flaub: The number of threads should be a parameter, not something that we probe on the current machine. That's because we want to be able to support AOT use cases.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -253,8 +295,210 @@ void Stencil::CollectStrideOneIndices() {
...
+
+bool Stencil::IndexExists(mlir::BlockArgument idx, unsigned tensor_idx) {
+  return tensorIdxs[tensor_idx].find(idx) != tensorIdxs[tensor_idx].end();
+}
+
+// Confirm if idx exists in the right place
+bool Stencil::ValidateIndexExistance(mlir::BlockArgument idx,
+                                     unsigned matrix_idx) {
+  switch (matrix_idx) {
+  case 0: {
```
flaub: Drop extra brace `{`, no variables are being declared so not needed.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -253,8 +295,210 @@ void Stencil::CollectStrideOneIndices() {
...
 
+// Test if idx in tensors[tensor_idx] is stride one index
+bool Stencil::IsStrideOne(mlir::BlockArgument idx, unsigned tensor_idx) {
+  return strideOne[tensor_idx].find(idx) != strideOne[tensor_idx].end();
+}
+
+// Test if idx in the tensors are stride one
+bool Stencil::ValidateStrideOne(mlir::BlockArgument idx, unsigned matrix_idx) {
+  switch (matrix_idx) {
+  case 0: {
```
flaub: Drop extra brace `{`

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Fixed.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Added a way to set it internally. Need to decide on passing the parameter from the outside. I'd say using a compiler option, but we can discuss this.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Taken the extra braces out.

pmlc/dialect/pxa/transforms/stencil.cc
flaub: We should try to use instance-specific options over global ones, see this: https://mlir.llvm.org/docs/WritingAPass/#instance-specific-pass-options

For testing it's probably not a big deal to use global options, provided there's a way to configure the option programmatically from a pipeline definition.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -253,8 +302,209 @@ void Stencil::CollectStrideOneIndices() {
...
+// Test if idx in tensors[tensor_idx] is stride one index
+bool Stencil::IsStrideOne(mlir::BlockArgument idx, unsigned tensor_idx) {
+  return strideOne[tensor_idx].find(idx) != strideOne[tensor_idx].end();
+}
+
+// Test if idx in the tensors are stride one
+bool Stencil::ValidateStrideOne(mlir::BlockArgument idx, unsigned matrix_idx) {
+  switch (matrix_idx) {
+  case 0:
+    // Test if M is stride one for B(3) and C(2)
```
jbruestle: This code seems correct, but the comment doesn't make sense to me.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -253,8 +302,209 @@ void Stencil::CollectStrideOneIndices() {
...
+    return IsStrideOne(idx, tensorsOrder[1]) &&
+           IsStrideOne(idx, tensorsOrder[2]);
+  case 1:
+    // N is not restricted for stride one
+    return true;
+  case 2:
+    // Test if K is stride one for A(0)
+    return IsStrideOne(idx, tensorsOrder[0]);
+  default:
+    assert(!"Wrong matrix_idx");
```
jbruestle: This should probably be llvm_unreachable along with the other assert.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -253,8 +302,209 @@ void Stencil::CollectStrideOneIndices() {
...
+  for (unsigned i = range; i > 0; --i) {
+    if (onlyEven[idx] && (range % i != 0)) {
+      continue;
+    }
+    tiles[idx] = i;
+    SearchTiles(idx + 1);
+  }
+}
+
+int64_t Stencil::idxRange(mlir::BlockArgument idx) {
```
jbruestle: We should probably cache or precompute this range.  Right now, we are recomputing this every time in the near innermost loop, and getConstantRanges is quite complex (involving affine simplification, etc).  I'm generally against early optimizations before measuring perf to be sure we care, but given that everything else is precomputed (is stride 1, etc), this one really stands out.  I think it's likely a relic of the fact the the range was just a simple constant in the pre-Affine version.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Added code to use thye instance options.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Fixed

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Fixed.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Caching in the Stencil object.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -253,8 +304,204 @@ void Stencil::CollectStrideOneIndices() {
...
+  if (opConstRanges != llvm::None) {
+    return (*opConstRanges)[idx.getArgNumber()];
+  }
+  return -1;
```
jbruestle: This should be an assertible condition, returning -1 here makes no sense.  Also, we probably shouldn't even use an optional as the member variable, since if the getConstantRanges() ranges fail, we exit early.  We should probably make opConstRanges just a SmallVec, and if getConstantRanges() succeeds, assign it by value.  Might also be good to assert that idx. getArgNumber() < opConstRanges.size().

LLITCHEV: Not using the Optional anymore.

pmlc/dialect/pxa/tests/stencil.mlir
```diff
@@ -47,7 +47,37 @@ module {
   // CHECK: affine.parallel
   // CHECK-NOT: is_gemm
 
-// CHECK-LABEL: @gemm_operation_rewrite_i32
+  // CHECK-LABEL: @no_gem_no_stride_one_1
+  func @no_gem_no_stride_one_1(%arg0: memref<100x100xf32>, %arg1: memref<100x100xf32>, %arg2: memref<100x100xf32>) {
```
flaub: typo: gemm

pmlc/dialect/pxa/tests/stencil.mlir
```diff
@@ -47,7 +47,37 @@ module {
...
+      pxa.reduce add %2, %arg2[%i, %j] : memref<100x100xf32>
+      "affine.terminator"() : () -> ()
+    }
+    return
+  }
+  // CHECK: affine.parallel
+  // CHECK-NOT: is_gemm
+
+  // CHECK-LABEL: @no_gem_no_stride_one_2
+  func @no_gem_no_stride_one_2(%arg0: memref<100x100xf32>, %arg1: memref<100x100xf32>, %arg2: memref<100x100xf32>) {
```
flaub: typo: gemm

pmlc/dialect/pxa/tests/stencil.mlir
```diff
@@ -47,7 +47,37 @@ module {
...
+      %1 = affine.load %arg0[%k, 2*%j] : memref<100x100xf32>
+      %3 = addf %0, %1 :f32
+      %2 = mulf %0, %3 : f32
+      pxa.reduce add %2, %arg2[%i, %j] : memref<100x100xf32>
+      "affine.terminator"() : () -> ()
+    }
+    return
+  }
+  // CHECK: affine.parallel
+  // CHECK-NOT: is_gemm
```
flaub: Can we use some other approach rather than adding an attribute? I don't think this is a good long-term approach. Since this is only for testing, you could use a remark or just actually lower this to a `xsmm.gemm` op.

pmlc/dialect/pxa/transforms/passes.h
```diff
@@ -13,10 +15,24 @@ class OpPassBase;
...
+      llvm::cl::desc("Specifies number of threads for the stencilling pass")};
+};
+
 struct StencilPass : public mlir::FunctionPass<StencilPass> {
+  StencilPass() = default;
+  StencilPass(const StencilPass &) {}
+  explicit StencilPass(const StencilPassOptions &options);
+
+  Option<int> numberOfThreadsOption{
+      *this, "affine-stencil-number-threads",
```
flaub: No need to prefix `affine-stencil`, just use `threads`.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -253,8 +304,202 @@ void Stencil::CollectStrideOneIndices() {
...
+
+void Stencil::SetNumberThreads() {
+  if (numThreads == 0) {
+    numThreads = std::thread::hardware_concurrency() + 1;
+  }
+}
+
 void Stencil::DoStenciling() {
   // Initialization
+  SetNumberThreads();
```
flaub: We should construct the `Stencil` object with whatever options it needs. And we should have a `Stencil` object per pass.

Basically, the `Stencil` object should be setup by construction.

pmlc/dialect/pxa/transforms/passes.h
```diff
@@ -13,10 +15,24 @@ class OpPassBase;
...
 
+struct StencilPassOptions
+    : public mlir::PassPipelineOptions<StencilPassOptions> {
+  Option<int> numberOfThreadsOption{
+      *this, "affine-stencil-number-threads",
+      llvm::cl::desc("Specifies number of threads for the stencilling pass")};
+};
+
 struct StencilPass : public mlir::FunctionPass<StencilPass> {
+  StencilPass() = default;
```
flaub: I don't think we should allow an 'unconfigured' pass. Why not force the number of threads to be specified always?

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -60,21 +60,72 @@ class Stencil {
...
+  // The throughput and startup cost of M*N*K matrix multiplication
+  std::pair<double, unsigned> Throughput(unsigned m, unsigned n, unsigned k);
+  // Evaluate the performance of the current searching state
+  double Evaluate();
+
+  // Set the number of threads to a default value if not set from outside.
+  void SetNumberThreads();
+
+  // Number of threads
+  unsigned numThreads;
```
flaub: There's no need for a separate variable, just use the `numberOfThreadsOption` directly (and rename it to `numThreads`)

flaub:
What's the difference between a `Stencil` and a `StencilPass`? Can we just put all the state needed on the `StencilPass`?

LLITCHEV:
Stencil pass is a pass object. Instantiated once per compiler (compilation). Stencil is an object that is instantiated for each AffineParallelOp. It keeps the state for one op only and does the stenciling analysis for that op only. Having said that, I think we should keep the current separation.

flaub:
Could we make a Pass that only works on AffineParallelOp?

https://mlir.llvm.org/docs/WritingAPass/#operationpass--op-specific

The example uses a `FuncOp` but there's no reason it can't be a `AffineParallelOp`

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: The option is on the compiler pass object and this var is on the Stencil POD that keeps the state for a single AffineParallelOp (see comment below).
I can't really use the same variable (unless I instantiate the whole option object or the option itself in the and as part of the Stencil object. I think it is much simpler and clear to keep simple objects asnd types in a POD object.
I renamed the option, but kept the variable in Stencil. I need something there anyway. The original one is on the StencilPass.

pmlc/dialect/pxa/transforms/passes.h
LLITCHEV: I can take it. If you don't instantiate the pass with explicit option value, the default is created, which is zero - or option not specified. SOeone might just want to let the RT of the OS or the stenciling pass decide what to do with that. I think that is how people will use it in most of the cases, but you still have an option to override the default.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: So, the design we talked over with Jeremy is that the Stencil object correspond to a AffineParallelOp. So, there are many Stencil-ers per pass - one for each AffineParralelOp. And each Stencil stencils a single AffineParallelOp.

pmlc/dialect/pxa/transforms/passes.h
LLITCHEV: Fixed

pmlc/dialect/pxa/tests/stencil.mlir
LLITCHEV: We probably will end up there. Once the XSMM is integrated. This is code that is needed to do that. I expect these tests to change.

flaub: SGTM

pmlc/dialect/pxa/tests/stencil.mlir
LLITCHEV: Fixed.

pmlc/dialect/pxa/transforms/passes.h
flaub: My feeling is that we should prefer explicit over implicit. Configuration of passes should all be explicitly defined someplace so that we don't 'accidentally' run into good or bad behavior.

pmlc/dialect/pxa/tests/stencil.mlir
LLITCHEV: Fixed and all the rest

pmlc/dialect/pxa/transforms/registration.cc
```diff
@@ -31,8 +31,13 @@ static mlir::PassRegistration<TestStrideInfoPass>
...
-static mlir::PassRegistration<StencilPass>
+static mlir::PassPipelineRegistration<StencilPassOptions>
     stencil_pass("affine-stencil",
-                 "Stencil the inputs to create \"micro\" GEMM operations.");
+                 "Stencil the inputs to create \"micro\" GEMM operations.",
+                 createStencilPass);
+// Lubo static mlir::PassRegistration<StencilPass>
+// Lubo     stencil_pass("affine-stencil",
+// Lubo                  "Stencil the inputs to create \"micro\" GEMM
+// operations.");
```
flaub: Remove this before you land

pmlc/dialect/pxa/transforms/stencil.cc
flaub: This seems orthogonal to the request to construct a `Stencil` explicitly with the numberOfThreads set instead of setting it after the `Stencil` object has been constructed.

pmlc/dialect/pxa/transforms/stencil.cc
flaub: This still seems to use the `std::thread::hardware_concurrency()`, is this intentional?

pmlc/dialect/pxa/transforms/registration.cc
LLITCHEV: Removed.

LLITCHEV:
About making the pass work on AffineParallelOp only... Just adding what we discussed off-line last night, since it could be useful for others.
The OperationPass has significant limitations due to the nature of optimizations being done. It operates only on operations attributed as IsolatedFromAbove. AffineParallelOp is not such operation, and so we can't use OperationPass in this case.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: The threads option is initialized from outside now.

pmlc/dialect/pxa/transforms/passes.h
LLITCHEV: Taken out.

LLITCHEV:
@jbruestle Oh... Misse that dup code completely. Removed and per your suggestion, added the thread init code to runOnFunction method.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -270,19 +509,46 @@ void Stencil::DoStenciling() {
...
 }
 
-std::unique_ptr<mlir::Pass> createStencilPass() {
-  return std::make_unique<StencilPass>();
+StencilPass::StencilPass(const StencilPassOptions &options) {
+  numThreads.setValue(options.numThreads);
+}
+
+void createStencilPass(mlir::OpPassManager &pm,
+                       const StencilPassOptions &options) {
```
flaub: I don't think we need a separate options object, can we just pass `numThreads` here as an argument?

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -270,19 +509,46 @@ void Stencil::DoStenciling() {
...
+  }
+
   op.setAttr("is_gemm", mlir::UnitAttr::get(op.getContext()));
 }
 
 void StencilPass::runOnFunction() {
   auto func = getFunction();
   func.walk([&](mlir::AffineParallelOp op) {
-    Stencil as(op);
+    unsigned threads = numThreads.getValue();
```
flaub: Let's do this in the constructor of the StencilPass, it should be constructed with a value (never empty).
This isn't an `Optional` (which is possible to hold no value), but an `Option` (which I argue should always hold a value, and I'd be surprised if it can even hold a nothing value).

pmlc/dialect/pxa/transforms/passes.h
```diff
@@ -13,10 +15,23 @@ class OpPassBase;
 
 namespace pmlc::dialect::pxa {
 
+struct StencilPassOptions
```
flaub: I'm not sure we even need a separate Options object. But even if we did, I wouldn't have it use the `Option<unsigned> numThreads`, this should only be confined to options that are directly on a pass so that the parse pipeline infrastructure can bind command line arguments to the pass.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -270,19 +509,46 @@ void Stencil::DoStenciling() {
...
   });
 }
 
-std::unique_ptr<mlir::Pass> createStencilPass() {
-  return std::make_unique<StencilPass>();
+StencilPass::StencilPass(const StencilPassOptions &options) {
+  numThreads.setValue(options.numThreads);
+}
+
+void createStencilPass(mlir::OpPassManager &pm,
```
flaub: See the other pass creation functions in MLIR (and our code). They don't take in a pass manager, instead they just return a unique pointer to the pass. I think we should be consistent here.

pmlc/dialect/pxa/transforms/stencil.cc
jbruestle: I think the idea here is that '0' threads means: please pick a number of threads for me, which is reasonable enough IMO.  Although, why don't we move this code outside of the loop?

flaub: I think we should pass `numThreads.getValue()` to the ctor of `Stencil` directly. If someone wants to use hardware_concurrency, they can do it at usage time.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -60,21 +60,74 @@ class Stencil {
...
+
+  // Set the number of threads to a default value if not set from outside.
+  void SetNumberThreads();
+
+  // Number of threads
+  unsigned numThreads;
+
 public:
-  explicit Stencil(mlir::AffineParallelOp opIn) : op(opIn) {}
+  explicit Stencil(mlir::AffineParallelOp opIn, int numThreadsIn)
```
flaub: no need to name is something different, just go with `numThreads` for the parameter name (even tho it 'collides' with the member name, C++ will sort it out).

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: I do foresee adding more options here. Passing the generic Options object allows us to pull all the things will need from it. Besides, the call to create the pass (creating a pass with instance Options is much different than the one with no options) is created like the following (much different than the pass with no options - for the options to be parsed and processed the pass is added to the OpPassManager)...
pm.addPass(std::make_unique<StencilPass>(options));
In no option pass, the pass is created only by doing std::make_unique<PadPass>();
The idiomatic boilerplate code for creating and instantiating a pass with instance specific options comes from the MLIR doc - https://github.com/llvm/llvm-project/blob/master/mlir/docs/WritingAPass.md#instance-specific-pass-options

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Moved.

pmlc/dialect/pxa/transforms/registration.cc
```diff
@@ -31,8 +31,9 @@ static mlir::PassRegistration<TestStrideInfoPass>
     regTestStrideInfo("test-stride-info",
                       "Report stride data for all loads/stores for unit tests");
 
-static mlir::PassRegistration<StencilPass>
+static mlir::PassPipelineRegistration<StencilPassOptions>
```
flaub: I think we can register a pass without having to register a pipeline...

pmlc/dialect/pxa/transforms/passes.h
LLITCHEV: The registration is taking an extra template parameter that is the Options object. The parser fills these struct with the parsed values.
The MLIR doc describing this is at https://github.com/llvm/llvm-project/blob/master/mlir/docs/WritingAPass.md#instance-specific-pass-options

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: In the MLIR code I could find a single pass that uses instance options (all the rest use static ones, which I mentioned up in this review). The pass is TestOptionsPass in TestPassManager.cpp.
From the MLIR doc on instance pass it looks like you need to add the pass to the PM and that would cause the options to be parsed.
The doc is at https://github.com/llvm/llvm-project/blob/master/mlir/docs/WritingAPass.md#instance-specific-pass-options

pmlc/dialect/pxa/transforms/stencil.cc
flaub: So I think we want to register a pass instead of a pipeline here. The first example in that doc is for passes, the second example is for pipelines.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Old habits die hard. :) To me it is more clear this way. But sure...

pmlc/dialect/pxa/transforms/registration.cc
LLITCHEV: It could be. I just followed the docs on Instance Pass Options at https://github.com/llvm/llvm-project/blob/master/mlir/docs/WritingAPass.md#instance-specific-pass-options
That is apparently how they do it.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -270,19 +509,46 @@ void Stencil::DoStenciling() {
...
+
   op.setAttr("is_gemm", mlir::UnitAttr::get(op.getContext()));
 }
 
 void StencilPass::runOnFunction() {
   auto func = getFunction();
+  unsigned threads = numThreads.getValue();
   func.walk([&](mlir::AffineParallelOp op) {
-    Stencil as(op);
+    if (threads == 0) {
```
jbruestle: Sorry to continue to nitpick, but why is the zero test/replace still inside thee loop?  It only needs to be done once...

pmlc/dialect/pxa/transforms/registration.cc
flaub: Unfortunately the docs to mention how to register a pass with instance-specific options, but it is a distinct thing from a pipeline with instance-specific options.

See the `TestPassManager.cpp` in the MLIR codebase

pmlc/dialect/pxa/transforms/stencil.cc
jbruestle: Yes, there are two different examples in that section of the docs, passes and pass pipelines, only the first one is relevant for us.  Certainly the wording could have been clearer in the docs, or perhaps a section break would be helpful, I misread that one the first time too...

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Yeah, and it will be hoisted out anyway. I don't worry too much about these things. Contemporary compiler have a very good hoisters (especially the SSA based ones). Will move it out, though...

pmlc/dialect/pxa/transforms/registration.cc
LLITCHEV: Using that now.

pmlc/dialect/pxa/transforms/stencil.cc
flaub: I think we should make it an error if the number of threads is not explicitly specified. This is because, if someone writes a `.mlir` test and starts checking that the tiling proceeds in a certain way, then it may work for them due to having say 8 cores, but it will fail on someone else's machine.

jbruestle: Personally, I think having a zero-threads == system # of threads is fine, and in production you almost certainly want a way to specify system # of threads.  Regarding hoisting though, this probably would not get hoisted since almost certainly the compiler doesn't know that hardware_concurrency always returns the same number and cannot return zero, esp since it's likely a system call and thus not applicable for inlining.  That said, I'm not actually concerned with perf optimization at a micro level, but just the fact the logically selecting the number of threads is a once per pass decision, so it should look like one.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: The checkin before the last one doesn't use the PipelinePass anymore.

pmlc/dialect/pxa/transforms/stencil.cc
flaub: I'm not saying we should never use hardware_concurrency. To be clear I think we should use it in the pipeline definition. But from the pass itself, here, I don't think we should use it. I fear that if someone writes a `.mlir` test and uses the command-line version (via `pmlc-opt`) to access this pass, then they should be forced to set `threads=`.

pmlc/dialect/pxa/transforms/passes.h
```diff
@@ -14,6 +18,13 @@ class OpPassBase;
 namespace pmlc::dialect::pxa {
 
 struct StencilPass : public mlir::FunctionPass<StencilPass> {
+  StencilPass() = default;
+  StencilPass(const StencilPass &) {}
+  // explicit StencilPass(const StencilPassOptions &options);
```
flaub: Remove this

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: Re: hoisting... This will be hoisted (at least by all 4 C++ compilers I worked on). It operates on a local variable and the compilers assume moving around is fine as long as the variable that is assigned on is not marked volatile. When measured in the past, the optimization is too good to pass it up. And it was widely adopted to allow moving and give a way out with the volatile keyword.

jbruestle: No it will *not* be hoisted unless it knows enough details about the hardware_concurrency call, which is non-const (https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency), and may be internally hit a syscall, or be in a different translation unit.  Imagine that hardware_concurrency returned the value 0 at random 90% of the time, and the value 100 10% of the time (which, unless the compiler can see it's definition, it might).  In that case, it would be reassigned to 100 only around the 10th loop on average and then it would retain it's value and never be called again, since it is inside the if clause.  So clearly it's illegal to hoist it since it would produce invalid behavior to call it only once (it may have side effects, being non-const).  Again, I don't really care about the runtime speed either way, since it's irrelevant in term of overall per, I just wanted it moves up since it's setup logic and not per-loop logic, so it's clearer that way.  Still, if you feel the need to argue the point twice, I can't let it go uncorrected.

PR #963, opened 2020-03-09
Add the heatmap to the xsmm stenciler

LLITCHEV:


.pre-commit-config.yaml
```diff
@@ -4,7 +4,7 @@ exclude: >
     bzl/version.tpl.py|
     ci/pipeline.yml|
     plaidbench/plaidbench/networks/keras/imdb_lstm.h5|
-    tile/targets/cpu/heatmap_skx_xeonplat_8180_1-7GHz_mblocked.csv.gz
+    pmlc/dialect/pxa/transforms/heatmap_skx_xeonplat_8180_1-7GHz_mblocked.csv.gz
```
flaub: Let's put this in `pmlc/target/x86`, since it's only useful for XSMM/x86 targets

LLITCHEV: Moved.

pmlc/target/x86/BUILD
```diff
@@ -7,7 +7,7 @@ load("//bzl:plaidml.bzl", "plaidml_cc_library")
 plaidml_cc_library(
     name = "x86",
     srcs = glob(["*.cc"]),
-    hdrs = glob(["*.h"]),
+    hdrs = glob(["*.h", "*.gz"]),
```
flaub: I don't think this is necessary

pmlc/dialect/pxa/transforms/heatmap.h
```diff
@@ -0,0 +1,14 @@
+// Copyright 2018, Intel Corporation
```
flaub: Can we move these files over to `pmlc/target/x86` too?

pmlc/dialect/pxa/transforms/BUILD
```diff
@@ -32,3 +36,10 @@ plaidml_cc_library(
     deps = [":lib"],
     alwayslink = 1,
 )
+
+heatmap(
```
flaub: I'd prefer to keep all of the x86 specific things over in `pmlc/target/x86`

pmlc/dialect/pxa/transforms/BUILD
```diff
@@ -23,6 +26,7 @@ plaidml_cc_library(
         "//pmlc/dialect/pxa/ir",
         "//pmlc/dialect/tile/ir",
         "//pmlc/dialect/eltwise/ir",
+        "//pmlc/target/x86",
```
flaub: I think we should consider how we're going to integrate the components.

We have the generic stencil pass, but it should be configured with a specific costing model. The stencil pass doesn't need to know about the costing model directly, it could take a `std::function` or even a class with virtual methods on it. That way we could have the `x86` pipeline decide that it wants to use the stencil pass with the x86-specific heatmap applied.

pmlc/target/x86/BUILD
LLITCHEV: It is. It didb't work without it. If the file is not "exported" from the package it is not accessible from other packages.
This seems to be describing it.
https://stackoverflow.com/questions/43921911/how-to-resolve-bazel-undeclared-inclusions-error

pmlc/dialect/pxa/transforms/heatmap.h
LLITCHEV: Will do.

pmlc/dialect/pxa/transforms/BUILD
LLITCHEV: OK. I had no idea what the overall structure is. I thought the xsmm will be use by this very specific XSMM/CPU stenciler. So I put it there.
I will move it.

LLITCHEV: Also, in the auto_stenciler, (the old code base) it is done the same way. This is just a call to a function from a bzl file, I think.

pmlc/target/x86/BUILD
flaub: We can export the file if needs be in another way, but not through this mechanism.

flaub: See: `exports_files`: https://docs.bazel.build/versions/master/be/functions.html#exports_files

pmlc/dialect/pxa/transforms/BUILD
LLITCHEV: We should. Let's talk about it. The heatmap is a standalone component, I think.

pmlc/target/x86/BUILD
LLITCHEV: No need to do this anymore. All is moved to the target/x86 package.

pmlc/dialect/pxa/transforms/BUILD
flaub: OK, I think we should configure the stencil pass with a cost model. You could have it be in the form of a function that takes in a tile and returns a cost.

The big deal is that I don't think we should have the PXA dialect depend on the x86 target. It should be the other way around.

LLITCHEV: Sure! I'll start prototyping things as soon as this gets in. This is just putting the heatmap in. I think we need this no matter what and this could go in... Small checkins is good...

PR #965, opened 2020-03-10
Multi gpu launch funcs to vulkan calls support (WIP)

YangleiZouIntel:
createConvertGpuLaunchFuncToVulkanCallsPass() support only single gpu/spirv module.
making changes based on this pass to support multi gpu launch funcs

### DIFF
--- a/pmlc/rt/vulkan/wrappers.cc
+++ b/pmlc/rt/vulkan/wrappers.cc
@@ -45,6 +45,10 @@ public:
     free(value[0]);
     free(value[1]);
     free(value[2]);
+
+    if (failed(vulkanRuntime.destroy())) {
+      llvm::errs() << "runOnVulkan failed";
+    }
   }
 
   void setResourceData(DescriptorSetIndex setIndex, BindingIndex bindIndex,
@@ -71,8 +75,7 @@ public:
   void runOnVulkan() {
     std::lock_guard<std::mutex> lock(mutex);
     if (failed(vulkanRuntime.initRuntime()) || failed(vulkanRuntime.run()) ||
-        failed(vulkanRuntime.updateHostMemoryBuffers()) ||
-        failed(vulkanRuntime.destroy())) {
+        failed(vulkanRuntime.updateHostMemoryBuffers())) {
       llvm::errs() << "runOnVulkan failed";
     }
   }
--- a/pmlc/rt/vulkan/wrappers.cc
+++ b/pmlc/rt/vulkan/wrappers.cc
@@ -28,10 +28,24 @@ static llvm::ManagedStatic<VulkanRuntimeManager> vkRuntimeManager;
 
 class VulkanRuntimeManager {
 public:
-  VulkanRuntimeManager() = default;
+  VulkanRuntimeManager() {
+    int64_t buffer_size = 9;
+    value[0] = reinterpret_cast<float *>(malloc(buffer_size * sizeof(float)));
+    value[1] = reinterpret_cast<float *>(malloc(buffer_size * sizeof(float)));
+    value[2] = reinterpret_cast<float *>(malloc(buffer_size * sizeof(float)));
+    for (int i = 0; i < 9; i++) {
+      *(value[0] + i) = i + 1;
+      *(value[1] + i) = i + 1;
+      *(value[2] + i) = 0;
+    }
+  }
   VulkanRuntimeManager(const VulkanRuntimeManager &) = delete;
   VulkanRuntimeManager operator=(const VulkanRuntimeManager &) = delete;
-  ~VulkanRuntimeManager() = default;
+  ~VulkanRuntimeManager() {
+    free(value[0]);
+    free(value[1]);
+    free(value[2]);
+  }
 
   void setResourceData(DescriptorSetIndex setIndex, BindingIndex bindIndex,
                        const VulkanHostMemoryBuffer &memBuffer) {
@@ -62,6 +76,7 @@ public:
       llvm::errs() << "runOnVulkan failed";
     }
   }
+  float *value[3];
 
 private:
   VulkanRuntime vulkanRuntime;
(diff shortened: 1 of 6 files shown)


pmlc/rt/vulkan/wrappers.cc
```diff
@@ -28,10 +28,28 @@ static llvm::ManagedStatic<VulkanRuntimeManager> vkRuntimeManager;
...
+    int64_t buffer_size = 9;
+    value[0] = reinterpret_cast<float *>(malloc(buffer_size * sizeof(float)));
+    value[1] = reinterpret_cast<float *>(malloc(buffer_size * sizeof(float)));
+    value[2] = reinterpret_cast<float *>(malloc(buffer_size * sizeof(float)));
+    for (int i = 0; i < 9; i++) {
+      *(value[0] + i) = i + 1;
+      *(value[1] + i) = i + 1;
+      *(value[2] + i) = 0;
+    }
+  }
```
flaub: This might be for testing, but it can't remain here long term.

flaub:
@YangleiZouIntel Here's what I was working on but I won't be able to work on it for awhile now that I'm pulled off to work on other things: https://github.com/plaidml/plaidml/pull/966

It doesn't yet work, and we might need to make changes upstream, but this is probably what we should probably try to get to work next. Specifically the pipeline for intel_gen.

brianretford:
> Why do we need to support multiple GPU modules? It seems like we just need to support multiple kernel functions in a single GPU module. I think we should focus on how to lower tile -> affine -> gpu -> SPIR-V.

I agree with the frank that we should close on basic execution as quickly as possible and start focusing on porting optimization passes and creating compiler pass configs for GEN9 until we reach parity with the pre-MLIR plaidml

pmlc/rt/vulkan/wrappers.cc
YangleiZouIntel: Yes, I was using this for test.

YangleiZouIntel:
> > Why do we need to support multiple GPU modules? It seems like we just need to support multiple kernel functions in a single GPU module. I think we should focus on how to lower tile -> affine -> gpu -> SPIR-V.
> 
> I agree with the frank that we should close on basic execution as quickly as possible and start focusing on porting optimization passes and creating compiler pass configs for GEN9 until we reach parity with the pre-MLIR plaidml

Thanks Brian and Frank, we will work with Frank on close the basic execution ASAP. 
The reason I added this multiple gpu launch call support is as follows. 
I checked the IR generated (in edsl tests) after converting to Gpu (The vulkan runtime has passes can turn the gpu with host module to SPIR-V and the run on vulkan, so it seems we only need to convet the IR to gpu module and then add passed used in vulkan rt). Right now most simple operations like addf, dotf are turned into 2 gpu modules each with single function, so I added this support to help close the basic execution of a single op. If we plan change the lowering to single gpu module with multiple funcs, then we can skip this or modify this to multi funcs support.
Thanks.

pmlc/rt/vulkan/wrappers.cc
YangleiZouIntel: I will delete this .

flaub:
We should double-check if this is still the case. I think the upstream code was changed so that the GpuToLoops pass will turn each outer loop into a `gpu.func` instead of a `gpu.module` with a `gpu.func` inside. If that's not the case, we should look into fixing that, if at all possible. Another approach would be to write a separate pass that merges multiple `gpu.module` into a single one.

YangleiZouIntel:
> We should double-check if this is still the case. I think the upstream code was changed so that the GpuToLoops pass will turn each outer loop into a `gpu.func` instead of a `gpu.module` with a `gpu.func` inside. If that's not the case, we should look into fixing that, if at all possible. Another approach would be to write a separate pass that merges multiple `gpu.module` into a single one.

Thanks, Frank. It seems plaidml/bridge/keras:backend_tests testAddElements is turned into single gpu module but tests like testConv2d is tunrned into 2 gpu modules. We will look into fixing it, maybe it is easier to merge them into a single module with multiple functions.

flaub:
OK. Could you add a `.mlir` test similar to https://github.com/plaidml/plaidml/pull/966/files#diff-432b5bf6c92a855a38df209799c4720d that demonstrates that multiple gpu modules are being generated?

YangleiZouIntel:
> OK. Could you add a `.mlir` test similar to https://github.com/plaidml/plaidml/pull/966/files#diff-432b5bf6c92a855a38df209799c4720d that demonstrates that multiple gpu modules are being generated?



> OK. Could you add a `.mlir` test similar to https://github.com/plaidml/plaidml/pull/966/files#diff-432b5bf6c92a855a38df209799c4720d that demonstrates that multiple gpu modules are being generated?

I added a conv2d.mlir test which seems to generate two gpu modules. Please double check, thanks.
https://github.com/plaidml/plaidml/pull/966/commits/07901147962a53a144400a710af376944935f2ec

flaub:
Thanks! I see it now. I'm a little surprised the `GpuKernelOutliningPass` creates multiple modules, perhaps the upstream authors don't have any tests that generate multiple kernels. It'd be interesting to see how IREE handles this situation. Perhaps they don't use the upstream outlining pass and have written their own custom one. Ideally we can just fix the exiting outlining pass to only generate a single module, I can't think of any reason this isn't possible. Perhaps it's a naming thing, in which case I'd just say we have a standard/static name for the `gpu.module` because the function name can still provide uniqueness.

YangleiZouIntel:
> Thanks! I see it now. I'm a little surprised the `GpuKernelOutliningPass` creates multiple modules, perhaps the upstream authors don't have any tests that generate multiple kernels. It'd be interesting to see how IREE handles this situation. Perhaps they don't use the upstream outlining pass and have written their own custom one. Ideally we can just fix the exiting outlining pass to only generate a single module, I can't think of any reason this isn't possible. Perhaps it's a naming thing, in which case I'd just say we have a standard/static name for the `gpu.module` because the function name can still provide uniqueness.

Thanks for your note. Yes, I think we can merge them in the outlining pass. I will see how to do that.
Since we will generate only single gpu.module, so we are going to close this draft PR and work to close the basic execution ASAP.

PR #966, opened 2020-03-10
E2E lowering for iGPU

flaub:


pmlc/target/intel_gen/tests/e2e.mlir
```diff
@@ -0,0 +1,8 @@
+// RUN: pmlc-opt | pmlc-vulkan-runner %s --entry-point-result=void | FileCheck %s
```
flaub: This line isn't right, it needs to be `pmlc-opt --target-intel_gen`.

pmlc/target/intel_gen/pipeline.cc
```diff
@@ -16,43 +19,52 @@
...
   pm.addPass(createGpuKernelOutliningPass());
-  // NOTE: canonicalizer/cse at this stage causes later passes to fail
+  // pm.addPass(createConvertStandardToSPIRVPass());
 
-  pm.addNestedPass<ModuleOp>(createConvertGPUToSPIRVPass());
+  // GPU to SPIR-V.
+  pm.addPass(createLegalizeStdOpsForSPIRVLoweringPass());
   pm.addPass(createCanonicalizerPass());
   pm.addPass(createCSEPass());
+  pm.addPass(createConvertGPUToSPIRVPass());
```
YangleiZouIntel: So should we add a new pass before this to add spv.entry_point_abi local_size for gpu.func and the host func(like @main in the vulkan rt tests)?

YangleiZouIntel: Basically the host func is allocating buffers&buffers binding and it will turned into llvm can call runonvulkan to execute spirv binary. We had some draft code and should be able to make it work soon. Please let me know your thoughts. Thanks.

flaub: I had a discussion with the Google team working on this. I think we've arrived at a consensus which is to add a `gpu.block_size` attribute to the `gpu.func`, so that when we lower from GPU to SPIR-V, it can migrate the `gpu.block_size` across to the `spv.entry_point_abi local_size=` attribute. This work would need to be done upstream, but we could do a PoC of it locally in our repo.

Here's a discussion of this topic:
https://llvm.discourse.group/t/using-attributes-to-specify-workgroup-configuration-when-lowering-to-gpu/496/34

You can also join the Discord channel to see the conversation I had with antiaganst about this.
https://discord.gg/xS7Z362

pmlc/target/intel_gen/pipeline.cc
```diff
@@ -16,43 +19,52 @@
...
+  pm.addPass(createCanonicalizerPass());
+  pm.addPass(createCSEPass());
 
+  // pm.addPass(std::make_unique<LoopsToGPUPass>());
   pm.addPass(createSimpleLoopsToGPUPass(1, 1));
-  pm.addNestedPass<FuncOp>(createCanonicalizerPass());
-  pm.addNestedPass<FuncOp>(createCSEPass());
-
+  pm.addPass(createCanonicalizerPass());
   pm.addPass(createGpuKernelOutliningPass());
```
flaub: @YangleiZouIntel This is the pass that will generate the host-side `main` function for us.

pmlc/target/intel_gen/pipeline.cc
```diff
@@ -16,43 +19,52 @@
...
+  // pm.addPass(std::make_unique<IREEGPUToSPIRVPass>());
+
+  // SPIR-V passes for lowering attributes.
+  pm.addNestedPass<spirv::ModuleOp>(spirv::createLowerABIAttributesPass());
+  pm.addNestedPass<spirv::ModuleOp>(createCanonicalizerPass());
+  pm.addNestedPass<spirv::ModuleOp>(createCSEPass());
 
-  // pm.addPass(createLowerToLLVMPass(true));
+  // GPU to Vulkan.
+  pm.addPass(createConvertGpuLaunchFuncToVulkanCallsPass());
```
flaub: @YangleiZouIntel  This is the pass that will translate `gpu.launch_func` into vulkan runtime calls.

pmlc/target/intel_gen/pipeline.cc
```diff
@@ -16,43 +19,52 @@
...
+  // pm.addPass(createConvertLinalgToLoopsPass());
+  pm.addPass(conversion::pxa_to_affine::createLowerPXAToAffinePass());
   pm.addPass(createLowerAffinePass());
-  pm.addNestedPass<FuncOp>(createCanonicalizerPass());
-  pm.addNestedPass<FuncOp>(createCSEPass());
+  pm.addPass(createCanonicalizerPass());
+  pm.addPass(createCSEPass());
 
+  // pm.addPass(std::make_unique<LoopsToGPUPass>());
   pm.addPass(createSimpleLoopsToGPUPass(1, 1));
```
flaub: @YangleiZouIntel  This pass should be converting outer loops into `gpu.func`

YangleiZouIntel: Thanks. It seems each outermost loop.for will be translated into a single gpu module. If the outermost loops have the same index and range, we can merge them so it will generate a single  gpu module with a single function.

pmlc/target/intel_gen/pipeline.cc
YangleiZouIntel: I see. Thanks.

pmlc/target/intel_gen/pipeline.cc
flaub: We have a fusion pass that operates on the affine dialect which would be used to merge compatible loops.

What we need in this situation is to either:
1. fix the upstream code so that it only ever creates a single `gpu.module`
2. write a new pass the merges all `gpu.module`s into a single one

flaub:
Thanks for this example. I do see it generating multiple `gpu.module`s. While it's true that a real fusion pass in this particular case would probably wind up causing these two kernels to be merged into one, at this stage I think we still need to consider being able to generate a single `gpu.module` or merge multiple ones into a single one for later lowerings.

pmlc/target/intel_gen/pipeline.cc
YangleiZouIntel: Thanks. I see. I agree it's better to merge compatible loops in fusion pass or as you mentioned, we can fix the existing outlining pass to generate single gpu.module.

flaub:
After some more investigation over the weekend, I think we might need to stick with a single function inside a `spv.module`. This is because the entry point ABI attribute seems to be applied to the `spv.module`, which implies there's only one entry point per module.

Looking thru the Vulkan API this appears to also be the way kernels are launched. It's a bit confusing, but this section talks about it: https://www.khronos.org/registry/vulkan/specs/1.2/html/chap9.html#pipelines-compute

The idea is that we construct a compute pipeline and then bind the pipeline to the command buffer. Just after this we call `vkCmdDispatch`, which will call the associated SPIR-V module. (See the `VkPipelineShaderStageCreateInfo`.

Really what we need to do is turn the host-side module into code that calls `VulkanRuntime::createSchedule` (a new method I added recently). Basically we need to turn a sequence of `std.alloc` and `gpu.launch_func` ops into `LaunchKernelAction` and `MemoryTransferAction`. I was thinking that instead of lowering `gpu.launch_func` into a `setBinaryShader`, we would instead lower into a call like `launch_kernel(memref_deps)`, which would then construct a `LaunchKernelAction`.

pmlc/rt/vulkan/vulkan_runtime.cc
```diff
@@ -719,3 +722,97 @@ LogicalResult VulkanRuntime::updateHostMemoryBuffers() {
...
+};
+
+struct MemoryTransferAction : Action {
+  VkBuffer src;
+  VkBuffer dst;
+  SmallVector<VkBufferCopy, 1> regions;
+};
+
+LogicalResult VulkanRuntime::createSchedule() {
+  std::vector<ActionPtr> schedule;
```
flaub: TODO: this needs to be filled in. We need to figure out a way to turn the host-side MLIR module into this schedule.

For each user input and output (the arguments to the main entrypoint of the program), we'd need to create a `MemoryTransferAction` that represents a copy from host memory to device memory (in the case of inputs) and a copy from device memory to host memory (in the case of outputs).

For each `gpu.launch_func`, we'd want to create a `LaunchKernelAction`. This is basically a tuple of: `vector<memref>` representing the memrefs that a kernel depends on (either they are user inputs or they are the outputs of other kernels). The `pipeline`, `pipelineLayout`, `descriptorSets`, and `workGroups` are all based off of the `gpu.launch_func` op itself. Each of the memrefs that a kernel uses should turn into an entry in the `deps` field.

YangleiZouIntel: Thanks. So we need to convert each `gpu.launch_func` to something like `launch_kernel(memref_deps)` in which we call `MemoryTransferAction` and `VulkanRuntime::createSchedule()`. (Please correct me if I misunderstood.) I think in this case we can add a new pass after gpu kernel outlining pass to do this work instead of making our own outlining pass.

YangleiZouIntel: so the host-side MLIR will be turned into  calls to VulkanRuntime::createSchedule() .

flaub: Or perhaps calls to something that builds up a schedule so that at the very end we say `runOnVulkan` which internally calls `createSchedule`.

flaub:
Looking good, I think this is headed in the right direction :)

PR #971, opened 2020-03-18
Fix pmlc/util:env ODR

earhart:


flaub:
Why do you need this change?

There's currently no way for this to be an ODR, because the only thing that uses the `:env` rule is the `conda_run`, which is due to bootstrapping issues.

earhart:
I'm trying to prevent future ODRs by making the code more reliable / harder to misuse, by making the libraries work together correctly.

earhart:
... to be clear -- this wasn't because I ran into an actual ODR issue; I just noticed it in the BUILD file, and it was a quick fix with little chance of regressions.

flaub:
I'd prefer that we just remove this odd dependency entirely. I can take a stab at that later, but in the meantime, could we avoid this change so that it doesn't make the BUILD file 'complicated'?

earhart:
Might want to ensure that by setting the visibility so that only `conda_run` can see it...

PR #974, opened 2020-03-20
Integrated the heat map into the stencil pass

LLITCHEV:
Added a costing function in the heatmap itself and passing it to the stencil pass.


LLITCHEV:
I thought I set the plaidml-v1 as base, but from all these conflicts maybe didn't stick.

PR #975, opened 2020-03-20
Integrated the heat map into the stencil pass

LLITCHEV:
Added a costing function in the heatmap itself and passing it to the stencil pass.

### DIFF
--- a/pmlc/dialect/pxa/transforms/passes.h
+++ b/pmlc/dialect/pxa/transforms/passes.h
@@ -2,18 +2,13 @@
 
 #pragma once
 
+#include <functional>
 #include <memory>
-#include <utility>
 
-#include "mlir/Pass/Pass.h"
-#include "mlir/Pass/PassManager.h"
-#include "mlir/Pass/PassOptions.h"
+#include "llvm/ADT/ArrayRef.h"
 
 namespace mlir {
-class FuncOp;
 class Pass;
-template <typename T>
-class OpPassBase;
 } // namespace mlir
 
 namespace pmlc::dialect::pxa {
--- a/pmlc/dialect/pxa/transforms/passes.h
+++ b/pmlc/dialect/pxa/transforms/passes.h
@@ -3,6 +3,7 @@
 #pragma once
 
 #include <memory>
+#include <utility>
 
 #include "mlir/Pass/Pass.h"
 #include "mlir/Pass/PassManager.h"
@@ -18,16 +19,28 @@ class OpPassBase;
 namespace pmlc::dialect::pxa {
 
 struct StencilPass : public mlir::FunctionPass<StencilPass> {
-  StencilPass() = default;
-  StencilPass(const StencilPass &) {}
-  // explicit StencilPass(const StencilPassOptions &options);
+  StencilPass() : costerSet(false) {}
+  StencilPass(const StencilPass &) : costerSet(false) {}
 
   Option<unsigned> numThreads{
       *this, "threads",
       llvm::cl::desc("Specifies number of threads for the stencilling pass")};
   void runOnFunction() final;
+  void setCoster(std::function<std::pair<double, unsigned>(const unsigned *,
+                                                           const unsigned)>
+                     costerIn) {
+    coster = costerIn;
+    costerSet = true;
+  }
+
+  bool costerSet;
+  std::function<std::pair<double, unsigned>(const unsigned *, const unsigned)>
+      coster;
 };
 
 std::unique_ptr<mlir::Pass> createStencilPass();
+std::unique_ptr<mlir::Pass> createStencilPassWithCoster(
+    std::function<std::pair<double, unsigned>(const unsigned *, const unsigned)>
+        coster);
 
 } // namespace pmlc::dialect::pxa
--- a/pmlc/dialect/pxa/transforms/passes.h
+++ b/pmlc/dialect/pxa/transforms/passes.h
@@ -18,29 +18,15 @@ class OpPassBase;
 
 namespace pmlc::dialect::pxa {
 
-struct StencilPass : public mlir::FunctionPass<StencilPass> {
-  StencilPass() : costerSet(false) {}
-  StencilPass(const StencilPass &) : costerSet(false) {}
-
-  Option<unsigned> numThreads{
-      *this, "threads",
-      llvm::cl::desc("Specifies number of threads for the stencilling pass")};
-  void runOnFunction() final;
-  void setCoster(std::function<std::pair<double, unsigned>(const unsigned *,
-                                                           const unsigned)>
-                     costerIn) {
-    coster = costerIn;
-    costerSet = true;
-  }
-
-  bool costerSet;
-  std::function<std::pair<double, unsigned>(const unsigned *, const unsigned)>
-      coster;
+struct StencilCost {
+  double throughput;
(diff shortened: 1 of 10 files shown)


pmlc/dialect/pxa/transforms/passes.h
```diff
@@ -18,16 +19,28 @@ class OpPassBase;
 namespace pmlc::dialect::pxa {
 
 struct StencilPass : public mlir::FunctionPass<StencilPass> {
-  StencilPass() = default;
-  StencilPass(const StencilPass &) {}
-  // explicit StencilPass(const StencilPassOptions &options);
+  StencilPass() : costerSet(false) {}
```
jbruestle: There is no need for 'costerSet', since std::function has a 'null' version, which is it's default constructor, so you can always just use if (coster) ...

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -123,9 +121,22 @@ class Stencil {
...
+  // Whether the copster function was set
+  bool costerSet;
+  // External coster function.
+  std::function<std::pair<double, unsigned>(const unsigned *, const unsigned)>
+      coster;
+
 public:
-  explicit Stencil(mlir::AffineParallelOp op, int numThreads)
-      : op(op), numThreads(numThreads) {
+  explicit Stencil(mlir::AffineParallelOp op, int numThreads, bool costerSet,
```
jbruestle: Definitely should not have 'costerSet' here, and in fact I'd imagine you should assert coster is a valid callable std::function, since constructing a stencil without a cost function makes no sense.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -546,4 +561,12 @@ std::unique_ptr<mlir::Pass> createStencilPass() {
   return std::make_unique<StencilPass>();
 }
 
+std::unique_ptr<mlir::Pass> createStencilPassWithCoster(
```
jbruestle: Why not just change the constructor to the original createStencilPass?  Does having one without a coster make any sense.

LLITCHEV: Maybe someone can point how to register a pass with constructor function with parameters. I have not seen anything like this in the code and looking at the PassRegistry.h, the only way to register a Pass is with no argument function.
Here is some code...
template <typename ConcretePass> struct PassRegistration {
  PassRegistration(StringRef arg, StringRef description,
                   const PassAllocatorFunction &constructor) {
    registerPass(arg, description, PassID::getID<ConcretePass>(), constructor);
  }
using PassAllocatorFunction = std::function<std::unique_ptr<Pass>()>;
Basically, the registration code fails to compile with the following erroir...
pmlc/dialect/pxa/transforms/registration.cc:35:5: error: no matching constructor for initialization of 'mlir::PassRegistration<StencilPass>'
    stencil_pass("affine-stencil",
    ^            ~~~~~~~~~~~~~~~~~
external/llvm-project/mlir/include/mlir/Pass/PassRegistry.h:139:3: note: candidate constructor not viable: no known conversion from 'std::unique_ptr<mlir::Pass> (std::function<std::pair<double, unsigned int> (const unsigned int *, const unsigned int)>)' to 'const mlir::PassAllocatorFunction' (aka 'const function<std::unique_ptr<Pass> ()>') for 3rd argument
  PassRegistration(StringRef arg, StringRef description,
  ^
external/llvm-project/mlir/include/mlir/Pass/PassRegistry.h:144:3: note: candidate constructor not viable: requires 2 arguments, but 3 were provided
  PassRegistration(StringRef arg, StringRef description)

There is currently a function in the code that is the "default" coster - return 1.0 for each tile. And then the user can privide it's own. If any of you have a snipped/doc describing how to register passes with parameters in the registration process, I'd really appreciate a pointer. I just could not find any in our/llvm code bases.

flaub: Keep in mind that PassRegistration is only needed for the `pmlc-opt` tool, which is really only intended for testing. So what I would do is have a named PassRegistration that builds a specific stencil pass with a specific coster. Call it the  `-xsmm-stencil-pass` or something. I don't think we need to worry about making the PassRegistration generic itself.

flaub: Also, I'd register this specific pass over in the `pmlc/target/x86` area, since it's X86 specific and will need access to the heatmap, which is also x86 specific.

pmlc/target/x86/heatmap.tpl.cc
```diff
@@ -1,9 +1,17 @@
...
 #include "pmlc/target/x86/heatmap.h"
 
 namespace pmlc::target::x86 {
 
+static const unsigned STARTUP_COST = 32;
+static const unsigned SPEC_STENCIL_TUPLES = 1;
+static const unsigned SPEC_STENCIL_INDICES = 3;
+// clang-format off
+// Need the spaces between braces, so the macro replacement logic doesnt trigger.
+static unsigned special_stencils[1][3] = { {64, 16, 3} };
```
flaub: Why do we need to modify the template? I would think that we could introduce any specializations in normal c++ code outside of the template. Generally, I think we should keep source code templates that turn into real code as small as possible due to syntax highlighting, and like this clang-format thing you had to do.

Also, I'm not up to sync with why we need to add this additional code to the heatmap. Is it to somehow speed up the search by introducing even further constraints?

pmlc/target/x86/heatmap.tpl.cc
```diff
@@ -22,4 +30,67 @@ float kHeatmapValues[] = {
     // clang-format on
 };
 
+static std::map<std::tuple<unsigned, unsigned, unsigned>, double> *
```
flaub: So instead of adding new helper functions to the c++ template file, I think we should add it to a normal c++ file.

pmlc/dialect/pxa/transforms/stencil.cc
LLITCHEV: @flaub Yes! That is exactly what I figured and implemented it this way. The creasteStencilPass function does exactly that... It register the pass and uses a default coster function (the default one). The createStencilPassWithCoster is not registered and is intended to be instantiated by the pm.addPass(createStencilPassWithCoster(function)) from the pipeline directly the same way ConstantTypesPass in the tile layer is instantiated (pm.addPass(createConstantTypesPass(floatx, intx));)

LLITCHEV: Moving the registration out of the stencil pass and in the x86 is an interesting idea but it could mean that the whole pass needs to be moved. We are testing its logic now with a "default" coster function. And to specify another is not feasible with no ability to register it - again the tests for the pass are running with opt, so it needs to be registered.

flaub:
@LLITCHEV I took a stab at fixing it, WDYT?

pmlc/target/x86/heatmap.tpl.cc
LLITCHEV: This is a specialization that is for xsmm only. It is needed to allow converting to xsmm one of the first convolution of resnet50. Otherwise we are much slower at run time. (I hope I understood what the question was about re: additional code).

LLITCHEV:
Just approve it, so it gets in. I can't approve my PRs.

PR #978, opened 2020-03-21
Make edsl.h more CMake friendly

tzerrell:
Probably don't want to land this as-is, but these changes make edsl.h more friendly to CMake compilers (i.e. to remove warnings/errors). The `ThunkTensorDeriv` in particular is a hack to deal with that function not being functional right now.


tzerrell:
Superseded by #982

PR #986, opened 2020-03-27
Add a pass that bounds checks the Load and Store Ops operations

LLITCHEV:


flaub:
Why is this in the `stdx` dialect? Do we plan to upstream this?

jbruestle:
> Why is this in the `stdx` dialect? Do we plan to upstream this?

Actually, this was my suggestion since it acts on std dialect operations.

flaub:
I guess that's fine for now, but I'd like keep `stdx` clean so it's easier to upstream things. I think this sort of transform probably belongs in the common area, so under `pmlc/transforms` would be my choice.

pmlc/dialect/stdx/transforms/boundscheck.cc
```diff
@@ -0,0 +1,89 @@
...
+#include "pmlc/dialect/stdx/transforms/boundscheck.h"
+#include "pmlc/util/logging.h"
+
+using namespace mlir; // NOLINT
+
+namespace pmlc::dialect::stdx {
+
+namespace {
+template <typename LoadStoreOp>
+class BoundsCheckGenerator {
```
jbruestle: I would probably not have made a class, but just used two free functions (one templated) for getOrInsertFunc and generateBoundsChecks, but that would only make the code a bit smaller, so probably more of a preference thing.  Or alternately, used an object to store pass-common stuff (probably caching results of generating the function declaration), and then had the op / loc be variables in a member function.

pmlc/dialect/stdx/transforms/boundscheck.cc
```diff
@@ -0,0 +1,89 @@
...
+  Location loc;
+  ModuleOp module;
+  Type i32Type;
+  Type indexType;
+
+private:
+  BoundsCheckGenerator(LoadStoreOp op, Builder builder)
+      : op(op), loc(op.getLoc()),
+        module(op.template getParentOfType<ModuleOp>()),
+        i32Type(builder.getIntegerType(32)), indexType(builder.getIndexType()) {
```
jbruestle: Minor point, no need to construct a builder just for this:  IntegerType::get(32, op.getContext()) also works.

pmlc/dialect/stdx/transforms/boundscheck.cc
```diff
@@ -0,0 +1,89 @@
...
+    std::array<Type, 2> inputs{indexType, i32Type};
+    ArrayRef<Type> results{};
+    auto funcType = builder.getFunctionType(inputs, results);
+    ArrayRef<NamedAttribute> attrs{};
+    builder.create<FuncOp>(loc, symbol, funcType, attrs);
+    return SymbolRefAttr::get(symbol, context);
+  }
+
+  Value createConstantIntOp(int64_t value, OpBuilder builder) {
+    return builder.create<ConstantIntOp>(loc, value, i32Type);
```
jbruestle: Not sure it's worth making a method for a one line function call.

PR #990, opened 2020-03-30
Added a runtime function for the boundschecks.

LLITCHEV:


pmlc/rt/boundscheck.cc
```diff
@@ -0,0 +1,16 @@
...
+
+#include <stdexcept>
+
+// This function is called from the BoundsCheckPass.
+// index - ndex for a dimension that is accessed by load/store operation
+// range - upper bound of the range for this dimension (it is always 0 -
+// range)
+extern "C" void plaidml_rt_bounds_check(intptr_t index, unsigned range) {
+  int64_t accessIndex = (int64_t)index;
+  int64_t upperRange = (int64_t)range;
```
flaub: Use c++ style casts (i.e. `static_cast` or `reinterpret_cast`)

pmlc/rt/boundscheck.cc
```diff
@@ -0,0 +1,16 @@
...
+
+#include <stdint.h>
+
+#include <stdexcept>
+
+// This function is called from the BoundsCheckPass.
+// index - ndex for a dimension that is accessed by load/store operation
+// range - upper bound of the range for this dimension (it is always 0 -
+// range)
+extern "C" void plaidml_rt_bounds_check(intptr_t index, unsigned range) {
```
flaub: Why are we using `unsigned` for the `range`? Doesn't that mean the bitwidth is machine-specific? I would have expected an `int64_t` here instead.

pmlc/rt/boundscheck.cc
```diff
@@ -0,0 +1,16 @@
+// Copyright 2020 Intel Corporation
+
+#include <stdint.h>
```
flaub: Use `<cstdint>` instead of C headers.

pmlc/rt/boundscheck.cc
```diff
@@ -0,0 +1,16 @@
...
+
+// This function is called from the BoundsCheckPass.
+// index - ndex for a dimension that is accessed by load/store operation
+// range - upper bound of the range for this dimension (it is always 0 -
+// range)
+extern "C" void plaidml_rt_bounds_check(intptr_t index, unsigned range) {
+  int64_t accessIndex = (int64_t)index;
+  int64_t upperRange = (int64_t)range;
+  if (accessIndex < 0 || accessIndex >= upperRange)
+    std::runtime_error("Out of bounds index for mlir::LoadOp or mlir::StoreOp");
```
flaub: Does throwing actually work inside the MLIR JIT? Have you tested this?

pmlc/rt/boundscheck.cc
```diff
@@ -0,0 +1,16 @@
+// Copyright 2020 Intel Corporation
+
+#include <stdint.h>
+
+#include <stdexcept>
+
+// This function is called from the BoundsCheckPass.
+// index - ndex for a dimension that is accessed by load/store operation
```
flaub: typo: "ndex"

pmlc/rt/boundscheck.cc
LLITCHEV: It was coming like this from the codegen side. Fixed to use int64_t now.

pmlc/rt/boundscheck.cc
LLITCHEV: Using static_cast now.

pmlc/rt/boundscheck.cc
jbruestle: So I think @flaub's point was that we shouldn't make any of our extern C functions take any parameters that are bitwidth specific (i.e. different for 64 bit or 32 bit compilers), which of course includes intptr_t as well.  Let's always used completely fixed bitwidth types for all interfacing.

pmlc/rt/boundscheck.cc
LLITCHEV: This is not in the Jit, I think. The code is already jitted. We are executing the Jitted code when the exception is thrown. I did test it by adding some code to the xsmm RT function and it did propagate the exception properly and failed the test.

pmlc/rt/boundscheck.cc
LLITCHEV: Fixed

pmlc/rt/boundscheck.cc
LLITCHEV: FIxed

pmlc/rt/boundscheck.cc
flaub: This will be executed by the JIT is what I mean.

pmlc/rt/boundscheck.cc
flaub: What I meant was that we shouldn't use `unsigned` in the C signature.

pmlc/rt/boundscheck.cc
```diff
@@ -0,0 +1,15 @@
...
+
+// This function is called from the BoundsCheckPass.
+// index - index for a dimension that is accessed by load/store operation
+// range - upper bound of the range for this dimension (it is always 0 -
+// range)
+extern "C" void plaidml_rt_bounds_check(intptr_t index, int64_t range) {
+  int64_t accessIndex = static_cast<int64_t>(index);
+  if (accessIndex < 0 || accessIndex >= range)
+    throw std::runtime_error(
+        "Out of bounds index for mlir::LoadOp or mlir::StoreOp");
```
flaub: I think we should have a test that purposefully goes out of bounds to ensure that this is all working as we expect. We should be able to do that by hand writing the MLIR code to go out of bounds.

LLITCHEV: This kind of test is certainly easy to do, but we can't run anything yet - need to split the conversion to Stdx and LLVM dialects and insert this pass there. I certainly am planing to add a test like this. Once this is integrated maybe? I just want to add the method in for now, so I can work on splitting the conversion and adding this pass to the pipeline.

flaub: Integration should involve the test, why add the integration without the test? In general, I think we should practice TDD (Test-Driven Development), where we first add a test that is broken and expected to fail because we haven't written any code. Then add the code to get the test to pass, and then finally commit and get it reviewed.

pmlc/rt/boundscheck.cc
flaub: You mention that it "failed the test", can you add the test to this PR? I don't understand how you have verified that the exception properly gets thrown. How could it? At the time of this writing, it didn't even have a `throw`, so it was clearly broken.

pmlc/rt/boundscheck.cc
flaub: FYI, the final pipeline for x86 does not need to be complete, that can be integrated later. What's great about MLIR is that you can piece together parts of passes using `pmlc-opt` and `pmlc-jit` command line utilities and then put this into a `.mlir` test. See the `xsmm_call.mlir` test for an example of how to put together a sequence of passes along with running it in the JIT and then verifying that we get the result we want to have.

flaub: The RUN line would look like:

```
// RUN: pmlc-opt -stdx-check-bounds -convert-std-to-llvm %s | pmlc-jit -entry-point-result=void | FileCheck %s
```

And then you'll need to add a main entry:

```
func @main() {
  %i = constant 0 : index
  %j = constant 10 : index // out of bounds
  %buf = alloc() : memref<20x10xf32>
  call @simpleLoad(%buf, %i, %j) -> (f32) : (memref<20x10xf32>, index, index) -> (f32)
  dealloc %buf : memref<20x10xf32>
}
```

Then we need to figure out how to check for the exception being thrown.

LLITCHEV: Added a test and support in plmc-jit for allowing to specify expected runtime exception.

pmlc/rt/boundscheck.cc
LLITCHEV: I talked to Jeremy about that. I vertifyed in the xsmm.cc, by adding tyo throw an exception and it worked as expected - exception propagated. So, that proved that mlir generated code beghaves properly in regard with exception handling.
This doesn't matter now, because I added the infra and test to allow exception case testing in the plmc jit.

pmlc/tools/pmlc-jit/pmlc-jit.cc
```diff
@@ -20,11 +27,47 @@
...
+      continue;
+    }
+
+    fixedUpCounter++;
+  }
+
+  // Zero up the fixed parameters that have been moved forward.
+  for (int i = fixedUpCounter; i < argc; i++) {
+    argv[i] = nullptr;
+  }
```
flaub, line 62: Why aren't we using LLVM's command line processing? If you want a new command line flag, just register one with LLVM, there's no need to do custom argv processing.

pmlc/tools/pmlc-jit/pmlc-jit.cc
```diff
@@ -20,11 +27,47 @@
...
+
+  try {
+    return mlir::JitRunnerMain(fixedUpCounter, argv, nullptr);
+  } catch (const std::runtime_error &e) {
+    if (e.what() != expectedString) {
+      throw e;
```
flaub, line 68: I think we should just print the exception, that way FileCheck can compare the result.

pmlc/tools/pmlc-jit/pmlc-jit.cc
```diff
@@ -10,8 +10,15 @@
 // CPU by translating MLIR to LLVM IR before JIT-compiling and executing the
 // latter.
 //
+// The implementation also supports specifying an expected runtime_error being
+// thrown and validates with no failure the expected string is correctly
+// thrown.
 //===----------------------------------------------------------------------===//
 
+#include <iostream>
```
flaub, line 18: BTW, `iostream` is banned for use in LLVM: https://llvm.org/docs/CodingStandards.html#include-iostream-is-forbidden

LLITCHEV: This was for debugging. I don't use it. Still debugging.

pmlc/tools/pmlc-jit/pmlc-jit.cc
LLITCHEV: This could work. Let me check to see if this could work. I worry that the process excecution will fail. And if I catch all the exceptions, that could be a problem. If I limit to runtime_error, maybe it could work.

pmlc/tools/pmlc-jit/pmlc-jit.cc
LLITCHEV: I did register a command and tried that, but the jit tool  that we delegate to still fails not recognizing the option. I had to take it out to make it work.

pmlc/dialect/stdx/tests/boundscheck-rt.mlir
```diff
@@ -0,0 +1,18 @@
+// RUN: pmlc-opt -stdx-check-bounds -convert-std-to-llvm %s | pmlc-jit -entry-point-result=void -expected-runtime-exception-string="Out of bounds index for mlir::LoadOp or mlir::StoreOp"
```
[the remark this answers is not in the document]
LLITCHEV: Ok. Let me try to see how would this work.

pmlc/tools/pmlc-jit/pmlc-jit.cc
```diff
@@ -15,58 +15,28 @@
...
-  }
-
   try {
-    return mlir::JitRunnerMain(fixedUpCounter, argv, nullptr);
+    return mlir::JitRunnerMain(argc, argv, nullptr);
   } catch (const std::runtime_error &e) {
-    if (e.what() != expectedString) {
-      throw e;
-    }
+    llvm::outs() << e.what();
```
flaub, line 39: Add a newline

pmlc/tools/pmlc-jit/pmlc-jit.cc
```diff
@@ -15,58 +15,28 @@
...
-
-  // Zero up the fixed parameters that have been moved forward.
-  for (int i = fixedUpCounter; i < argc; i++) {
-    argv[i] = nullptr;
-  }
-
   try {
-    return mlir::JitRunnerMain(fixedUpCounter, argv, nullptr);
+    return mlir::JitRunnerMain(argc, argv, nullptr);
   } catch (const std::runtime_error &e) {
```
flaub, line 38: Catch `std::exception` here, since `std::runtime_error` derives from it.

pmlc/tools/pmlc-jit/pmlc-jit.cc
flaub: Prefix the error with `"ERROR: "`

ci/pipeline.yml
```diff
@@ -13,6 +13,7 @@ steps:
       PLATFORM: '{{name}}'
     artifact_paths:
       - "tmp/build/**/*"
+      - "bazel-testlogs/**/*.log"
```
flaub, line 16: Cool!

pmlc/dialect/stdx/tests/boundscheck-rt.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: pmlc-opt -stdx-check-bounds -convert-std-to-llvm %s | pmlc-jit -entry-point-result=void | FileCheck %s
+
+module {
+  func @simpleLoad(%A: memref<20x10xf32>, %i: index, %j: index) -> (f32) {
+    %0 = load %A[%i, %j] : memref<20x10xf32>
+    return %0: f32
+  }
+  
+  func @main() {
+    // Lubo %4 = alloca %"class.std::runtime_error", align 8
```
flaub, line 10: Remove comment

pmlc/tools/pmlc-jit/pmlc-jit.cc
```diff
@@ -10,12 +10,18 @@
 // CPU by translating MLIR to LLVM IR before JIT-compiling and executing the
 // latter.
 //
+// The implementation also supports specifying an expected runtime_error being
+// thrown and validates with no failure the expected string is correctly
+// thrown.
```
flaub, line 15: This comment is now obsolete.

LLITCHEV: Taken it out.

pmlc/dialect/stdx/tests/boundscheck-rt.mlir
LLITCHEV: Taken out

pmlc/tools/pmlc-jit/pmlc-jit.cc
LLITCHEV: Added "ERROR: "

pmlc/tools/pmlc-jit/pmlc-jit.cc
LLITCHEV: Caught std::exception.
The reason I went with the expected exception was that with this implementation there is a possibiolity we get false positive passing tests - the jit can throw a RT exception and the test can fail. We would not know about it, since the exception will be silently "swallowed" - not for this tests, but for all the rest that there is no CHECK/CHECK-NOT check performed by the FileCheck.

flaub: That's a really good point about silently swallowing exceptions. I could see having a command line flag to toggle between these when an exception is caught.
* Return 0: only print exceptions
* Return non-zero

PR #991, opened 2020-04-01
Make generic stenciling pass

tzerrell:
Make a generic stenciling pass `StencilBase`, and rewrite the XSMM stenciling pass to derive from it. See documentation in `pmlc/dialect/pxa/transforms/stencil.h`.

--- [Original description] ---
Draft for making a more generic stenciling pass. Very much a work in progress, but it does pass the stencil lit test. Missing things include:
* A cost function beyond `return 3;`
* Tensor-index requirements that actually check stride properties instead of just passing
* The tile size generator is minimal and I think maybe just wrong?
* Many style and maintainability cleanups needed.


pmlc/dialect/pxa/transforms/stencil_generic.h
```diff
@@ -0,0 +1,114 @@
...
+
+struct LoadStoreOps {
+  // The load and store ops of an AffineParallel
+  llvm::SmallVector<mlir::AffineLoadOp, 2> loads;
+  llvm::SmallVector<AffineReduceOp, 1>
+      stores; // TODO: Might be either store or reduce
+};
+
+// TODO: May want something more like an autotile.h UnionGenerator?
+using TileSizeGenerator = std::function<std::vector<size_t>(size_t)>;
```
jbruestle, line 54: UnionGenerator is a TileSizeGenerator already! Specifically, you can pass a functor object to an std::function... See the following example code (which prints 15)

#include
#include

struct Foo {
double add;
double operator()(double val) { return val + add; }
};

int main() {
Foo x = {5};
std::function<double(double)> f = x;
std::cout << f(10) << "\n";
}

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,164 @@
...
+      llvm::SmallVector<mlir::BlockArgument, 8> new_bound_idxs(bound_idxs);
+      new_bound_idxs.push_back(block_arg);
+      RecursiveBindIndex(new_bound_idxs, tensors);
+    }
+  }
+}
+
+void StencilGeneric::DoStenciling() {
+  // Initialization
+  // TODO: Something like?: tensors.clear();
```
jbruestle, line 78: Since the stencil objects are single use, you shoudn't need to do any 'reseting' of things.

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,164 @@
...
+  {
```
jbruestle, line 79: Is there a reason for this scope?

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,164 @@
...
+      loads_and_stores = maybe_loads_and_stores.getValue();
+    } else {
+      IVLOG(4, "Cannot Stencil: Operations fail to pattern-match.");
+      return;
+    }
+  }
+
+  llvm::SmallVector<mlir::Value, 3> tensors;
+  for (auto &load_op : loads_and_stores.loads) {
+    // TODO: It's probably fine, but I'm a little worried `getMemRef` isn't
```
jbruestle, line 91: All 'Values' in MLIR should remain valid unless they defining op / block arg is destroyed.  Types are valid forever as they are interned, so no need to worry.

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,164 @@
...
+    // } else if (auto true_store_op =
+    // llvm::dyn_cast_or_null<mlir::AffineStoreOp>(store_op)) {
+    //   tensors.push_back(true_store_op.getMemRef());
+    // } else {
+    //   // TODO: throw?
+    //   IVLOG(1, "Unexpected failure to load tensors from ops in stenciling");
+    //   return;
+    // }
+  }
+  std::sort(tensors.begin(), tensors.begin() + first_store_idx);
```
jbruestle, line 111: Is there a reason not to store the input and output tensors in separate vectors to avoid some of this complexity?

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,164 @@
...
+      RecursiveBindIndex(llvm::SmallVector<mlir::BlockArgument, 8>(), tensors);
+    } while (std::next_permutation(tensors.begin() + first_store_idx,
+                                   tensors.end()));
+  } while (std::next_permutation(tensors.begin(),
+                                 tensors.begin() + first_store_idx));
+
+  // TODO: If we desire more complex requirements than pairwise tensor-to-index
+  // stride requirements, that function could go here
+
+  for (const auto &perm : legal_permutations) {
```
jbruestle, line 125: We should call this on the interior-most call of RecursiveBindIndex probably, since there is no reason I can think of to collect the valid permutations before costing each one.  And doing the tile sizing inline prevents storage of the options.

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,164 @@
...
+    // TODO: Clean, try to get everything on the same unsigned type/width
+    auto ranges = op.getConstantRanges();
```
jbruestle, line 127: This is constant over the entire search (i.e. it could even be done in the constructor, or certainly at the dawn of DoStenciling if you are worried about failure during constructions.  This call is not trivial in terms of the work required.

pmlc/dialect/pxa/transforms/stencil_generic.h
tzerrell: Oh cool, thanks! (technically there's a `size_t`/`int64_t` problem, but that's not what my concern was and also something I need to be generally resolving with this code.)

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,164 @@
...
+    if (!ranges) {
+      // Can't stencil without constant ranges TODO: Check this earlier, not in
+      // the middle of some late loop!
+      IVLOG(1, "NO CONSTANT RANGES TODO");
+      break;
+    }
+    for (size_t i = 0; i < perm.indexes.size(); i++) {
```
jbruestle, line 134: I dont thing this correctly goes over all combinations of tiles sizes without recursion. You might want to consider reusing the code and/or logic in autotile.

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,164 @@
...
+      llvm::SmallVector<size_t, 8> tile_size;
+      size_t idx_in_block = perm.indexes[i].getArgNumber();
+      try { // TODO: probably don't keep long term
+        for (const auto &size :
+             tiling_generators[i](ranges.getValue()[idx_in_block])) {
+          tile_size.push_back(size);
+        }
+      } catch (const std::bad_function_call &e) {
```
jbruestle, line 142: What is this about?

pmlc/dialect/pxa/transforms/stencil_generic.cc
tzerrell: nope, just a typo/oversight

pmlc/dialect/pxa/transforms/stencil_generic.cc
tzerrell: No. I think I was trying to reduce the number of parameters on `RecursiveBindIndex`? But that's not actually a good reason.

pmlc/dialect/pxa/transforms/stencil_generic.cc
tzerrell: :+1: I had a TODO that I must have accidentally deleted that I was doing that in the wrong place, thanks for flagging.

pmlc/dialect/pxa/transforms/stencil_generic.cc
tzerrell: WIP code -- if sufficient `tiling_generators` aren't initialized that's the error that comes up. In finalized form, I suspect we want to test early that the requirements/tiling_generators are of consistent sizes and not do a `try` block here?

pmlc/dialect/pxa/transforms/stencil_generic.cc
tzerrell: I'll take a look, thanks

tzerrell: Also, yes, agreed that this does not go over the tile sizes correctly.

pmlc/dialect/pxa/transforms/stencil_generic.cc
tzerrell: Actually doing the related refactor, it is doing something a little nontrivial: either we need separate `requirements` for loads and stores, or we need to at some point join the loads & stores into one vector. While perhaps this is premature optimization, it feels better to do that out here than in the guts of the recursive function -- especially since the complexity has to go _somewhere_.

For the moment, I'm just going to make this look a little cleaner with a iterator to the split location.

jbruestle: Fair point, but another way to go is to use ArrayRef.  Basically, you can do:

SmallVector<Value, 3> all = <something>;
auto inputs = ArrayRef<Value>(x).slice(0, num_inputs);
auto outputs =  ArrayRef<Value>(x).slice(num_inputs, num_outputs);

for (Value x : inputs) {
  <goo>
}

or:

Value v = outputs[i];  // Acesses all[num_inputs + i]

or whatever.  So it's easy to make a 'slice' of a vector that works like a full vector for all std calls, etc.

tzerrell: It appears `ArrayRef`s don't have `swap`, which precludes the necessary `std::sort` (and presumably also `std::next_permutation`)

pmlc/dialect/pxa/transforms/BUILD
```diff
@@ -9,13 +9,16 @@ plaidml_cc_library(
     srcs = [
         "autotile.cc",
         "stencil.cc",
+        "stencil_generic.cc",
+        "stencil_xsmm.cc",
```
flaub, line 13: XSMM is specific to the x86 target, they should not go into `pxa/transforms`.

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,164 @@
...
+// #include "pmlc/util/logging.h"
+
+namespace pmlc::dialect::pxa {
+
+// TODO: Better to maintain bound_idxs as both set & vector, or just vector?
+//   Or could maintain as map from BlockArg to numeric index (i.e. where it
+//   would have been were it a vector)
+// TODO: Also, should probably at least be a small vector (how small?)
+void StencilGeneric::RecursiveBindIndex(
+    const llvm::SmallVector<mlir::BlockArgument, 8> &bound_idxs,
```
flaub, line 33: nit: the LLVM/MLIR style is to use `camelBack` instead of `snake_case`.

pmlc/dialect/pxa/transforms/stencil_generic.cc
flaub: Actually, I think we should use the original logic from the stencil pass in v0:
https://github.com/plaidml/plaidml/blob/master/tile/codegen/stencil.cc#L26

I think we should basically port this to MLIR. It recursively works on N-dimensions, it handles ordered permutations, and it should be fairly generic already, so shouldn't require much tweaking for MLIR (but I might be wrong about that).

pmlc/dialect/pxa/transforms/stencil_generic.cc
flaub: We should try not to use exceptions at all. The MLIR way is to return errors. If we are introducing new exceptions, then we must have a catch just before the pass returns control to MLIR. MLIR cannot handle exceptions (in fact they disable exceptions in upstream builds).

pmlc/dialect/pxa/transforms/stencil_xsmm.cc
```diff
@@ -0,0 +1,176 @@
...
+      rhs = llvm::dyn_cast_or_null<mlir::AffineLoadOp>(
+          mulfOp.rhs().getDefiningOp());
+    } else if (auto muliOp = llvm::dyn_cast_or_null<mlir::MulIOp>(defOp)) {
+      lhs = llvm::dyn_cast_or_null<mlir::AffineLoadOp>(
+          muliOp.lhs().getDefiningOp());
+      rhs = llvm::dyn_cast_or_null<mlir::AffineLoadOp>(
+          muliOp.rhs().getDefiningOp());
+    } else {
+      IVLOG(5, "The source of the reduce is not a multiplication operation");
+      return llvm::Optional<LoadStoreOps>(); // i.e. fail to pattern-match
```
flaub, line 92: Use `llvm::None`

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,164 @@
...
+
+// #include "pmlc/util/logging.h"
+
+namespace pmlc::dialect::pxa {
+
+// TODO: Better to maintain bound_idxs as both set & vector, or just vector?
+//   Or could maintain as map from BlockArg to numeric index (i.e. where it
+//   would have been were it a vector)
+// TODO: Also, should probably at least be a small vector (how small?)
+void StencilGeneric::RecursiveBindIndex(
```
flaub, line 32: nit: the LLVM/MLIR style for function names is `camelBack`.

pmlc/dialect/pxa/transforms/stencil_xsmm.cc
```diff
@@ -0,0 +1,176 @@
...
+    const unsigned kNumValidInstrInGemmRegion = 5;
+    auto *body = op.getBody();
+
+    // Verify the number of ops
+    if (body->getOperations().size() != kNumValidInstrInGemmRegion) {
+      IVLOG(5, "The AffineParallelOp region didn't have the right number of "
+               "instructions for a GEMM");
+      return llvm::Optional<LoadStoreOps>(); // i.e. fail to pattern-match
+    }
+
```
flaub, line 51: I think it's possible to use `m_Op` or a variant of it that uses `detail::RecursivePatternMatcher` from https://github.com/llvm/llvm-project/blob/master/mlir/include/mlir/IR/Matchers.h#L263.  I think this would make the pattern matching and capturing a single line of code.

pmlc/dialect/pxa/transforms/stencil_generic.cc
tzerrell: Are asserts also reasonable in MLIR land? This is a situation that should never happen (it comes from calling an uninitialized `std::function`)

pmlc/dialect/pxa/transforms/stencil_generic.cc
tzerrell: Ok, I'll take a look at that

pmlc/dialect/pxa/transforms/stencil_xsmm.cc
tzerrell: Thanks, I'll take a look

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,265 @@
...
+      // Verify the requirements for this index with each tensor are all met
+      bool reqsMet = true;
+      for (unsigned i = 0; i < tensors.size(); i++) {
+        try { // TODO: probably don't keep long term
+          if (!requirements.at(std::make_pair(i, currIdx))(tensors[i],
+                                                           blockArg)) {
+            reqsMet = false;
+            break;
+          }
+        } catch (const std::out_of_range &e) {
```
jbruestle, line 141: Is this throw caused my requirements.at(....)?  Seems incorrect handling.  I'd imagine if we don't have any specific requirements for a tensor/index pair, that would mean that any mapping is OK, so we should probably do:

auto it = requriements.find(...)
if (it != requirements.end() && !it->second(tensors[i], blockArg) {
  reqsMet = false;
  break;
}

tzerrell: :+1: yes I agree that seems like the right way to handle this

pmlc/dialect/pxa/transforms/stencil_xsmm.cc
```diff
@@ -0,0 +1,203 @@
...
+             }},
+        };
+    tilingGenerators.push_back(PowerOfTwoGenerator());
+    tilingGenerators.push_back(PowerOfTwoGenerator());
+    tilingGenerators.push_back(PowerOfTwoGenerator());
+  }
+};
+
+struct XSMMStencilPass
+    : public mlir::PassWrapper<XSMMStencilPass, mlir::FunctionPass> {
```
flaub, line 184: BTW, to use the new `passes.td`, replace the base class with the generated one from tablegen: `NewXSMMStencilBase`

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,197 @@
...
+    llvm::SmallVector<mlir::BlockArgument, 8> *boundIdxs,
+    const llvm::SmallVector<unsigned, 3> &tensorIDs) {
+  auto currIdx = boundIdxs->size();
+  if (currIdx == semanticIdxCount) {
+    // This is a legal binding, go find a tiling for it
+    llvm::SmallVector<int64_t, 8> currTileSize(semanticIdxCount);
+    RecursiveTileIndex(TensorAndIndexPermutation(tensorIDs, *boundIdxs),
+                       &currTileSize, 0);
+  } else {
+    for (const auto &blockArg : blockArgs) {
```
flaub, line 89: Always treat `BlockArgument` with by-value semantics

pmlc/dialect/pxa/transforms/stencil_generic.cc
```diff
@@ -0,0 +1,197 @@
...
+    ranges = maybeRanges.getValue(); // TODO: Is this how to use Optional?
+  } else {
+    IVLOG(4, "Cannot Stencil: Requires constant ranges");
+    return;
+  }
+
+  auto maybeLoadsAndStores = capture();
+  if (maybeLoadsAndStores) {
+    loadsAndStores =
+        maybeLoadsAndStores.getValue(); // TODO: Is this how to use Optional?
```
flaub, line 166: Yes, or there's a `operator*` overload that you can use, or `operator->`, so you can say (just examples or optional usage):

```
loadAndStores = *maybeLoadAndStores;
maybeLoadAndStores->loads.size();
```

tzerrell:
Just two outstanding design questions from my perspective, both minor and flagged in the code with `TODO`:
* This includes `autotile.h` to get access to the Tile Size Generators. Maybe we should have a separate file containing the generators that both autotile and stencil include, rather than having stencil depend on autotile?
* Should we make `requirements` and/or `tilingGenerators` configurable at the pass level for the XSMM stencil pass? (i.e. the same way the XSMM cost heatmap/StencilCostFunction is)

tzerrell:
After discussing with @flaub, I'm working on making the `requirements` interface cleaner.

tzerrell:
This is ready for (hopefully) final review! The two minor design details mentioned above are still outstanding: Should we move the tile size generators out of `autotile.h`? And should we make `requirements` and/or `tilingGenerators` configurable at the pass level for the XSMM stencil pass?

flaub:
LGTM, might want @jbruestle  to weigh in.

PR #992, opened 2020-04-02
implementation of tile.reshape via memcpy

mars1026:
Currently supports f32 only. Will expand for other element types if the basic approach is good.

### DIFF
--- a/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
+++ b/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
@@ -38,6 +38,8 @@ using dialect::tile::ContractionOp;
 using dialect::tile::ContractionOpOperandAdaptor;
 using dialect::tile::getPaddingInfo;
 using dialect::tile::IndexOp;
+using dialect::tile::ReshapeOp;
+using dialect::tile::ReshapeOpOperandAdaptor;
 using dialect::tile::ShapeOp;
 using dialect::tile::ShapeOpOperandAdaptor;
 using dialect::tile::TraceOp;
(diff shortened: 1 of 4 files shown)


flaub:
I think we can implement a memcpy with the affine dialect. That way we have an impl for all backends, not just x86.

earhart:
On the flip side -- it might be useful to have a higher-level memcpy operation abstraction; some platforms may have an optimized implementation.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -727,6 +729,79 @@ struct IndexOpConversion : public OpConversionPattern<IndexOp> {
...
+
+    // The source is adaptor.tensor()
+    auto source =
+        rewriter.create<MemRefCastOp>(loc, adaptor.tensor(), unrankedType);
+    // The dest is op.result()
+    auto dest = rewriter.create<MemRefCastOp>(loc, op.result(), unrankedType);
+
+    // The new shape is adaptor.dims()
+    auto i32Type = rewriter.getIntegerType(32);
+    Value count = adaptor.dims()[0];
```
jbruestle, line 766: Rank 0 tensors do exist.  If someone does a reshape on one, this will probably crash.  Probably better to make a constant value of 1 and then multiply by every dim.

pmlc/rt/runtime.cc
```diff
@@ -8,3 +10,22 @@ extern "C" void plaidml_rt_trace(const char *msg) {
...
+                     StridedMemRefType<T, 0> *src, int32_t count) {
+  auto destPtr = dest->data + dest->offset;
+  auto srcPtr = src->data + src->offset;
+  size_t bytes = count * sizeof(T);
+  memcpy(destPtr, srcPtr, bytes);
+}
+} // namespace
+
+extern "C" void plaidml_rt_copy_f32(size_t destRank,
+                                    StridedMemRefType<float, 0> *dest,
```
jbruestle, line 26: Windows apparently wants these to be rank 1 at a minimum.  Shouldn't actually matter here since we never access any of the per-rank array elements.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -727,6 +729,79 @@ struct IndexOpConversion : public OpConversionPattern<IndexOp> {
...
+    // Multiply all the dims together to get total number of elements
+    for (size_t i = 1; i < adaptor.dims().size(); ++i) {
+      count = rewriter.create<MulIOp>(loc, count, adaptor.dims()[i]);
+    }
+    count = rewriter.create<IndexCastOp>(loc, count, i32Type);
+
+    // Value count = rewriter.create<ConstantIntOp>(loc, totalDims, i32Type);
```
jbruestle, line 773: Left over code.

flaub:
@jbruestle  Yes I was thinking of approach 3, either use the `idenitity` lowering or just write a new one that does load/store appropriately.

flaub:
@earhart For progressive lowering, we'd want to introduce a `affine.copy` op. This seems fine to me, but it still wouldn't look like going from `tile` to `std.call` directly. I also think we should eventually pattern match copies (there are a few other cases where they get generated or are the result of a transformation).

flaub:
Agreed with @jbruestle  about the need to write `.mlir` lit tests. For the lowering to affine approach, I'd have a test which ensures that a `tile.reshape` turns into an `affine.parallel` loop with `affine.load` and `affine.store` in the body.

PR #997, opened 2020-04-06
Make convolution easier to use

flaub:
* Use a fluent builder for convolution
* Replace string enums with integer enums


flaub:
@tzerrell WDYT?

tzerrell:
> @tzerrell WDYT?

This is excellent, thank you!

PR #999, opened 2020-04-06
WIP - 1sanz edsl examples v0

1sanz:
Reorgyolo example in eDSL 
still working on issues with channel increase once done will write up neater testing mechanism using python unit tests and complete the PR 
TODO: handle cases where the spatial dims aren't evenly divided by stride
TODO: write python unittests
TODO: write backend_test style test against pytorch implementation
TODO: remove test functions and python test code

### DIFF
--- a/networks/edsl_examples/BUILD
+++ b/networks/edsl_examples/BUILD
@@ -14,3 +14,17 @@ py_binary(
         "//plaidml2/op:py",
     ],
 )
+
+py_test(
+    name = "tests",
+    srcs = [
+        "tests.py",
+        "reorgyolo.py",
+    ],
+    deps = [
+        "//plaidml2:py",
+        "//plaidml2/edsl:py",
+        "//plaidml2/exec:py",
+        "//plaidml2/op:py",
+    ],
+)
--- a/networks/edsl_examples/BUILD
+++ b/networks/edsl_examples/BUILD
@@ -1,0 +1,16 @@
+# Copyright 2018 Intel Corporation.
+
+load("//bzl:plaidml.bzl", "plaidml_cc_binary","plaidml_cc_test")
+
+package(default_visibility = ["//visibility:public"])
+
+py_binary(
+    name = "reorgyolo",
+    srcs = ["reorgyolo.py"],
+    deps = [
+        "//plaidml2:py",
+        "//plaidml2/edsl:py",
+        "//plaidml2/exec:py",
+        "//plaidml2/op:py",
+    ],
+)
--- a/networks/edsl_examples/reorgyolo.py
+++ b/networks/edsl_examples/reorgyolo.py
@@ -38,13 +38,13 @@ import plaidml2.op as plaidml_op
 def reorgyolo_comparison(arrayIn, batch, C, H, W, stride, forward=False):
     arrayLen = len(arrayIn)
     arrayOut = np.zeros(arrayLen)
-    out_c = C // (stride * stride)
+    out_c = C // (stride * stride)  #1
     for b in range(batch):
-        for k in range(C):
-            for j in range(H):
-                for i in range(W):
+        for k in range(C):  #0-4
+            for j in range(H):  #0-6
+                for i in range(W):  #0-6
                     in_index = i + W * (j + H * (k + C * b))
-                    c2 = k % out_c
+                    c2 = k % out_c  #
                     offset = k // out_c
                     w2 = i * stride + offset % stride
                     h2 = j * stride + offset // stride
@@ -60,18 +60,17 @@ def reorgyolo_comparison_nodivmod(arrayIn, batch, C, H, W, stride, forward=False
     arrayLen = len(arrayIn)
     arrayOut = np.zeros(arrayLen)
     out_c = C // (stride * stride)
-    _c1_quotient_range = int(C // (out_c))
     for n1 in range(batch):
         for w1 in range(W):
             for h1 in range(H):
                 for c2 in range(out_c):
-                    for _w2_quotient in range(_c1_quotient_range // stride):
-                        for _w2 in range(stride):
-                            _c1 = _w2 + _w2_quotient * stride
+                    for h_jump in range(stride):
+                        for w_jump in range(stride):
+                            _c1 = w_jump + h_jump * stride
                             c1 = c2 + (_c1 * out_c)
                             in_index = w1 + W * (h1 + H * (c1 + C * n1))
-                            w2 = w1 * stride + _w2
-                            h2 = h1 * stride + _w2_quotient
+                            w2 = w1 * stride + w_jump
+                            h2 = h1 * stride + h_jump
                             out_index = int(w2 + W * stride * (h2 + H * stride *
                                                                (c2 + out_c * n1)))
                             if forward:
@@ -101,12 +100,12 @@ def reorgyolo(I, stride, forward):
 
 
 def main():
-    n_i = 2
+    n_i = 1
     c_i = 4
     h_i = 6
     w_i = 6
     stride = 2
-    forward = True
+    forward = False
 
     I_data_linear = np.array(list(range(n_i * c_i * h_i * w_i))).astype(np.int)
     I_data = np.reshape(I_data_linear, (n_i, c_i, h_i, w_i))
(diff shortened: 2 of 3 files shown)


networks/edsl_examples/reorgyolo.py
```diff
@@ -0,0 +1,189 @@
...
+    H = dims[2]
+    W = dims[3]
+
+    out_c = C // (stride * stride)
+
+    N_out = N
+    C_decrease = int(C // (stride * stride))
+    C_increase = int(C * (stride * stride))
+    _c1_quotient_range = int(C // (C_decrease))
+    _w2_quotient_range = int(_c1_quotient_range // stride)
```
tzerrell, line 99: There are a bunch of weird cases here that are hard to reason about. I think this is right, but still: Ensure at a minimum you test cases where stride does not divide C, where stride divides C but stride^2 doesn't, where stride^2 divides C but stride^3 doesn't, and where stride^3 divides C. Probably would also be good to test where stride^2 doesn't evenly divide C and moreover `C // stride^2` is itself not divisible by C.

(Unless any of those cases are expected to not be supported by reorg YOLO -- which maybe they aren't? I'm not entirely clear on what _should_ happen in a non-even division.)

networks/edsl_examples/reorgyolo.py
```diff
@@ -0,0 +1,189 @@
...
+    N_in, C_in, H_in, W_in = edsl.TensorDims(4)
+    n1, w1, h1, c2, _w2_quotient, _w2 = edsl.TensorIndexes(6)
+    I.bind_dims(N_in, C_in, H_in, W_in)
+
+    #if forward:
+    O = edsl.TensorOutput(N_out, C_decrease, int(H * stride), int(W * stride))
+    O[n1, c2, h1 * stride + _w2_quotient, w1 * stride +
+      _w2] = I[n1, c2 + ((_w2 + _w2_quotient * stride) * out_c), h1, w1]
+    # else:
+    #     O = edsl.TensorOutput(N_out, C_increase, int(H/stride), int(W/stride))
```
tzerrell, line 109: This `H//stride` and `W//stride` code might be a place to look at for `forward == false` problems

tzerrell: To clarify: I mean this in cases where the spatial dims aren't evenly divided by the stride. If stride divides H and W evenly I would not expect these to cause problems.

networks/edsl_examples/reorgyolo.py
```diff
@@ -0,0 +1,189 @@
...
+    #     O[n1, c2+((_w2 + _w2_quotient*stride)*out_c), h1, w1] = I[n1,c2, h1*stride + _w2_quotient, w1*stride + _w2]
+
+    O.add_constraint(_w2 < stride)
+    O.add_constraint(_w2_quotient < _w2_quotient_range)
+    O.add_constraint(c2 < C_decrease)
+    O.add_constraint(h1 < H)
+    O.add_constraint(w1 < W)
+    O.add_constraint(n1 < N)
```
flaub, line 117: @tzerrell  Are these constraints necessary? Is it because we have a complicated LHS?

networks/edsl_examples/reorgyolo.py
flaub: I don't see `_w2_quotient_range` actually being used by anything, is that intentional?

flaub: Oh, I see it's for the constraint... Is the constraint required?

networks/edsl_examples/BUILD
```diff
@@ -0,0 +1,16 @@
+# Copyright 2018 Intel Corporation.
+
+load("//bzl:plaidml.bzl", "plaidml_cc_binary","plaidml_cc_test")
+
+package(default_visibility = ["//visibility:public"])
+
+py_binary(
+    name = "reorgyolo",
+    srcs = ["reorgyolo.py"],
```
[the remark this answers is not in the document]
1sanz: yes ... this maybe a temporary home for some ops. I was thinking we could move ops into op lib once we have thoroughly tested them out here. Also I feel like detailed examples demonstrating op conversion into eDSL might be a good thing to have even if they are for ops that already exist in the op lib.

networks/edsl_examples/reorgyolo.py
```diff
@@ -0,0 +1,189 @@
...
+#    output:
+#            Output Tensor:
+#                   dimentions [N, C*(s^2), H/s, W/s] OR [N, C/(s^2), H*s, W*s]
+#   -------------------------- Testing method-------------------------------------
+# The reorg function provided here:
+#   https://gist.github.com/leimao/ece7217b5d07fe4e685c47af5e76744a
+# is currently being used for local testing
+# TODO: iron out issues with channel decrease (forward = false )
+# TODO: write python unittests
+# TODO: write backend_test style test against pytorch implementation
```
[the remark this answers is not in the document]
1sanz: agreed.

networks/edsl_examples/BUILD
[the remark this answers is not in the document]
flaub: I plan to port this over to `plaidml-v1` and in so doing will add it to the op lib so I can use it elsewhere. Thanks!

networks/edsl_examples/reorgyolo.py
1sanz: @flaub these might actually be the reason behind forward=false failing... I am working on a fix

networks/edsl_examples/reorgyolo.py
tzerrell: good point, see my response to your other comment on this

networks/edsl_examples/reorgyolo.py
tzerrell: I would expect the constraint on `_w2` to be necessary and the others to be redundant

networks/edsl_examples/reorgyolo.py
```diff
@@ -0,0 +1,189 @@
...
+                                arrayOut[in_index] = arrayIn[out_index]
+    return arrayOut
+
+
+def reorgyolo(I, stride, forward):
+    dims = I.shape.int_dims
+    N = dims[0]
+    C = dims[1]
+    H = dims[2]
+    W = dims[3]
```
flaub, line 90: These aren't needed, `I.bind_dims()` does the same thing. BTW, this won't work in plaidml-v1, `shape` has been dropped from `Tensor`.

networks/edsl_examples/reorgyolo.py
```diff
@@ -84,39 +84,28 @@ def reorgyolo_comparison_nodivmod(arrayIn, batch, C, H, W, stride, forward=False
...
     else:
-        O = edsl.TensorOutput(N_out, C_increase, int(H / stride), int(W / stride))
+        C_increase = C * (stride * stride)
+        O = edsl.TensorOutput(N, C_increase, H // stride, W // stride)
         O[n1, c2 +
-          ((_w2 + _w2_quotient * stride) * out_c), h1, w1] = I[n1, c2, h1 * stride +
-                                                               _w2_quotient, w1 * stride + _w2]
+          ((_w2 + _w2_quotient * stride) * C), h1, w1] = I[n1, c2, h1 * stride +
+                                                           _w2_quotient, w1 * stride + _w2]
+        O.add_constraint(c2 < C)
```
tzerrell, line 106: This constraint should be implied by the `C` dimension of `I` being accessed by the affine formula `c2`. Does it work if you remove the constraint?

networks/edsl_examples/reorgyolo.py
```diff
@@ -84,39 +84,28 @@ def reorgyolo_comparison_nodivmod(arrayIn, batch, C, H, W, stride, forward=False
...
     if forward:
-        O = edsl.TensorOutput(N_out, C_decrease, H * stride, W * stride)
+        C_decrease = C // (stride * stride)
+        _c1_quotient_range = C // (C_decrease)
+        _w2_quotient_range = _c1_quotient_range // stride
+        O = edsl.TensorOutput(N, C_decrease, H * stride, W * stride)
         O[n1, c2, h1 * stride + _w2_quotient, w1 * stride +
-          _w2] = I[n1, c2 + ((_w2 + _w2_quotient * stride) * out_c), h1, w1]
+          _w2] = I[n1, c2 + ((_w2 + _w2_quotient * stride) * C_decrease), h1, w1]
+        O.add_constraint(c2 < C_decrease)
```
tzerrell, line 98: Oh, this one also should be unnecessary (implied by the `C_decrease` dimension of `O`)

flaub:
Now that this is in the op-lib of `plaidml-v1`, do we still want to get this PR landed? @brianretford?

brianretford:
I don't think so, probably fine to abandon it

PR #1000, opened 2020-04-07
upload test logs as artifacts in pipeline

drgone:


ci/pipeline.yml
```diff
@@ -13,6 +13,7 @@ steps:
       PLATFORM: '{{name}}'
     artifact_paths:
       - "tmp/build/**/*"
+      - "bazel-testlogs/**/*.log"
```
flaub, line 16: Actually, let's do `"bazel-testlogs/**/test.log"`, since I don't think anyone needs to look at `test.xml.log` files.

PR #1007, opened 2020-04-09
Fix an assertion that stop crashing tests

LLITCHEV:
Always set the AffineMap to empty one in case there are empty lower and upper bounds.


flaub:
So, I wonder if we should only have ranges on tile dialect, because that's all we need for the affine dialect.

flaub:
@LLITCHEV Could you add a `.mlir` based test that shows this error? In general, as we go and fix backend tests, I think we should try to write `.mlir` tests for each failure so that we can get coverage via `bazelisk test //...`

PR #1009, opened 2020-04-10
Put RangeConstraints in canonical form

tzerrell:
The same RangeConstraint can be expressed in two directions (e.g. `0 <= x < 5` is equivalent to `0 <= 4 - x < 5` -- recall that `x` is integral); this canonicalizes which direction is chosen.

### DIFF
--- a/pmlc/dialect/tile/transforms/contraction.cc
+++ b/pmlc/dialect/tile/transforms/contraction.cc
@@ -150,7 +150,7 @@ BoundsAndConstraints Constraints::ComputeBounds() {
 }
 
 static IndexPoly MakePoly(ContractionOp op, AffineExpr expr) {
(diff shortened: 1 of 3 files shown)


pmlc/dialect/tile/transforms/contraction.cc
```diff
@@ -209,6 +209,10 @@ static IndexAccess ConvertAffineMap(ContractionOp op, mlir::AffineMap map) {
 }
 
 Contraction::Contraction(ContractionOp op) {
+  if (VLOG_IS_ON(2)) {
+    auto op_result = op.result();
+    IVLOG(2, "Processing: " << mlir::debugString(op_result));
+  }
```
flaub, line 215: `op.result()` is basically a zero-cost thing to do, so no need to wrap this in `VLOG_IS_ON(2)`

PR #1011, opened 2020-04-13
GPU-SPIRV-Vulkan support

YangleiZouIntel:


pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -0,0 +1,466 @@
...
+                                                             Value &buffer) {
+  auto type = buffer.getType();
+  if (auto memRefType = type.dyn_cast_or_null<MemRefType>()) {
+    auto rank = memRefType.getRank();
+    auto elementType = memRefType.getElementType();
+    if (elementType.isF32()) {
+      auto dynamicBuffer = builder.create<mlir::MemRefCastOp>(
+          loc, buffer, getDynamicMemRefType(rank, elementType));
+      auto unrankedBuffer = builder.create<mlir::MemRefCastOp>(
+          loc, dynamicBuffer, getUnrankedMemRefType(elementType));
```
flaub, line 256: Can we cast directly to an unranked type?

flaub:
Is this PR meant to supersede #966?

pmlc/conversion/gpu/gpu_to_vulkan.cc
YangleiZouIntel: Yes. But one thing is that the mlir llvm lowing pass does not implement unranked memref types for function signatures, so I added that in pmlc/conversion/gpu/llvm_lowering.cc, converting unranked memref to a rank variable and a pointer to descriptor and then it works.

pmlc/conversion/gpu/llvm_lowering.cc
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  if (type.isa<UnrankedMemRefType>()) {
+    mlir::Type llvmInt64Type =
+        LLVM::LLVMType::getInt64Ty(converter.getDialect());
+    mlir::Type llvmPointerType =
+        LLVM::LLVMType::getInt8PtrTy(converter.getDialect());
+    result.append({llvmInt64Type, llvmPointerType});
+    return success();
+  }
```
YangleiZouIntel, line 80: This is that I added for converting unranked memref to a rank variable and a pointer. This works for now, we should replace this pass when upstream implemented this thing.

YangleiZouIntel:
> Is this PR meant to supersede #966?

Yes. In #966 I can not invite you to review, so I created this new PR and also squashed the commits.

pmlc/conversion/gpu/gpu_to_vulkan.cc
flaub: I'm not sure why you think that, because we use unranked memrefs in other places at the LLVM level.

pmlc/conversion/gpu/llvm_lowering.cc
flaub: We shouldn't need this, it is already handled upstream.

flaub: OK, I see now, we did a similar thing in the x86 pipeline. See the `mixedPtrFuncArgTypeConverter`

pmlc/conversion/gpu/gpu_to_vulkan.cc
flaub: I think we can go from a static ranked memref to an unranked memref in a single cast.

pmlc/conversion/gpu/passes.td
```diff
@@ -0,0 +1,24 @@
...
+  let constructor = "pmlc::conversion::gpu::createConvertGpuLaunchFuncToVulkanCallsPass()";
+}
+
+#endif // __PMLC_CONVERSION_GPU_TO_VULKAN_PASSES__
+
+
+#ifndef __PMLC_CONVERSION_KERNEL_OUTLINING_PASSES__
+#define __PMLC_CONVERSION_KERNEL_OUTLINING_PASSES__
+
+include "mlir/Pass/PassBase.td"
```
flaub, line 17: Remove these lines, this almost looks like multiple files were copy/pasted into one.

pmlc/rt/vulkan/tests/BUILD
```diff
@@ -2,8 +2,26 @@
...
+        "//plaidml/edsl",
+        "//plaidml/exec",
+        "//pmlc/rt/vulkan:link",
+    ],
+)
+
+lit_test(
+    name = "edsl_test_gpu.cc",
+    data = [":cc_test_gpu"],
+)
```
flaub, line 27: I was hoping we could put this in the other directory because now we need a special `lit.local.cfg` which uses `vkprobe` to enable/disable this specific test.

pmlc/rt/vulkan/wrappers.cc
```diff
@@ -168,6 +168,15 @@ void bindBufferFloat32(void *vkRuntimeManager, DescriptorSetIndex setIndex,
...
     break;
+  case 4:
+    bindBuffer<float, 4>(vkRuntimeManager, setIndex, bindIndex, ptr);
+    break;
+  case 5:
+    bindBuffer<float, 5>(vkRuntimeManager, setIndex, bindIndex, ptr);
+    break;
+  case 6:
+    bindBuffer<float, 6>(vkRuntimeManager, setIndex, bindIndex, ptr);
+    break;
```
flaub, line 179: Can we cast to an unranked memref so that we don't need all these variations?

pmlc/rt/vulkan/tests/CPPLINT.cfg
```diff
@@ -0,0 +1 @@
+filter=-whitespace/line_length
```
flaub, line 1: Why is this needed?

pmlc/rt/vulkan/vulkan_runtime.cc
```diff
@@ -160,25 +304,38 @@ LogicalResult VulkanRuntime::run() {
...
+  updateHostMemoryBuffers();
   return success();
 }
 
 LogicalResult VulkanRuntime::createInstance() {
-  RETURN_ON_VULKAN_ERROR(volkInitialize(), "volkInitialize");
+  if (volkInitialized != VK_SUCCESS) {
+    volkInitialized = volkInitialize();
+    RETURN_ON_VULKAN_ERROR(volkInitialized, "volkInitialize");
+  }
```
flaub, line 338: Why do we call `createInstance` more than once? That seems like the real issue here.

pmlc/conversion/gpu/passes.td
YangleiZouIntel: Yes, I will remove these.

pmlc/rt/vulkan/tests/BUILD
YangleiZouIntel: I see, will use vkprobe as a trigger.

pmlc/rt/vulkan/tests/CPPLINT.cfg
YangleiZouIntel: This is added because when I git commit, some lines of the cc_test_gpu.cc are too long.

pmlc/rt/vulkan/vulkan_runtime.cc
YangleiZouIntel: Yes, I am still WIP to make the changes per our discussion earlier this week to split the main into init and main so vulkan will be initialized once and not destroyed every run.

pmlc/rt/vulkan/wrappers.cc
YangleiZouIntel: Actually the static memref has been casted to unranked memref. Every unranked memref is lowered into two variables (int rank, void ptr) after llvm lowering pass. The thing is that here I call a template function(bindbuffer), which has to know the rank at compilation time, but the rank is a variable not a constant(so I use switch case here). I will check if there is a better way to do this.

pmlc/rt/vulkan/tests/CPPLINT.cfg
flaub: Oh, I see this is in the `vulkan/tests` directory, ok great.

pmlc/rt/vulkan/wrappers.cc
```diff
@@ -21,111 +21,207 @@
...
-    if (failed(vulkanRuntime.initRuntime()) || failed(vulkanRuntime.run()) ||
-        failed(vulkanRuntime.updateHostMemoryBuffers()) ||
-        failed(vulkanRuntime.destroy())) {
+    if (failed(vulkanRuntime.setLaunchKernelAction())) {
       llvm::errs() << "runOnVulkan failed";
     }
   }
 
+  void addLaunchActionToSchedule() {
+    std::lock_guard<std::mutex> lock(mutex);
```
flaub, line 70: What's the thinking behind the locking? Do we anticipate this to occur from multiple threads? It seems like the host-function we generate will be executed on a single thread.

YangleiZouIntel: This lock is tend to avoid possible action disorder in the future. Yes, it should run on a single thread now and we can safely remove it. Will we consider multiple threads sharing the same schedule and generating actions concurrently in the future?

flaub: I don't think so since generating the schedule should only happen once during the `init` function eventually.

pmlc/rt/vulkan/tests/BUILD
```diff
@@ -2,8 +2,31 @@
...
+load("//pmlc:lit.bzl", "glob_lit_tests", "lit_test")
+load("//bzl:plaidml.bzl", "plaidml_cc_test")
 
 glob_lit_tests(
     data = ["//pmlc/rt/vulkan:vkprobe"],
 )
+
+plaidml_cc_test(
+    name = "cc_test_gpu",
+    srcs = ["edsl_test_gpu.cc"],
```
flaub, line 14: Could we turn `edsl_test.cc` into a library so that we can use it for both CPU and GPU tests?

YangleiZouIntel: I would like to. But for now, one thing is the PLAIDML_DEVICE and PLAIDML_TARGET are hardcoded in the env, we might need two testenv.cc. Another thing is many tests can pass now but there are still a few tests fails on gpu(which I commented in the file) due to limited support in mlir passes (like spv pointer to int1 is not supported, unnested loop et al). In addition, last time I saw the Linux CI server fails for some Uint64 tests, so I commented the Uint64 tests. Also some tests commented in cc_tests.cc passed in cc_test_gpu.cc like GradientDot.
So the passed test set is slightly different for gpu and cpu right now.

pmlc/rt/vulkan/tests/multi-kernel.mlir
```diff
@@ -0,0 +1,73 @@
+// RUN: pmlc-vulkan-runner %s | FileCheck %s
+
+// CHECK: [23,   23,   23],  
+// CHECK: [23,   23,   23],  
+// CHECK: [23,   23,   23]
+
+
+module attributes {gpu.container_module, spv.target_env = #spv.target_env<#spv.vce<v1.0, [Shader], [SPV_KHR_storage_buffer_storage_class]>, {max_compute_workgroup_invocations = 128 : i32, max_compute_workgroup_size = dense<[128, 128, 64]> : vector<3xi32>}>} {
```
flaub, line 8: Can you limit the columns to 80, or 120 at least?

pmlc/rt/vulkan/wrappers.cc
```diff
@@ -21,111 +21,207 @@
...
 
-  void setEntryPoint(const char *entryPoint) {
+  void createLaunchKernelAction(uint8_t *shader, uint32_t size,
+                                const char *entryPoint,
+                                NumWorkGroups numWorkGroups) {
     std::lock_guard<std::mutex> lock(mutex);
+    vulkanRuntime.createLaunchKernelAction();
+    vulkanRuntime.setShaderModule(shader, size);
     vulkanRuntime.setEntryPoint(entryPoint);
+    vulkanRuntime.setNumWorkGroups(numWorkGroups);
```
flaub, line 46: Why not just have the `createLaunchKernelAction` take the `shader`, `entryPoint` and `numWorkGroups`?

pmlc/rt/vulkan/tests/multi-kernel.mlir
YangleiZouIntel: Yes, sure.

pmlc/rt/vulkan/tests/BUILD
```diff
@@ -2,8 +2,31 @@
...
+    tags = ["manual"],  # the `edsl_test.cc.test` rule will run this test via lit
+    deps = [
+        "//plaidml:testenv",
+        "//plaidml/edsl",
+        "//plaidml/exec",
+        "//pmlc/rt/vulkan",
+    ],
+    linkstatic = 1,
```
flaub, line 22: Why do we need this?

pmlc/rt/vulkan/wrappers.cc
YangleiZouIntel: Yes, I will add them to a single function. It is succinct and better. I thought to leave these functions as they are quite self-explanatory.

pmlc/rt/vulkan/tests/BUILD
YangleiZouIntel: It seems the native.cc_test could not find vk symbols (while native.cc_binary can). I found some ref online saying Bazel by default links binaries statically, but tests dynamically. Adding this static link flag can help native.cc_test find vk symbols. 
Ref:
https://github.com/bazelbuild/bazel/issues/2944

pmlc/rt/vulkan/tests/edsl_test_gpu.cc
```diff
@@ -0,0 +1,937 @@
...
+Tensor Dot(const Tensor &X, const Tensor &Y) {
+  TensorDim I, J, K;
+  TensorIndex i("i"), j("j"), k("k");
+  X.bind_dims(I, K);
+  Y.bind_dims(K, J);
+  auto R = TensorOutput(I, J);
+  R(i, j) += X(i, k) * Y(k, j);
+  return R;
+}
+/*
```
flaub, line 51: Remove all these commented out tests, we can re-add them when they start working as needed.

YangleiZouIntel: Ah, I see. I will remove them.

pmlc/rt/vulkan/tests/BUILD
YangleiZouIntel: Could you please suggest how can I set PLAIDML_DEVICE and PLAIDML_TARGET to intel_gen for this test? I am considering adding a copy of testenv.h testenv.cc to pmlc/rt/vulkan/tests for cc_test_gpu only.

YangleiZouIntel:
@flaub The Linux CI server right now seems does not support UINT64. I know we will be reading these kind of capabilities from device, should I remove all the UINT64 tests (which passed on my local machine)?

PR #1017, opened 2020-04-15
Use Dialect decl from tablegen

flaub:


flaub:
I'm following MLIR core. Apparently TF uses a static initializer for auto-registration. So far it doesn't look like we need it yet.

earhart:
SGTM.  (Well, with the caveat that there's no way for a caller of PlaidML to safely initialize some of these dialects but not others; I really wish multiple registration calls per-dialect were okay.  Not really a PlaidML issue, tho.)

PR #1020, opened 2020-04-15
Fix for calculating buffer sizes when calling from Python to plaidml

LLITCHEV:
Thanks, JB!

### DIFF
--- a/plaidml/bridge/keras/backend_test.py
+++ b/plaidml/bridge/keras/backend_test.py
@@ -408,7 +408,6 @@ class TestBackendOps(unittest.TestCase):
     def testArgmaxWithAxis(self, b):
         return b.argmax(b.variable(n(2, 3, 4) % 3), axis=-2)
 
-    @unittest.skip("Cull crashing tests")
     @compareForwardExact()
     def testArgmaxUnequal(self, b):
         x = b.variable(m(3, 2))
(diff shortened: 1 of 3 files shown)


plaidml/core/ffi.cc
```diff
@@ -369,7 +369,8 @@ uint64_t plaidml_shape_get_nbytes(  //
         total += (sizes[i] - 1) * strides[i];
       }
     }
-    return (total + 1) * (shape->type.getElementTypeBitWidth() / 8);
+    unsigned elem_bytes = (shape->type.getElementTypeBitWidth() + 7) / 8;
```
flaub, line 372: Could we use `llvm::divideCeil` here?

plaidml/bridge/keras/backend_test.py
```diff
@@ -589,6 +587,7 @@ def testDivConstant(self, b, x, c):
             c / x,
         ]
 
+    @unittest.skip("Failing on llvm.cpu")
```
flaub, line 590: Does this crash or just fail? If it fails, why do we skip it?

LLITCHEV: Didn't know that we skip only hard crashing tests. Fixed.

plaidml/core/ffi.cc
LLITCHEV: Yes, I think we can. Using it now.

PR #1021, opened 2020-04-16
Move mlir pointer forward

jbruestle:


pmlc/dialect/pxa/transforms/test_analysis.cc
```diff
@@ -32,7 +32,7 @@ struct TestStrideInfoPass : public TestStrideInfoBase<TestStrideInfoPass> {
   void runOnOperation() final {
     auto op = getOperation();
     op->walk([&](Operation *op) {
-      TypeSwitch<Operation *>(op)
+      llvm::TypeSwitch<Operation *>(op)
```
flaub, line 35: I'm surprised this is actually needed, usually llvm symbols are aliased into the mlir namespace.

jbruestle: Yeah, the problem here is that they are both present and it can't decide which one to use.

PR #1023, opened 2020-04-18
switch to latest master revision of libxsmm which incl. bug fixes for…

alheinecke:
… batch reduce gemm and fixes python3 compatibility


hfp:
As this PR refers to a particular commit/master, I want to promise we will test with PlaidML prior to future releases (starting with v1.16). Perhaps we can also add PlaidML to our Buildkite based CI (dedicated infrastructure). For the latter, it would be useful to know what to test and expect. Perhaps someone can follow up and let us know.

PR #1028, opened 2020-04-21
1sanz edsl examples v0

1sanz:
Reorgyolo example for master with conversion to eDSL notes added. 

This serves as an example on how to convert loops into eDSL.


flaub:
I think since this has already been ported to plaidml-v1, we can close this.

PR #1037, opened 2020-05-01
Fix all of the BoundCheck failing tests

LLITCHEV:
Disable temporarily the do_grad option in the backend_tests, since it is not implemented yet.
Fix an issue with getting out of bounds in case of lowerBounds not 0 for ContractOp.

### DIFF
--- a/plaidml/bridge/keras/backend_test.py
+++ b/plaidml/bridge/keras/backend_test.py
@@ -194,7 +194,7 @@ def compareMultiple(arguments):
 def opTest(in_data,
            tol=DEFAULT_TOL,
            atol=DEFAULT_ATOL,
-           do_grads=True,
+           do_grads=False,
            skip_theano=True,
            skip_tensorflow=False,
            verbose=False,
@@ -1093,7 +1093,6 @@ class TestBackendOps(unittest.TestCase):
             b.conv2d(im, km, padding='same', dilation_rate=(2, 2), data_format=df),
         ]
 
-    @unittest.skip("Cull crashing tests")
     @opTest(
         [[m(1, 1, 3, 1),
           m(1, 4, 1, 1), (1, 1, 9, 1), (1, 4), 'same', 'channels_last', (1, 1)],
--- a/plaidml/bridge/keras/backend_test.py
+++ b/plaidml/bridge/keras/backend_test.py
@@ -279,14 +279,15 @@ def opTest(in_data,
                             atol=atol,
                             err_msg='ERR: datum={}, test={}, x=plaidml, y=theano'.format(
                                 didx, idx))
-                        for x in range(0, len(pmlr[1])):
-                            npt.assert_allclose(
-                                pmlr[1][x],
-                                thr[1][x],
-                                rtol=tol,
-                                atol=atol,
-                                err_msg='ERR: datum={}, test={}, grad, x=plaidml, y=theano'.format(
-                                    didx, idx))
+                        if do_grads:
+                            for x in range(0, len(pmlr[1])):
+                                npt.assert_allclose(
+                                    pmlr[1][x],
+                                    thr[1][x],
+                                    rtol=tol,
+                                    atol=atol,
+                                    err_msg='ERR: datum={}, test={}, grad, x=plaidml, y=theano'.
+                                    format(didx, idx))
                 if not skip_tensorflow:
                     for idx, (pmlr, tfr) in enumerate(zip(plaidml_results, tensorflow_results)):
                         idx = idx + 1
--- a/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
+++ b/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
@@ -640,12 +640,12 @@ struct ContractionOpConversion : public OpConversionPattern<ContractionOp> {
 
     auto ubMap = AffineMap::get(0, 0, {ubExprs}, op.getContext());
     // Make the outer loops
-    auto forOp = rewriter.create<AffineParallelOp>(
-        loc,
-        /*lbMap=*/op.lowerBounds().getValue(),
-        /*lbArgs=*/llvm::SmallVector<Value, 0>{},
-        /*ubMap=*/ubMap,
-        /*ubArgs=*/llvm::ArrayRef<Value>{});
+    auto forOp =
+        rewriter.create<AffineParallelOp>(loc,
+                                          /*lbMap=*/op.lowerBounds().getValue(),
+                                          /*lbArgs=*/llvm::ArrayRef<Value>{},
+                                          /*ubMap=*/ubMap,
+                                          /*ubArgs=*/llvm::ArrayRef<Value>{});
 
     auto body = forOp.getBody();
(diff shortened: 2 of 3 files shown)


pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -626,20 +626,25 @@ struct ContractionOpConversion : public OpConversionPattern<ContractionOp> {
     fillBuffer(rewriter, loc, cionAdaptor.init(), alloc.resultMemRef,
                alloc.rankedTensorType.getShape());
 
-    // Determine ranges
-    SmallVector<int64_t, 8> ranges;
+    // Determine lower and upper bounds.
+    SmallVector<AffineExpr, 8> ubExpr;
```
flaub, line 630: Rename this `ubExprs`

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -626,20 +626,25 @@ struct ContractionOpConversion : public OpConversionPattern<ContractionOp> {
...
     auto lowerBounds = op.lowerBounds().getValue();
     auto upperBounds = op.upperBounds().getValue();
     assert(lowerBounds.getNumResults() == upperBounds.getNumResults() &&
            "mismatched dims for lower and upper bounds");
     for (unsigned i = 0; i < lowerBounds.getNumResults(); i++) {
-      auto rangeExpr = upperBounds.getResult(i) - lowerBounds.getResult(i) + 1;
-      auto range = rangeExpr.cast<AffineConstantExpr>().getValue();
-      ranges.emplace_back(range);
+      auto uExpr = upperBounds.getResult(i) + 1;
```
flaub, line 636: Name this `ubExpr`

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -626,20 +626,25 @@ struct ContractionOpConversion : public OpConversionPattern<ContractionOp> {
...
+      auto upper = uExpr.cast<AffineConstantExpr>().getValue();
+      ubExpr.push_back(rewriter.getAffineConstantExpr(upper));
     }
 
+    auto ubMap =
+        AffineMap::get(0, 0, (ArrayRef<AffineExpr>)ubExpr, op.getContext());
     // Make the outer loops
-    auto forOp = rewriter.create<AffineParallelOp>(loc, ranges);
+    auto forOp = rewriter.create<AffineParallelOp>(
+        loc, op.lowerBounds().getValue(), llvm::SmallVector<Value, 0>{}, ubMap,
```
flaub, line 645: Name these params like:

```
/*lbMap=*/op.lowerBounds().getValue(),
/*lbArgs=*/ArrayRef<Value>{},
/*ubMap=*/ubMap,
/*ubArgs=*/ArrayRef<Value>{});
```

Also, don't use `SmallVector<Value, 0>`, use `ArrayRef<Value>` instead.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -626,20 +626,25 @@ struct ContractionOpConversion : public OpConversionPattern<ContractionOp> {
...
-      auto rangeExpr = upperBounds.getResult(i) - lowerBounds.getResult(i) + 1;
-      auto range = rangeExpr.cast<AffineConstantExpr>().getValue();
-      ranges.emplace_back(range);
+      auto uExpr = upperBounds.getResult(i) + 1;
+      auto upper = uExpr.cast<AffineConstantExpr>().getValue();
+      ubExpr.push_back(rewriter.getAffineConstantExpr(upper));
     }
 
+    auto ubMap =
+        AffineMap::get(0, 0, (ArrayRef<AffineExpr>)ubExpr, op.getContext());
```
flaub, line 642: Use `{ubExpr}` instead of using a c-style cast.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
LLITCHEV: Renamed.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
LLITCHEV: Renamed.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
flaub: Actually, why do we need a cast at all? Won't `ubExprs` work here?

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -626,20 +626,27 @@ struct ContractionOpConversion : public OpConversionPattern<ContractionOp> {
...
+      ubExprs.push_back(rewriter.getAffineConstantExpr(upper));
     }
 
+    auto ubMap = AffineMap::get(0, 0, {ubExprs}, op.getContext());
     // Make the outer loops
-    auto forOp = rewriter.create<AffineParallelOp>(loc, ranges);
+    auto forOp = rewriter.create<AffineParallelOp>(
+        loc,
+        /*lbMap=*/op.lowerBounds().getValue(),
+        /*lbArgs=*/llvm::SmallVector<Value, 0>{},
```
flaub, line 646: Use ArrayRef here

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
LLITCHEV: Casting was needed because the compiler got a error with unable to deduce type. Using the C++ style helped the compiler along and cast is not needed.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
LLITCHEV: Changed the type.

plaidml/bridge/keras/backend_test.py
```diff
@@ -279,14 +279,15 @@ def output(self, *args):
...
                                 didx, idx))
-                        for x in range(0, len(pmlr[1])):
-                            npt.assert_allclose(
-                                pmlr[1][x],
-                                thr[1][x],
-                                rtol=tol,
-                                atol=atol,
-                                err_msg='ERR: datum={}, test={}, grad, x=plaidml, y=theano'.format(
-                                    didx, idx))
+                        if do_grads:
```
flaub, line 282: I don't think this is right, don't we want to always compare results between backends no matter if we're performing grads or not?

LLITCHEV: I just got the code from the tensorflow case - few lines below. It is doing the same thing. The issue is that if do_grad is not true, pmlr[1] is not populated and we get a RT exception with no element at index.

flaub: OK, I see, fair enough.

PR #1038, opened 2020-05-01
Enable some conv tests

LLITCHEV:


tzerrell:
oh huh, apparently those parts of CI are non-blocking?

flaub:
That's the 'public' bazel farm builds. The only builds we block PRs on are the internal ones.

flaub:
BTW, the build is broken because OpenVINO just moved to their own org:

https://github.com/opencv/dldt
is now
https://github.com/openvinotoolkit/openvino

PR #1041, opened 2020-05-07
Autotiling prototype change

shiyao00:
The cost model needs more information including the original AffineParallelOp and target info. So we need to add these info to findBestTileSize and cost model.

### DIFF
--- a/pmlc/dialect/pxa/transforms/autotile.h
+++ b/pmlc/dialect/pxa/transforms/autotile.h
@@ -70,11 +70,15 @@ private:
   Single single;
 };
 
+// Empty target info. Subclasses for different targets.
+class TargetInfo {
+};
+
 // A tile cost model is a functor from an array of tile sizes (i.e.
 // ArrayRef<int64_t>) to a double, which is 'inf' for infeasible tilings.  For
 // example:
 
-inline double DummyCostModel(mlir::AffineParallelOp op, ArrayRef<int64_t> tile, void* args) {
+inline double DummyCostModel(mlir::AffineParallelOp op, ArrayRef<int64_t> tile, const TargetInfo& info) {
   return 1.0;
 }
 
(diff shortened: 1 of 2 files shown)


pmlc/dialect/pxa/transforms/autotile.h
```diff
@@ -74,14 +74,18 @@ class UnionGenerator<Single> {
...
 // Given a generator and cost model, find the best tile size, return empty
 // tiling when all tiles are infeasible
 template <typename Generator, typename CostModel>
-llvm::SmallVector<int64_t, 8> findBestTileSize(const Generator &generator,
+llvm::SmallVector<int64_t, 8> findBestTileSize(mlir::AffineParallelOp op,
+                                               const Generator &generator,
                                                const CostModel &costModel,
-                                               ArrayRef<int64_t> ranges) {
+                                               ArrayRef<int64_t> ranges,
+                                               void *args) {
```
flaub, line 88: I'd like to avoid using `void*` for context, what do you want to pass here? Couldn't we use a lambda that captures any required context? Or even a functor object?

shiyao00: args is for the target info (memory size, #processors...). Do you have any better idea?

flaub: Yes, the thing that plugs in the cost model could contain all this extra info. Either as captures to a lambda or as a Functor object.

shiyao00: I want to pass a structure including the above info. However, the structures may differ for different Autotile examples.

pmlc/dialect/pxa/transforms/autotile.h
```diff
@@ -70,17 +70,25 @@ class UnionGenerator<Single> {
...
+}
 
 // Given a generator and cost model, find the best tile size, return empty
 // tiling when all tiles are infeasible
 template <typename Generator, typename CostModel>
-llvm::SmallVector<int64_t, 8> findBestTileSize(const Generator &generator,
+llvm::SmallVector<int64_t, 8> findBestTileSize(mlir::AffineParallelOp op,
+                                               const Generator &generator,
                                                const CostModel &costModel,
+                                               const TargetInfo &targetInfo,
```
flaub: I don't think we need to pass anything extra here, the `CostModel` callback can contain any target-specific context you need.

Imagine this:

```
struct MyCostModel {
  MyCostModel(unsigned numCores) : numCores(numCores) {}

  double operator()(ArrayRef<int64_t> tile) {
    // use numCores to compute cost
    return numCores * 3;
  }

  unsigned numCores;
};
```

Then to use it:
```
auto tileSize = findBestTileSize(MyGenerator{}, MyCostModel{10}, *ranges);
```

PR #1044, opened 2020-05-11
Initial implementation of PRNG function using MLIR

LLITCHEV:
This PR implements Prng function implementation using the MLIR.

### DIFF
--- a/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
+++ b/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
@@ -537,7 +537,7 @@ struct BufferAllocator {
   }
 };
 
-struct TilePrngOpConversion : public OpConversionPattern<PrngOp> {
+struct PrngOpConversion : public OpConversionPattern<PrngOp> {
   using OpConversionPattern<PrngOp>::OpConversionPattern;
 
   LogicalResult
@@ -968,8 +968,7 @@ struct LowerTileToPXAPass : public LowerTileToPXABase<LowerTileToPXAPass> {
     patterns.insert<
         TileConstantOpConversion, CastOpConversion, FuncOpConversion,
         IndexOpConversion, ReturnOpConversion, ScalarConstantOpConversion,
-        ShapeOpConversion, TraceOpConversion, TilePrngOpConversion,
-        // TODO: PrngOpConversion
+        ShapeOpConversion, TraceOpConversion, PrngOpConversion,
         // TODO: SpecialOpConversion (GatherOp, ReshapeOp,
         // ScatterOp, ZeroOp)
         ContractionOpConversion<CombinationKind::none, FirstOperand>,
--- a/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
+++ b/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
@@ -34,6 +34,7 @@ using dialect::tile::ContractionOp;
 using dialect::tile::ContractionOpOperandAdaptor;
 using dialect::tile::getPaddingInfo;
 using dialect::tile::IndexOp;
+using dialect::tile::PrngOp;
 using dialect::tile::ShapeOp;
 using dialect::tile::ShapeOpOperandAdaptor;
 using dialect::tile::TraceOp;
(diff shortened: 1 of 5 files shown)


pmlc/rt/prng.cc
```diff
@@ -0,0 +1,58 @@
+// Copyright 2020 Intel Corporation
+#include "mlir/ExecutionEngine/RunnerUtils.h"
+#include "pmlc/compiler/registry.h"
+#include "pmlc/util/logging.h"
+
+extern "C" void
+plaidml_rt_prng(size_t stateRank, UnrankedMemRefType<float> *state,
+                size_t resultRank, UnrankedMemRefType<float> *result,
+                size_t newStateRank, UnrankedMemRefType<float> *newState) {
+  if (resultRank <= 0) {
```
flaub, line 10: Since you're using a `size_t`, `<= 0` should never be possible.

On another note, use `unsigned` rather than `size_t`.

pmlc/rt/prng.cc
```diff
@@ -0,0 +1,58 @@
+// Copyright 2020 Intel Corporation
+#include "mlir/ExecutionEngine/RunnerUtils.h"
+#include "pmlc/compiler/registry.h"
+#include "pmlc/util/logging.h"
+
+extern "C" void
+plaidml_rt_prng(size_t stateRank, UnrankedMemRefType<float> *state,
```
flaub, line 7: `state` and `newState` should be `UnrankedMemRefType<int32_t>`

pmlc/rt/prng.cc
```diff
@@ -0,0 +1,58 @@
...
+                size_t resultRank, UnrankedMemRefType<float> *result,
+                size_t newStateRank, UnrankedMemRefType<float> *newState) {
+  if (resultRank <= 0) {
+    // Nothing to do or a bad parameter.
+    return;
+  }
+
+  // From the StridedMemRef types below we use only the sizes[]. Their offset is
+  // independent of the value of the second type parameter.
+  StridedMemRefType<int, 1> *stateStrided = (StridedMemRefType<int, 1> *)state;
```
flaub, line 17: Don't use C-style casts, please use a C++ one, like `static_cast<>()` or `reinterpret_cast<>()`

pmlc/rt/prng.cc
```diff
@@ -0,0 +1,58 @@
...
+  int *in_state = stateStrided->data + stateStrided->offset;
+  float *buf = resultStrided->data + resultStrided->offset;
+  int *out_state = newStateStrided->data + newStateStrided->offset;
+
+  // A reimplementation of the PRNG from tile/lang/gen_special.cc.
+  // x_n = (s1_n ^ s2_n ^ s3_n)
+  // s1_{n+1} = (((s1_n & 4294967294) <<12) ^ (((s1_n <<13) ^ s1_n) >>19))
+  // s2_{n+1} = (((s2_n & 4294967288) << 4) ^ (((s2_n << 2) ^ s2_n) >>25))
+  // s3_{n+1} = (((s3_n & 4294967280) <<17) ^ (((s3_n << 3) ^ s3_n) >>11))
+  for (size_t i = 0; i < count; ++i) {
```
flaub, line 37: Use `unsigned` rather than `size_t`

pmlc/rt/prng.cc
LLITCHEV: Resolved

pmlc/rt/prng.cc
LLITCHEV: Resolved

pmlc/rt/prng.cc
LLITCHEV: resolved

pmlc/rt/prng.cc
LLITCHEV: Resolved

pmlc/rt/prng.cc
```diff
@@ -0,0 +1,59 @@
+// Copyright 2020 Intel Corporation
+#include "mlir/ExecutionEngine/RunnerUtils.h"
+#include "pmlc/compiler/registry.h"
+#include "pmlc/util/logging.h"
+
+extern "C" void
+plaidml_rt_prng(unsigned stateRank, UnrankedMemRefType<float> *state,
+                unsigned resultRank, UnrankedMemRefType<float> *result,
+                unsigned newStateRank, UnrankedMemRefType<float> *newState) {
```
flaub, line 9: These actually aren't `UnrankedMemRefType<float>*`, they are actually `StridedMemRefType<T, N>*`.

So I would replace this with:

```
plaidml_rt_prng(unsigned stateRank, StridedMemRefType<int32_t, 1> *state,
                unsigned resultRank, StridedMemRefType<float, 1> *result,
                unsigned newStateRank, StridedMemRefType<int32_t, 1> *newState) {
```

And then you can drop the casts below.

LLITCHEV: Resolved

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -536,6 +537,67 @@ struct BufferAllocator {
   }
 };
 
+struct TilePrngOpConversion : public OpConversionPattern<PrngOp> {
```
flaub, line 540: Rename to `PrngOpConversion`

LLITCHEV: Resolved

flaub:
LGTM, I'll let @jbruestle have the final say.

pmlc/rt/prng.cc
```diff
@@ -0,0 +1,52 @@
...
+  // s3_{n+1} = (((s3_n & 4294967280) <<17) ^ (((s3_n << 3) ^ s3_n) >>11))
+  for (unsigned i = 0; i < count; ++i) {
+    buf[i] = (in_state[0] ^ in_state[1] ^ in_state[2]) / 4294967296.0;
+    out_state[0] = (((in_state[0] & 4294967294) << 12) ^
+                    (((in_state[0] << 13) ^ in_state[0]) >> 19));
+    out_state[1] = (((in_state[1] & 4294967288) << 4) ^
+                    (((in_state[1] << 2) ^ in_state[1]) >> 25));
+    out_state[2] = (((in_state[2] & 4294967280) << 17) ^
+                    (((in_state[2] << 3) ^ in_state[2]) >> 11));
+    in_state = out_state;
```
jbruestle, line 39: This pointer reassignment is a bit wierd and seems likely to confuse the optimizer, in a perfect world I probably would load in_state into 3 locals, do all the PRNG on those locals, and then write to out_state, but not a blocker...

LLITCHEV: That was just the code in thde S\stripe version of plaidml. But I can change it.

LLITCHEV: Hoisted the in_state out of loop and changed the rest of the code not to do pointer assignment.

LLITCHEV: Actually, looking a bit more carefully, this is the right way the code should work. The assignment of in_state_to out_state makes sure the next iteration through the loop uses the new state, without actually modifying the incoming in_state collection.

PR #1046, opened 2020-05-11
tile.reshape -> stdx.reshape -> view rewrite

mars1026:

### DIFF
--- a/pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
+++ b/pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
@@ -4,6 +4,7 @@
 #include "mlir/Conversion/StandardToLLVM/ConvertStandardToLLVM.h"
 #include "mlir/Conversion/StandardToLLVM/ConvertStandardToLLVMPass.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/Dialect/StandardOps/EDSC/Intrinsics.h"
 #include "mlir/Dialect/StandardOps/IR/Ops.h"
 #include "mlir/IR/AffineMap.h"
 #include "mlir/Pass/Pass.h"
@@ -17,6 +18,7 @@ using namespace mlir; // NOLINT[build/namespaces]
 namespace pmlc::conversion::stdx_to_llvm {
 
 namespace stdx = dialect::stdx;
+namespace edsc = mlir::edsc;
 
 namespace {
 
@@ -98,6 +100,77 @@ struct FPToUILowering : public LLVMLegalizationPattern<stdx::FPToUIOp> {
   }
 };
 
+class BaseViewConversionHelper {
+public:
+  explicit BaseViewConversionHelper(Type type)
+      : d(MemRefDescriptor::undef(rewriter(), loc(), type)) {}
+
+  explicit BaseViewConversionHelper(Value v) : d(v) {}
+
+  /// Wrappers around MemRefDescriptor that use EDSC builder and location.
+  Value allocatedPtr() { return d.allocatedPtr(rewriter(), loc()); }
+  void setAllocatedPtr(Value v) { d.setAllocatedPtr(rewriter(), loc(), v); }
+  Value alignedPtr() { return d.alignedPtr(rewriter(), loc()); }
+  void setAlignedPtr(Value v) { d.setAlignedPtr(rewriter(), loc(), v); }
+  Value offset() { return d.offset(rewriter(), loc()); }
+  void setOffset(Value v) { d.setOffset(rewriter(), loc(), v); }
+  Value size(unsigned i) { return d.size(rewriter(), loc(), i); }
+  void setSize(unsigned i, Value v) { d.setSize(rewriter(), loc(), i, v); }
+  void setConstantSize(unsigned i, int64_t v) {
+    d.setConstantSize(rewriter(), loc(), i, v);
+  }
+  Value stride(unsigned i) { return d.stride(rewriter(), loc(), i); }
+  void setStride(unsigned i, Value v) { d.setStride(rewriter(), loc(), i, v); }
+  void setConstantStride(unsigned i, int64_t v) {
+    d.setConstantStride(rewriter(), loc(), i, v);
+  }
+
+  operator Value() { return d; }
+
+private:
+  OpBuilder &rewriter() { return edsc::ScopedContext::getBuilder(); }
+  Location loc() { return edsc::ScopedContext::getLocation(); }
+
+  MemRefDescriptor d;
+};
+
+struct ReshapeLowering : public LLVMLegalizationPattern<stdx::ReshapeOp> {
+  using LLVMLegalizationPattern<stdx::ReshapeOp>::LLVMLegalizationPattern;
+  using Base = LLVMLegalizationPattern<stdx::ReshapeOp>;
+
+  LogicalResult
+  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto reshapeOp = cast<stdx::ReshapeOp>(op);
+    MemRefType dstType = reshapeOp.getResult().getType().cast<MemRefType>();
+
+    if (!dstType.hasStaticShape())
+      return failure();
+
+    int64_t offset;
+    SmallVector<int64_t, 4> strides;
+    auto res = getStridesAndOffset(dstType, strides, offset);
+    if (failed(res) || llvm::any_of(strides, [](int64_t val) {
+          return ShapedType::isDynamicStrideOrOffset(val);
+        }))
+      return failure();
+
+    edsc::ScopedContext context(rewriter, op->getLoc());
+    stdx::ReshapeOpOperandAdaptor adaptor(operands);
+    BaseViewConversionHelper baseDesc(adaptor.tensor());
+    BaseViewConversionHelper desc(typeConverter.convertType(dstType));
+    desc.setAllocatedPtr(baseDesc.allocatedPtr());
+    desc.setAlignedPtr(baseDesc.alignedPtr());
+    desc.setOffset(baseDesc.offset());
+    for (auto en : llvm::enumerate(dstType.getShape()))
+      desc.setConstantSize(en.index(), en.value());
+    for (auto en : llvm::enumerate(strides))
+      desc.setConstantStride(en.index(), en.value());
+    rewriter.replaceOp(op, {desc});
+    return success();
+  }
+};
+
 /// A pass converting MLIR operations into the LLVM IR dialect.
 struct LowerToLLVMPass : public LowerToLLVMBase<LowerToLLVMPass> {
   // Run the dialect converter on the module.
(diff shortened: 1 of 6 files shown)


pmlc/conversion/tile_to_pxa/tests/reshape.mlir
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// CHECK-LABEL: func @reshaper
+// CHECK: stdx.reshape
+// CHECK: memref<2x20xf32>
+
+// -----
+
+module {
+  func @reshaper(%arg0: tensor<5x6xf32>, %arg1: tensor<5x2x3xf32>) -> tensor<5x6xf32> {
```
flaub, line 33: Doesn't look like `%arg0` is used here, if so drop it.

pmlc/conversion/tile_to_pxa/tests/reshape.mlir
```diff
@@ -0,0 +1,59 @@
...
+module {
+  func @reshaper(%arg0: tensor<2x4x5xf32>) -> tensor<2x20xf32> {
+    %c2 = tile.constant 2
+    %c20 = tile.constant 20
+    %0 = "tile.reshape"(%arg0, %c2, %c20) : (tensor<2x4x5xf32>, index, index) -> tensor<2x20xf32>
+    return %0 : tensor<2x20xf32>
+  }
+}
+
+// CHECK-LABEL: func @reshaper
```
flaub, line 26: Use different func names and labels so that FileCheck won't accidentally validate the wrong function.

pmlc/dialect/stdx/ir/ops.cc
```diff
@@ -85,6 +85,26 @@ static void buildFPToUIOp(Builder *builder, OperationState &result,
...
 
+// ---- ReshapeOp ----
+
+void printReshapeOp(OpAsmPrinter &p, ReshapeOp op) {
+  p << op.getOperation()->getName() << " (" << op.getOperand() << ") : ("
+    << op.getOperand().getType() << ") -> "
+    << op.getOperation()->getResult(0).getType();
+}
+
+ParseResult parseReshapeOp(OpAsmParser &parser, OperationState &result) {
```
flaub, line 96: If we're going to have a custom printer we need to have a custom parser.

flaub:
For any new ops that we add that have custom printer/parsers, we need to have a `roundtrip.mlir` to test this. Search for `roundtrip.mlir` for examples (both in our code and upstream MLIR).

Additionally, for any ops that have custom verifiers, we need to have an `invalid.mlir`.

pmlc/conversion/tile_to_pxa/tests/reshape.mlir
```diff
@@ -0,0 +1,59 @@
...
+    %c5 = tile.constant 5
+    %c6 = tile.constant 6
+    %0 = "tile.reshape"(%arg1, %c5, %c6) : (tensor<5x2x3xf32>, index, index) -> tensor<5x6xf32>
+    return %0 : tensor<5x6xf32>
+  }
+}
+
+// CHECK-LABEL: func @reshaper
+// CHECK: stdx.reshape
+// CHECK: memref<5x6xf32>
```
flaub, line 43: Could we add more to the test so that it verifies that the strides are being manipulated? Or am I misunderstanding what the `BaseViewConversionHelper` is doing?

flaub: Oh, perhaps what I'm missing is that this is a tile -> stdx test. How about we have a stdx -> LLVM test as well?

pmlc/conversion/tile_to_pxa/tests/reshape.mlir
```diff
@@ -1,44 +1,44 @@
 // RUN: pmlc-opt -convert-tile-to-pxa -canonicalize -cse %s | FileCheck %s
 
 module {
-  func @reshaper(%arg0: tensor<1x1x60xf32>) -> tensor<60xf32> {
+  func @reshaper0(%arg0: tensor<1x1x60xf32>) -> tensor<60xf32> {
     %c60 = tile.constant 60
     %0 = "tile.reshape"(%arg0, %c60) : (tensor<1x1x60xf32>, index) -> tensor<60xf32>
```
flaub, line 6: We added a custom printer/parser (via the declarative assembly) but this is using the generic format.

flaub: Oops, nevermind :)

pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
```diff
@@ -98,6 +100,77 @@ struct FPToUILowering : public LLVMLegalizationPattern<stdx::FPToUIOp> {
...
+    d.setConstantStride(rewriter(), loc(), i, v);
+  }
+
+  operator Value() { return d; }
+
+private:
+  OpBuilder &rewriter() { return edsc::ScopedContext::getBuilder(); }
+  Location loc() { return edsc::ScopedContext::getLocation(); }
+
+  MemRefDescriptor d;
```
flaub, line 134: Would prefer not to use single letter variable names

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -802,6 +804,28 @@ struct IndexOpConversion : public OpConversionPattern<IndexOp> {
...
+    IVLOG(2, "ReshapeOpConversion::matchAndRewrite>");
+
+    // Create an adaptor, to interpret the operands
+    ReshapeOpOperandAdaptor adaptor(operands);
+
+    auto tensor = adaptor.tensor();
+
+    TypeConverter typeConverter;
+    auto resultType =
+        typeConverter.convertType(op.result().getType()).cast<MemRefType>();
```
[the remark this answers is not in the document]
mars1026: good catch! the custom builder function needed it but the declarative-assembly version doesn't

pmlc/conversion/stdx_to_llvm/tests/reshape.mlir
```diff
@@ -0,0 +1,85 @@
...
+module {
+  func @reshaper0(%arg0: memref<1x1x60xf32>) -> memref<60xf32> {
+    %0 = stdx.reshape(%arg0) : (memref<1x1x60xf32>) -> memref<60xf32>
+    return %0 : memref<60xf32>
+  }
+}
+
+// CHECK-LABEL: func @reshaper0
+// CHECK: undef : !llvm<"{ float*, float*, i64, [3 x i64], [3 x i64] }">
+// CHECK: [[index0:%[0-9]+]] = llvm.mlir.constant(60 : index)
```
flaub, line 12: Usually we do `%[[index0:.*]]` because we have no guarantee that the printer will continue to use numbers, but it does always prefix with `%`

pmlc/conversion/stdx_to_llvm/tests/reshape.mlir
```diff
@@ -0,0 +1,85 @@
...
+// CHECK: insertvalue [[index0]], %{{[0-9]+}}[3, 0]
+// CHECK: [[index1:%[0-9]+]] = llvm.mlir.constant(1 : index)
+// CHECK: insertvalue [[index1]], %{{[0-9]+}}[4, 0]
+// CHECK: return {{%[0-9]+}} : !llvm<"{ float*, float*, i64, [1 x i64], [1 x i64] }">
```
flaub, line 16: Likewise, for wildcards/don't care, use: `%{{.*}}`

pmlc/conversion/stdx_to_llvm/tests/reshape.mlir
```diff
@@ -9,11 +9,11 @@ module {
 
 // CHECK-LABEL: func @reshaper0
 // CHECK: undef : !llvm<"{ float*, float*, i64, [3 x i64], [3 x i64] }">
-// CHECK: [[index0:%[0-9]+]] = llvm.mlir.constant(60 : index)
-// CHECK: insertvalue [[index0]], %{{[0-9]+}}[3, 0]
-// CHECK: [[index1:%[0-9]+]] = llvm.mlir.constant(1 : index)
-// CHECK: insertvalue [[index1]], %{{[0-9]+}}[4, 0]
-// CHECK: return {{%[0-9]+}} : !llvm<"{ float*, float*, i64, [1 x i64], [1 x i64] }">
+// CHECK: [[index0:%.+]] = llvm.mlir.constant(60 : index)
```
flaub, line 12: But you can put `%` outside the regex, because MLIR vars will always start with `%`

pmlc/dialect/stdx/ir/ops.td
```diff
@@ -43,4 +44,17 @@ def FPToUIOp :
...
+class StdX_Plain_Op<string mnemonic, list<OpTrait> traits = []> :
+    Op<StdX_Dialect, mnemonic, traits> {
+}
+
+def ReshapeOp : StdX_Plain_Op<"reshape"> {
+  let summary = "tensor reshape operation";
+  let arguments = (ins AnyMemRef:$tensor);
+  let results = (outs AnyMemRef:$result);
+  let assemblyFormat = [{
+    `(` $tensor `)` attr-dict `:` functional-type($tensor, results)
```
flaub, line 56: We need an `roundtrip.mlir` test for this custom printer/parser.

pmlc/dialect/stdx/ir/ops.td
```diff
@@ -43,4 +44,17 @@ def FPToUIOp :
   }]>];
 }
 
+class StdX_Plain_Op<string mnemonic, list<OpTrait> traits = []> :
+    Op<StdX_Dialect, mnemonic, traits> {
+}
```
flaub, line 49: Why do we need a different base class?

mars1026: I wanted to use assemblyFormat and therefore do not want the declarations for custom print/parse/verify functions supplied in StdX_Op.  If I define ReshapeOp based on StdX_Op instead, the compiler complains that I have not provided implementations.

pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
```diff
@@ -98,6 +100,79 @@ struct FPToUILowering : public LLVMLegalizationPattern<stdx::FPToUIOp> {
...
+
+struct ReshapeLowering : public LLVMLegalizationPattern<stdx::ReshapeOp> {
+  using LLVMLegalizationPattern<stdx::ReshapeOp>::LLVMLegalizationPattern;
+  using Base = LLVMLegalizationPattern<stdx::ReshapeOp>;
+
+  LogicalResult
+  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto reshapeOp = cast<stdx::ReshapeOp>(op);
+    MemRefType dstType = reshapeOp.getResult().getType().cast<MemRefType>();
```
LLITCHEV, line 147: I'd just add here a defensive check to make sure reshapeOp is not nullptr. Other than that it looks fine to me.

flaub: We're already ensuring the type is `stdx::ReshapeOp` because the base class is templated on this type. The `cast` already has an assertion inside of it, so no need for additional checks.

PR #1048, opened 2020-05-12
Canonicalize with partially static shape

tzerrell:
Adds canonicalization of SymbolicContractions with partially static shapes. I think it's forward progress, but this does not fully resolve this issues with PlaidBench ResNet50:
* At the end of compiling the network, PlaidBench crashes with unclear cause (not yet investigated).
* Network compilation performance is bad (like 5 minutes maybe on my Mac). I think (but haven't looked closely) that when computeShape is called it rebuilds & re-canonicalizes the entire network up to the selected op, and gets pretty bad when computeShape is called a bunch.

I'm uncertain whether the crash is related to this PR, or whether it has always existed and we're only now getting far enough to hit it. I'm therefore also uncertain whether to try to land this immediately or wait until investigating the crash.

### DIFF
--- a/pmlc/dialect/tile/ir/ops.cc
+++ b/pmlc/dialect/tile/ir/ops.cc
@@ -387,9 +387,27 @@ struct SymbolicContractionCanonicalizer
     auto sourceType = op.result().getType().cast<RankedTensorType>();
     auto resultType = RankedTensorType::get(shape, sourceType.getElementType());
     if (!resultType.hasStaticShape()) {
-      return failure();
+      if (resultType == sourceType) {
+        return failure();
+      }
+      // Can't rewrite to a non-symbolic contraction, but we can at least update
+      // the result type
+      auto newOp = rewriter.create<SymbolicContractionOp>(
+          op.getLoc(), resultType, op.init(), op.cons(), op.size(), op.sink(),
+          op.srcs(), op.agg(), op.combo(), rewriter.getUnitAttr(),
+          rewriter.getStringAttr(op.name().getValueOr("")));
+      // TODO: This seems like not the right way to handle no_reduce
+      if (op.no_reduce().hasValue()) {
+        newOp.setAttr("no_reduce", rewriter.getUnitAttr());
+      } else {
+        newOp.removeAttr("no_reduce");
+      }
+      rewriter.replaceOp(op, newOp.result());
+      util::UpdateFuncOpType(newOp.getOperation());
+      return success();
     }
 
+    // TODO: Do I need to verify foldability first?
     IsFoldableVisitor foldable_checker;
     if (!foldable_checker.is_foldable(op)) {
       return failure();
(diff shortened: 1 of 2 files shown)


pmlc/dialect/tile/ir/ops.cc
```diff
@@ -387,9 +387,27 @@ struct SymbolicContractionCanonicalizer
...
+        newOp.setAttr("no_reduce", rewriter.getUnitAttr());
+      } else {
+        newOp.removeAttr("no_reduce");
+      }
+      rewriter.replaceOp(op, newOp.result());
+      util::UpdateFuncOpType(newOp.getOperation());
+      return success();
     }
 
+    // TODO: Do I need to verify foldability first?
```
flaub, line 410: Yes, we can't fold min/max.

tzerrell: That seems to me like it means I _don't_ need to move the foldability check above the partially static shape code. The partially static shape code is just rewriting the result type, not changing from a symbolic contraction to a non-symbolic contraction. So it seems like inability to fold min/max shouldn't matter. Or am I missing something?

flaub: OK, let's remove the comment then.

pmlc/dialect/tile/ir/ops.cc
```diff
@@ -387,9 +387,27 @@ struct SymbolicContractionCanonicalizer
...
+      auto newOp = rewriter.create<SymbolicContractionOp>(
+          op.getLoc(), resultType, op.init(), op.cons(), op.size(), op.sink(),
+          op.srcs(), op.agg(), op.combo(), rewriter.getUnitAttr(),
+          rewriter.getStringAttr(op.name().getValueOr("")));
+      // TODO: This seems like not the right way to handle no_reduce
+      if (op.no_reduce().hasValue()) {
+        newOp.setAttr("no_reduce", rewriter.getUnitAttr());
+      } else {
+        newOp.removeAttr("no_reduce");
+      }
```
flaub, line 404: I think you can say:

```
      auto nameAttr = rewriter.getStringAttr(op.name().getValueOr(""));
      auto reduceAttr = op.no_reduce().hasValue() ? rewriter.getUnitAttr() : UnitAttr{};
      auto newOp = rewriter.create<SymbolicContractionOp>(
          op.getLoc(), resultType, op.init(), op.cons(), op.size(), op.sink(),
          op.srcs(), op.agg(), op.combo(), reduceAttr,
          nameAttr);
```
```

File: pmlc/dialect/tile/ir/ops.cc [hunk 3]
```diff
@@ -387,9 +387,27 @@ struct SymbolicContractionCanonicalizer
...
+      rewriter.replaceOp(op, newOp.result());
```
flaub, line 405: I think you can just do `newOp` here (should cast to value), or `{newOp}`.

pmlc/dialect/tile/ir/ops.cc
tzerrell: Great, thanks!

I realize I don't currently have an easy way to confirm this does the same thing, so I'll be adding a lit test with these changes as well.

pmlc/dialect/tile/ir/ops.cc
tzerrell: It apparently needs the braces

flaub:
Nice test. Could we throw in one that has min/max so that we can see that everything is ok?

PR #1051, opened 2020-05-12
Use unsigned ints instead of ints

LLITCHEV:
Keras expects only positove random numbers.

### DIFF
--- a/pmlc/rt/prng.cc
+++ b/pmlc/rt/prng.cc
@@ -4,11 +4,11 @@
 #include "pmlc/util/logging.h"
 
 extern "C" void plaidml_rt_prng(unsigned stateRank,
-                                StridedMemRefType<int32_t, 1> *state,
+                                StridedMemRefType<uint32_t, 1> *state,
                                 unsigned resultRank,
                                 StridedMemRefType<float, 1> *result,
                                 unsigned newStateRank,
-                                StridedMemRefType<int32_t, 1> *newState) {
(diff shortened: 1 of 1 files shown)


flaub:
Hang on, this should have no effect from keras's point of view.

flaub:
This only changes the state data type, which is completely unrelated to the result data type.

tzerrell:
Huh. It very much appears to change the result of calls to `K.random_uniform` (producing values from `[0, 1)`, whereas w/out this change values were from `[-0.5,0.5)`). We're only manually checking what the output appears to be since the backend_test tests for these have additional errors, but the pattern was pretty clear.

tzerrell:
Perhaps something somewhere is (incorrectly?) retaining the signedness of the state when producing the result values?

flaub:
That's what I'm worried about, that perhaps there's some other issue.

jbruestle:
Two things are going on here.  1) Using int32 is fundamentally incorrect and changes how the shifts are interpreted, possibly causing all sorts of issues (sorry to have overlooked that, see the code here for reference https://github.com/spotify/linux/blob/master/lib/random32.c). 2) Additionally, the output will natively produce 32 bits of 'random' bits, and to convert that to a number [0, 1), we  do a cast / division (on line 35 above), which does two different things depending on the source type, which in fact does take signedness into account exactly as one would imagine.

PR #1053, opened 2020-05-12
Add prng signness test ex

LLITCHEV:


flaub:
We should probably also test that with a particular seed we always get a particular result.

jbruestle:
Hmm, CI is failing.  Very interesting.

LLITCHEV:
Yeah, that is weird...
The error is :


It probbaly has to do with these lines in the test. Although why does it fail on Linux only?
auto S = Placeholder(DType::UINT32, {3, 3});
  ..........
  std::vector<int> input = {
      -5, -6, -7,  //
      4,  5,  6,   //
      7,  8,  9,   //
  };
............
binder.input(S).copy_from(input.data());

LLITCHEV:
I traced the failure...
This issue happens only in optimized Linux builds - the dbg ones are fine.
The issue is an optimization that the C++ compiler makes on Linux that generates the wrong code. It has to do with the StridedMemRef<XXX, 1> in the signature of the RT PRNG function.
In this particular case, the compiler uses the value of the type parameter to optimize the loop for calculating the count of outputs and doesn't fill all the values.
A bigger value for the. type parameter (like 20) should be sufficient enough and after talking to @tzerrell we agree on that. To note is that since the parameter is a pointer, there is no increase in memory used, but ony a bigger value for the optimizer to allow fo accessing properly the sizes[N] elemets.

PR #1054, opened 2020-05-12
Fix contractions returning zeroes

tzerrell:
Fixes an ordering issue in PXA to Affine where reduce ops with no indexes were having their bodies moved to the end of their containing block. In particular, this meant that reductions initializing the output of a contraction to zero were being moved after reductions doing the core work of a contraction.

This fixes the bug where most backend tests using a contraction returned tensors of all 0s.

### DIFF
--- a/pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
+++ b/pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
@@ -78,13 +78,13 @@ struct AffineParallelOpConversion : public LoweringBase<AffineParallelOp> {
     // instead put `op`'s body where `op` is.)
     auto &innerLoopOps = rewriter.getInsertionBlock()->getOperations();
     auto &stripeBodyOps = op.region().front().getOperations();
-    mlir::Block::iterator insertion_loc;
+    mlir::Block::iterator insertionLoc;
     if (op.lowerBoundsMap().getNumResults() > 0) {
-      insertion_loc = std::prev(innerLoopOps.end());
+      insertionLoc = std::prev(innerLoopOps.end());
     } else {
-      insertion_loc = mlir::Block::iterator(op);
+      insertionLoc = mlir::Block::iterator(op);
     }
-    innerLoopOps.splice(insertion_loc, stripeBodyOps, stripeBodyOps.begin(),
+    innerLoopOps.splice(insertionLoc, stripeBodyOps, stripeBodyOps.begin(),
                         std::prev(stripeBodyOps.end()));
     // Replace all uses of old values
     size_t idx = 0;
--- a/pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
+++ b/pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
@@ -57,6 +57,38 @@ struct LoweringBase : public OpConversionPattern<OpType> {
   LogicalResult match(Operation *op) const override { return mlir::success(); }
 };
 
+// This pattern removes affine.parallel ops with no induction variables
+struct AffineParallelRank0Remover
+    : public mlir::OpRewritePattern<AffineParallelOp> {
+  using mlir::OpRewritePattern<AffineParallelOp>::OpRewritePattern;
+
+  // Default benefit of 2 since this generally needs to run before
+  // AffineParallelOp conversion
+  explicit AffineParallelRank0Remover(MLIRContext *ctx,
+                                      mlir::PatternBenefit benefit = 2)
+      : OpRewritePattern(ctx, benefit) {}
+
+  LogicalResult
+  matchAndRewrite(AffineParallelOp op,
+                  mlir::PatternRewriter &rewriter) const override {
+    // Check that there are no induction variables
+    if (op.lowerBoundsMap().getNumResults() != 0)
+      return mlir::failure();
+    // Remove the affine.parallel wrapper, retain the body in the same location
+    auto &parentOps = rewriter.getInsertionBlock()->getOperations();
+    auto &parallelBodyOps = op.region().front().getOperations();
+    parentOps.splice(mlir::Block::iterator(op), parallelBodyOps,
+                     parallelBodyOps.begin(), std::prev(parallelBodyOps.end()));
+    // Replace outputs with values from yield
+    auto termIt = std::prev(parallelBodyOps.end());
+    for (size_t i = 0; i < op.getNumResults(); i++) {
+      op.getResult(i).replaceAllUsesWith(termIt->getOperand(i));
+    }
+    rewriter.eraseOp(op);
+    return mlir::success();
+  }
+};
+
 struct AffineParallelOpConversion : public LoweringBase<AffineParallelOp> {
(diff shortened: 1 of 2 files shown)


pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -73,10 +73,18 @@ struct AffineParallelOpConversion : public LoweringBase<AffineParallelOp> {
...
     // Move ParallelOp's operations (single block) to Affine innermost loop.
+    // (But when `op` has no indexes, there are no Affine loops and we need to
+    // instead put `op`'s body where `op` is.)
     auto &innerLoopOps = rewriter.getInsertionBlock()->getOperations();
     auto &stripeBodyOps = op.region().front().getOperations();
-    innerLoopOps.splice(std::prev(innerLoopOps.end()), stripeBodyOps,
-                        stripeBodyOps.begin(), std::prev(stripeBodyOps.end()));
+    mlir::Block::iterator insertion_loc;
+    if (op.lowerBoundsMap().getNumResults() > 0) {
+      insertion_loc = std::prev(innerLoopOps.end());
```
flaub, line 82: nit: use `camelBack` style

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -73,10 +73,18 @@ struct AffineParallelOpConversion : public LoweringBase<AffineParallelOp> {
...
+    } else {
+      insertion_loc = mlir::Block::iterator(op);
+    }
+    innerLoopOps.splice(insertion_loc, stripeBodyOps, stripeBodyOps.begin(),
```
flaub, line 86: I think it might be cleaner just to say:

```
if (op.lowerBoundsMap().getNumResults()) {
  innerLoopOps.splice(std::prev(innerLoopOps.end()), stripeBodyOps, stripeBodyOps.begin(), std::prev(stripeBodyOps.end()));
} else {
  innerLoopOps.splice(op, stripeBodyOps, stripeBodyOps.begin(), std::prev(stripeBodyOps.end()));
}
```

Also, we should probably excise the word `stripe`

flaub:
I'm having a little trouble visualizing the bug and the fix here. Could we add a `.mlir` test the illustrates the issue?

tzerrell:
> Could we add a `.mlir` test the illustrates the issue?

Good idea, I added a lit test that shows that even when the output rank is 0, we still need to perform the initialization of the output tensor prior to performing the main contraction (this is what was going wrong in the old version, when the rank was 0 it was putting the initialization of the output tensor at the_end_ of the "current innermost affine.for block". Which, since there are _no_ affine.for blocks for initialization in the rank 0 output case, ended up moving the initialization to the end of the main body of the code, i.e. _after_ the rest of the contraction).

That said, I discussed briefly with @jbruestle and we think it would be better to instead remove all AffineParallelOps that have 0 indexes prior to lowering from PXA to Affine. We then no longer need this special case code (perhaps retain an assert that the AffineParallelOp has rank >0?). Thus I've switched this PR to a draft for now while I make that change.

We could accomplish this by either adding the "remove 0-index AffineParallelOps" pattern to the canonicalizer and canonicalizing before we lower pxa_to_affine, or we could just add that pattern as an early step of pxa_to_affine lowering. Let me know if you have an opinion on which.

tzerrell:
Split out a rank 0 affine parallel op remover

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -57,6 +57,38 @@ struct LoweringBase : public OpConversionPattern<OpType> {
...
 };
 
+// This pattern removes affine.parallel ops with no induction variables
+struct AffineParallelRank0Remover
+    : public mlir::OpRewritePattern<AffineParallelOp> {
+  using mlir::OpRewritePattern<AffineParallelOp>::OpRewritePattern;
+
+  // Default benefit of 2 since this generally needs to run before
+  // AffineParallelOp conversion
+  explicit AffineParallelRank0Remover(MLIRContext *ctx,
```
jbruestle, line 67: I think rather than use benefit (which does work but eventually becomes complex to reason about since you need to order things carefully and it's an integer), we should modify the AffineParallelOpConversion to have a match which fails when the number of dims is == 0 (since in fact it *doesnt* work in that case).  Then only one of the two patterns will be applicable at any given time.

flaub: Oh, clever, let’s add a clear comment for this since it’s not entirely obvious. But seems reasonable.

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -57,11 +57,47 @@ struct LoweringBase : public OpConversionPattern<OpType> {
...
+      op.getResult(i).replaceAllUsesWith(termIt->getOperand(i));
+    }
+    rewriter.eraseOp(op);
+    return mlir::success();
+  }
+};
+
+struct AffineParallelOpConversion
+    : public OpConversionPattern<AffineParallelOp> {
+  MLIRContext *ctx;
```
flaub, line 91: Is this needed?

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -57,11 +57,47 @@ struct LoweringBase : public OpConversionPattern<OpType> {
...
+    for (size_t i = 0; i < op.getNumResults(); i++) {
+      op.getResult(i).replaceAllUsesWith(termIt->getOperand(i));
+    }
+    rewriter.eraseOp(op);
+    return mlir::success();
+  }
+};
+
+struct AffineParallelOpConversion
+    : public OpConversionPattern<AffineParallelOp> {
```
flaub, line 90: The other uses OpRewritePattern, maybe we should be consistent.

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
tzerrell: I'm mirroring the use in `LoweringBase`, which is what `AffineParallelOpConversion` had used prior to my change to the `matchAndRewrite` formulation. I don't actually know why it's in `LoweringBase`, so perhaps it isn't needed here.

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
tzerrell: The `AffineParallelRank0Remover` using `OpRewritePattern` is the inconsistent one; everything else uses `OpConversionPattern` via `LoweringBase`. I had originally thought of the rank 0 remover possibly seeing additional use outside of lowering, hence the use of a rewrite pattern, but we don't currently do that so perhaps I should investigate switching it to an OpConversionPattern?

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
tzerrell: I'll put together another PR to revise w/ this comment

PR #1059, opened 2020-05-13
Add vulkan integer and float datatype support

YangleiZouIntel:
Add vulkan rt support to integer and float types.
Refactored buffer binding functions.

### DIFF
--- a/pmlc/conversion/gpu/gpu_to_vulkan.cc
+++ b/pmlc/conversion/gpu/gpu_to_vulkan.cc
@@ -50,8 +50,27 @@ static constexpr const char *kCreateVulkanMemoryTransferAction =
     "createVulkanMemoryTransferAction";
 static constexpr const char *kAddVulkanLaunchActionToSchedule =
     "addVulkanLaunchActionToSchedule";
+
+static constexpr const char *kBindBufferBFloat16 = "bindBufferBFloat16";
+static constexpr const char *kBindBufferFloat16 = "bindBufferFloat16";
 static constexpr const char *kBindBufferFloat32 = "bindBufferFloat32";
-static constexpr const char *kBindBufferInt64 = "bindBufferInt64";
+static constexpr const char *kBindBufferFloat64 = "bindBufferFloat64";
+
+// These functions are signless, meaning they apply to both signed and unsigned
+// integers
+static constexpr const char *kBindBufferInteger8 = "bindBufferInteger8";
+static constexpr const char *kBindBufferInteger16 = "bindBufferInteger16";
+static constexpr const char *kBindBufferInteger32 = "bindBufferInteger32";
+static constexpr const char *kBindBufferInteger64 = "bindBufferInteger64";
+
+static constexpr const int byteBits = 8;
+static constexpr const int intMinBytes = 1;
+static constexpr const int intMaxBytes = 8;
+
+static constexpr const int bFloat16Bytes = 2;
+static constexpr const int float16Bytes = 2;
+static constexpr const int float32Bytes = 4;
+static constexpr const int float64Bytes = 8;
 
 /// A pass to convert gpu launch op to vulkan launch call op, by creating a
 /// SPIR-V binary shader from `spirv::ModuleOp` using `spirv::serialize`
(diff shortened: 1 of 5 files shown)


vendor/openvino/repo.bzl
```diff
@@ -32,8 +32,8 @@ def openvino_workspace():
 
     http_archive(
         name = "openvino",
-        sha256 = "7b416d45765aea2cdd42edcdfc4e6ab4d7c4d37587530bf51a43b4c261e24a16",
-        strip_prefix = "dldt-2020.1",
+        sha256 = "0d641b158fff868ae7a969c02bf73462e335c15c40b42e3fe65ac6ea7cffea66",
+        strip_prefix = "openvino-2020.1",
         url = "https://github.com/opencv/dldt/archive/2020.1.zip",
```
YangleiZouIntel, line 37: This is not related to the buffer binding. But the openvino zip file seems to have changed, I have to update the sha256 to be able to compile and pass CI. Please double check.

flaub: Good catch, could you make this into a separate PR?

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -109,15 +126,61 @@ class ConvertGpuLaunchFuncToVulkanCalls
   }
 
   const char *getBufferBindingFunc(Type &elementType) {
```
flaub, line 128: Don't pass `Type` by-reference, always use by-value semantics.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -109,15 +126,61 @@ class ConvertGpuLaunchFuncToVulkanCalls
...
     if (elementType.isF32()) {
       return kBindBufferFloat32;
     }
+    if (elementType.isF64()) {
+      return kBindBufferFloat64;
+    }
     return nullptr;
   }
 
+  void getElementTypeSize(Type &elementType, uint32_t &size) {
```
flaub, line 156: I'm not sure why we need this method. Isn't there a `type.getIntOrFloatBitWidth()` that you can use?

vendor/openvino/repo.bzl
YangleiZouIntel: Sure

pmlc/conversion/gpu/gpu_to_vulkan.cc
YangleiZouIntel: Thanks for pointing to this function, I will replace it by this function.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -50,8 +50,20 @@ static constexpr const char *kCreateVulkanMemoryTransferAction =
...
+static constexpr const char *kBindBufferFloat64 = "bindBufferFloat64";
+
+// These functions are signless, meaning they apply to both signed and unsigned
+// integers
+static constexpr const char *kBindBufferInteger8 = "bindBufferInteger8";
+static constexpr const char *kBindBufferInteger16 = "bindBufferInteger16";
+static constexpr const char *kBindBufferInteger32 = "bindBufferInteger32";
+static constexpr const char *kBindBufferInteger64 = "bindBufferInteger64";
+
+static constexpr const int byteBits = 8;
```
flaub, line 66: Use `kByteBits` for constant values.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -210,14 +235,27 @@ ConvertGpuLaunchFuncToVulkanCalls::bindBuffers(Location loc, OpBuilder &builder,
...
     if (auto memRefType = buffer.getType().dyn_cast_or_null<MemRefType>()) {
+      auto shape = memRefType.getShape();
+      uint32_t numElement = 1;
+      for (auto dim : shape) {
+        numElement *= dim;
+      }
+
       auto elementType = memRefType.getElementType();
+      bufferElementTypes.push_back(elementType);
+      uint32_t elementTypeSize = elementType.getIntOrFloatBitWidth() / byteBits;
```
flaub, line 246: Use `llvm::divideCeil` to account for `i1` types.

YangleiZouIntel: It is a good point, thanks.

PR #1061, opened 2020-05-14
Fix an issue where the conversion map was not updated

LLITCHEV:
The issue was that calling op.replaceAllUses is not updating the rewriter map of new/old operations. Use the rewriter.replaceOp(...) instead.

### DIFF
--- a/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
+++ b/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
@@ -570,8 +570,7 @@ struct PrngOpConversion : public OpConversionPattern<PrngOp> {
         loc, symbol, ArrayRef<Type>{},
         ArrayRef<Value>{stateCast, resultCast, newStateCast});
 
-    rewriter.replaceOp(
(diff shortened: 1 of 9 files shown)


pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -570,10 +570,8 @@ struct PrngOpConversion : public OpConversionPattern<PrngOp> {
         loc, symbol, ArrayRef<Type>{},
         ArrayRef<Value>{stateCast, resultCast, newStateCast});
 
-    op.result().replaceAllUsesWith(allocResult.resultMemRef);
-    op.new_state().replaceAllUsesWith(allocState.resultMemRef);
-
-    rewriter.eraseOp(op);
+    rewriter.replaceOp(
+        op, ValueRange({allocResult.resultMemRef, allocState.resultMemRef}));
```
flaub, line 574: drop the `ValueRange(`

LLITCHEV: Dropped.

flaub:
LGTM! Thanks @jbruestle  for explaining it to me

PR #1063, opened 2020-05-14
Implement min/max aggregations.

LLITCHEV:


flaub:
nice

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -414,6 +415,16 @@ static Value createInit(OpBuilder &builder, Location loc, Type type,
...
       return builder.create<mlir::ConstantFloatOp>(loc, value, floatType);
     }
+    case AggregationKind::min: {
+      auto value = convertFloatUsingType(
+          llvm::APFloat(std::numeric_limits<float>::max()), floatType);
+      return builder.create<mlir::ConstantFloatOp>(loc, value, floatType);
+    }
+    case AggregationKind::max: {
+      auto value = convertFloatUsingType(
+          llvm::APFloat(std::numeric_limits<float>::min()), floatType);
```
tzerrell, line 425: This should be the most negative value of a float, i.e. `-std::numeric_limits<float>::max()`. See https://stackoverflow.com/questions/3529394/obtain-minimum-negative-float-value-in-c and https://en.cppreference.com/w/cpp/types/climits

flaub: Oops, good catch.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -414,6 +415,16 @@ static Value createInit(OpBuilder &builder, Location loc, Type type,
       auto value = convertFloatUsingType(llvm::APFloat(1.0), floatType);
       return builder.create<mlir::ConstantFloatOp>(loc, value, floatType);
     }
+    case AggregationKind::min: {
+      auto value = convertFloatUsingType(
+          llvm::APFloat(std::numeric_limits<float>::max()), floatType);
```
flaub, line 420: Also, we probably want to use max of `double` rather than `float`. Actually, let's see if there's an LLVM way to create a `APFloat` max and min.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
jbruestle: Actually, we might want to use +- inf....

jbruestle: That's nice because then the type (double/float) doesn't matter, since inf is inf regardless of precisison.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
tzerrell: There is an `APFloat::getInf`, but it requires you to specify the float semantics; I went with just constructing from a double in PR #1064 because that seemed clearer, but perhaps I'm wrong.

PR #1064, opened 2020-05-14
Fix max contractions of negatives

tzerrell:
Fixes the smallest positive vs most negative problem identified in https://github.com/plaidml/plaidml/pull/1063 by switching to +/-infinity for the initialization of min and max contractions respectively. Updates the test for max pool to find the previous error.

### DIFF
--- a/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
+++ b/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
@@ -417,12 +417,12 @@ static Value createInit(OpBuilder &builder, Location loc, Type type,
     }
     case AggregationKind::min: {
       auto value = convertFloatUsingType(
-          llvm::APFloat(std::numeric_limits<float>::max()), floatType);
+          llvm::APFloat(std::numeric_limits<double>::infinity()), floatType);
       return builder.create<mlir::ConstantFloatOp>(loc, value, floatType);
     }
     case AggregationKind::max: {
       auto value = convertFloatUsingType(
-          llvm::APFloat(std::numeric_limits<float>::min()), floatType);
+          llvm::APFloat(-std::numeric_limits<double>::infinity()), floatType);
       return builder.create<mlir::ConstantFloatOp>(loc, value, floatType);
     }
     default:
--- a/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
+++ b/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
@@ -417,12 +417,12 @@ static Value createInit(OpBuilder &builder, Location loc, Type type,
     }
     case AggregationKind::min: {
       auto value = convertFloatUsingType(
-          llvm::APFloat(std::numeric_limits<double>::infinity()), floatType);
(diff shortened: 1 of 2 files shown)


pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -417,12 +417,12 @@ static Value createInit(OpBuilder &builder, Location loc, Type type,
     }
     case AggregationKind::min: {
       auto value = convertFloatUsingType(
-          llvm::APFloat(std::numeric_limits<float>::max()), floatType);
+          llvm::APFloat(std::numeric_limits<double>::infinity()), floatType);
```
flaub, line 420: What if we used `llvm::APFloat::getInf()`? I'm not exactly sure what to pass to the `Sem` argument, but it seems like generating an `APFloat` directly might work for all data types better than this. It's not clear to me what happens when `floatType` is 16, 32, 64, or even larger bitwidths.

tzerrell: I switched to `APFloat` with `APFloat::IEEEdouble()` semantics; does that seem better than the numeric_limits approach?

(Looking in `llvm/unittests/ADT/APFloatTest.cpp` suggests IEEEsingle() and IEEEquad() are also options for semantics, although those don't seem better. Presumably other variants of semantics exist as well.)

tzerrell: I suppose we could also add a switch for semantics based on the type we get.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -417,12 +417,12 @@ static Value createInit(OpBuilder &builder, Location loc, Type type,
     }
     case AggregationKind::min: {
       auto value = convertFloatUsingType(
-          llvm::APFloat(std::numeric_limits<double>::infinity()), floatType);
+          llvm::APFloat::getInf(llvm::APFloat::IEEEdouble(), false), floatType);
```
flaub, line 420: Ah, now try:
```
auto value = llvm::APFloat::getInf(type.getFloatSemantics(), false);
```

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
flaub: Ya, I think the `type` has the target float semantics that we want (no need to 'convert')

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
tzerrell: Doesn't immediately work, but I'll poke around (I get `error: no member named 'getFloatSemantics' in 'mlir::Type'`)

flaub: Sorry, I meant `floatType.getFloatSemantics()` (it's on the `FloatType`, but not on `Type`)

tzerrell: That did it. Thanks for figuring that out!

PR #1065, opened 2020-05-15
Constant API

dgkutnic:


flaub:
Looks like debug builds are failing (and Windows), for some reason.

dgkutnic:
@flaub I think I see the issue - I created a program in my `ConstAdd` test, but didn't add any `CHECK` labels to it -- thus, the next test is comparing against the program created in ConstAdd. I'll fix that.

tzerrell:
> @flaub I think I see the issue - I created a program in my `ConstAdd` test, but didn't add any `CHECK` labels to it -- thus, the next test is comparing against the program created in ConstAdd. I'll fix that.

I'm surprised that the CHECK_LABEL in  the subsequent `CppEdsl.Dot` test isn't preventing such problems.

plaidml/edsl/__init__.py
```diff
@@ -816,6 +816,26 @@ def TensorIndexes(count):
     """Creates multiple ``TensorIndex`` objects based on ``count``."""
     return [TensorIndex() for i in range(count)]
 
+def Constant(dtype_or_shape, buffer, dims=[], name=''):
+    """Creates a tensor with constant values.
+
+    Args:
+        dtype_or_shape (DType | LogicalShape): A data type or a shape can be
+            specified. If a shape is specified, the `dims` parameter is ignored.
+        dims (list, optional): Specifies the dimensions of the ``Constant``.
```
tzerrell, line 825: Documentation for the `buffer` parameter is missing

dgkutnic: Good catch, thanks!

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -55,6 +55,20 @@ Tensor Softmax(const Tensor& X) {
   return E / N;
 }
 
+Tensor ConstAdd() {
+  std::string DEFAULT_DEVICE = "llvm_cpu.0";
+  std::vector<int64_t> shape = {4};
+  std::vector<int> a = {4,3,2,1};
```
flaub, line 61: Doesn't look like `clang-format` has been used.

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -55,6 +55,20 @@ Tensor Softmax(const Tensor& X) {
   return E / N;
 }
 
+Tensor ConstAdd() {
+  std::string DEFAULT_DEVICE = "llvm_cpu.0";
```
flaub, line 59: We should drop this, why is it needed?

dgkutnic: I see these two constructors for `Buffer`:
* `Buffer(const std::string& device, const TensorShape& shape)`
* `Buffer(plaidml_buffer* ptr, const TensorShape& shape)`

The `DEFAULT_DEVICE` string exists because I was using the first constructor, should I use the second constructor instead?

dgkutnic: This is used to choose which device to create the `Buffer` on. I've modified the code so that it grabs the current device name instead of using a hardcoded device name.

dgkutnic:
I thought that adding `CHECK` statements to `CppEdsl.ConstAdd` would eliminate the debug build errors, but the debug builds are still failing on `CppEdsl.Dot` (right after ConstAdd), and I'm not sure why:

```
/var/lib/buildkite-agent/bazel/lab-13-1/plaidml-plaidml/plaidml-v1/execroot/com_intel_plaidml/bazel-out/linux_x86_64-dbg/bin/plaidml/edsl/tests/edsl_test.cc.test.runfiles/com_intel_plaidml/plaidml/edsl/tests/edsl_test.cc:337:18: error: CHECK-LABEL: expected string not found in input
 // CHECK-LABEL: CppEdsl.Dot
                 ^
<stdin>:125:1: note: scanning from here
module {
```

Could the `IVLOG` statement that I have in `ConstAdd` be throwing things off? FWIW, I grabbed that statement directly from another test.

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -55,6 +55,20 @@ Tensor Softmax(const Tensor& X) {
   return E / N;
 }
 
+Tensor ConstAdd() {
+  std::vector<int64_t> shape = {4};
+  std::vector<int> a = {4, 3, 2, 1};
+  std::vector<int> b = {1, 2, 3, 4};
+  std::string current_device = plaidml::Settings::get("PLAIDML_DEVICE");
```
flaub, line 62: Use const ref so we don't make a copy of the string. 
```
const auto& curDevice = 
```

nit: use `camelBack` instead of `snake_case`.

flaub:
My guess is that there's an assertion firing or something. 

Try:
```
bazelisk run run //plaidml/edsl/tests:cc_test -c dbg
```

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -288,6 +302,29 @@ TEST_F(CppEdsl, Add) {
...
+  auto program = makeProgram("const_add", {O});
+
+  // clang-format off
+  // CHECK-LABEL: CppEdsl.ConstAdd
+  // CHECK: func @const_add
+  // CHECK: %{{.*}} = "eltwise.add"(%{{.*}}, %{{.*}}) : (tensor<4xsi32>, tensor<4xsi32>) -> tensor<4xsi32>
+  // CHECK: return %{{.*}} : tensor<4xsi32>
+  // clang-format on
+
+  exec::Binder binder(program);
```
flaub, line 316: You can use:

```
  std::vector<int32_t> expected = {5, 5, 5, 5};
  checkProgram(program, {{}, {{O, expected}});
```

flaub:
When I tried this, I got:

```
Invalid plaidml_datatype
UNREACHABLE executed at plaidml/core/ffi.cc:106!
```

flaub:
Well that ended up being tricky, so I pushed the fixes :)

dgkutnic:
Thanks @flaub for making those changes! The code is much cleaner now 😊

PR #1071, opened 2020-05-17
Fix bug in tile transform

shiyao00:


pmlc/dialect/pxa/transforms/tile.cc
```diff
@@ -31,11 +32,23 @@ void performTiling(AffineParallelOp op, llvm::ArrayRef<int64_t> tileSizes) {
...
-  innerLoopOps.splice(std::prev(innerLoopOps.end()), outerLoopOps,
-                      std::next(outerLoopOps.begin(), 1), outerLoopOps.end());
+  for (auto& origOp : outerLoopOps) {
+    if (&origOp == inner.getOperation()) {
+      continue;
+    }
+    Operation* newOp = origOp.clone(mapper);
+    innerLoopOps.push_back(newOp);
+  }
+  outerLoopOps.erase(std::next(outerLoopOps.begin(), 1), outerLoopOps.end());
```
flaub, line 51: I think we can use `Region::cloneInto` here.

shiyao00: I tried Region::clone. It appends the src blocks after the dest region blocks. In our case, the entry block, bb0, in inner loop still exists. We have to remove bb0 and reset bb1's arguments and adjust cfg. It's not that convenient.

pmlc/dialect/pxa/transforms/tile.cc
```diff
@@ -31,11 +32,23 @@ void performTiling(AffineParallelOp op, llvm::ArrayRef<int64_t> tileSizes) {
...
+  // Clone new instructions into the interior
   auto &innerLoopOps = inner.getBody()->getOperations();
   auto &outerLoopOps = outerBody->getOperations();
-  innerLoopOps.splice(std::prev(innerLoopOps.end()), outerLoopOps,
-                      std::next(outerLoopOps.begin(), 1), outerLoopOps.end());
+  for (auto& origOp : outerLoopOps) {
+    if (&origOp == inner.getOperation()) {
+      continue;
+    }
+    Operation* newOp = origOp.clone(mapper);
```
jbruestle, line 48: This does a deep copy, can we instead 1) splice the interior in 2) walk the uses of the outer args and replace them with the inners args (except the uses in the actual inner PF begin/end).  Or alternately 1) Make the inner AP with fake ranges (to get values for the inner args constructed), 2) splice the interior in 3) For each dim, call outerArg.replaceAllUsesWith(innerArg) 4) Update the inner AP to have ranges defined by the outer args.  This is a pattern I have seen before in MLIR code.  Either of these makes the big-O for the transform be the number of uses of the index rather than the entire interior of the loop.  

It's probably not critical for performance yet, and the code currently is readable and correct, so if you would rather me just  OK this PR, we can always get to this if/when we run into perf issues, if so just let me know.

pmlc/dialect/pxa/tests/autotile.mlir
```diff
@@ -12,5 +11,16 @@ func @dot(%arg0: memref<100x100xf32>, %arg1: memref<100x100xf32>) -> memref<100x
   }
   return %out : memref<100x100xf32>
 }
-// CHECK: affine.parallel
-// CHECK: affine.parallel
+
+// CHECK-LABEL: func @dot
+// CHECK-SAME: (%arg0: memref<100x100xf32>, %arg1: memref<100x100xf32>) -> memref<100x100xf32>
+// CHECK: %1 = affine.parallel (%arg2, %arg3, %arg4) = (0, 0, 0) to (100, 100, 100) step (10, 10, 10) {
```
flaub, line 17: Use `%[[FOO:.*]]` instead of `%argX` for captures, and then `%[[FOO]]` for uses.

See: https://mlir.llvm.org/getting_started/TestingGuide/#filecheck-best-practices

pmlc/dialect/pxa/transforms/tile.cc
```diff
@@ -36,6 +37,12 @@ void performTiling(AffineParallelOp op, llvm::ArrayRef<int64_t> tileSizes) {
   auto &outerLoopOps = outerBody->getOperations();
   innerLoopOps.splice(std::prev(innerLoopOps.end()), outerLoopOps,
                       std::next(outerLoopOps.begin(), 1), outerLoopOps.end());
+  auto innerIdxs = inner.getBody()->getArguments();
+  for (auto& innerOp : innerLoopOps) {
+    for (unsigned i = 0; i < outerIdxs.size(); ++i) {
+      innerOp.replaceUsesOfWith(outerIdxs[i], innerIdxs[i]);
```
jbruestle, line 43: Hmm, I think this does not do the correct thing for any sort of deep loop structure (i.e. tiling the outer loops of an already 2 level affine.parallel).  I checked the implementation of op.replaceUsesWith and it doesn't walk into the regions inside of an Op.  I think you need to walk all the uses of the outerIdx and replace all of them with innerIdx, except the uses in the inner affine.for's definition.

pmlc/dialect/pxa/tests/autotile.mlir
```diff
@@ -12,5 +11,16 @@ func @dot(%arg0: memref<100x100xf32>, %arg1: memref<100x100xf32>) -> memref<100x
   }
   return %out : memref<100x100xf32>
 }
-// CHECK: affine.parallel
-// CHECK: affine.parallel
+
+// CHECK-LABEL: func @dot
+// CHECK-SAME: (%arg0: memref<100x100xf32>, %arg1: memref<100x100xf32>) -> memref<100x100xf32>
```
flaub, line 16: %arg0 should also be captured, or just use `%{{.*}}` if you don't need a capture

pmlc/dialect/pxa/tests/autotile.mlir
```diff
@@ -12,5 +11,16 @@ func @dot(%arg0: memref<100x100xf32>, %arg1: memref<100x100xf32>) -> memref<100x
...
+// CHECK: %1 = affine.parallel (%[[arg2:.*]], %[[arg3:.*]], %[[arg4:.*]]) = (0, 0, 0) to (100, 100, 100) step (10, 10, 10) {
```
flaub, line 17: We should capture the results so that `%1` isn't used directly.

flaub: Another possibility is that if you don't need `%1` from anything, it's fine to just not check for it.

pmlc/dialect/pxa/transforms/tile.cc
flaub: Let's add a deeply nested loop test case to check for this case.

pmlc/dialect/pxa/transforms/tile.cc
```diff
@@ -2,12 +2,29 @@
...
 #include <iostream>
 
+#include "mlir/include/mlir/IR/BlockAndValueMapping.h"
 #include "pmlc/dialect/pxa/transforms/tile.h"
 
 namespace pmlc::dialect::pxa {
 
 using mlir::AffineParallelOp;
 
+void replaceAllUses(Operation* op, mlir::BlockAndValueMapping& mapper) {
```
jbruestle, line 12: So my whole point of not calling clone was to avoid recursion.  Basically, what I'm say is: For a given value (such as the outer index variable) you can in O(uses) time walk over all the uses of the value, and fix them in place, so then it's not depending on how many ops are in the inner body, but just how many times the index is used (which will typically be much smaller).  Also, since you are just updating the value in place, no allocation is done at all.  The entire point of MLIR's complex use-def tracking system is largely to avoid walking over ops at all.

shiyao00: ok, I'll try.

pmlc/dialect/pxa/transforms/tile.cc
```diff
@@ -36,6 +36,16 @@ void performTiling(AffineParallelOp op, llvm::ArrayRef<int64_t> tileSizes) {
...
+  // Replace old indices with new indices
+  auto innerIdxs = inner.getBody()->getArguments();
+  for (unsigned i = 0; i < outerIdxs.size(); ++i) {
+    outerIdxs[i].replaceAllUsesWith(innerIdxs[i]);
+  }
+  unsigned numIdxs = inner.lowerBoundsMap().getNumInputs();
+  for (unsigned i = 0; i < numIdxs; ++i) {
+    inner.setOperand(i, outerIdxs[i]);
+    inner.setOperand(i + numIdxs, outerIdxs[i]);
+  }
```
jbruestle, line 48: This looks great!  I'm slightly surprised the MLIR allows the inner AP to even temporarily have uses that are first defined in it's interior, but perhaps the check for recursive relations happens only during validation, and of course by that time it's been fixed.  Presuming that is the correct semantics (which if the debug build doesn't complain, I assume it is), that's a very clean solution!

pmlc/dialect/pxa/tests/autotile.mlir
```diff
@@ -1,16 +1,34 @@
...
+// CHECK:       %5 = affine.load %[[arg0]][%[[arg10]], %[[arg9]]] : memref<100x100xf32>
+// CHECK:       %6 = mulf %4, %5 : f32
+// CHECK:       %7 = pxa.reduce add %6, %[[obuf]][%[[arg8]], %[[arg9]]] : memref<100x100xf32>
+// CHECK:       affine.yield %7 : memref<100x100xf32>
+// CHECK:     }
+// CHECK:     affine.yield %3 : memref<100x100xf32>
+// CHECK:   }
+// CHECK:   affine.yield %2 : memref<100x100xf32>
+// CHECK: }
+// CHECK: return %1 : memref<100x100xf32>
```
jbruestle, line 34: So in general, we should never use bare %x, since nothing requires the printer to use a specific naming convention. 
Either capture and use is (and when capturing and using, no reason not to use meaningful names, since they are symbolic), or just put in a regex.  Or better yet, ignore some of the code that's not relevant to what you are testing.  You could probably remove mulf, yield, and %1, %2, etc from the test.  Check will just skip over anything it needs to until it hits what it wants, so being more complete than needed to test the thing you care about (in this case, index mechanics) doesn't help anything and actually makes the tests more brittle.

jbruestle: For example, you could just write CHECK: affine.load %[[arg1]][%[[arg8]], %[[arg10]]], since you don't care about the return or the memref type probably.

PR #1073, opened 2020-05-18
Fix system library lookup for MacOS

LLITCHEV:
The compiler adds an underscore, but on MacOS system symbols have no one. The Stripe code did similar thing.
Thanks Mars!

### DIFF
--- a/plaidml/edsl/tests/edsl_test.cc
+++ b/plaidml/edsl/tests/edsl_test.cc
@@ -1090,6 +1090,30 @@ TEST_F(CppEdsl, PrngResultNotNegative) {
   }
 }
 
+TEST_F(CppEdsl, Cos) {
+  auto S = Placeholder(DType::FLOAT32, {3, 3});
+  TensorDim I, J;
+  TensorIndex i("i"), j("j");
+  S.bind_dims(I, J);
+  auto O = cos(S);
+  auto program = makeProgram("cos", {O});
+  std::cout << program << std::endl;
+  // clang-format off
+  // CHECK-LABEL: CppEdsl.Cos
+  // clang-format on
+
+  std::vector<float> input = {
+      5.0, 6.0, 7.0,  //
+      4.0, 5.0, 6.0,  //
+      7.0, 8.0, 9.0,  //
+  };
+
+  auto binder = exec::Binder(program);
+  auto executable = binder.compile();
+  binder.input(S).copy_from(input.data());
+  executable->run();
+}
+
 TEST_F(CppEdsl, ConvI8) {
   auto I = Placeholder(DType::INT8, {1, 224, 224, 3});
   auto K = Placeholder(DType::INT8, {3, 3, 1, 32});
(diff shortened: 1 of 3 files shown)


pmlc/compiler/executable.cc
```diff
@@ -164,7 +164,19 @@ static void *tryResolveSymbol(StringRef symbol) {
     if (auto ptr = resolveSymbol(symbol.drop_front()))
       return ptr;
   }
-  return llvm::sys::DynamicLibrary::SearchForAddressOfSymbol(symbol.str());
+
```
flaub, line 167: Isn't this already handled in the few lines above?

flaub:
Do you have `.mlir` test case where this currently fails?

flaub:
Perhaps this was needed to find symbols for `sin`? If so, I'd like to explicitly register symbols in `pmlc/rt/runtime.cc` instead of trying to do an automatic lookup.

pmlc/compiler/executable.cc
LLITCHEV: The code just above handles resolving symbols from our library. There is no code to handle system library symbols like llvm::sys::DynamicLibrary::SearchForAddressOfSymbol("_cos"); for example. I merely added code to handle this case (as it was in the Stripe version).

LLITCHEV:
Added an edsl test.

LLITCHEV:
I tried registering a function, but it was always making a call - the cos function is intrinsic and the compiler replaces them in the LLVM phases with a CPU instruction. If we register a function, the function is always called (the compiler doesn't know how to replace it - that is what it did with all the attempts I tried to push it through) and thus creating a call overhead. Calling the intrinsic directly removes the need to have the call - better for perf. Maybe there is a way to register an intrinsic function which I just didn't see?

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -1090,6 +1090,30 @@ TEST_F(CppEdsl, PrngResultNotNegative) {
...
+  std::vector<float> input = {
+      5.0, 6.0, 7.0,  //
+      4.0, 5.0, 6.0,  //
+      7.0, 8.0, 9.0,  //
+  };
+
+  auto binder = exec::Binder(program);
+  auto executable = binder.compile();
+  binder.input(S).copy_from(input.data());
+  executable->run();
```
flaub, line 1114: Use `checkProgram()`

pmlc/testing/BUILD
```diff
@@ -1,3 +1,5 @@
+# Copyright 2019 Intel Corporation.
```
flaub, line 1: 2020

flaub:
I think the long term resolution to all of this is to use the `llvm-project/rt` library directly. That's where these intrinsics are defined. But for now, we can use this approach.

plaidml/edsl/tests/edsl_test.cc
LLITCHEV: Using checkProgram.

pmlc/testing/BUILD
LLITCHEV: Fixed.

PR #1074, opened 2020-05-19
i1 Storage To i32 pass added

abialas1:
Since there is no i1 storage supported in spirv dialect, this pass is to add pass to handle such cases including conversions from\to i8 for load and op stores in standard dialect


pmlc/dialect/stdx/tests/i1StorageToi8.mlir
```diff
@@ -0,0 +1,30 @@
...
+    loop.for %arg2 = %c0 to %c10 step %c1 {
+      %c0_3 = constant 0 : index
+      %c20 = constant 20 : index
+      %c1_4 = constant 1 : index
+      loop.for %arg3 = %c0_3 to %c20 step %c1_4 {
+        %1 = load %arg0[%arg2, %arg3] : memref<10x20xf32>
+        %2 = cmpf "olt", %1, %cst : f32
+        store %2, %0[%arg2, %arg3] : memref<10x20xi1>
+		%3 = load %0[%arg2, %arg3] : memref<10x20xi1>
+		store %3, %0[%arg2, %arg3] : memref<10x20xi1>
```
flaub, line 19: Missing indentation

pmlc/dialect/stdx/tests/i1StorageToi8.mlir
```diff
@@ -0,0 +1,30 @@
+// RUN: pmlc-opt -stdx-i1-storage-to-i8 %s | FileCheck %s
+
+module {
```
flaub, line 3: nit: We usually don't need the `module` in a `.mlir` test (but it doesn't hurt either).

pmlc/dialect/stdx/tests/i1StorageToi8.mlir
```diff
@@ -0,0 +1,30 @@
...
+  func @eltwise_add(%arg0: memref<10x20xf32>, %arg1: memref<10x20xf32>) {
+    %cst = constant 0.000000e+00 : f32
+    %0 = alloc() : memref<10x20xi1>
+    %c0 = constant 0 : index
+    %c10 = constant 10 : index
+    %c1 = constant 1 : index
+    loop.for %arg2 = %c0 to %c10 step %c1 {
+      %c0_3 = constant 0 : index
```
flaub, line 11: You can reuse `%c0`, no need to create a new constant with the same value. It looks like maybe this code came from the output of something else? If so, try grabbing this sample after the CSE and Canonicalize passes were used.

pmlc/dialect/stdx/transforms/BUILD
```diff
@@ -23,6 +23,7 @@ plaidml_cc_library(
     name = "transforms",
     srcs = [
         "boundscheck.cc",
+        "i1StorageToi8.cc",
```
flaub, line 26: Currently, all our filenames are `snake_case`, can you name this `i1_to_i8.cc` for consistency?

pmlc/dialect/stdx/transforms/i1StorageToi8.cc
```diff
@@ -0,0 +1,98 @@
...
+#include "mlir/Pass/Pass.h"
+
+using namespace mlir; // NOLINT[build/namespaces]
+
+namespace pmlc::dialect::stdx {
+
+namespace {
+
+/// Changes loadOp from i1 memref to loadOp i8 followed by trunci i8->i1
+class LoadOpi1Toi8 final : public OpRewritePattern<LoadOp> {
```
flaub, line 15: Can we rename this to `LoadOpI1ToI8`?

pmlc/dialect/stdx/transforms/i1StorageToi8.cc
```diff
@@ -0,0 +1,98 @@
...
+    return success();
+  }
+
+  auto destType = rewriter.getIntegerType(8);
+  auto loc = loadOp.getLoc();
+
+  loadOp.getMemRef().setType(
+      MemRefType::get(loadOp.getMemRefType().getShape(), destType));
+
+  auto loadOp_new =
```
flaub, line 47: Rename to `newLoadOp`

pmlc/dialect/stdx/transforms/i1StorageToi8.cc
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/// Hook for adding patterns.
+void populatei1StorageToi8(MLIRContext *context,
+                           OwningRewritePatternList &patterns) {
+  patterns.insert<LoadOpi1Toi8, StoreOpi1Toi8>(context);
+}
+
+struct i1StorageToi8Pass
+    : public mlir::PassWrapper<i1StorageToi8Pass, mlir::FunctionPass> {
```
flaub, line 84: Use `I1StorageToI8Base< I1StorageToI8Pass>` as the base type. `I1StorageToI8Base` is generated by tablegen from the `passes.td`.

pmlc/dialect/stdx/transforms/passes.td
```diff
@@ -8,4 +8,9 @@ def BoundsCheck : FunctionPass<"stdx-check-bounds"> {
   let constructor = "pmlc::dialect::stdx::createBoundsCheckPass()";
 }
 
+def i8StorageToi8 : FunctionPass<"stdx-i1-storage-to-i8"> {
```
flaub, line 11: Rename to `I1StorageToI8`

pmlc/dialect/stdx/transforms/i1StorageToi8.cc
```diff
@@ -0,0 +1,98 @@
...
+  rewriter.replaceOp(loadOp, {truncOp});
+  return success();
+}
+
+/// Changes storeOp to i1 memref to zexti i1->i8 followed by storeOp to i8
+LogicalResult StoreOpi1Toi8::matchAndRewrite(StoreOp storeOp,
+                                             PatternRewriter &rewriter) const {
+  auto elementType = storeOp.value().getType();
+  if (!elementType.isInteger(1)) {
+    return success();
```
jbruestle, line 61: This should return 'failure' since you "failed" to match.  If you return 'success' the patter rewriter presumes you may have made changes and will run all of the patterns again on the result.  It has 10 time limit or something like that for continual rewrites of the same pattern, which is why this isn't causing an infinite loop.

pmlc/dialect/stdx/tests/i1StorageToi8.mlir
```diff
@@ -0,0 +1,30 @@
...
+        %1 = load %arg0[%arg2, %arg3] : memref<10x20xf32>
+        %2 = cmpf "olt", %1, %cst : f32
+        store %2, %0[%arg2, %arg3] : memref<10x20xi1>
+		%3 = load %0[%arg2, %arg3] : memref<10x20xi1>
+		store %3, %0[%arg2, %arg3] : memref<10x20xi1>
+      }
+    }
+    return
+  }
+  // CHECK: %{{.*}} = alloc() : memref<10x20xi8>
```
jbruestle, line 24: So it's a matter of style and both are allowable, but I tend to prefer putting the check lines 'inline' with the code, so for example, you would have the alloc, and then the CHECK of the modified alloc, and then the store and the check of the modified store, etc.  Filecheck (which lit runs) doesn't actually care about the placement of the check lines, just the order, and I find the inlined version easier to read.

pmlc/dialect/stdx/tests/i1StorageToi8.mlir
abialas1: I will remove the module part if it is redundant. Seems like there is a bit inconsistency with using this, other mlir tests from this dialect has it, also pxa have those, and for example tile does not.

pmlc/dialect/stdx/tests/i1StorageToi8.mlir
abialas1: Yes it was generated without CSE and Canonicalize, will update the code with usage of these passes.

pmlc/dialect/stdx/tests/i1_to_i8.mlir
```diff
@@ -0,0 +1,31 @@
...
+    %0 = alloc() : memref<10x20xi1>
+    // CHECK: %{{.*}} = alloc() : memref<10x20xi8>
+    %c0 = constant 0 : index
+    %c10 = constant 10 : index
+    %c1 = constant 1 : index
+    loop.for %arg2 = %c0 to %c10 step %c1 {
+      %c0_3 = constant 0 : index
+      %c20 = constant 20 : index
+      %c1_4 = constant 1 : index
+      loop.for %arg3 = %c0_3 to %c20 step %c1_4 {
```
flaub, line 15: Use `%c0` and drop `%c0_3`. There seems to be redundant constants.

flaub: Do we actually need loops for this test? I would shrink this test down to the bare essentials.

pmlc/dialect/stdx/tests/i1_to_i8.mlir
```diff
@@ -0,0 +1,31 @@
+// RUN: pmlc-opt -stdx-i1-storage-to-i8 %s | FileCheck %s
+
+module {
+  func @i1_to_i8_test(%arg0: memref<10x20xf32>, %arg1: memref<10x20xf32>) {
+    %cst = constant 0.000000e+00 : f32
+    %0 = alloc() : memref<10x20xi1>
+    // CHECK: %{{.*}} = alloc() : memref<10x20xi8>
```
flaub, line 7: If you don't to check or retain the return value, you can just leave it off:

```
// CHECK: alloc() : memref<10x20xi8>
```

pmlc/dialect/stdx/tests/i1_to_i8.mlir
```diff
@@ -1,31 +1,23 @@
...
-        // CHECK: %{{.*}} = zexti %{{.*}} : i1 to i8
-        store %3, %0[%arg2, %arg3] : memref<10x20xi1>
-        // CHECK: store %{{.*}}, %{{.*}}[%arg2, %arg3] : memref<10x20xi8>
-      }
-    }
-    return
-  }
+func @simpleStore(%i: index, %j: index) {
+  %0 = alloc() : memref<20x10xi1>
+  // alloc() : memref<20x10xi8>
```
flaub, line 5: I think you missed the `CHECK: `

pmlc/dialect/stdx/tests/i1_to_i8.mlir
```diff
@@ -1,31 +1,23 @@
...
+  store %c0, %0[%i, %j] : memref<20x10xi1>
+  // CHECK: store %1, %0[%arg0, %arg1] : memref<20x10xi8>
+  return
+}
+
+func @simpleLoadStore(%i: index, %j: index) {
+  %0 = alloc() : memref<20x10xi1>
+  // alloc() : memref<20x10xi8>
+  %1 = load %0[%i, %j] : memref<20x10xi1>
+  // CHECK: load %0[%arg0, %arg1] : memref<20x10xi8>
```
flaub, line 17: We don't want to use `%arg0` directly in `CHECK` lines, because there's no guarantee that the asm printer will continue to use this format for arguments. Basically, all MLIR variables should either be ignored with `%{{.*}}` or they should be captured and used with `%[[FOO:.*]]` and `[[FOO]]` respectively.

abialas1:
Please do not merge this one yet. Seems like this solution is not working in spirv dialect (not allowed trunci and zexti on i1 types). I will change this pass a bit to be working with spv

abialas1:
I believe all comments from review were addressed, can we have this pull request merged?

flaub:
LGTM, @jbruestle ?

pmlc/dialect/stdx/transforms/i1_to_i32.cc
```diff
@@ -0,0 +1,113 @@
...
+#include "mlir/Pass/Pass.h"
+#include "pmlc/dialect/stdx/transforms/pass_detail.h"
+
+using namespace mlir; // NOLINT[build/namespaces]
+
+namespace pmlc::dialect::stdx {
+
+namespace {
+
+/// Changes loadOp from i1 memref to loadOp i32 followed by creting constant
```
flaub, line 15: typo: creating

pmlc/dialect/stdx/transforms/i1_to_i32.cc
```diff
@@ -0,0 +1,113 @@
...
+class StoreOpI1ToI32 final : public OpRewritePattern<StoreOp> {
+public:
+  using OpRewritePattern<StoreOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(StoreOp storeOp,
+                                PatternRewriter &rewriter) const override;
+};
+} // namespace
+
+/// Changes loadOp from i1 memref to loadOp i32 followed by creting constant
```
flaub, line 37: typo: creating

pmlc/dialect/stdx/transforms/i1_to_i32.cc
```diff
@@ -0,0 +1,113 @@
...
+  auto elementType = storeOp.value().getType();
+  if (!elementType.isInteger(1)) {
+    return failure();
+  }
+
+  auto destType = rewriter.getIntegerType(32);
+  auto loc = storeOp.getLoc();
+
+  auto const0 =
+      rewriter.create<ConstantOp>(loc, rewriter.getIntegerAttr(destType, 0));
```
flaub, line 78: We can use `ConstantIntOp`

abialas1: thanks, switched to use ConstantIntOp for all integer constantOp in file

pmlc/dialect/stdx/transforms/i1_to_i32.cc
```diff
@@ -0,0 +1,110 @@
...
+  auto destType = rewriter.getIntegerType(32);
+  auto loc = storeOp.getLoc();
+
+  auto const0 = rewriter.create<ConstantIntOp>(loc, 0, destType);
+  auto const1 = rewriter.create<ConstantIntOp>(loc, 1, destType);
+  auto selOp = rewriter.create<SelectOp>(
+      loc, storeOp.value(), const1.getResult(), const0.getResult());
+
+  storeOp.getMemRef().setType(
+      MemRefType::get(storeOp.getMemRefType().getShape(), destType));
```
flaub, line 82: Why do we use `setType` when we then use `replaceOpWithNewOp`? It seems like we can drop the `setType` unless I'm missing something.

pmlc/dialect/stdx/transforms/i1_to_i32.cc
```diff
@@ -0,0 +1,110 @@
...
+  auto elementType = loadOp.result().getType();
+  if (!elementType.isInteger(1)) {
+    return failure();
+  }
+
+  auto destType = rewriter.getIntegerType(32);
+  auto loc = loadOp.getLoc();
+
+  loadOp.getMemRef().setType(
+      MemRefType::get(loadOp.getMemRefType().getShape(), destType));
```
flaub, line 51: Same goes here, it seems like if we are using `replaceOp`, then we don't need to use `setType`

pmlc/dialect/stdx/transforms/i1_to_i32.cc
abialas1: This memref is used when creating the new storeOp, and its type needs to be changed. I am just reusing it instead of creating a new one with proper type.

PR #1075, opened 2020-05-20
Add folder for ReshapeOp

dgkutnic:


flaub:
This is great, but let's write a `.mlir` test that proves the folding is working. Similar to `constant-fold.mlir`

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -877,6 +877,17 @@ TEST_F(CppEdsl, Reciprocal) {
   checkProgram(program, {{A, input}}, {{R, expected}});
 }
 
+TEST_F(CppEdsl, ReshapeFold) {
+  auto A = Placeholder(DType::INT32, {3, 3});
+  auto R = reshape(A, {3, 3});
+  auto program = makeProgram("reshape_fold", {R});
```
flaub, line 883: `program` is unused. Also, can you call `checkProgram()` so that it actually runs?

pmlc/dialect/tile/ir/ops.cc
```diff
@@ -681,6 +681,15 @@ Type PrngOp::getResultType(ArrayRef<Value> operands) {
 
 // ---- ReshapeOp ----
 
+OpFoldResult ReshapeOp::fold(ArrayRef<Attribute> operands) {
+  IVLOG(5, "ReshapeOp::fold");
+  // reshape(x, x.shape) -> x
+  if (getOperand(0).getType() == getType()) {
```
flaub, line 687: Might be nicer to use `tensor()` instead of `getOperand(0)` here

dgkutnic:
Hi @flaub,

I tried sending you a message on Teams but I'm undergoing the Office 365 migration today so I can't access any messages you may have sent through any Office products. I'm pushing the changes I have so far, but I know that `checkProgram` is failing on the reshape folding test I created, even though the generated program seemingly looks correct.

No matter what values I pass in as an input, the returned output is all 0s, so it seems like the values are getting lost somewhere. The program looks like it's just returning the input, so I'm not sure where the failure could be. Here's the program for reference:

```
module {
  func @reshape_fold(%arg0: tensor<3x3xsi32>) -> tensor<3x3xsi32> {
    return %arg0 : tensor<3x3xsi32>
  }
}
```

Any ideas on what could be causing the unexpected behavior of this program?

flaub:
There's a special case for empty programs like this one, where the entire program is returning one of the inputs. In this case we need to create an `identity()` op to induce a copy of the tensor. This was handled by https://github.com/plaidml/plaidml/blob/plaidml-v1/pmlc/dialect/tile/builder.cc#L643. It seems like this new folder happens after this, so now I wonder if we need to try to deal with this issue at the end of `MakeProgram` instead of at the beginning.

flaub:
To see what's going on, turn up the logging verbosity and look at how the tile code is lowered into affine and beyond, you should be able to see that there's nothing to copy the input to the output, which is why you're always seeing 0s.

pmlc/dialect/tile/ir/ops.cc
```diff
@@ -681,6 +681,15 @@ Type PrngOp::getResultType(ArrayRef<Value> operands) {
 
 // ---- ReshapeOp ----
 
+OpFoldResult ReshapeOp::fold(ArrayRef<Attribute> operands) {
+  IVLOG(5, "ReshapeOp::fold");
+  // reshape(x, x.shape) -> x
+  if (tensor().getType() == getType()) {
+    return getOperand(0);
```
flaub, line 688: Can you return `tensor()` here?

pmlc/dialect/tile/tests/reshape-fold.mlir
```diff
@@ -0,0 +1,21 @@
+// RUN: pmlc-opt -canonicalize %s | FileCheck %s
+
+// CHECK-LABEL: @basic
+func @basic(%arg0: tensor<4x4xsi32>) -> tensor<4x4xsi32> {
+  %c4 = tile.constant 4
+  %0 = "tile.reshape"(%arg0, %c4, %c4) : (tensor<4x4xsi32>, index, index) -> tensor<4x4xsi32>
+  return %0 : tensor<4x4xsi32>
+  // CHECK-NEXT: return %arg0
```
flaub, line 8: CHECK lines should not use `%arg0` directly, instead we need to capture the argument up above, so use this:

```
// CHECK-LABEL: @basic
// CHECK-SAME: %[[arg0:.*]]: tensor<4x4xsi32>
// CHECK-NEXT: return %[[arg0]]
```

pmlc/dialect/tile/builder.cc
```diff
@@ -769,6 +769,10 @@ TileBuilder::MakeProgram(StringRef name, const ProgramMutations &mutations,
   for (unsigned i = 0; i < returnOp.getNumOperands(); i++) {
     auto userValue = outputs[i];
     auto finalValue = returnOp.getOperand(i);
+    if (!finalValue.getDefiningOp()) {
+      IVLOG(2, "reached condition where return operand has no defining op");
+      userValue = MakePrimitiveOp("ident", {outputs[i]});
+    }
```
flaub, line 775: This is a great solution if it works. Could we remove the special handling at the top of `MakeProgram` now that we have this?

dgkutnic, line 775: Thanks for the explanation!

I've been looking into ways to add an `ident` op around any outputs that need copies of the input tensor once optimization passes are complete. One of the ways I've been exploring is to check (at the end of `MakeProgram`) whether each `ReturnOp` argument has a defining op, and wrap the argument with `ident` if that defining op is not found.

I've added an `IVLOG` to see where this condition is reached, and out of all the cc_tests, the only time it ever gets reached is during `ReshapeFold` (which is expected, because that's the only place where we are returning an input value that hasn't already got an `ident` wrapper around it). However, I know that my attempt to wrap the output with `ident` isn't working, since the `ident` statement isn't added when the condition is reached. This is a partial thing while I figure that out (or if you have any ideas/feedback, let me know)

flaub: I think there were a bunch of keras backend tests that would trip this condition. @tzerrell might recall better which tests those might be (tests that would produce empty program bodies).

flaub: Actually, we might need to say:
```
finalValue = MakePrimitiveOp("ident", {userValue});
```

dgkutnic: Thanks for the suggestion. I tried this, and it still doesn't seem to be affecting the behavior of the builder:

```
2020-05-21 17:56:14,593 VERBOSE-5 [default] ReshapeOp::fold
2020-05-21 17:56:14,594 VERBOSE-2 [default] reached condition where return operand has no defining op
2020-05-21 17:56:14,594 VERBOSE-2 [default] TileBuilder::MakeProgram>


module {
  func @reshape_fold(%arg0: tensor<3x3xsi32> {tile.name = "A"}) -> tensor<3x3xsi32> {
    return %arg0 : tensor<3x3xsi32>
  }
}
```

tzerrell: I think there might have been a few other tests that happened to hit this, but the one that is attempting to test this is `TestBackendOps.testPassthrough`.

flaub: Oh, we can't use `MakePrimitiveOp`, we need to create it directly. This is because there are actually 2 modules here. One is the 'workspace' module that the EDSL directly inserts new ops into. But by the end of `MakeProgram` we've actually cloned a subgraph based on the trace into a new module. We need to insert the `ident` into this new module.

tzerrell: Also possibly related: `testConstant`, `testUpdate`, `testSelfMult`, and `testDupOutputs`.

dgkutnic: I tried this and it segfaulted. Perhaps I'm misunderstanding what I need to do here?

```
    if (!finalValue.getDefiningOp()) {
      IVLOG(2, "reached condition where return operand has no defining op");
      auto ident = builder.create<eltwise::IdentOp>(impl->loc, userValue);
      module.push_back(ident);
    }
```

flaub: No need to do `module.push_back(ident)` (this would effectively add an op at the top-level, outside of the `FuncOp`). When you use `builder.create`, this should create the op in the location of the current insertion point. So you'll likely need to adjust the insertion point. I would actually create a new `OpBuilder` and use the constructor that takes an `Operation*` since that ctor will initialize the insertion point to be just before the specified operation. In this case, you'd want to insert this just before `return` op.

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -1018,6 +1018,15 @@ TEST_F(CppEdsl, DupOut) {
...
   auto R = Dot(Dot(A, B), C);
   auto program = makeProgram("dup_out", {R, R, R});
+  // clang-format off
+  // CHECK: func @dup_out
+  // CHECK:  %[[cst:.*]] = "eltwise.sconst"() {value = 0.000000e+00 : f64} : () -> tensor<f32>
+  // CHECK:  %{{.*}} = tile.contract add, mul, %[[cst]], %{{.*}}, %{{.*}} {idxs = ["{{.*}}", "{{.*}}", "{{.*}}"], sink = #map{{[0-9]+}}, srcs = [#map{{[0-9]+}}, #map{{[0-9]+}}]} : tensor<f32>, tensor<10x20xf32>, tensor<20x30xf32> -> tensor<10x30xf32>
+  // CHECK:  %[[out:.*]] = tile.contract add, mul, %[[cst]], %{{.*}}, %{{.*}} {idxs = ["{{.*}}", "{{.*}}", "{{.*}}"], sink = #map{{[0-9]+}}, srcs = [#map{{[0-9]+}}, #map{{[0-9]+}}]} : tensor<f32>, tensor<10x30xf32>, tensor<30x40xf32> -> tensor<10x40xf32>
+  // CHECK:  %{{.*}} = "eltwise.ident"(%[[out]]) : (tensor<10x40xf32>) -> tensor<10x40xf32>
+  // CHECL:  %{{.*}} = "eltwise.ident"(%[[out]]) : (tensor<10x40xf32>) -> tensor<10x40xf32>
+  // CHECK:  return %{{.*}}, %{{.*}}, %{{.*}} : tensor<10x40xf32>, tensor<10x40xf32>, tensor<10x40xf32>
```
flaub, line 1028: Might be good to actually check the operands of the return to ensure we're returning one copy each.

dgkutnic: there's a typo here too -- I've fixed it now.

PR #1081, opened 2020-05-20
Upgrade Convolution Fluent API Defaults

tzerrell:
Add more defaults to the convolution fluent API. This includes rank-dependent defaults. With this, it is possible to define a convolution with just the input & filter tensors if all the defaults are correct for the use case.

~Note that this is a work in progress. Outstanding work includes upgrading existing convolution builders to use these defaults, and adding logging & reviewing testing. Putting up a draft PR now b/c I know @1sanz would benefit from this API being available promptly.~


plaidml/op/op.h
```diff
@@ -5,8 +5,11 @@
 #include <string>
 #include <vector>
 
+#include "llvm/Support/FormatVariadic.h"
```
flaub, line 8: Unfortunately I think this is not a good dependency to introduce to the EDSL public API. This means that all users of the EDSL would have to pull in LLVM.

plaidml/op/op.h
```diff
@@ -5,8 +5,11 @@
...
+
 #include "plaidml/edsl/edsl.h"
 #include "plaidml/op/ffi.h"
+#include "pmlc/util/logging.h"
```
flaub, line 12: This is another dependency I'd like to avoid for the public API (it's fine for WIP testing/etc).

plaidml/op/op.h
tzerrell: Ok, I'll back it out

plaidml/op/op.h
tzerrell: For landed code, should I avoid logging altogether? I had been thinking logging all the parameters of a failed convolution would be a nice feature.

I suppose I could construct a complex exception message with stringstream if that's more appropriate than either of these headers.

tzerrell: Alternatively, I could move the default stride/dilation/data_dilation value inference into the oplib, and make logging/exception code there analogous to the code for e.g. autopadding

flaub: Yes, the more we can push into the oplib itself the better I think. That way all language bindings benefit from it.

tzerrell:
Updated to move the bulk of the work into the oplib. This unfortunately fails lint due to `readability/fn_size`. I'm not sure shortening the `convolution` function will improve readability that much, but perhaps it will, so I'm looking into how best to break it into pieces.

tzerrell:
@dgkutnic I added you as a reviewer due to your previous work on the fluent API. Most of the changes are in the op library instead though, so this may not actually be super relevant to you.

tzerrell:
@1sanz and @cnamrata15 note in particular the new default values for the convolution parameters given in plaidml/op/op.h

flaub:
Seems like a reasonable way to split up the function.

PR #1086, opened 2020-05-27
add workaround for return of reshape views; enable tests which now pass

mars1026:
Insert an elementwise copy when returning a value which is the result of a reshape, since the output needs to become a user-supplied buffer.

### DIFF
--- a/pmlc/dialect/tile/builder.cc
+++ b/pmlc/dialect/tile/builder.cc
@@ -765,9 +765,16 @@ TileBuilder::MakeProgram(StringRef name, const ProgramMutations &mutations,
   for (unsigned i = 0; i < returnOp.getNumOperands(); i++) {
     auto userValue = outputs[i];
     auto finalValue = returnOp.getOperand(i);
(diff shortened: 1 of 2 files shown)


pmlc/dialect/tile/builder.cc
```diff
@@ -765,9 +765,16 @@ TileBuilder::MakeProgram(StringRef name, const ProgramMutations &mutations,
...
+    // Determine whether an elementwise copy is required
+    bool mustCopy = false;
+    if (auto defOp = finalValue.getDefiningOp()) {
+      // Output is the result of a reshape (view)
+      mustCopy = mlir::dyn_cast<ReshapeOp>(defOp);
+    } else {
       IVLOG(2, "Reached condition: output that refers directly to input");
+      mustCopy = true;
+    }
+    if (mustCopy) {
```
flaub, line 777: I think this can be simplified to:

```
auto defOp = finalValue.getDefiningOp();
if (!defOp || isa<ReshapeOp>(defOp)) {
```

PR #1087, opened 2020-05-27
add testBatchNormAndUpdate to CI, since it passes OK

mars1026:
Compared the list of passing tests to the list in ci/plan.yml; this was the only one missing.

### DIFF
--- a/ci/plan.yml
+++ b/ci/plan.yml
@@ -219,9 +219,11 @@ SUITES:
           - TestBackendOps.testIsPlaceholder
           - TestBackendOps.testUpdate
           - TestBackendOps.testMovingAverageUpdate
+          - TestBackendOps.testBatchNormAndUpdate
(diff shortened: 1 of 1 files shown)


flaub:
Are we now at a majority of passing tests? If so, maybe it's time to flip the plan around so that we assume to run all tests and only mark the ones we want to skip/ignore (blacklist vs whitelist).

tzerrell:
> Are we now at a majority of passing tests? If so, maybe it's time to flip the plan around so that we assume to run all tests and only mark the ones we want to skip/ignore (blacklist vs whitelist).

We're definitely at majority of tests passing. Switching back to a skip methodology would also let us take advantage of the sharding CI code again (if that's something we need). It might make the outstanding failures a little harder to notice on local builds (where they fail by default right now, instead of skipping by default in a skip methodology), but conversely would make regressions easier to notice on local builds.

PR #1088, opened 2020-05-28
Switch OpenVINO bridge to newer nGraph-based API

tzerrell:


plaidml/bridge/openvino/src/plaidml_plugin/ops/relu.cpp
```diff
@@ -4,23 +4,21 @@
...
 using namespace plaidml;          // NOLINT[build/namespaces]
 using namespace InferenceEngine;  // NOLINT[build/namespaces]
 
 namespace PlaidMLPlugin {
 
-IE_SUPPRESS_DEPRECATED_START
-
 static OpRegistration reg("relu", [](const Context& ctx) {
-  auto* layer = dynamic_cast<ReLULayer*>(ctx.layer);
+  // Note that nGraph ReLU does not include an alpha parameter
```
brianretford, line 18: hopefully this would be dealt with somewhere in Model Optimizer by adding a multiplication, doesn't mean we can safely ignore but would be good to confirm

tzerrell: nGraph has a PReLU layer for cases with an alpha parameter, presumably that will be called instead.

tzerrell:
@flaub and I discussed a couple of revisions to this I will follow up with:
* Improving the keys we use for the tensor map to avert naming issues
* Updating to the newer OpenVINO tests if they are stable enough to use (they're under active development). Doing this will reduce testing code duplication, and the newer tests are built with the nGraph-based API in mind, thereby avoiding oddities like the ReLU case discussed above.

PR #1090, opened 2020-05-29
Fix ReshapeOp operand checking for scalars

dgkutnic:


plaidml/edsl/tests/edsl_test.cc
```diff
@@ -895,6 +895,29 @@ TEST_F(CppEdsl, ReshapeFold) {
...
+  int32_t input = 2;
+  // checkProgram doesn't work on scalars
+  auto binder = exec::Binder(program);
+  auto executable = binder.compile();
+  binder.input(A).copy_from(&input);
+  executable->run();
+  auto view = binder.output(R).mmap_current();
+  auto data = reinterpret_cast<const int32_t*>(view.data());
+  ASSERT_THAT(view.size(), sizeof(int32_t));
+  EXPECT_THAT(data[0], 2);
```
flaub, line 918: `checkProgram` should work, but you have to put the scalar into a buffer, like this:

```
  std::vector<int32_t> data = {2};
  checkProgram(program, {{A, data}}, {{R, data}});
```

pmlc/dialect/tile/ir/ops.cc
```diff
@@ -719,12 +719,20 @@ void ReshapeOp::getCanonicalizationPatterns(OwningRewritePatternList &results,
...
+  // In the case of scalars, return the original type
+  if (!tensorType.getRank()) {
+    return tensorType;
+  }
+  auto dims = operands.drop_front();
+  if (!dims.size()) {
+    throw std::runtime_error(
+        "ReshapeOp with Tensor of rank > 0 requires dims to be set");
+  }
   auto elementType = tensorType.getElementType();
```
flaub, line 736: I don't think this is exactly right, imagine if the user wanted to reshape a `tensor<1xf32>` into a `tensor<f32>`, that should be legal shouldn't it? What we could do though is check if both the source and target are both rank 0, then return the operand. However, we should have a test for the N-rank to 0-rank reshape when all the dims of the N-rank tensor are 1.

flaub: Likewise, we should be able to reshape a 0-rank tensor into an N-rank tensor, provided the `dims` are all 1s.

pmlc/dialect/tile/ir/ops.cc
```diff
@@ -719,12 +719,20 @@ void ReshapeOp::getCanonicalizationPatterns(OwningRewritePatternList &results,
 
 Type ReshapeOp::getResultType(ArrayRef<Value> operands) {
   IVLOG(5, "ReshapeOp::getResultType>")
-  if (operands.size() < 2) {
-    throw std::runtime_error("ReshapeOp requires at least 2 operands");
+  if (!operands.size()) {
```
flaub, line 722: Prefer `operand.empty()`

pmlc/dialect/tile/ir/ops.cc
dgkutnic: That makes sense. I wasn't thinking of the N-rank to 0-rank reshape case when I wrote this. I agree that `ComputeShape` should handle this.

dgkutnic:
Thanks for improving this and adding new testcases!

PR #1091, opened 2020-05-30
EDSL Fluent API: convolution default parameters fix

1sanz:
Issue with initialization was observed while working on onnxRT

### DIFF
--- a/plaidml/op/op.h
+++ b/plaidml/op/op.h
@@ -140,7 +140,17 @@ inline edsl::Tensor concatenate(const std::vector<edsl::Tensor>& tensors, int ax
 
 class convolution {
  public:
-  explicit convolution(edsl::Tensor I, edsl::Tensor F) : I_(I), F_(F) {}
+  explicit convolution(edsl::Tensor I, edsl::Tensor F)
+      : I_(I),
+        F_(F),
+        groups_(1),
+        autopad_mode_(AutoPadMode::SAME_UPPER),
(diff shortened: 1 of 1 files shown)


flaub:
Interesting, what was the problem you were running into? I don't see how these changes would fix anything unless you were perhaps using an older compiler which doesn't properly handle C++11 initialization (or maybe this is a C++14 feature?). I've seen older compilers that also failed here, so perhaps we should try to make the client-side EDSL code as backwards compatible as we can (basically only use C++11 features).

plaidml/op/op.h
```diff
@@ -248,15 +258,15 @@ class convolution {
   std::vector<int> filter_shape_;    // Default: empty (i.e. no filter dim check)
   int groups_ = 1;
```
flaub, line 259: Possibly missed a spot here.

tzerrell:
> Interesting, what was the problem you were running into?

Unset defaults manifesting as the enum validation code throwing `std::runtime_error("Invalid enumeration value")`. This went away if the default value was explicitly passed by calling the setter function.

plaidml/op/op.h
1sanz: fixed.

1sanz:
> LGTM except the overlooked `groups_` default nit @flaub mentioned.
> 
> (I take it this approach did ultimately resolve the issue?)

yes. At first I thought it was not helping but turned out that was a cmake cache issue.

PR #1094, opened 2020-06-01
Add openmp to the unix conda environment

earhart:


flaub:
Thanks! I think we'll need to add the lib itself to `conda/unix.BUILD` too

earhart:
Adding it to the `.yml` seems to fix the build issue I was running into; no need for another rule, as near as I can tell.

flaub:
Can you run `ldd` on the resulting binary and verify that it indeed uses the version from the conda env?

PR #1096, opened 2020-06-02
XSMM JIT split

flaub:


flaub:
Unless we want to handle dynamic shapes, it seems like we could potentially call the dispatch to JIT a function and then store the function someplace. Later, we just need to stitch this function pointer up to the runtime call. In this case, I don't see us having two runtime calls exposed at runtime. Rather, at compile time, we should call the dispatch method (during a pass) to generate the function pointer.

Another idea we've been kicking around for other purposes is to have our programs generate two top level functions instead of one right now. One function would be for initialization, it would be called once at startup. The second function would be the function we generate now, the actual program where I/O buffers are passed as parameters.

With 2 functions, doing the dispatch could happen in the initialization function. As it is right now, it looks like you're going to call the dispatch each time we execute the program.

alheinecke:
The hoisting should be part of the Canonicalizer and it's unclear to me, why it's not pulling it up.

I was also pondering to do it in the stencil pass and I thought I want to do the split in the lowering pass and the  Canonicalizer will fix it.... 

`
// *** IR Dump After Canonicalizer ***
#map0 = affine_map<(d0, d1) -> (d0, d1)>
#map1 = affine_map<() -> (0)>
#map2 = affine_map<() -> (32)>
#map3 = affine_map<() -> (8)>
#map4 = affine_map<(d0, d1)[s0] -> (d0 * 16 + s0 + d1)>
#map5 = affine_map<(d0, d1)[s0] -> (d0 * 32 + s0 + d1)>
#map6 = affine_map<() -> (16)>


module {
  func @plaidml_rt_xsmm_exec_gemm_f32(memref<*xf32>, memref<*xf32>, memref<*xf32>, memref<*xi64>)
  func @plaidml_rt_xsmm_dispatch_gemm_f32(i32, i32, i32, i32, i32, i32, memref<*xi64>)
  func @dot(%arg0: memref<16x32xf32>, %arg1: memref<8x16xf32>, %arg2: memref<8x32xf32>) {
    %cst = constant 0.000000e+00 : f32
    %c8_i32 = constant 8 : i32
    %c32_i32 = constant 32 : i32
    %c16_i32 = constant 16 : i32
    affine.for %arg3 = 0 to 8 {
      affine.for %arg4 = 0 to 32 {
        affine.store %cst, %arg2[%arg3, %arg4] : memref<8x32xf32>
      }
    }
    affine.for %arg3 = 0 to 8 step 8 {
      affine.for %arg4 = 0 to 32 step 32 {
        affine.for %arg5 = 0 to 16 step 16 {
          %0 = subview %arg1[%arg3, %arg5] [8, 16] [1, 1]  : memref<8x16xf32> to memref<8x16xf32, #map4>
          %1 = memref_cast %0 : memref<8x16xf32, #map4> to memref<*xf32>
          %2 = subview %arg0[%arg5, %arg4] [16, 32] [1, 1]  : memref<16x32xf32> to memref<16x32xf32, #map5>
          %3 = memref_cast %2 : memref<16x32xf32, #map5> to memref<*xf32>
          %4 = subview %arg2[%arg3, %arg4] [8, 32] [1, 1]  : memref<8x32xf32> to memref<8x32xf32, #map5>
          %5 = memref_cast %4 : memref<8x32xf32, #map5> to memref<*xf32>
          %6 = alloca() : memref<1xi64>
          %7 = memref_cast %6 : memref<1xi64> to memref<*xi64>
          call @plaidml_rt_xsmm_dispatch_gemm_f32(%c16_i32, %c32_i32, %c32_i32, %c8_i32, %c32_i32, %c16_i32, %7) : (i32, i32, i32, i32, i32, i32, memref<*xi64>) -> ()
          call @plaidml_rt_xsmm_exec_gemm_f32(%1, %3, %5, %7) : (memref<*xf32>, memref<*xf32>, memref<*xf32>, memref<*xi64>) -> ()
        }
      }
    }
    return
  }
}
`

%6, %7 and the call can be perfectly pulled up and the constant for the dipatch call are hoisted.

flaub:
It might be that CSE (Common Subexpression Elimination) is actually the pass you're looking for.

alheinecke:
CSE is also not doing it and there seem to be issues in MLIR and LLVM around this:
https://github.com/tensorflow/mlir/issues/65

At least the unit test work now, when comparing the full line.

I failed to create a MemRef to an int64 (too dumb and new:-)). Perhaps then it would be hoisted ?!

alheinecke:
FWIW, I have created another branch were I had to "fix-hack" some pxa transforms of function-rewrite (the problem from last night). But still it doesn't get hoisted :-( https://github.com/plaidml/plaidml/tree/aheineck-xsmm-jitsplit_verb 

`module {
  func @plaidml_rt_xsmm_exec_gemm_f32(memref<*xf32>, memref<*xf32>, memref<*xf32>, i64)
  func @plaidml_rt_xsmm_dispatch_gemm_f32(i32, i32, i32, i32, i32, i32) -> i64
  func @dot(%arg0: memref<16x32xf32>, %arg1: memref<8x16xf32>, %arg2: memref<8x32xf32>) {
    %cst = constant 0.000000e+00 : f32
    %c8_i32 = constant 8 : i32
    %c32_i32 = constant 32 : i32
    %c16_i32 = constant 16 : i32
    %c1 = constant 1 : index
    %c8 = constant 8 : index
    %c32 = constant 32 : index
    %c0 = constant 0 : index
    %c16 = constant 16 : index
    scf.for %arg3 = %c0 to %c8 step %c1 {
      scf.for %arg4 = %c0 to %c32 step %c1 {
        store %cst, %arg2[%arg3, %arg4] : memref<8x32xf32>
      }
    }
    scf.for %arg3 = %c0 to %c8 step %c8 {
      scf.for %arg4 = %c0 to %c32 step %c32 {
        scf.for %arg5 = %c0 to %c16 step %c16 {
          %0 = subview %arg1[%arg3, %arg5] [8, 16] [1, 1]  : memref<8x16xf32> to memref<8x16xf32, #map0>
          %1 = memref_cast %0 : memref<8x16xf32, #map0> to memref<*xf32>
          %2 = subview %arg0[%arg5, %arg4] [16, 32] [1, 1]  : memref<16x32xf32> to memref<16x32xf32, #map1>
          %3 = memref_cast %2 : memref<16x32xf32, #map1> to memref<*xf32>
          %4 = subview %arg2[%arg3, %arg4] [8, 32] [1, 1]  : memref<8x32xf32> to memref<8x32xf32, #map1>
          %5 = memref_cast %4 : memref<8x32xf32, #map1> to memref<*xf32>
          %6 = call @plaidml_rt_xsmm_dispatch_gemm_f32(%c16_i32, %c32_i32, %c32_i32, %c8_i32, %c32_i32, %c16_i32) : (i32, i32, i32, i32, i32, i32) -> i64
          call @plaidml_rt_xsmm_exec_gemm_f32(%1, %3, %5, %6) : (memref<*xf32>, memref<*xf32>, memref<*xf32>, i64) -> ()
        }
      }
    }
    return
  }
}`

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -231,6 +231,10 @@ struct FuncOpConversion : public OpConversionPattern<FuncOp> {
   matchAndRewrite(FuncOp op, ArrayRef<Value> operands,
                   ConversionPatternRewriter &rewriter) const final {
     FunctionType type = op.getType();
+    /* do not rewrite xsmm dispatch functions */
+    if (op.getName().str() == "plaidml_rt_xsmm_dispatch_gemm_f32") {
+      return mlir::success();
+    }
```
flaub, line 237: Might be better/more general to have this rewrite only work on FuncOps with a body .

alheinecke: I believe the whole pass right now is a bit weird, why making a functions output part of the inputs?

flaub: What we are doing here is generally named 'buffer assignment'. The idea is to convert from a functional/value-like semantic into one with buffers and side-effects where we write results into output buffers provided by the user. However, this only needs to take place on the top-level function, not any declarations for external functions provided at runtime (like `plaidml_rt_xsmm_dispatch_gemm_f32`).

@jbruestle WDYT?

flaub: I think we'll also need to adjust the 
```
target.addDynamicallyLegalOp<FuncOp>(
      [](FuncOp op) { return op.getType().getNumResults() == 0; });
```

So that it also allows for empty function body declarations to be considered legal regardless the number of results.

alheinecke:
@flaub : I would need some help with this as I seem to miss something here :-( 
I have added an operator split with custom ops for xsmm dispatch and invoke (we might want to re-thing in future how we split functionality between stencil and lowering). 

But when enabling it with:
https://github.com/plaidml/plaidml/blob/aheineck-xsmm-jitsplit/pmlc/target/x86/stencil_xsmm.cc#L309 
making sure we are going into the new code, the xsmm lowering fails with the subview creation for "C":
```
error: 'std.subview' op operand #1 must be index, but got 'memref<16x32xf32>'
// *** IR Dump After mlir::detail::VerifierPass Failed *** ('func' operation: @dot)
```
Can you help?

PR #1101, opened 2020-06-06
Add failing XSMM test case

flaub:
Hey @alheinecke, could you try this with:

```
bazelisk run plaidml/edsl/tests:cc_test -c dbg -- --gtest_filter=CppEdsl.BigDot
```

Note that it needs to print this, otherwise it crashed:

```
[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (193 ms total)
[  PASSED  ] 1 test.
```


flaub:
To follow up, this appears to be due to my stack size being limited to 8k. If I use `ulimit -S -s 131092`, then it will run without crashing.

flaub:
This should now be resolved

PR #1105, opened 2020-06-09
Add split to OpenVINO plugin

tzerrell:


plaidml/bridge/openvino/src/plaidml_plugin/ops/split.cpp
```diff
@@ -0,0 +1,51 @@
...
+  auto splits = layer->get_num_splits();
+  ngraph::AxisSet axes;
+  auto axis_ngraph_op = std::dynamic_pointer_cast<ngraph::op::Constant>(layer->input_value(1).get_node_shared_ptr());
+  if (axis_ngraph_op) {
+    axes = axis_ngraph_op->get_axis_set_val();
+  } else {
+    THROW_IE_EXCEPTION << "Dynamic axis not currently supported by PlaidML plugin";
+  }
+  IE_ASSERT(axes.size() == 1);
+  auto axis = axes.to_vector()[0];
```
flaub, line 31: Might be good to throw this into a utility since this appears to be something you'll need to do fairly often.

plaidml/bridge/openvino/src/plaidml_plugin/ops/split.cpp
```diff
@@ -0,0 +1,51 @@
...
+  std::vector<edsl::TensorIndex> I_idxs(ndims);
+  std::vector<edsl::Tensor> Os;
+  I.bind_dims(I_dims);
+  auto O_dims = I_dims;
+  auto split_size = I_dims[axis] / splits;
+  O_dims[axis] = split_size;
+  for (size_t i = 0; i < splits; i++) {
+    auto thisO = edsl::TensorOutput(O_dims);
+    auto O_idxs = I_idxs;
+    O_idxs[axis] = I_idxs[axis] + i * split_size;
```
flaub, line 44: I'm a bit puzzled by this, what's the intent? And what tile code is actually generated?

tzerrell: > It's unclear whether the crash is in our code or the test, because nothing uses the test with `num_splits != 1`. I'm now investigating how to fix this.

The problem is related to multiple network outputs from a single node. It's at least partially related to our code, but the full scope of the problem is unclear. This will be an involved fix, so I'm tabling it for now. I'll extract the constant axis set reading code to a different PR.

tzerrell:
Under the new architecture this will not land here.

PR #1106, opened 2020-06-09
Add reshape to OpenVINO plugin

tzerrell:


plaidml/bridge/openvino/src/plaidml_plugin/ops/reshape.cpp
```diff
@@ -0,0 +1,42 @@
...
+  IE_ASSERT(ctx.operands.size() == 2);
+  auto I = ctx.operands.at(0);
+  // operands.at(1) is unused, just read the Constant instead
+  ngraph::Shape shape;
+  auto shape_ngraph_op = std::dynamic_pointer_cast<ngraph::op::Constant>(layer->input_value(1).get_node_shared_ptr());
+  if (shape_ngraph_op) {
+    shape = shape_ngraph_op->get_shape_val();
+  } else {
+    THROW_IE_EXCEPTION << "Dynamic reshaping not currently supported by PlaidML plugin";
+  }
```
flaub, line 28: Might be good to throw this into a utility if multiple ops will need it.

PR #1111, opened 2020-06-11
Add MVN to OpenVINO plugin

tzerrell:
Depends on https://github.com/plaidml/openvino/pull/5, will be a draft until that lands.

### DIFF
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -40,6 +40,7 @@ Value maximum(const Value&);
 Value mean(const Value&);
 Value min(const Value&);
 Value minimum(const Value&);
+Value mvn(const Value&);
 Value l2norm(const Value&);
 Value pool(const Value&);
 Value prod(const Value&);
@@ -1971,6 +1972,39 @@ Value minimum(const Value& value) {
   return Value{O};
 }
 
+Value mvn(const Value& value) {
+  IVLOG(1, "mvn");
+  auto args = value.as_tuple();
+  if (args.size() != 6) {
+    throw std::runtime_error("mvn expects 6 arguments");
+  }
+  auto I_raw = args[0];
+  auto I = args[0].as_tensor();
+  auto axes = args[1];
+  auto normalize_variance = args[2].as_bool();
+  auto epsilon = args[3].as_float();
+  auto across_channels = args[4].as_bool();
+  auto layout = args[5].as_str();
+  if (axes.is_none()) {
+    std::vector<int64_t> raw_axes;
+    for (size_t i = 0; i < layout.size(); i++) {
+      auto dim = layout[i];
+      if (dim == 'N') continue;
+      if (dim == 'C' && !across_channels) continue;
+      raw_axes.push_back(i);
+    }
+    axes = edsl::make_tuple(raw_axes);
+  }
+  auto R = I - mean(edsl::make_tuple(I_raw, axes, /*keepdims=*/true)).as_tensor();
+
+  if (normalize_variance) {
+    auto stdev = edsl::sqrt(variance(edsl::make_tuple(I_raw, axes, /*keepdims=*/true)).as_tensor());
+    R = R / maximum(edsl::make_tuple(stdev, Tensor(epsilon))).as_tensor();
+  }
+
+  return Value{R};
+}
+
 Value l2norm(const Value& value) {
   IVLOG(1, "l2norm");
   auto args = value.as_tuple();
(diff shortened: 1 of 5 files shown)


plaidml/op/op.h
```diff
@@ -337,6 +337,64 @@ inline edsl::Tensor minimum(const edsl::Tensor& X, const edsl::Tensor& Y) {
...
+    // negative axes interpreted as in numpy
+    axes_ = edsl::make_tuple(axes);
+    return *this;
+  }
+
+  mvn& remove_axes_parameter() {
+    // Restores axes to the default of being unused (meaning use `across_channels` & `layout` instead)
+    axes_ = edsl::None();
+    return *this;
+  }
```
flaub, line 360: Hmm, do we need this? What's the thinking here?

tzerrell: Eh, probably not really necessary. The thinking was that without this, once `axes` was set there was no way to un-set it except creating a new instance of the `mvn` class. That said, I don't really see why anyone would need to set and then un-set the `axes` parameter.

plaidml/op/op.h
```diff
@@ -337,6 +337,64 @@ inline edsl::Tensor minimum(const edsl::Tensor& X, const edsl::Tensor& Y) {
...
+
+  mvn& epsilon(double value) {
+    epsilon_ = value;
+    return *this;
+  }
+
+  mvn& layout(const std::string& value) {
+    // only used if `axes` is `None`
```
flaub, line 379: Maybe we could add a check here and throw if the user specified `axes`. And also we could do the same for `axes` (if `layout` was set, then throw).

plaidml/op/op.h
```diff
@@ -337,6 +337,65 @@ inline edsl::Tensor minimum(const edsl::Tensor& X, const edsl::Tensor& Y) {
...
+      : I_(I),
+        axes_(edsl::None()),
+        normalize_variance_(true),
+        epsilon_(1e-9),
+        across_channels_(true),
+        layout_("NHWC") {}
+
+  mvn& axes(const std::vector<int64_t>& axes) {
+    // negative axes interpreted as in numpy
+    if (!across_channels_ || layout_ != "NHWC") {
```
flaub, line 352: Hmm, the only actual requirement is that `N` and `C` are in the layout, but I don't think a specific fully formed layout is required.

flaub: Sorry, actually just `C` must exist, and then if `N` exists we skip it.

tzerrell: This check is only implementing "And also we could do the same for axes (if layout was set, then throw)." (since "NHWC" is the default value of layout, this throws when layout is set.) Do you want a requirement checking that `C` exists? I currently don't do any such check at all, just skip "N" (and "C" if `across_channels_ == true`)

flaub: Hmm, maybe we shouldn't have a default for layout then... 

It seems like we want two exclusive groups of options:
- group 1:
  - axes
- group 2:
  - across_channels
  - layout

flaub: OK, maybe we just say `if (!layout_.empty()) throw`

plaidml/op/op.h
```diff
@@ -340,16 +340,11 @@ inline edsl::Tensor minimum(const edsl::Tensor& X, const edsl::Tensor& Y) {
 class mvn {
  public:
   explicit mvn(const plaidml::edsl::Tensor& I)
-      : I_(I),
-        axes_(edsl::None()),
-        normalize_variance_(true),
-        epsilon_(1e-9),
-        across_channels_(true),
-        layout_("NHWC") {}
+      : I_(I), axes_(edsl::None()), normalize_variance_(true), epsilon_(1e-9), across_channels_(true), layout_("") {}
```
flaub, line 343: No need to default to `""` (that's what the default ctor of a `std::string` does already)

plaidml/op/lib/ops.cc
```diff
@@ -1741,6 +1742,39 @@ Value minimum(const Value& value) {
...
+  }
+  auto I_raw = args[0];
+  auto I = args[0].as_tensor();
+  auto axes = args[1];
+  auto normalize_variance = args[2].as_bool();
+  auto epsilon = args[3].as_float();
+  auto across_channels = args[4].as_bool();
+  auto layout = args[5].as_str();
+  if (axes.is_none()) {
+    std::vector<int64_t> raw_axes;
```
flaub, line 1759: Should probably check if `layout` is specified here.

tzerrell:
Rebased forward and ready for review

cnamrata15:
Seems like some conversations might be unresolved. 
Looks good to me otherwise.

tzerrell:
I've marked as resolved the outstanding inline comments, as I believe they have all been resolved.

PR #1119, opened 2020-06-15
Add broadcast to op lib

dgkutnic:
I've been working with @tzerrell regarding some weirdness I've been seeing in my implementation of the `broadcast` op.

Essentially, I wrote a `broadcast` op, the tile code looks fine, but I run into the following errors when trying to test the code with `runProgram` or `checkProgram`.

```
error: 'func' op unsupported module-level operation
unknown file: Failure
C++ exception with description "could not convert to LLVM IR" thrown in the test body.
```

To try and debug, I created a test `CppEdsl.BroadcastNonOpLib` in `edsl_test.cc` that does a broadcast using _just_ EDSL code, and replicated the same code using the `broadcast` op in `op_test.cc`.

Program generated from EDSL code:
```
#map0 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
#map1 = affine_map<(d0, d1, d2, d3) -> (d0, d2, d3)>


module {
  func @broadcast(%arg0: tensor<1x10x10xf32>) -> tensor<1x7x10x10xf32> {
    %cst = "eltwise.sconst"() {value = 0.000000e+00 : f64} : () -> tensor<f32>
    %0 = tile.contract assign, none, %cst, %arg0 {sink = #map0, srcs = [#map1]} : tensor<f32>, tensor<1x10x10xf32> -> tensor<1x7x10x10xf32>
    return %0 : tensor<1x7x10x10xf32>
  }
}
```

Program generated from the op lib:
```
#map0 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
#map1 = affine_map<(d0, d1, d2, d3) -> (d0, d2, d3)>


module {
  func @broadcast_numpy(%arg0: tensor<1x10x10xf32>) -> tensor<1x7x10x10xf32> {
    %cst = "eltwise.sconst"() {value = 0.000000e+00 : f64} : () -> tensor<f32>
    %0 = tile.contract assign, none, %cst, %arg0 {sink = #map0, srcs = [#map1]} : tensor<f32>, tensor<1x10x10xf32> -> tensor<1x7x10x10xf32>
    return %0 : tensor<1x7x10x10xf32>
  }
}
```

Why would basically the same tile code pass in one case and fail in the other case?

### DIFF
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -488,11 +488,12 @@ Value broadcast(const Value& value) {
   // Define broadcast indices
   for (int i = 0; i < target_shape.size(); i++) {
     O_dims.emplace_back(TensorDim{target_shape[i]});
-    auto tidx = TensorIndex(llvm::formatv("x{0}", i));
+    auto tidx = TensorIndex();
+    // auto tidx = TensorIndex(llvm::formatv("x{0}", i));
     if (std::find(broadcast_axes.begin(), broadcast_axes.end(), i) != broadcast_axes.end()) {
       I_idxs.emplace_back(tidx);
     }
-    O_idxs.emplace_back(TensorIndex(tidx));
+    O_idxs.emplace_back(tidx);
   }
   auto O = TensorOutput(O_dims);
   O(O_idxs) = I(I_idxs);
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -487,7 +487,7 @@ Value broadcast(const Value& value) {
   std::vector<TensorIndex> O_idxs;
 
   // Define broadcast indices
-  for (int i = 0; i < target_shape.size(); i++) {
+  for (size_t i = 0; i < target_shape.size(); i++) {
     O_dims.emplace_back(TensorDim{target_shape[i]});
     auto tidx = TensorIndex(llvm::formatv("x{0}", i));
     if (input_shape[i] == 1) {
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -23,6 +23,7 @@ Value all(const Value&);
 Value any(const Value&);
 Value argmax(const Value&);
 Value binary_crossentropy(const Value&);
+Value broadcast(const Value&);
 Value clip(const Value&);
 Value concatenate(const Value&);
 Value convolution(const Value&);
@@ -464,6 +465,39 @@ Value binary_crossentropy(const Value& value) {
   };
   // Safe to use P without copy since it is built internally to this function
   return Value{OverrideGrads(deriv, std::vector<Tensor>{T, P}, O)};
+}
+
+Value broadcast(const Value& value) {
+  IVLOG(1, "broadcast");
+  auto args = value.as_tuple();
+  if (args.size() != 3) {
+    throw std::runtime_error(llvm::formatv("PlaidML broadcast op expects 3 arguments (received {0})", args.size()));
+  }
+
+  auto I = args[0].as_tensor();
+  auto target_shape = args[1].as_int_tuple();
+  auto broadcast_axes = args[2].as_int_tuple();
+
+  std::vector<TensorDim> O_dims;
+  std::vector<TensorDim> I_dims(I.rank());
+  I.bind_dims(I_dims);
+
+  std::vector<TensorIndex> I_idxs;
+  std::vector<TensorIndex> O_idxs;
+
+  // Define broadcast indices
+  for (int i = 0; i < target_shape.size(); i++) {
+    O_dims.emplace_back(TensorDim{target_shape[i]});
+    auto tidx = TensorIndex(llvm::formatv("x{0}", i));
+    if (std::find(broadcast_axes.begin(), broadcast_axes.end(), i) != broadcast_axes.end()) {
+      I_idxs.emplace_back(tidx);
+    }
+    O_idxs.emplace_back(TensorIndex(tidx));
+  }
+  auto O = TensorOutput(O_dims);
+  O(O_idxs) = I(I_idxs);
+
+  return Value{O};
 }
 
 Value clip(const Value& value) {
@@ -2788,6 +2822,7 @@ void RegisterOps() {
   registry->Register("any", any);
   registry->Register("argmax", argmax);
   registry->Register("binary_crossentropy", binary_crossentropy);
+  registry->Register("broadcast", broadcast);
   registry->Register("clip", clip);
   registry->Register("concatenate", concatenate);
   registry->Register("convolution", convolution);
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -488,8 +488,7 @@ Value broadcast(const Value& value) {
   // Define broadcast indices
   for (int i = 0; i < target_shape.size(); i++) {
     O_dims.emplace_back(TensorDim{target_shape[i]});
-    auto tidx = TensorIndex();
-    // auto tidx = TensorIndex(llvm::formatv("x{0}", i));
+    auto tidx = TensorIndex(llvm::formatv("x{0}", i));
     if (std::find(broadcast_axes.begin(), broadcast_axes.end(), i) != broadcast_axes.end()) {
       I_idxs.emplace_back(tidx);
     }
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -475,6 +475,7 @@ Value broadcast(const Value& value) {
   }
 
   auto I = args[0].as_tensor();
+  auto input_shape = I.compute_shape().sizes();
   auto target_shape = args[1].as_int_tuple();
   auto broadcast_axes = args[2].as_int_tuple();
 
@@ -489,7 +490,9 @@ Value broadcast(const Value& value) {
   for (int i = 0; i < target_shape.size(); i++) {
     O_dims.emplace_back(TensorDim{target_shape[i]});
     auto tidx = TensorIndex(llvm::formatv("x{0}", i));
-    if (std::find(broadcast_axes.begin(), broadcast_axes.end(), i) != broadcast_axes.end()) {
+    if (input_shape[i] == 1) {
+      I_idxs.emplace_back(TensorIndex(0));
+    } else if (std::find(broadcast_axes.begin(), broadcast_axes.end(), i) != broadcast_axes.end()) {
       I_idxs.emplace_back(tidx);
     }
     O_idxs.emplace_back(tidx);
--- a/plaidml/op/op_test.cc
+++ b/plaidml/op/op_test.cc
@@ -29,7 +29,6 @@ namespace {
 class OpTest : public TestFixture {};
 
 Program makeProgram(const std::string& name, const std::vector<Tensor>& outputs) {
-  // TODO: remove empty target once all of these are passing.
   return ProgramBuilder(name, outputs).compile();
 }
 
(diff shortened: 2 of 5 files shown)


flaub:
Basically, the OpTest was broken because it wasn't setting the `target` for the when compiling a program. This was due to other failures that have all been fixed now.

plaidml/op/op_test.cc
```diff
@@ -30,7 +30,7 @@ class OpTest : public TestFixture {};
 
 Program makeProgram(const std::string& name, const std::vector<Tensor>& outputs) {
   // TODO: remove empty target once all of these are passing.
```
flaub, line 32: oops, forgot to remove this comment

dgkutnic: Done

tzerrell:
> * Broadcasting where dimension values are 1 - the test broadcasting `[3, 1] → [3, 4]` has incorrect results

Ah yes, this would need to fill in a 1 instead of the index for that dimension in `I_idxs` in such a case.

Not sure offhand what's causing the other problem

dgkutnic:
Hi @jbruestle @flaub,

I've been working on the explicit broadcast op for the op lib, and need some help with debugging affine map correctness issues. 

I've created a few different test cases for a few of the broadcasting semantics I've encountered, and in each of the cases, the affine map looks correct. However, in all cases _besides_ the no-op broadcast, it seems as if the row/column ordering of the broadcast gets mixed up (for instance, if we want the result to be `000011112222`, it becomes `012012012012`, and vice versa). Tim and I have both looked at the corresponding tile dialect code and can't see any issues.

I've attached the test cases and tile dialect modules below for your reference. Can you take a look and let me know if you can see anything obviously wrong with the affine maps or elsewhere?

## Broadcast Results for each Broadcast

<table>
<tr>
  <td><div align="center"><b>Broadcast Shapes</b></div></td>
  <td><div align="center"><b>Input Value</b></div></td>
  <td><div align="center"><b>Expected Value</b></div></td>
  <td><div align="center"><b>Actual Value</b></div></td>
</tr>
<tr>
  <td><div align="center"><p>No-Op</p><code>[3, 4] -> [3, 4]</code></div></td>
  <td><pre><code>[[0, 1, 2],
[0, 1, 2],
[0, 1, 2],
[0, 1, 2]]</code></pre></td>
  <td><div align="center"><pre><code>[[0, 1, 2],
[0, 1, 2],
[0, 1, 2],
[0, 1, 2]]</code></pre></div></td>
  <td><div align="center"><pre><code>[[0, 1, 2],
[0, 1, 2],
[0, 1, 2],
[0, 1, 2]]</code></pre></div></td>
</tr>
<tr>
  <td><div align="center"><p>Numpy Legal Broadcast</p><code>[3] -> [4, 3]</code></div></td>
  <td><div align="center"><code>[0, 1, 2]</code></div></td>
  <td><div align="center"><pre><code>[[0, 0, 0, 0],
[1, 1, 1, 1],
[2, 2, 2, 2]]</code></pre></div></td>
  <td><div align="center"><pre><code>[[0, 1, 2, 0], 
[1, 2, 0, 1],
[2, 0, 1, 2]]</code></pre></div></td>
</tr>
<tr>
  <td><div align="center"><p>Numpy Illegal Broadcast</p><code>[3] -> [3, 4]</code></div></td>
  <td><div align="center"><code>[0, 1, 2]</code></div></td>
  <td><div align="center"><pre><code>[[0, 1, 2],
[0, 1, 2],
[0, 1, 2],
[0, 1, 2]]</code></pre></div></td>
  <td><div align="center"><pre><code>[[0, 0, 0],
[0, 1, 1],
[1, 1, 2],
[2, 2, 2]]</code></pre></div></td>
</tr>
<tr>
  <td><div align="center"><p>Legal Broadcast with any input dim=1</p><code>[3, 1] -> [3, 4]</code></div></td>
  <td><div align="center"><code>[0, 1, 2]</code></div></td>
  <td><div align="center"><pre><code>[[0, 1, 2],
[0, 1, 2],
[0, 1, 2],
[0, 1, 2]]</code></pre></div></td>
  <td><div align="center"><pre><code>[[0, 0, 0],
[0, 1, 1],
[1, 1, 2],
[2, 2, 2]]</code></pre></div></td>
</tr>
</table>

## Tile Code for each Broadcast

No-Op: `[3, 4] -> [3, 4]`
```
#map0 = affine_map<(d0, d1) -> (d0, d1)>


module {
  func @broadcast_nop(%arg0: tensor<3x4xf32>) -> tensor<3x4xf32> {
    %cst = "eltwise.sconst"() {value = 0.000000e+00 : f64} : () -> tensor<f32>
    %0 = tile.contract assign, none, %cst, %arg0 {idxs = ["x0", "x1"], sink = #map0, srcs = [#map0]} : tensor<f32>, tensor<3x4xf32> -> tensor<3x4xf32>
    return %0 : tensor<3x4xf32>
  }
}
```

Numpy Legal Broadcast: `[3] -> [4, 3]`
```
#map0 = affine_map<(d0, d1) -> (d0, d1)>
#map1 = affine_map<(d0, d1) -> (d1)>


module {
  func @broadcast_numpy_legal(%arg0: tensor<3xf32>) -> tensor<4x3xf32> {
    %cst = "eltwise.sconst"() {value = 0.000000e+00 : f64} : () -> tensor<f32>
    %0 = tile.contract assign, none, %cst, %arg0 {idxs = ["x0", "x1"], sink = #map0, srcs = [#map1]} : tensor<f32>, tensor<3xf32> -> tensor<4x3xf32>
    return %0 : tensor<4x3xf32>
  }
}
```

Numpy Illegal Broadcast: `[3] -> [3, 4]`
```
#map0 = affine_map<(d0, d1) -> (d0, d1)>
#map1 = affine_map<(d0, d1) -> (d0)>


module {
  func @broadcast_non_numpy(%arg0: tensor<3xf32>) -> tensor<3x4xf32> {
    %cst = "eltwise.sconst"() {value = 0.000000e+00 : f64} : () -> tensor<f32>
    %0 = tile.contract assign, none, %cst, %arg0 {idxs = ["x0", "x1"], sink = #map0, srcs = [#map1]} : tensor<f32>, tensor<3xf32> -> tensor<3x4xf32>
    return %0 : tensor<3x4xf32>
  }
}
```

Numpy Legal Broadcast with any input dim=1: `[3, 1] -> [3, 4]`
```
#map0 = affine_map<(d0, d1) -> (d0, d1)>
#map1 = affine_map<(d0, d1) -> (d0, 0)>


module {
  func @broadcast_numpy_legal_2(%arg0: tensor<3x1xf32>) -> tensor<3x4xf32> {
    %cst = "eltwise.sconst"() {value = 0.000000e+00 : f64} : () -> tensor<f32>
    %0 = tile.contract assign, none, %cst, %arg0 {idxs = ["x0", "x1"], sink = #map0, srcs = [#map1]} : tensor<f32>, tensor<3x1xf32> -> tensor<3x4xf32>
    return %0 : tensor<3x4xf32>
  }
}
```

flaub:
I think you've got row-major and column-major mixed up. In C++, we assume row-major, and so a shape like `[4, 3]` I would expect to have elements like:
```
[[x11, x12, x13],
 [x21, x22, x23],
 [x31, x32, x33],
 [x41, x42, x43]]
```

AFAICT, your expected results are simply transposed.

dgkutnic:
Thanks Frank for clarifying! Yes, that was a mix-up between row- and column-major ordering on my end.

tzerrell:
If we have a row-/column-major mixup, wouldn't that mean that the no-op results are wrong? If you 2 have it sorted out I suppose my confusion doesn't matter, but I remain confused.

dgkutnic:
The issue resolving the no-op was resolved offline. I've modified the test case to be clearer.

Old test case:
```
[[0, 1, 2, 0],
 [1, 2, 0, 1],
 [2, 0, 1, 2]]
```

New test case:
```
[[0, 1, 2, 3],
 [0, 1, 2, 3],
 [0, 1, 2, 3]]
```

plaidml/op/lib/ops.cc
```diff
@@ -466,6 +467,42 @@ Value binary_crossentropy(const Value& value) {
...
 
+Value broadcast(const Value& value) {
+  IVLOG(1, "broadcast");
+  auto args = value.as_tuple();
+  if (args.size() != 3) {
+    throw std::runtime_error(llvm::formatv("PlaidML broadcast op expects 3 arguments (received {0})", args.size()));
+  }
+
+  auto I = args[0].as_tensor();
+  auto input_shape = I.compute_shape().sizes();
```
tzerrell, line 478: The compile-time perf impact of computing shape each time we broadcast is probably going to be pretty bad. Do we care about that currently in the context of this commit?

If so, switching to an `expand_dims` + autobroadcast implementation should be able to avoid this. The costs of that would be reimplementing the entire thing & the new version would have many more contractions (which should all be optimized away, as they're just reinterpreting the same underlying data with a new shape; but I'm not sure whether we have the relevant optimizations implemented yet).

dgkutnic: I want to hear what Frank says, but here's my perspective:
The `broadcast` op is primarily to be used in TensorFlow for now, and perhaps in other framework implementations later. Right now, our main focus is getting a few networks up and running, with performance being a secondary priority. It's important to note performance hiccups like this, but I don't think it's our top priority to fix them now.

flaub: OK, `compute_shape` indeed should be avoided, however, how could we make this work without it? I have a feeling we need a new primitive. For now I'm fine with letting it slide and dealing with this later. But I do wonder how we'd make this work if we wanted to support 'deferred' shapes and it so happens that the `?` dim resolves to `1`.

plaidml/op/op_test.cc
```diff
@@ -157,6 +156,112 @@ module {
...
+
+TEST_F(OpTest, BroadcastNumpy2) {
+  auto A = Placeholder(DType::FLOAT32, {3, 1});
+  std::vector<int> rshape = {3, 4};
+  std::vector<int> bdims = {0, 1};
+  auto C = broadcast(A, rshape, bdims);
+  auto program = makeProgram("broadcast_numpy_2", {C});
+
+  std::vector<float> A_input = {0, 1, 2};
+  std::vector<float> C_output = {0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 2};
```
flaub, line 215: I find it more readable to put in newlines for each 'row'.

plaidml/op/lib/ops.cc
tzerrell: My idea was the following:
Use `expand_dims` to add size 1 dimensions where necessary (to the front of the lower-dimensional tensor in Numpy-style, to the back of the lower-dimensional tensor in that style, or where explicitly requested in the explicit style). Now the tensors are the same rank, and we can rely on autobroadcasting, since when all tensors have the same number of dimensions there is only one legal broadcast so it doesn't matter what the broadcast style is.

tzerrell: But it sounds like there's consensus around deferring this to later.

tzerrell: The exception to "only one legal broadcast" is if you're not broadcasting 2 tensors to their minimal shared shape, but instead explicitly "upbroadcasting" a tensor to a specific size, e.g. explicitly broadcasting a `(1, 3)` tensor to shape `(4, 3)` (or e.g. a `(?, 3)` tensor where `?` = `1` or `4` to `(4, 3)`); we _could_ do this with an autobroadcast technique by e.g. adding `zeroes_like(target_shape)` to the tensor with the expectation that optimization passes will remove everything but the broadcast. Not sure if that's a good idea.

flaub: Oh I see. Yes I was wondering what the `broadcast_axes` parameter was for, but it seems like we could just have explicit mode for now and then allow users to specify where the `1`s should be inserted.

plaidml/op/op_test.cc
dgkutnic: Sorry, must have missed this one while reformatting.

plaidml/op/op_test.cc
```diff
@@ -157,6 +156,114 @@ module {
...
+
+TEST_F(OpTest, BroadcastNumpy3) {
+  auto A = Placeholder(DType::FLOAT32, {3});
+  std::vector<int> rshape = {4, 3};
+  std::vector<int> bdims = {1};
+  auto C = broadcast(A, rshape, bdims);
+  auto program = makeProgram("broadcast_numpy_3", {C});
+
+  std::vector<float> A_input = {0, 1, 2};
+  std::vector<float> C_output = {0, 1, 2, 0, 1, 2, 0, 1, 2, 0, 1, 2};
```
tzerrell, line 229: Also here

plaidml/op/op_test.cc
```diff
@@ -157,6 +156,114 @@ module {
...
+
+TEST_F(OpTest, BroadcastNonNumpy) {
+  auto A = Placeholder(DType::FLOAT32, {3});
+  std::vector<int> rshape = {3, 4};
+  std::vector<int> bdims = {0};
+  auto C = broadcast(A, rshape, bdims);
+  auto program = makeProgram("broadcast_non_numpy", {C});
+
+  std::vector<float> A_input = {0, 1, 2};
+  std::vector<float> C_output = {0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 2};
```
tzerrell, line 241: Also here

plaidml/op/op_test.cc
dgkutnic: hmm, seems like clang-format is doing something to these. I remember explicitly putting newlines here.

tzerrell: it will format back to one line if you don't put `//` at the end of the line if it thinks it all fits on one line.

dgkutnic: ok, that makes sense. fixed these now

Issue #1122, opened 2020-06-16
the tensor which pass to plaidml::op::concatenate have to cast first.


XingHongChenIntel:
In a very simple tile example, such like below:    
 
    auto order = mv::Order("NCHW");
    std::vector<int64_t> I_shape{1, 1, 16, 16};
    auto I = Placeholder(DType::FLOAT16, I_shape);
    auto B = I * 1.5;
    auto R = plaidml::op::concatenate({I, B}, 2);
    auto program = ProgramBuilder("test_concatenate", {R}) 
                       .floatx(DType::FLOAT16)
                       .compile();

we will get the compile **error**:

**error: 'affine.store' op first operand must have same type memref element type**
terminate called after throwing an instance of 'std::runtime_error'
  what():  conversion to the LLVM IR dialect failed

but if we cast tensor(B) to DType::FLOAT16, then it would compile successfully.

   `auto B = cast( I * 1.5, DType::FLOAT16);`

because the ProgramBuilder::floatx() method is used for specifying which data type constants will actually have. 
since we already set it to FLOAT16. why we still need to cast tensor(B)?

flaub:
I'll have a fix for this shortly.

PR #1124, opened 2020-06-16
Fix for #1122

flaub:
With the current implementation of floatx/intx, it's possible for the initial element type inference of a contraction to compute a higher-precision type. Later, once the constant types are materialized, we need to re-infer the element type as a symbolic contraction is being converted to a regular one.

### DIFF
--- a/pmlc/dialect/tile/tests/build-cion.mlir
+++ b/pmlc/dialect/tile/tests/build-cion.mlir
@@ -1,5 +1,8 @@
-// RUN: pmlc-opt -canonicalize -cse -split-input-file %s | FileCheck %s
+// RUN: pmlc-opt -tile-make-program -canonicalize -cse -split-input-file %s | FileCheck %s
 
+// CHECK: #[[MAP0:map[0-9]+]] = affine_map<(d0, d1, d2) -> (d0, d1)>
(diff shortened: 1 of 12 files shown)


pmlc/dialect/tile/tests/build-cion.mlir
```diff
@@ -16,17 +19,11 @@ func @dot(%arg0: tensor<1x2xf32>, %arg1: tensor<2x3xf32>) -> tensor<?x?xf32> {
   return %10 : tensor<?x?xf32>
 }
 
-// CHECK: #[[MAP0:map[0-9]+]] = affine_map<(d0, d1, d2) -> (d0, d1)>
-// CHECK: #[[MAP1:map[0-9]+]] = affine_map<(d0, d1, d2) -> (d0, d2)>
-// CHECK: #[[MAP2:map[0-9]+]] = affine_map<(d0, d1, d2) -> (d2, d1)>
```
tzerrell, line 21: The specific structure of the affine maps is encoding some semantic information, it seems a little weird to remove these checks. That said, if that's the typical MLIR style, then I don't object.

flaub: Not removed, just moved, look up above at line 3.

tzerrell: whoops, thanks!

PR #1126, opened 2020-06-17
Broadcast op

dgkutnic:
A continuation of #1119

### DIFF
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -475,9 +475,15 @@ Value broadcast(const Value& value) {
   }
 
   auto I = args[0].as_tensor();
+  auto broadcast_axes = args[2].as_int_tuple();
+  if (I.rank() != broadcast_axes.size()) {
+    throw std::runtime_error(
+        llvm::formatv("Mismatched broadcast axes (received {0} broadcast axes for input tensor of rank {1})",
+                      broadcast_axes.size(), I.rank()));
+  }
+
   auto input_shape = I.compute_shape().sizes();
   auto target_shape = args[1].as_int_tuple();
-  auto broadcast_axes = args[2].as_int_tuple();
 
   std::vector<TensorDim> O_dims;
   std::vector<TensorDim> I_dims(I.rank());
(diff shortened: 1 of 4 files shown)


plaidml/op/lib/ops.cc
```diff
@@ -466,6 +467,52 @@ Value binary_crossentropy(const Value& value) {
...
+  // Define output dims and indexes
+  for (size_t i = 0; i < target_shape.size(); i++) {
+    O_dims.emplace_back(TensorDim{target_shape[i]});
+    O_idxs.emplace_back(TensorIndex(llvm::formatv("x{0}", i)));
+  }
+
+  // Define input dims and indexes
+  for (size_t i = 0; i < broadcast_axes.size(); i++) {
+    if (input_shape[i] == 1) {
+      I_idxs.emplace_back(TensorIndex(0));
```
flaub, line 504: No need to use `TensorIndex(0)`, there's a ctor for integers.

plaidml/op/lib/ops.cc
```diff
@@ -466,6 +467,52 @@ Value binary_crossentropy(const Value& value) {
...
+
+  auto input_shape = I.compute_shape().sizes();
+  auto target_shape = args[1].as_int_tuple();
+
+  std::vector<TensorDim> O_dims;
+  std::vector<TensorDim> I_dims(I.rank());
+  I.bind_dims(I_dims);
+
+  std::vector<TensorIndex> I_idxs;
+  std::vector<TensorIndex> O_idxs;
```
flaub, line 493: You could just construct this vector with the right size.

plaidml/op/lib/ops.cc
```diff
@@ -466,6 +467,52 @@ Value binary_crossentropy(const Value& value) {
...
+  if (I.rank() != broadcast_axes.size()) {
+    throw std::runtime_error(
+        llvm::formatv("Mismatched broadcast axes (received {0} broadcast axes for input tensor of rank {1})",
+                      broadcast_axes.size(), I.rank()));
+  }
+
+  auto input_shape = I.compute_shape().sizes();
+  auto target_shape = args[1].as_int_tuple();
+
+  std::vector<TensorDim> O_dims;
```
flaub, line 488: Can we say:

```
std::vector<TensorDim> O_dims(target_shape.begin(), target_shape.end());
```
```

dgkutnic: Fixed

PR #1127, opened 2020-06-17
Add transpose to OpenVino plugin

1sanz:
transpose op added to OpenVino plugin
get_axes_vector_from_constant_operand function added to plaidml_util.hpp to access unsorted axes 
 get_axes_from_constant_operand renamed to get_axes_set_from_constant_operand for clarity


plaidml/bridge/openvino/src/plaidml_plugin/plaidml_util.hpp
```diff
@@ -19,7 +20,8 @@
 
 namespace PlaidMLPlugin {
 
-ngraph::AxisSet get_axes_from_constant_operand(size_t operand_idx, ngraph::Node* layer);
+ngraph::AxisSet get_axes_set_from_constant_operand(size_t operand_idx, ngraph::Node* layer);
+ngraph::AxisVector get_axes_vector_from_constant_operand(size_t operand_idx, ngraph::Node* layer);
```
flaub, line 24: I think we should rename to:

`get_axis_set` and `get_axis_vector`, because to me, `axes` is the plural of `axis` and a set or a vector is valid plurality of `axis`. So `get_axes_set` makes it feel like it should be returning a `vector<AxisSet>`.

plaidml/bridge/openvino/tests/functional/plaidml/single_layer_tests/BUILD
```diff
@@ -10,7 +10,7 @@ TAGS = [
 
 cc_test(
     name = "single_layer_tests",
-    srcs = glob(["*.cpp"]),
+    srcs = glob(["transpose.cpp"]),
```
tzerrell, line 13: Leave this as `"*.cpp"`

1sanz: oops. sorry.

flaub: BTW, if you want to run a single test, look at the docs for gtest here: 
https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#running-a-subset-of-the-tests

basically:

```
bazelisk run //some/test -- --gtest_filter=Fixture.Case
```

PR #1133, opened 2020-06-17
Add OV layer tanh

cnamrata15:


tzerrell:
Updating the title to give a little more context, since it will be the commit message when it lands

tzerrell:
> By including the `abs` code in both its own PR and this PR, you're likely to get merge conflicts when the second PR tries to land (It also makes more work for reviewers).

Looks like I was wrong about this and git is smart enough to do the merge without help. Still creates a bit more work for reviewers, but not a big deal.

cnamrata15:
My bad, I forgot to change the base branch to plaidml-v1 and created the tanh branch. Noticed it afterwards and was kind of curious about how git deals with it.

Will keep this in mind next time onwards.

PR #1134, opened 2020-06-17
Add OV layer sigmoid

cnamrata15:


plaidml/bridge/openvino/src/plaidml_plugin/ops/sigmoid.cpp
```diff
@@ -0,0 +1,24 @@
+/ Copyright(C) 2020 Intel Corporation
```
tzerrell, line 1: Missing `/`

plaidml/bridge/openvino/src/plaidml_plugin/ops/sigmoid.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include "plaidml/op/op.h"
+
+    using namespace plaidml;      // NOLINT[build/namespaces]
+using namespace InferenceEngine;  // NOLINT[build/namespaces]
+
+namespace PlaidMLPlugin {
+
+static OpRegistration reg("sigmoid", [](const Context& ctx) {
+  auto* layer = dynamic_cast<ngraph::opset1::Sigmoid*>(ctx.layer);
```
tzerrell, line 18: Don't include the creation of a `layer` variable if you never need to use it (e.g. on ops with no attributes). It causes compiler errors due to the unused variable.

tzerrell:
Edited title for more context

plaidml/bridge/openvino/src/plaidml_plugin/ops/sigmoid.cpp
cnamrata15: I see. It usually gives a warning though.
I tried removing it, but the errors persist :(

tzerrell: Yes, we have warnings as errors turned on, so the warning becomes an error.

Hmm, the same error persists? And you fixed the `/` -> `//` comment above?

PR #1146, opened 2020-06-18
OpenVino : Pool - max and avg

cnamrata15:


cnamrata15:
> Cool! Do the tests pass?

Yes :)

plaidml/bridge/openvino/src/plaidml_plugin/ops/avg_pool.cpp
```diff
@@ -33,8 +33,7 @@ static OpRegistration reg("avgpool", [](const Context& ctx) {
   bool include_padding_in_avg = false;
   bool use_ceil_for_output_shape = false;
   std::vector<int> padding;
-  auto result = op::pool(I, pool_type, kernel, strides, autopad_mode, padding, input_layout, include_padding_in_avg,
-                         use_ceil_for_output_shape);
+  auto result;
```
flaub: This is invalid c++

cnamrata15: Please ignore this.

plaidml/bridge/openvino/src/plaidml_plugin/ops/avg_pool.cpp
```diff
@@ -0,0 +1,54 @@
...
+    for (auto pad : layer->get_pads_begin()) {
+      padding.push_back(pad);
+    }
+    for (auto pad : layer->get_pads_end()) {
+      padding.push_back(pad);
+    }
+    result = op::pool(I, pool_type, kernel, strides, autopad_mode, padding, input_layout, include_padding_in_avg,
+                      use_ceil_for_output_shape);
+  } else {
+    result = op::pool(I, pool_type, kernel, strides, autopad_mode, padding, input_layout, include_padding_in_avg,
```
flaub, line 47: Since this is identical to the code inside the true case of the if, I think we should put the common code just after the if.

So:

```
if (autopad_mode == plaidml::op::AutoPadMode::EXPLICIT) {
  for (auto pad : layer->get_pads_begin()) {
    padding.push_back(pad);
  }
  for (auto pad : layer->get_pads_end()) {
    padding.push_back(pad);
  }
}

auto result = op::pool(I, pool_type, kernel, strides, autopad_mode, padding, input_layout, include_padding_in_avg,
                      use_ceil_for_output_shape);
```

cnamrata15: Oh, yes. That looks cleaner. Thanks.

PR #1148, opened 2020-06-19
add sqrt op

jd-bartlett96:
Added sqrt operation & test


jd-bartlett96:
Bad styling; redoing request

PR #1149, opened 2020-06-19
Add OpenVINO plugin ConvolutionBackpropData op

tzerrell:
Adds the ConvolutionBackpropData op to the PlaidML OpenVINO plugin. Includes several parts:

1. Actually adding the op
2. Reformatting the tests to make the linter happy
3. Extending the oplib convolution to allow it to infer the result size for ConvDerivMode::DATA. Add a new parameter to the oplib conv to enable this (IMO we don't want it on by default; it's forced to make assumptions about the possible size as multiple input data sizes correspond to the same output size and I don't think we should make such assumptions without an explicit request)
4. Add some fancy template metaprogramming code to the Convolution Fluent API to allow multiple integer vector types to be used as inputs.
5. Advances the OpenVINO reference

Sorry that these are bundled in one PR, ~but no part of this is tested without the other parts~ turns out some of the edsl cc_tests do test the fluent API changes, I could split that off if it's important but I'd prefer not to.

### DIFF
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -325,6 +325,42 @@ std::pair<TensorDim, TensorDim> compute_padding_and_output_size(  //
   throw std::runtime_error(llvm::formatv("Unexpected autopadding mode: {0}", to_string(autopad_mode)));
 }
 
+std::pair<TensorDim, TensorDim> compute_padding_and_input_size(  //
+    const TensorDim& output_size,                                //
+    const TensorDim& filter_size,                                //
+    int64_t stride,                                              //
+    AutoPadMode autopad_mode,                                    //
+    int64_t pad_lo,                                              //
+    int64_t pad_hi,                                              //
+    int64_t dilation,                                            //
+    int64_t data_dilation) {
+  // Note that this computes the smallest input_size that would produce the given output_size
+
+  if (data_dilation != 1) {
+    throw std::runtime_error("Cannot infer input data size for transposed convolution with data dilation");
+  }
+
+  auto O_eff = output_size;
+  auto F_eff = (dilation * (filter_size - 1)) + 1;  // Effective Filter Size
+  if (autopad_mode == AutoPadMode::NONE) {
+    TensorDim pad_before(pad_lo);
+    TensorDim input_size(O_eff * stride - pad_lo - pad_hi + F_eff - stride);
+    return std::pair<TensorDim, TensorDim>(pad_before, input_size);
+  }
+  if (autopad_mode == AutoPadMode::VALID) {
+    TensorDim pad_before(0);
+    TensorDim input_size(O_eff * stride + F_eff - stride);
+    return std::pair<TensorDim, TensorDim>(pad_before, input_size);
+  }
+  if (autopad_mode == AutoPadMode::SAME_LOWER || autopad_mode == AutoPadMode::SAME_UPPER) {
+    TensorDim input_size(O_eff * stride - stride + 1);
+    int64_t lower_term = (autopad_mode == AutoPadMode::SAME_LOWER) ? 1 : 0;
+    TensorDim pad_before((max(0, (O_eff - 1) * stride + F_eff - input_size) + lower_term) / 2);
+    return std::pair<TensorDim, TensorDim>(pad_before, output_size);
+  }
+  throw std::runtime_error(llvm::formatv("Unexpected autopadding mode: {0}", to_string(autopad_mode)));
+}
+
 std::vector<int64_t>* extend_manual_padding(std::vector<int64_t>* pads, size_t rank) {
   // TODO: Perhaps we should throw for sizes != 0, rank, 2*rank?
   if (pads->size() > 2 * rank) {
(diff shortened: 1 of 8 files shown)


plaidml/op/op.h
```diff
@@ -156,25 +156,54 @@ class convolution {
...
-    data_dilations_ = data_dilations;
+  template <typename T, typename std::enable_if<std::is_integral<T>::value>::type* = nullptr>
+  convolution& dilations(const std::vector<T>& dilations) {
+    dilations_ = edsl::make_tuple(dilations);
     return *this;
   }
 
-  convolution& filter_shape(const std::vector<int>& filter_shape) {
-    filter_shape_ = filter_shape;
+  template <typename T, typename std::enable_if<std::is_integral<T>::value>::type* = nullptr>
```
flaub, line 180: Cool!

tzerrell:
Turns out we're getting build failures due to errors like
```
error: no matching function for call to 'plaidml::op::convolution::strides(llvm::ArrayRef<int>&)'
```
It looks like with the old version, the compiler was fine implicitly casting an `ArrayRef<int>` to a `vector<int>`, but in the templated format, it can't figure out that it can cast `ArrayRef<int>` to a `vector<T>` with `T` == `int`.

We can't just make a version of the function taking an `ArrayRef` because we don't want to force anyone including `op.h` to include LLVM (I presume). It would work to make anyone passing in an `llvm::ArrayRef` explicitly cast it to a `std::vector`.

@flaub do you have an opinion on whether it's better to back out the templated convolution builder functions or better to require explicit casting from `llvm::ArrayRef` to `std::vector` when building a convolution?

flaub:
I thought there was a `to_vector` or something that ArrayRef or LLVM already implements.

flaub:
Try: `ArrayRef<T>::vec()`

flaub:
I don't see any ArrayRef in this code...

plaidml/op/lib/ops.cc
```diff
@@ -816,30 +855,33 @@ Value convolution(const Value& value) {
...
-    throw std::runtime_error("Convolution op expects 17 arguments");
+  if (args.size() != 18) {
+    throw std::runtime_error("Convolution op expects 18 arguments");
   }
   auto I_or_O = args[0].as_tensor();  // O if deriv_mode is DATA, else I
   auto F_or_O = args[1].as_tensor();  // O if deriv_mode is FILTER, else F
-  auto strides = args[2].as_int_tuple();
-  auto dilations = args[3].as_int_tuple();
-  auto data_dilations = args[4].as_int_tuple();
+  auto strides = args[2].is_none() ? std::vector<int64_t>{} : args[2].as_int_tuple();
```
flaub, line 867: Let's make a `Value::as_int_tuple_or_empty()`

tzerrell:
> I don't see any ArrayRef in this code...

There isn't an ArrayRef here. The problem is in https://github.com/plaidml/plaidml/blob/plaidml-v1/networks/oplib/resnet50.cc#L27 continuing through line 36. That code compiles without this PR and fails to compile with this PR, with an error that reads in part
```
ERROR: /home/tim/plaidml/networks/oplib/BUILD:7:1: C++ compilation of rule '//networks/oplib:oplib' failed (Exit 1)
networks/oplib/resnet50.cc: In function 'plaidml::edsl::Tensor networks::oplib::{anonymous}::block(const plaidml::edsl::Tensor&, llvm::ArrayRef<plaidml::edsl::Tensor>, llvm::ArrayRef<plaidml::edsl::Tensor>, llvm::ArrayRef<int>, bool, llvm::StringRef)':
networks/oplib/resnet50.cc:36:38: error: no matching function for call to 'plaidml::op::convolution::strides(llvm::ArrayRef<int>&)'
                      .strides(strides)
                                      ^
In file included from networks/oplib/resnet50.cc:9:
./plaidml/op/op.h:163:16: note: candidate: 'template<class T, typename std::enable_if<std::is_integral<_Tp>::value>::type* <anonymous> > plaidml::op::convolution& plaidml::op::convolution::strides(const std::vector<T>&)'
   convolution& strides(const std::vector<T>& strides) {
                ^~~~~~~
./plaidml/op/op.h:163:16: note:   template argument deduction/substitution failed:
networks/oplib/resnet50.cc:36:38: note:   'llvm::ArrayRef<int>' is not derived from 'const std::vector<T>'
                      .strides(strides)
                                      ^
```
ArrayRef does indeed have a conversion to std::vector: https://llvm.org/doxygen/classllvm_1_1ArrayRef.html#a745a0df36a633dcbc4acfd0bf9dffe17. I don't know why it was able to implicitly convert in the non-template form but unable to implicitly convert in the templated form; my knowledge of templates isn't very deep.

tzerrell:
I found a way to minimize the changes needed to use this code with ArrayRefs. We do need to explicitly give the type when using an ArrayRef, but if the code I added in the latest commit to `networks/oplib/resnet50.cc` looks reasonable then that works.

flaub:
LGTM, provided we figure out what the macos test failures are all about.

tzerrell:
> LGTM, provided we figure out what the macos test failures are all about.

macos failures went away when I rebased on to a more recent `plaidml-v1`

PR #1150, opened 2020-06-19
Added sqrt op

jd-bartlett96:
Added sqrt op & test, with correct styling


plaidml/bridge/openvino/tests/functional/plugin/plaidml/shared_tests_instances/single_layer_tests/sqrt.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include "single_layer_tests/sqrt.hpp"
+#include "common_test_utils/test_constants.hpp"
+#include <vector>
+
+using LayerTestsDefinitions::SqrtLayerTest;
+
+namespace {
+
+std::vector<std::vector<size_t>> inShapes = {
+    {2}, {1, 1, 1, 3}, {1, 2, 4}, {1, 4, 4}, {1, 4, 4, 1}, {1, 1, 1, 1, 1, 1, 3}, {1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1}};
```
tzerrell, line 14: It's better to have newlines between these, i.e. like
```
{2},
{1, 1, 1, 3},
```
By default the linter removes these newlines, but if you put `//` (i.e. an empty comment) at each newline, the linter will let you keep them

jd-bartlett96: Good to know! Had it as multi-lined originally but I kept the changes linter made. Should be all set now.

PR #1153, opened 2020-06-22
Add OpenVINO plugin grouped convolution

tzerrell:
This needs #1149 to land first


tzerrell:
Updated onto latest `plaidml-v1` with #1149 landed, ready for review.

PR #1154, opened 2020-06-22
slice op changes

dgkutnic:
* Handle scalar slice op
* C++ API: Add tuple overload for slice op


flaub:
Does this API work for you?

Also I added some tests (which we should try to get into the habit of when we modify the API)

dgkutnic:
Yes, this API works well. Thanks for making these changes.

PR #1158, opened 2020-06-23
Added sign op

jd-bartlett96:


plaidml/bridge/openvino/src/plaidml_plugin/ops/sign.cpp
```diff
@@ -0,0 +1,25 @@
...
+using namespace plaidml;          // NOLINT[build/namespaces]
+using namespace InferenceEngine;  // NOLINT[build/namespaces]
+
+namespace PlaidMLPlugin {
+
+static OpRegistration reg("sign", [](const Context& ctx) {
+  IE_ASSERT(ctx.operands.size() == 1);
+  auto I = ctx.operands.at(0);
+  auto Z = edsl::Tensor(0.0) * I;
+  auto O = Z + 1;
```
tzerrell, line 21: Does this work if you instead use
```
auto Z = edsl::Tensor(0.0);
auto O = edsl::Tensor(1.0);
```
?

I would expect that to autobroadcast correctly, and I think we should prefer to autobroadcast where possible, but I could be overlooking something.

jd-bartlett96: Yes, this change works. Done!

PR #1159, opened 2020-06-23
Added negative op

jd-bartlett96:


flaub:
It should, but there's a bug in the lowering (or perhaps we just haven't added a lowering for `NegOp` yet). Would be great to get that fixed/implemented :)

jd-bartlett96:
-I appears to work fine (not sure why I didn't start with that)

tzerrell:
> It should, but there's a bug in the lowering (or perhaps we just haven't added a lowering for `NegOp` yet). Would be great to get that fixed/implemented :)

I believe the bug with `NegOp` only exists for integer types, and this only tests floats.

PR #1160, opened 2020-06-23
Add an LLVM interpreter backend to the executable API

flaub:
Use `LLVM_JIT=Interpreter` env var to activate. 

It currently crashes early on during the wrapper that deals with user-supplied I/O buffers.


flaub:
This is based on #1215

PR #1161, opened 2020-06-23
Add OpenVINO ops DepthToSpace & SpaceToDepth

tzerrell:
Note that these have somewhat different implementations, one using a more complex contraction, one using a simpler contraction followed by a reshape; this works around a bug that was showing up in defractionalization.

~Can't land until https://github.com/plaidml/openvino/pull/14 is in.~

### DIFF
--- a/plaidml/bridge/openvino/src/plaidml_plugin/ops/depth_to_space.cpp
+++ b/plaidml/bridge/openvino/src/plaidml_plugin/ops/depth_to_space.cpp
@@ -11,7 +11,6 @@
 #include "ngraph/opsets/opset2.hpp"
 
 #include "plaidml/op/op.h"
-#include "pmlc/util/logging.h"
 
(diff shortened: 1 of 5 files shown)


plaidml/bridge/openvino/src/plaidml_plugin/ops/depth_to_space.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <math.h>
+
+#include "plaidml_ops.hpp"
+#include "plaidml_util.hpp"
+
+#include "ngraph/opsets/opset.hpp"
+#include "ngraph/opsets/opset2.hpp"
+
+#include "plaidml/op/op.h"
+#include "pmlc/util/logging.h"
```
cnamrata15, line 14: Should we remove this?

PR #1162, opened 2020-06-24
WIP OpenVino plugin tile, squeeze

1sanz:
- tile: unregistered op error

### DIFF
--- a/plaidml/bridge/openvino/src/plaidml_plugin/ops/tile.cpp
+++ b/plaidml/bridge/openvino/src/plaidml_plugin/ops/tile.cpp
@@ -1,0 +1,27 @@
+// Copyright (C) 2020 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "plaidml_ops.hpp"
+#include "plaidml_util.hpp"
+
+#include "ngraph/opsets/opset.hpp"
+#include "ngraph/opsets/opset1.hpp"
+
+#include "plaidml/op/op.h"
+
+using namespace plaidml;          // NOLINT[build/namespaces]
+using namespace InferenceEngine;  // NOLINT[build/namespaces]
+
+namespace PlaidMLPlugin {
+
+static OpRegistration reg("tile", [](const Context& ctx) {
+  IE_ASSERT(ctx.operands.size() == 2);
+  auto I = ctx.operands.at(0);
+  auto reps = get_shape_from_constant_operand(1, ctx.layer);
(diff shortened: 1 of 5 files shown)


plaidml/bridge/openvino/tests/functional/plugin/plaidml/shared_tests_instances/single_layer_tests/tile.cpp
```diff
@@ -0,0 +1,24 @@
...
+using LayerTestsDefinitions::TileLayerTest;
+
+namespace {
+
+const std::vector<InferenceEngine::Precision> netPrecisions = {
+    InferenceEngine::Precision::FP32,
+};
+
+INSTANTIATE_TEST_CASE_P(TileCheck, TileLayerTest,
+                        ::testing::Combine(::testing::Values(std::vector<int64_t>({1})),         //
```
tzerrell, line 19: If the tiling factor has only ones, this is a no-op; use something larger here, like {3} or {2, 3}

tzerrell:
There appears to be an error in the tests for Tile unrelated to our changes. It's an nGraph interpreter error, which is what the testing compares against; Tile is a fused op, and nGraph expects it to be decomposed by the time it reaches the interpreter, but for some reason this isn't happening. This isn't a problem with other fused ops, so I'm not sure what's wrong here.

I propose we land this anyway and just have a known error in the test, since the problem doesn't really seem like it's on ouur side.

plaidml/bridge/openvino/src/plaidml_plugin/ops/tile.cpp
```diff
@@ -0,0 +1,27 @@
...
+using namespace InferenceEngine;  // NOLINT[build/namespaces]
+
+namespace PlaidMLPlugin {
+
+static OpRegistration reg("tile", [](const Context& ctx) {
+  IE_ASSERT(ctx.operands.size() == 2);
+  auto I = ctx.operands.at(0);
+  auto reps = get_shape_from_constant_operand(1, ctx.layer);
+  std::vector<int> reps_ints;
+  reps_ints.assign(reps.begin(), reps.end());
```
tzerrell, line 23: The `reps_ints` is here because you need the vector to use a different type of integer than the one returned from `get_shape_from_constant_operand`, correct? We should upgrade the oplib to be nicer about this at some point, but no need to do that in this PR.

1sanz: yes.

PR #1163, opened 2020-06-24
Adding selu op

jd-bartlett96:
Testing incomplete; need fix on OV side

### DIFF
--- a/plaidml/bridge/openvino/src/plaidml_plugin/ops/selu.cpp
+++ b/plaidml/bridge/openvino/src/plaidml_plugin/ops/selu.cpp
@@ -1,0 +1,26 @@
+// Copyright (C) 2020 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "plaidml_ops.hpp"
+
+#include "ngraph/opsets/opset.hpp"
+#include "ngraph/opsets/opset1.hpp"
+
+#include "plaidml/op/op.h"
+
+using namespace plaidml;          // NOLINT[build/namespaces]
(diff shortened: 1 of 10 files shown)


plaidml/bridge/openvino/src/plaidml_plugin/ops/selu.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+namespace PlaidMLPlugin {
+
+static OpRegistration reg("selu", [](const Context& ctx) {
+  IE_ASSERT(ctx.operands.size() == 3);
+  auto I = ctx.operands.at(0);
+  auto alpha = ctx.operands.at(1);
+  auto lambda = ctx.operands.at(2);
+
+  return edsl::make_tuple(lambda * edsl::select(I > 0, I, alpha * (edsl::exp(I) - 1)));
```
tzerrell, line 23: I think you could replace the entire `edsl::select(...)` with `op::elu(I, alpha)`

jd-bartlett96: It looks like `op::elu` is expecting alpha to be a float instead of a Tensor (and does not automatically cast the 1d Tensor to a float)...is it worth fixing this (adding elu functionality to accept alpha as a Tensor)?

jd-bartlett96: (The tests all pass as it currently stands)

tzerrell: That would be a nice upgrade, but it's a separate issue; IMO leave it as is for now.

PR #1165, opened 2020-06-25
Add OpenVINO op StridedSlice

tzerrell:
@dgkutnic and @flaub I've included you as reviewers just so you can see the small extension to the oplib Slice API that's in this PR (there are unique semantics from passing `None`s to Slice which I needed, so this PR adds a way to do so).

### DIFF
--- a/plaidml/bridge/openvino/src/plaidml_plugin/ops/strided_slice.cpp
+++ b/plaidml/bridge/openvino/src/plaidml_plugin/ops/strided_slice.cpp
@@ -1,0 +1,106 @@
+// Copyright (C) 2020 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "plaidml_ops.hpp"
+#include "plaidml_util.hpp"
+
+#include "ngraph/opsets/opset.hpp"
+#include "ngraph/opsets/opset1.hpp"
+
+#include "plaidml/op/op.h"
+#include "pmlc/util/logging.h"
+
+using namespace plaidml;          // NOLINT[build/namespaces]
+using namespace InferenceEngine;  // NOLINT[build/namespaces]
+
+namespace {
+
+template <typename T>
+std::vector<T> cast_constant_operand(size_t operand_idx, ngraph::Node* layer) {
+  auto ngraph_const =
+      std::dynamic_pointer_cast<ngraph::op::Constant>(layer->input_value(operand_idx).get_node_shared_ptr());
+  if (ngraph_const) {
+    return ngraph_const->cast_vector<T>();
+  } else {
+    THROW_IE_EXCEPTION << "Dynamic slicing not currently supported by PlaidML plugin; all of begin, end, and stride "
+                          "must be Constants.";
+  }
+}
+
+}  // namespace
+
+namespace PlaidMLPlugin {
+
+static OpRegistration reg("stridedslice", [](const Context& ctx) {
+  auto* layer = dynamic_cast<ngraph::opset1::StridedSlice*>(ctx.layer);
+  IE_ASSERT(ctx.operands.size() <= 4);
+  IE_ASSERT(ctx.operands.size() >= 3);
+  auto I = ctx.operands.at(0);
+  auto starts = cast_constant_operand<int64_t>(1, layer);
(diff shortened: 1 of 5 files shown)


tzerrell:
(Marked as draft because this needs https://github.com/plaidml/openvino/pull/18 to work correctly)

plaidml/bridge/openvino/src/plaidml_plugin/ops/strided_slice.cpp
```diff
@@ -0,0 +1,106 @@
...
+    end_mask.push_back(0);
+  }
+  while (shrink_axis_mask.size() < I.rank()) {
+    shrink_axis_mask.push_back(0);
+  }
+
+  IVLOG(1, "Starts: " << starts << ", stops: " << stops << ", steps: " << steps);
+  for (size_t i = 0; i < I.rank(); i++) {
+    if (i >= starts.size()) {
+      result.add_dim(edsl::None(), edsl::None(), edsl::None());
```
dgkutnic, line 86: This slice is essentially a no-op, right? Perhaps I'm missing something here - is it possible/easy to call `add_dim` with `I_dims[i]` as the argument?

flaub: I suggest we add a `add_dim()` that takes no arguments, which then internally sets the `start`, `stop`, and `step` to none. I'd rather not expose `Value` as arguments to the API, since it's much easier to make a mistake that way (Value basically removes static type checking via type erasure).

tzerrell: That's a reasonable addition for this case, but it doesn't address cases where only some of the arguments are None; see for example lines 98-100:
```
    result.add_dim(begin_mask[i] ? edsl::None() : edsl::Value(starts[i]),  //
                   end_mask[i] ? edsl::None() : edsl::Value(stops[i]),     //
                   edsl::Value(steps[i]));
```
Fundamentally, the arguments are an int-or-none type, not an int type. This exactly mirrors what happens with a NumPy slice op -- we intentionally use the same syntax. NumPy slicing has semantics for omitted `start` and `stop` values ("go all the way") that is not directly duplicated by any integer value: `::-1` is different from `3::-1` is different from `:0:-1` is different from `3:0:-1`. If you have the size of the dimension you're slicing (i.e. you've called compute_shape), it's generally possible to replace the Nones with integers based on the size, e.g. `:0:-1` can be rewritten as `N-1:0:-1`. But adding a call to compute_shape seems very expensive, and the logic here is complex and not always intuitive (`::-1`, for example, isn't `N-1:0:-1` (that excludes `0`), it's not `N-1:-1:-1` (`-1` means the last element, this is an empty slice), it's `N-1:-N-1:-1`) -- that's complexity I don't want to push past our API to the user. We need some way to represent passing a "there's no int in this position" argument.

Using a `Value` is one way to accomplish this, and it matches how the oplib currently works under the hood (see e.g. https://github.com/plaidml/plaidml/blob/tzerrell-ov-slice/plaidml/op/lib/ops.cc#L2300, https://github.com/plaidml/plaidml/blob/tzerrell-ov-slice/plaidml/op/lib/ops.cc#L2314, https://github.com/plaidml/plaidml/blob/tzerrell-ov-slice/plaidml/op/lib/ops.cc#L2335). There are other options; I'm not yet sold on any of the ones I've come up with but they at least seem plausibly reasonable:
* We could create an int-or-none type distinct from Value
* We could restrict the start/stop/step arguments to integers, and have separate "mask" arguments for where a None is used instead. This is the approach OpenVINO uses, and perhaps I'm just used to NumPy but I personally find it unintuitive.
* We could have a separate `add_dim` function for each pattern of None locations. Since `step` doesn't need to take `None` (it's equivalent to `1`), there would need to be 4 of these functions: add_dim_ints, add_dim_start_none, add_dim_stop_none, add_dim_two_nones. Hopefully we could come up with better names.

Do these changes need to go into this PR? They seem fairly substantial and fairly decoupled from the core thrust of this PR.

flaub: Thanks for the detailed rationale. I think this is a fine solution for now (using Value in the API).

PR #1167, opened 2020-06-25
OpenVino plugin reduce logical and, or

1sanz:


plaidml/bridge/openvino/src/plaidml_plugin/ops/reduce_logical_or.cpp
```diff
@@ -0,0 +1,26 @@
...
+using namespace plaidml;          // NOLINT[build/namespaces]
+using namespace InferenceEngine;  // NOLINT[build/namespaces]
+
+namespace PlaidMLPlugin {
+
+static OpRegistration reg("reducelogicalor", [](const Context& ctx) {
+  IE_ASSERT(ctx.operands.size() == 2);
+  auto I = ctx.operands.at(0);
+  std::vector<size_t> axes = get_axis_vector_from_constant_operand(1, ctx.layer);
+  auto* layer = dynamic_cast<ngraph::opset1::ReduceLogicalAnd*>(ctx.layer);
```
flaub, line 22: Is this right? Should it be `ReduceLogicalOr`?

tzerrell:
Note: These are untested, as the OpenVINO test infrastructure in 2020.2 does not allow tests with boolean inputs and nGraph throws if these ops are given non-boolean inputs.

plaidml/bridge/openvino/src/plaidml_plugin/ops/reduce_logical_or.cpp
tzerrell: It should be `ReduceLogicalOr`

cnamrata15: Oops, my bad.

1sanz: thanks @flaub I'll put in a PR to fix this

PR #1169, opened 2020-06-25
implement logical and & or (&&, ||) for C++ edsl

mars1026:


flaub:
Looking good, could you add some `.mlir` tests for the lowering?

flaub:
We should also consider what it would take to support python. Can we just overload operators like in C++?

flaub:
Also, I only see tests for integer types, might be good to have a floating point test as well as a heterogenous one.

pmlc/conversion/tile_to_pxa/tests/eltwise_logical.mlir
```diff
@@ -0,0 +1,66 @@
+// RUN: pmlc-opt -convert-tile-to-pxa -canonicalize -cse %s | FileCheck %s
+
+func @eltwise_and_f32(
+  %arg0: tensor<3x3xf32>,
+  %arg1: tensor<3x3xf32>
+) -> tensor<3x3xf32> {
+  %0 = "eltwise.logical_and"(%arg1, %arg0) : (
+    tensor<3x3xf32>,
+    tensor<3x3xf32>
+  ) -> tensor<3x3xf32>
```
flaub, line 10: I don't think we should have the logical ops ever return anything other than `i1`

flaub: Actually this doesn't seem legal, the definition of `EW_BinaryBoolOp` says that the result should be `EltwiseBool`. Perhaps the `EltwiseBool` predicate isn't right?

pmlc/dialect/eltwise/ir/ops.td
```diff
@@ -144,6 +144,12 @@ class EW_BinaryIntOp<string mnemonic, list<OpTrait> traits = [NoSideEffect]> :
     Arguments<(ins EltwiseInteger:$lhs, EltwiseInteger:$rhs)>,
     Results<(outs EltwiseInteger:$result)>;
 
+class EW_BinaryBoolOp<string mnemonic, list<OpTrait> traits = [NoSideEffect]> :
+    EW_BinaryOp<mnemonic, traits>,
+    Arguments<(ins EltwiseBool:$lhs, EltwiseBool:$rhs)>,
```
flaub, line 149: I think we want to allow any type here?

pmlc/dialect/eltwise/ir/ops.td
```diff
@@ -144,6 +144,25 @@ class EW_BinaryIntOp<string mnemonic, list<OpTrait> traits = [NoSideEffect]> :
     Arguments<(ins EltwiseInteger:$lhs, EltwiseInteger:$rhs)>,
     Results<(outs EltwiseInteger:$result)>;
 
+class EW_BinaryBoolOp<string mnemonic, list<OpTrait> traits = [NoSideEffect]> :
```
flaub, line 147: Let's drop this if nothing uses it

PR #1173, opened 2020-06-25
WIP OpenVino plugin one_hot

1sanz:


tzerrell:
Context: this is failing tests because we do not track name changes that appear to arise from OV optimizations

flaub:
Is this PR still valid?

tzerrell:
This is not valid to land as-is due to API changes and the move into the OV fork repo. It contains code that could still be useful for re-developing this op. Note that the mentioned error in my above comment is now documented on the OV fork as https://github.com/plaidml/openvino/issues/88

Closing as this is no longer the right target repo.

tzerrell:
Oh, so followup: Since the code is still useful for the other repo, please don't delete the branch.

tzerrell:
ah, mergify does that automatically? Fun :/

PR #1174, opened 2020-06-25
Add OpenVINO ops Convert, ConvertLike

tzerrell:
Depends on https://github.com/plaidml/openvino/pull/21 (which has now landed)


tzerrell:
Note: ConvertLike tests fail due to the same problem mentioned here: https://github.com/plaidml/plaidml/pull/1162#issuecomment-649032170

Issue #1131, opened 2020-06-17
Cannot legalize operation 'eltwise.neg'


LiyangLingIntel:
When I'm trying to get a bitwise not of a Tensor like this
```cpp
std::vector<int64_t> I_shape{1, 3, 16, 16};
auto I = Placeholder(DType::INT16, I_shape);
auto O = -I;
auto program = ProgramBuilder("tensor_neg", {O}).intx(DType::INT16).compile();
```
It gets compile error:

error: failed to legalize operation 'eltwise.neg'
terminate called after throwing an instance of 'std::runtime_error'
  what():  conversion to the LLVM IR dialect failed

Instead, We have to use `auto O = 0 - I;` , which could get the same result.

LiyangLingIntel:
It seems operation 'eltwise.neg' works for FLOAT DType, but goes wrong with INT DType, as there is a  NegFOp after tile lowering, but no NegIOp.

Issue #1130, opened 2020-06-17
Cannot legalize operation 'eltwise.bit_not'


LiyangLingIntel:
When I'm trying to get a bitwise not of a Tensor like this
```cpp
std::vector<int64_t> I_shape{1, 3, 16, 16};
auto I = Placeholder(DType::FLOAT16, I_shape);
auto O = ~I;
auto program = ProgramBuilder("tensor_round", {O}).floatx(DType::FLOAT16).compile();
```
It gets compile error:

error: failed to legalize operation 'eltwise.bit_not'
terminate called after throwing an instance of 'std::runtime_error'
  what():  conversion to the LLVM IR dialect failed
 
It seems it's not well supported.

PR #1183, opened 2020-06-26
Support for neg and bit_not ops

flaub:
Fix for #1130 
Fix for #1131


pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -336,6 +336,27 @@ static Type promoteTypes(ConversionPatternRewriter &rewriter, Location loc,
...
+};
+
+struct NotOp {
+  Value create(ConversionPatternRewriter &rewriter, Location loc,
+               Type resultType, ArrayRef<Value> operands,
+               ArrayRef<Type> types) {
+    // -(x + 1) = -1 - x
+    auto negOne = rewriter.create<mlir::ConstantIntOp>(loc, -1, resultType);
+    auto sub = rewriter.create<mlir::SubIOp>(loc, negOne, operands[0]);
+    return sub.getResult();
```
tzerrell, line 356: We know the signed ints will always be represented with 2's complement?

I guess there would be ambiguity around signedness with the int ops otherwise...

flaub: AFAIK all machines seem to use 2's complement. I think LLVM is based on that assumption. Also, there doesn't seem to be a `std.negi` or ` std.bit_not`. It looks like LLVM does have intrinsics that we could plumb thru at some point.

PR #1187, opened 2020-06-26
Openvino ops : ceil, greater, greater_eq, tan

cnamrata15:
Tan has been implemented using Cos as a temporary workaround.
Sin fixes using Cos for tests to work.

**Note**
Tests on the comparison ops fail with the error " C++ exception with description "Output 0 type 'f32' does not match Result type 'boolean'" thrown in the test body." This error is on the ngraph interpreter side.


plaidml/bridge/openvino/src/plaidml_plugin/ops/sin.cpp
```diff
@@ -17,7 +19,7 @@ namespace PlaidMLPlugin {
 static OpRegistration reg("sin", [](const Context& ctx) {
   IE_ASSERT(ctx.operands.size() == 1);
   auto I = ctx.operands.at(0);
-  return edsl::make_tuple(sin(I));
+  return edsl::make_tuple(cos(I - M_PI / 2));
```
tzerrell, line 22: Copy your workaround comment from `tan` to here

PR #1197, opened 2020-06-30
OpenVino plugin NormalizeL2, GRN

1sanz:
GRN: 4D Tensor input test passes. tests for 3D and 2D tensors failed. 
NormalizeL2: corner case (empty axis) is untested because of exception in ngraph 

These tests have been removed for now.


plaidml/bridge/openvino/src/plaidml_plugin/ops/normalize_l2.cpp
```diff
@@ -20,13 +20,15 @@ static OpRegistration reg("normalizel2", [](const Context& ctx) {
...
   std::vector<size_t> axes = get_axis_vector_from_constant_operand(1, ctx.layer);
   auto* layer = dynamic_cast<ngraph::opset1::NormalizeL2*>(ctx.layer);
-  // check if axes is empty
+  auto eps = layer->get_eps();
+  auto eps_mode = layer->get_eps_mode();
+
   if (axes.empty()) {
-    return edsl::make_tuple(I / I);
+    auto norm = I + eps;
+    return edsl::make_tuple(I / norm);
```
tzerrell, line 28: It's unclear to me if this is the desired behavior; it may be intended to be a broadcast 1 instead. Once the testing infra supports this we should re-examine it. For now it's fine.

1sanz: I agree.

tzerrell:
The GRN op only passes tests for rank 4 tensors (and a few very specific shapes of rank 3 and rank 2 tensors). We found that it was possible to get the op to pass tests in all cases by summing over axis 0 in the rank 3 case and not summing over any axis in the rank 2 case. This does not match the documentation and we believe it is incorrect behavior, so we stayed with this version which always sums over axis 1. However, it would be straightforward to change to the tested behavior if it is indeed correct.

PR #1200, opened 2020-06-30
implement eltwise.logical_not and add operator! to C++ EDSL

mars1026:


pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -458,6 +458,27 @@ struct LogicalOp {
   }
 };
 
+struct LogicalNotOp {
+  Value create(ConversionPatternRewriter &rewriter, Location loc,
+               Type resultType, ArrayRef<Value> operands,
+               ArrayRef<Type> types) {
+    auto &input = operands[0];
```
flaub, line 465: `Value` types should always be treated with value-like semantics (since it's implemented as a smart-pointer). No need to take a reference here.

flaub:
Other than the minor taking a `Value` by ref, LGTM!

pmlc/conversion/tile_to_pxa/tests/logical.mlir
```diff
@@ -60,4 +60,34 @@ func @eltwise_xor_mixed(
...
+  %arg0: tensor<3x3xf32>
+) -> tensor<3x3xi1> {
+  %0 = "eltwise.logical_not"(%arg0) : (
+    tensor<3x3xf32>
+  ) -> tensor<3x3xi1>
+  return %0 : tensor<3x3xi1>
+}
+
+// CHECK-LABEL: func @eltwise_not_f32
+// CHECK: constant 0 : f32
```
flaub, line 89: You could make this a fuzzy check like:
```
// CHECK: constant 0{{.*}} : f32
```

flaub:
Resolves PLA-6

PR #1201, opened 2020-06-30
Openvino : Logical Ops

cnamrata15:
Tests fail as inputs need to be boolean.


tzerrell:
Hmm, looks like we do have an `xor` op at the eltwise dialect level, maybe it's not plumbed through at the edsl level? Probably worth looking into in a followup

PR #1203, opened 2020-07-01
add missing lowering from eltwise.sin to MLIR SinOp

mars1026:


plaidml/edsl/tests/edsl_test.cc
```diff
@@ -1190,6 +1190,32 @@ TEST_F(CppEdsl, Cos) {
   checkProgram(program, {{S, A_input}}, {{O, C_output}});
 }
 
+TEST_F(CppEdsl, Sin) {
+  auto S = Placeholder(DType::FLOAT32, {3, 3});
+  TensorDim I, J;
+  TensorIndex i("i"), j("j");
+  S.bind_dims(I, J);
```
flaub, line 1197: drop these, they're unused.

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -1190,6 +1190,32 @@ TEST_F(CppEdsl, Cos) {
...
+  auto O = sin(S);
+  auto program = makeProgram("sin", {O});
+  std::cout << program << std::endl;
```
flaub, line 1200: `makeProgram` already prints the program to `stdout`

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -1190,6 +1190,32 @@ TEST_F(CppEdsl, Cos) {
...
+  // clang-format off
+  // CHECK-LABEL: CppEdsl.Sin
+  // clang-format on
```
flaub, line 1203: Not sure what value we have with this if we aren't going to check the actual code.

PR #1205, opened 2020-07-01
Updating oplib testing to lit

jd-bartlett96:
Modified necessary files to run lit testing on oplib, including a cpplint config file to ignore line length limits.

Implemented lit tests for Relu, Abs, and All

### DIFF
--- a/plaidml/op/BUILD
+++ b/plaidml/op/BUILD
@@ -3,6 +3,7 @@
 load("@rules_python//python:defs.bzl", "py_test")
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
 load("//bzl:plaidml.bzl", "plaidml_cc_library", "plaidml_cc_test")
+load("//pmlc:lit.bzl", "lit_test")
 
 SDK_HDRS = [
     "ffi.h",
@@ -63,8 +64,15 @@ py_library(
 plaidml_cc_test(
     name = "cc_test",
     srcs = ["op_test.cc"],
+    tags = ["manual"],  # the `edsl_test.cc.test` rule will run this test via lit
     deps = [
-        ":op",
         "//plaidml:testenv",
+        "//plaidml/edsl",
+        "//plaidml/exec",
     ],
+)
+
+lit_test(
+    name = "op_test.cc",
+    data = [":cc_test"],
 )
--- a/plaidml/op/CPPLINT.cfg
+++ b/plaidml/op/CPPLINT.cfg
@@ -1,0 +1,1 @@
+filter=-whitespace/line_length
--- a/plaidml/op/op_test.cc
+++ b/plaidml/op/op_test.cc
@@ -1,4 +1,5 @@
 // Copyright 2019 Intel Corporation.
+// RUN: cc_test | FileCheck %s
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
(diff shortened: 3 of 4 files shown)


plaidml/op/BUILD
```diff
@@ -63,8 +64,15 @@ py_library(
 plaidml_cc_test(
     name = "cc_test",
     srcs = ["op_test.cc"],
+    tags = ["manual"],  # the `edsl_test.cc.test` rule will run this test via lit
```
tzerrell, line 67: If we keep the non-lit aspects of the tests (e.g. the `runProgram(program);` lines) we should not mark this `manual`, as both tests will be needed.

plaidml/op/CPPLINT.cfg
```diff
@@ -0,0 +1 @@
+filter=-whitespace/line_length
```
tzerrell, line 1: Hmm... Does FileCheck have troubles with NOLINT directives? (Meaning: append `// NOLINT[whitespace/line_length]` to lines that are too long). If that works, it's preferable to turning off that part of the linter for all of `plaidml/op`.

plaidml/op/op_test.cc
```diff
@@ -32,48 +33,42 @@ Program makeProgram(const std::string& name, const std::vector<Tensor>& outputs)
...
+  std::cout << program << std::endl;
+
+  // clang-format off
+  // CHECK-LABEL: OpTest.All
+  // CHECK: func @all
+  // CHECK: %[[cst1:.*]] = "eltwise.sconst"() {value = 1 : i64} : () -> tensor<si32>
+  // CHECK: %[[cst0:.*]] = "eltwise.sconst"() {value = 0 : i64} : () -> tensor<si32>
+  // CHECK: %{{.*}} = "eltwise.cmp_eq"(%{{.*}}, %[[cst0]]) : (tensor<1x224x224x3xf32>, tensor<si32>) -> tensor<1x224x224x3xi1>
+  // CHECK: %{{.*}} = "eltwise.select"(%{{.*}}, %[[cst0]], %[[cst1]]) : (tensor<1x224x224x3xi1>, tensor<si32>, tensor<si32>) -> tensor<1x224x224x3xsi32>
+  // CHECK: %{{.*}} = tile.contract mul, none, %[[cst1]], %{{.*}} {sink = #map0, srcs = [#map1]} : tensor<si32>, tensor<1x224x224x3xsi32> -> tensor<si32>
```
tzerrell, line 66: Don't check for specific map names, which could easily change; just `sink = #{{.*}}, srcs = [#{{.*}}]`. You can name them like you do with `cst0` if you think it's important (but I don't think it is -- certainly it's not in this case where you don't include the definitions of the maps or reuse them).

plaidml/op/CPPLINT.cfg
jd-bartlett96: I'll try that out; I'd just been following what had been done in the edsl testing

plaidml/op/BUILD
jd-bartlett96: So just remove this line? Or should it be changed to something else?

flaub: No, it needs to be marked as manual, because lit will run it already.

plaidml/op/CPPLINT.cfg
flaub: Yep, this is correct for lit tests

plaidml/op/BUILD
tzerrell: thanks for the correction!

plaidml/op/op_test.cc
```diff
@@ -32,48 +33,42 @@ Program makeProgram(const std::string& name, const std::vector<Tensor>& outputs)
...
-TEST(Op, All) {
+TEST_F(OpTest, All) {
   auto I = Placeholder(DType::FLOAT32, {1, 224, 224, 3}, "I");
   auto program = makeProgram("all", {op::all(I)});
-  IVLOG(1, program);
-  EXPECT_THAT(program, Eq(R"#(
-#map0 = affine_map<() -> ()>
-#map1 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
-
+  std::cout << program << std::endl;
```
flaub, line 57: `makeProgram` should already print this to stdout

plaidml/op/op_test.cc
```diff
@@ -615,26 +610,19 @@ module {
...
-    %4 = "eltwise.cmp_lt"(%3, %arg0) : (tensor<10x20xf32>, tensor<10x20xf32>) -> tensor<10x20xi1>
-    %5 = "eltwise.select"(%4, %3, %arg0) : (tensor<10x20xi1>, tensor<10x20xf32>, tensor<10x20xf32>) -> tensor<10x20xf32>
-    return %5 : tensor<10x20xf32>
-  }
-}
-)#"));
+TEST_F(OpTest, Relu) {
+  auto A = Placeholder(DType::FLOAT32, {10, 20});
+  auto program = makeProgram("relu", {op::relu(A)});
+  std::cout << program << std::endl;
```
flaub, line 616: drop all of these

plaidml/op/op_test.cc
jd-bartlett96: I assume this means I should insert the `std::cout << program << std::endl;` into `makeProgram` since it's not there yet (unlike the `makeProgram` in the edsl test)

flaub: Oh, sorry, I didn't realize it was copied. Actually it should live in `testenv.h` and both the EDSL and Op tests should use a common function.

jd-bartlett96:
Closing due to ambiguity of how to make an appropriately flexible lit test for oplib ops (many of which could be implemented in different ways) while still having a meaningful test.

PR #1206, opened 2020-07-01
implement nine remaining eltwise numeric functions in C++ EDSL

mars1026:
acos, atan, cosh, erf, floor, pow, round, sinh, tan


pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
```diff
@@ -97,23 +101,78 @@ struct FPToUILowering : public LLVMLegalizationPattern<stdx::FPToUIOp> {
   }
 };
 
-struct ASinLowering : public LLVMLegalizationPattern<stdx::ASinOp> {
-  using LLVMLegalizationPattern<stdx::ASinOp>::LLVMLegalizationPattern;
-  using Base = LLVMLegalizationPattern<stdx::ASinOp>;
+template <typename T>
+struct LibMCallLowering : public LLVMLegalizationPattern<T> {
```
flaub, line 105: Nice

flaub:
I think the lowerings only work for `f32`, what happens if you try `f16` or `f64`? 

We have a few options:

1. Allow the lowering class to specify different runtime calls based on the type: (i.e. `f32 -> sinf` & `f64 -> sin`)
2. Modify the allowed types at the eltwise dialect level. So if we don't have `f16` versions for a function, then we need to make sure the type system doesn't allow it to be specified.

pmlc/dialect/stdx/ir/ops.td
```diff
@@ -46,13 +46,58 @@ def ReshapeOp : StdX_Plain_Op<"reshape"> {
...
+def ATanOp : StdX_UnaryFuncOp<"atan"> {
+  let summary = "arctangent";
+}
+
+def CosHOp : StdX_UnaryFuncOp<"cosh"> {
+  let summary = "hyperbolic cosine";
+}
+
+def ErfOp : StdX_UnaryFuncOp<"erf"> {
+  let summary = "error function";
```
tzerrell, line 75: Not critical, but for extreme clarity perhaps call this the `Gauss error function`

mars1026: I like it

mars1026:
> I think the lowerings only work for `f32`, what happens if you try `f16` or `f64`?

Yes, that's right; with f16 or f64, LLVM aborts with the message "Calling a function with a bad signature!".

flaub:
OK, reminder that the ultimate goal of these OpenVINO ops is to target HW with only fp16.

flaub:
Also, don't we need changes to edsl.h to expose some of these new ops? (erf comes to mind)

tzerrell:
I'm working on confirming right now by re-enabling tests, but I modified edsl.h in anticipation of these ops in https://github.com/plaidml/plaidml/pull/1189

flaub:
Thanks! Didn't see that one, LGTM.

PR #1207, opened 2020-07-02
add norm to oplib 

1sanz:
@tzerrell is this what you had in mind?


plaidml/op/lib/ops.cc
```diff
@@ -1917,6 +1918,28 @@ Value minimum(const Value& value) {
...
+    throw std::runtime_error("norm expects 4 arguments");
+  }
+
+  auto I = args[0].as_tensor();
+  auto axis = args[1].as_int();
+  auto eps = args[2].as_float();
+  auto eps_mode = validate<EpsMode>(args[3].as_int());
+
+  auto X = op::sum((I * I), edsl::make_tuple(axis), 1);
+  if (eps_mode == EpsMode::ADD) {
```
flaub, line 1934: Use a `switch` with enums.

plaidml/op/lib/ops.cc
```diff
@@ -40,6 +40,7 @@ Value maximum(const Value&);
 Value mean(const Value&);
 Value min(const Value&);
 Value minimum(const Value&);
+Value norm(const Value&);
```
tzerrell, line 43: Rename `norm` -> `l2norm`; there are too many "Norm"s in ML.

plaidml/op/op.h
```diff
@@ -437,6 +443,11 @@ inline edsl::Tensor minimum(const edsl::Tensor& X, const edsl::Tensor& Y) {
   return details::op("minimum", args).as_tensor();
 }
 
+inline edsl::Tensor norm(const edsl::Tensor& I, int axis, float eps, EpsMode eps_mode) {
```
tzerrell, line 446: Not critical but I think `epsilon` would be a clearer variable name than `eps`. EpsMode can probably stay as-is (it's clear from context when there's a variable named `epsilon`). Also it's probably sufficient to change it at the API level (i.e. here in op.h); in the context of the implementation it's pretty clear what `eps` is.

This matters in part because an epsilon term is only needed when this is used in the denominator, and this code covers more cases than that.

1sanz:
@tzerrell regarding the Fluent API, could you please take a look at the defaults? Should default axis be 0? Or should there be no default for axis? (forgot to change axis to tuple - will do)

PR #1210, opened 2020-07-02
WIP: Updating oplib testing

jd-bartlett96:
- Added unsqueeze to oplib
- When testing unsqueeze, noticed EXPECT_THAT testing statements in oplib testing were not being used correctly
- After discussion with @tzerrell and @flaub, decided to remove EXPECT_THAT tests due to brittleness (rather than fixing the statements)
- Added a "runProgram" statement to each op test as a minimal test of functionality, unless a more detailed "checkProgram" test was already used (long term goal is to write comprehensive checkProgram tests for all ops)
- Wrote a "checkProgram" test for squeeze & unsqueeze
- Moved makeProgram function to testenv.h so it can be shared by oplib and edsl tests

Issues: the BroadcastNumpy and Convolution tests do not currently pass. Still have to investigate broadcastnumpy (seems like probably a smaller fix) but convolution is causing a segfault before the operation is even run; have not been able to find root of that issue.


tzerrell:
Incomplete analysis: Running the convolution test with ASan reports a stack overflow in XSMM.

tzerrell:
It may be an off-by-one error in the padding; I tried to reproduce in backend_test, and same padding there used only 2 padding, not the 3 I was expecting. Changing this test to 2 makes it pass. That said ... we shouldn't be segfaulting when the padding is too large (indeed, we should still produce correct results, even if it creates a silly number of 0s, no matter how high the manual padding goes).

plaidml/op/lib/ops.cc
```diff
@@ -2914,6 +2915,53 @@ Value transpose(const Value& value) {
...
+
+  std::vector<int64_t> raw_axes;
+  if (args[1].is_int()) {
+    raw_axes.push_back(args[1].as_int());
+  } else {
+    raw_axes = args[1].as_int_tuple();
+  }
+  std::set<size_t> axes;
+  for (auto& raw_axis : raw_axes) {
+    axes.insert(normalize_axis(raw_axis, ndims + 1, "unsqueeze"));
```
tzerrell, line 2936: This should be `ndims + axes.size()`: The axes numbers are relative to the output tensor, which has as many more dimensions than `I` as the size of the `axes` set.

plaidml/op/lib/ops.cc
```diff
@@ -2914,6 +2915,53 @@ Value transpose(const Value& value) {
...
+  std::vector<TensorIndex> I_idxs;
+  std::vector<TensorDim> O_dims;
+  std::vector<TensorIndex> O_idxs;
+  I.bind_dims(I_dims);
+  size_t src_loc = 0;
+  size_t new_rank = ndims + axes.size();
+  for (size_t i = 0; i < new_rank; i++) {
+    if (axes.count(i)) {
+      O_dims.push_back(edsl::TensorDim(1));
+      O_idxs.emplace_back("a");
```
tzerrell, line 2949: This will give every unsqueeze index the same name. IIRC they are semantically meaningless, so I think this runs correctly, but it will be confusing to anyone using the name to figure out what's happening. Using `llvm::formatv("a{0}", i)` analogously to the `else` clause is probably the simplest way to clean this.

plaidml/op/op_test.cc
```diff
@@ -887,140 +476,79 @@ TEST(Op, SpatialPadding) {
...
-  }
-}
-)#"));
+
+  std::vector<uint64_t> A_input = {0, 1, 2,  3,  //
+                                   4, 5, 6,  7,  //
+                                   8, 9, 10, 11};
+  std::vector<uint64_t> C_output = {0, 1, 2,  3,  //
+                                    4, 5, 6,  7,  //
+                                    8, 9, 10, 11};
```
tzerrell, line 509: It might be clearer to use literally the same vector for both the input and the output; since these vectors do not include shape information, Squeeze ought to be a no-op. This is currently reflected by A_input and C_output having the same contents, but it might be more obvious if they weren't different variables.

plaidml/op/op_test.cc
```diff
@@ -887,140 +476,79 @@ TEST(Op, SpatialPadding) {
...
-    return %0 : tensor<20x10xf32>
-  }
-}
-)#"));
+  std::vector<uint64_t> A_input = {0, 1, 2,  3,  //
+                                   4, 5, 6,  7,  //
+                                   8, 9, 10, 11};
+  std::vector<uint64_t> C_output = {0, 1, 2,  3,  //
+                                    4, 5, 6,  7,  //
+                                    8, 9, 10, 11};
```
tzerrell, line 543: Same comment as on Squeeze

tzerrell:
Python API had not been updated; I've made what I believe are the necessary changes. @flaub can you review my commit on this PR?

flaub:
LGTM

tzerrell:
Looks like a test for L2Norm landed while this was being drafted; @jd-bartlett96 can you do the merge & get that tested as well?

tzerrell:
Also while you're in there for the merge, I suspect the `==` op on lines 16-24 of the test is no longer needed and can be removed.

PR #1211, opened 2020-07-03
Add CastOp folder and tests

dgkutnic:
* Added a `CastOp` folder which eliminates redundant `cast` ops in the form of `cast(x, x.type)`
* Added a couple of tests for `cast`: checking behavior with scalars and testing the folder


pmlc/dialect/eltwise/ir/ops.cc
```diff
@@ -142,6 +142,16 @@ DEFINE_CANONICALIZER(TanHOp);
...
 DEFINE_CANONICALIZER(SelectOp);
 
+OpFoldResult CastOp::fold(ArrayRef<Attribute> operands) {
+  auto oldType = getRankedTensorType(tensor().getType());
+  auto newType = getRankedTensorType(result().getType());
+  /// cast(x, x.type) -> x
+  if (oldType == newType) {
+    return tensor();
+  }
+  return constFoldBinaryOp(operands, [](double a, double b) { return a + b; });
```
flaub, line 152: We should return `{}` here to indicate that there was no folding to be done.

plaidml/edsl/tests/edsl_test.py
```diff
@@ -518,6 +518,40 @@ def test_skip(self):
...
+module {
+  func @cast(%arg0: tensor<si32>) -> tensor<f32> {
+    %0 = "eltwise.cast"(%arg0) : (tensor<si32>) -> tensor<f32>
+    return %0 : tensor<f32>
+  }
+}
+'''
+        self.checkProgram(program, [
+            (I, np.array([(3)])),
+        ], [3.0])
```
flaub, line 536: This is where using c++ would be a better test, because python might automatically cast for you, hiding the actual data type.

pmlc/dialect/eltwise/ir/ops.cc
```diff
@@ -142,6 +142,16 @@ DEFINE_CANONICALIZER(TanHOp);
 DEFINE_CANONICALIZER(TanOp);
 DEFINE_CANONICALIZER(SelectOp);
 
+OpFoldResult CastOp::fold(ArrayRef<Attribute> operands) {
+  auto oldType = getRankedTensorType(tensor().getType());
+  auto newType = getRankedTensorType(result().getType());
+  /// cast(x, x.type) -> x
```
flaub, line 148: I don’t understand this comment

plaidml/edsl/tests/edsl_test.py
dgkutnic: Good point. I'll switch both tests over to c++.

pmlc/dialect/eltwise/ir/ops.cc
dgkutnic: I was trying to say "in the case where you cast an input tensor `x` to a result type that matches `x`'s initial type, fold `x`"

Any ideas to make the comment more legible? I was just trying to match the comment format that the rest of the folders  were using

flaub: How about:
```
// cast(x).type == type -> x 
```

flaub:
Actually, I think we have a `const-fold.mlir` in the tile dialect tests.

dgkutnic:
> Actually, I think we have a const-fold.mlir in the tile dialect tests.

We have a `const-fold.mlir` for eltwise dialect too, isn't that a better place since `cast` is part of the eltwise dialect?

flaub:
Yes! Sorry, eltwise is much better :)

dgkutnic:
I added some lit tests to `const-fold.mlir` in the eltwise dialect. Let me know if what I have now is sufficient, or if we need more tests.

pmlc/dialect/eltwise/tests/constant-fold.mlir
```diff
@@ -212,3 +212,28 @@ func @fold_div_f32_0(%arg0: f32) -> f32 {
   // CHECK-NEXT: %0 = "eltwise.div"(%arg0, %cst) : (f32, f32) -> tensor<f32>
   // CHECK-NEXT: return %0 : tensor<f32>
 }
+
+// Expected behavior in this case of cast is to fold
+// CHECK-LABEL: @fold_cast_si32
+func @fold_cast_si32(%arg0: tensor<3xsi32>) -> tensor<3xsi32> {
+  %0 = "eltwise.cast"(%arg0) : (tensor<3xsi32>) -> tensor<3xsi32>
+  return %0 : tensor<3xsi32>
+  // CHECK-NEXT: return %arg0 : tensor<3xsi32>
```
flaub, line 221: Please don't use `%arg0` directly in a test:
https://mlir.llvm.org/getting_started/TestingGuide/#filecheck-best-practices

Use:
```
// CHECK-NEXT: return %{{.*}} : tensor<3xf32>
```

pmlc/dialect/eltwise/tests/constant-fold.mlir
```diff
@@ -212,3 +212,28 @@ func @fold_div_f32_0(%arg0: f32) -> f32 {
   // CHECK-NEXT: %0 = "eltwise.div"(%arg0, %cst) : (f32, f32) -> tensor<f32>
   // CHECK-NEXT: return %0 : tensor<f32>
 }
+
+// Expected behavior in this case of cast is to fold
+// CHECK-LABEL: @fold_cast_si32
+func @fold_cast_si32(%arg0: tensor<3xsi32>) -> tensor<3xsi32> {
+  %0 = "eltwise.cast"(%arg0) : (tensor<3xsi32>) -> tensor<3xsi32>
+  return %0 : tensor<3xsi32>
+  // // CHECK-NEXT: return %{{.*}} : tensor<3xf32>
```
flaub, line 221: There seem to be two sets of comments

dgkutnic: My bad

PR #1212, opened 2020-07-06
storti integral examples 

1sanz:
These examples now run fine, required eltwise functions are now available. 

TODO: 
- documentation 
- references for used research


flaub:
I'm considering closing this because we plan to drop support for bazel. Any comments @brianretford or @tzerrell?

tzerrell:
The examples are interesting, but without being able to run them my strong opinion is they should not be landed. My vote is to close the PR (since we don't currently want to land it and aren't currently working on it) but retain the branch (since the examples are interesting and could in principle be resurrected & run if we decide we want to come back to them).

@flaub do you know of a way to tell mergify to not delete the branch when a PR is closed?

flaub:
AFAIK, mergify only manages PRs/branches that are being automatically merged. Closing a PR should not be under its control. But perhaps github has a default setting for this, I'll check.

tzerrell:
I was concerned that the branch might be automatically deleted by mergify based on having just looked at this PR: https://github.com/plaidml/plaidml/pull/1738

flaub:
OK, I see that mergify does do that, interesting. We could just 'restore branch' after mergify deletes it. But if we really want to keep this around, I think we should only keep branches that have PRs. I'd like to eventually remove old branches that are no longer maintained in general.

tzerrell (commented):
With the updates for updated PlaidML, this looks generally good to me. It will also need to be updated for the Bazel -> CMake changes. I suggest reaching out to @flaub about that if it's not clear how to do so.

networks/scitile/storti_integrals/BUILD
```diff
@@ -0,0 +1,52 @@
+# Copyright 2018 Intel Corporation.
```
flaub, line 1: Drop this file

networks/scitile/storti_integrals/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+pml_py_test(
+  NAME storti_py_test
+  SRC storti_integral_tests.py
+  CHECKS
```
flaub, line 4: Need to add a:
```
DEPS
  plaidml::py
```

networks/scitile/storti_integrals/CMakeLists.txt
```diff
@@ -18,3 +18,20 @@ pml_py_test(
       smoke
       test
 )
+
+# the rules below are used to generate graphs for torus and hypersphere experiemnts. 
+# these are disables because they take a while to genrate and are onyl useful when the graphs need to be analyzed.
```
tzerrell, line 23: Typos: genrate -> generate and onyl -> only

networks/scitile/storti_integrals/hypersphere.py
```diff
@@ -85,7 +85,7 @@ def check_integral(type, N, minval, maxval, eps, frep, vars):
     ax1.plot(x2, y_line2, label=label2)
 
     plt.legend(loc='lower left')
-    plt.show()
+    plt.savefig('hypershere_integral_graph.pdf')
```
tzerrell, line 88: typo: hypershere -> hypersphere

networks/scitile/storti_integrals/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+pml_py_library(
+  NAME py
+  SRC
+    hypershere.py
```
flaub, line 4: Filename has a typo: hypershere.py -> hypersphere.py

PR #1215, opened 2020-07-09
Use new affine canonicalizations

flaub:


pmlc/compiler/executable.cc
```diff
@@ -125,6 +113,7 @@ class MemRefDescriptor {
     void *basePtr;
     void *data;
     int64_t offset;
+    int64_t sizesAndStrides[];
```
tzerrell, line 116: I'm probably just missing something, but I don't see the benefit of having sizes & strides concat'd into one variable like this.

flaub: This is a esoteric feature of C++. The idea is that the sizes and strides, in memory, appear in this position in the struct. It's actually dynamically sized, based on the rank. So we want to provide a way for the compiler to refer to this position in a way that doesn't imply a particular rank.

See the usage of `sizesAndStrides` in the constructor.

pmlc/compiler/executable.cc
```diff
@@ -133,6 +122,15 @@ class MemRefDescriptor {
...
     base->basePtr = data;
     base->data = data;
+    auto rank = type.getRank();
+    auto shape = type.getShape();
+    auto memRefType = MemRefType::get(shape, type.getElementType());
+    SmallVector<int64_t, 8> strides;
+    getStridesAndOffset(memRefType, strides, base->offset);
+    for (unsigned i = 0; i < rank; i++) {
+      base->sizesAndStrides[i] = shape[i];
+      base->sizesAndStrides[i + rank] = strides[i];
```
flaub, line 132: @tzerrell Here is the usage

pmlc/compiler/executable.cc
flaub: Oh, the reason it's concated into one is because how could I reasonably access just the strides portion? I can't statically define this at compile time (since its dependent on rank).

PR #1221, opened 2020-07-16
Add more convolution tests

flaub:


tzerrell:
I suspect the test errors may have the same root cause as discussed in https://github.com/plaidml/plaidml/pull/1210#issuecomment-653268471. Perhaps now is the time to dig into what that root cause is? Let me know if I should help with that

PR #1225, opened 2020-07-21
Fix for strided convolutions in XSMM.  

jbruestle:
Note: These convolutions should have a valid XSSM lowering, but currently there is a bug in lowering, so we simply skip optimizations for this case.

### DIFF
--- a/plaidml/op/op_test.cc
+++ b/plaidml/op/op_test.cc
@@ -196,6 +196,15 @@ TEST_F(OpTest, Convolution) {
   runProgram(program);
 }
 
+TEST_F(OpTest, Convolution_Padding3) {
+  auto I = Placeholder(DType::FLOAT32, {1, 224, 224, 3}, "I");
+  auto K = Placeholder(DType::FLOAT32, {7, 7, 3, 64}, "K");
+  auto O = op::convolution(I, K).strides({2, 2}).autopad_mode(AutoPadMode::EXPLICIT).manual_padding({3, 3});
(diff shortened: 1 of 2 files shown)


pmlc/target/x86/xsmm_stencil.cc
```diff
@@ -163,6 +163,36 @@ class StencilXSMM : public StencilBase {
...
+      }
+      return true;
+    };
+    for (auto op : perm.ioOps) {
+      bool good = false;
+      if (auto loadOp = mlir::dyn_cast<AffineLoadOp>(op)) {
+        good = verifyStride(loadOp);
+      } else if (auto reduceOp = mlir::dyn_cast<AffineReduceOp>(op)) {
+        good = verifyStride(reduceOp);
+      }
```
tzerrell, line 189: Do we need to consider `AffineStoreOp`s too?

jbruestle: Not for XSMM, since it's a requirement that it have 2 loads + 1 reduce (in the capture() function).

plaidml/op/op_test.cc
```diff
@@ -196,6 +196,15 @@ TEST_F(OpTest, Convolution) {
   runProgram(program);
 }
 
+TEST_F(OpTest, Convolution_Padding3) {
```
flaub, line 199: I was trying to move our tests to something a little more parameterized (like OpenVINO tests):
https://github.com/plaidml/plaidml/pull/1221

PR #1228, opened 2020-07-24
3D conv tests

dgkutnic:
`OpTest.3dConvolution` has a really long compile time (around 200000ms)
`OpTest.3dConvolution4` is currently segfaulting

### DIFF
--- a/plaidml/op/op_test.cc
+++ b/plaidml/op/op_test.cc
@@ -179,7 +179,7 @@ TEST_F(OpTest, Concatenate) {
 TEST_F(OpTest, 3dConvolution) {
   auto I = Placeholder(DType::FLOAT32, {1, 32, 224, 224, 3}, "I");
   auto K = Placeholder(DType::FLOAT32, {7, 7, 7, 3, 64}, "K");
-  edsl::TensorRef O = edsl::Tensor(op::convolution(I, K).strides({2, 2, 2}).autopad_mode(AutoPadMode::EXPLICIT).manual_padding({2, 3, 2, 3, 2, 3}));
+  edsl::TensorRef O = edsl::Tensor(op::convolution(I, K).strides({2, 2, 2}).autopad_mode(op::AutoPadMode::EXPLICIT).manual_padding({2, 3, 2, 3, 2, 3}));
   auto program = makeProgram("convolution", {O});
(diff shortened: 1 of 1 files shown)


plaidml/op/op_test.cc
```diff
@@ -176,6 +176,71 @@ TEST_F(OpTest, Concatenate) {
...
+  auto K = Placeholder(DType::FLOAT32, {7, 7, 7, 3, 64}, "K");
+  edsl::TensorRef O = edsl::Tensor(op::convolution(I, K).strides({2, 2, 2}).autopad_mode(op::AutoPadMode::EXPLICIT).manual_padding({2, 3, 2, 3, 2, 3}));
+  auto program = makeProgram("convolution", {O});
+  IVLOG(1, program);
+  std::vector<float> I_input(1*32*224*224*3, 0.0);
+  std::vector<float> K_filter(7*7*7*3*64, 0.0);
+  std::vector<float> O_output(1*16*112*112*64, 0.0);
+  edsl::TensorBuffers input_buffers{{I, I_input}, {K, K_filter}};
+  edsl::TensorBuffers output_buffers{{O, O_output}};
+  checkProgram(program, input_buffers, output_buffers);
```
flaub, line 190: Only use `checkProgram` if you have actual inputs/outputs to validate against. `runProgram` is sufficient when you don't want to check correctness, it will supply the properly sized input/output buffers for you.

plaidml/op/op_test.cc
```diff
@@ -176,6 +176,71 @@ TEST_F(OpTest, Concatenate) {
   runProgram(program);
 }
 
+TEST_F(OpTest, 3dConvolution) {
```
flaub, line 179: @tzerrell @jbruestle  This is actually the really slow test case. All the rest of these are passing for me, but this one runs for several minutes. I haven't profiled yet, but my guess is that it's either BILP or XSMM stenciling.

plaidml/op/op_test.cc
flaub: Actually, I think it's because the workaround is to avoid using XSMM, and in this case, due to the strides, this one will run very slowly. So I don't think this is a compiler perf issue.

dgkutnic:
addressed in #1230

PR #1229, opened 2020-07-24
Enable Localize and ResizeTmps passes for x86

adstraw:


pmlc/dialect/pxa/transforms/resize_tmps.cc
```diff
@@ -38,6 +38,14 @@ struct ResizeTmpsPass : public ResizeTmpsBase<ResizeTmpsPass> {
     Block *opBlock = op.getOperation()->getBlock();
     IVLOG(2, "Considering: " << debugString(*op.getOperation()));
 
+    for (auto &use : getIndirectUses(op)) {
+      IVLOG(2, "Found indicrect use: " << debugString(*use.getOwner()));
```
flaub, line 42: Typo: `indirect`

pmlc/target/x86/pipeline.cc
```diff
@@ -89,8 +89,8 @@ void addToPipeline(OpPassManager &pm) {
   // pm.addPass(createCanonicalizerPass());
   // pm.addPass(pmlc::dialect::pxa::createMemRefDataFlowOptPass());
   // pm.addPass(createCanonicalizerPass());
-  // pm.addPass(pmlc::dialect::pxa::createLocalizePass());
-  // pm.addPass(pmlc::dialect::pxa::createResizeTmpsPass());
+  pm.addPass(pmlc::dialect::pxa::createLocalizePass());
+  pm.addPass(pmlc::dialect::pxa::createResizeTmpsPass());
   // pm.addPass(createCanonicalizerPass());
   // pm.addPass(createCSEPass());
```
flaub, line 95: We should probably enable these passes as well (Canonicalizer and CSE), now that Localize and ResizeTmps are running.

flaub:
Minor comments, but overall LGTM, thanks for this!

pmlc/dialect/pxa/tests/resize_tmps.mlir
```diff
@@ -44,3 +44,37 @@ func @inner_indexes(%I: memref<100x100xf32>) -> (memref<100x100xf32>) {
   }
   return %O4 : memref<100x100xf32>
 }
+
+func @no_resize_return() -> memref<10x10xf32> {
+// CHECK-LABEL: func @no_resize_return
+  %cst = constant 0.000000e+00 : f32
+  %0 = alloc() : memref<10x10xf32>
+  // CHECK: %0 = alloc() : memref<10x10xf32>
```
flaub, line 52: Don't use `%0` in lit tests because the name of vars are subject to change and cause the test to fail. Instead, use:

```
// CHECK: %{{.*}} = alloc() : memref<10x10xf32>
```

Or even better, since the test doesn't need to use this var:

```
// CHECK: alloc() : memref<10x10xf32>
```

pmlc/dialect/pxa/tests/resize_tmps.mlir
```diff
@@ -44,3 +44,37 @@ func @inner_indexes(%I: memref<100x100xf32>) -> (memref<100x100xf32>) {
...
+  }
+  return %1 : memref<10x10xf32>
+}
+
+#set0 = affine_set<(d0, d1, d2, d3) : (d0 * -30 - d1 + 221 >= 0, d2 * -30 - d3 + 221 >= 0)>
+func @no_resize_expand(%arg0: memref<32x30xf32>) -> memref<1x222x222x32xf32> {
+// CHECK-LABEL: func @no_resize_expand
+  %0 = alloc() : memref<1x30x32x8x8x32xf32>
+  %1 = alloc() : memref<1x222x222x32xf32>
+  // CHECK: %1 = alloc() : memref<1x222x222x32xf32>
```
flaub, line 65: Same here, avoid the use of `%1`

pmlc/dialect/pxa/transforms/resize_tmps.cc
adstraw: Done.

pmlc/target/x86/pipeline.cc
adstraw: Done.

pmlc/dialect/pxa/tests/resize_tmps.mlir
adstraw: Done.

pmlc/dialect/pxa/tests/resize_tmps.mlir
adstraw: Done.

pmlc/dialect/pxa/tests/resize_tmps.mlir
```diff
@@ -44,3 +44,37 @@ func @inner_indexes(%I: memref<100x100xf32>) -> (memref<100x100xf32>) {
...
+      %7 = mulf %5, %6 : f32
+      %8 = pxa.reduce add %7, %1[0, %arg1 * 30 + %arg2, %arg3 * 30 + %arg4, %arg5] : memref<1x222x222x32xf32>
+      affine.yield %8 : memref<1x222x222x32xf32>
+    } else {
+      affine.yield %1 : memref<1x222x222x32xf32>
+    }
+    affine.yield %4 : memref<1x222x222x32xf32>
+  }
+  %3 = alloc() : memref<1xf32>
+  return %3 : memref<1xf32>
```
flaub, line 79: Seems like you'd want to return `%2` here.

adstraw: The reason I'm doing this is so that this test hits the "no expansion" case and not the "do not resize the return" case.

jbruestle: Can you return void instead?  That would be more clear.

pmlc/dialect/pxa/transforms/resize_tmps.cc
```diff
@@ -38,6 +38,14 @@ struct ResizeTmpsPass : public ResizeTmpsBase<ResizeTmpsPass> {
     Block *opBlock = op.getOperation()->getBlock();
     IVLOG(2, "Considering: " << debugString(*op.getOperation()));
 
+    for (auto &use : getIndirectUses(op)) {
+      IVLOG(2, "Found indirect use: " << debugString(*use.getOwner()));
```
jbruestle, line 42: Log level 2 is probably a too high for what is mostly debugging output unlikely to be generally useful, maybe log level 4 or just remove for the commit. The other logs are fine as they are describe why the pass did what it did.

pmlc/dialect/pxa/tests/resize_tmps.mlir
adstraw: Done.

pmlc/dialect/pxa/transforms/resize_tmps.cc
adstraw: Deleted.  I agree it is not needed.

PR #1230, opened 2020-07-24
Add tests, fix correctness

flaub:


ci/plan.yml
```diff
@@ -148,168 +146,7 @@ SUITES:
...
-          - TestBackendOps.testTruncatedNormalDev
-          - TestBackendOps.testRandomNormalDev
-          - TestBackendOps.testRandomUniformMean
-          - TestBackendOps.testRandomUniformDev
-          - TestBackendOps.testRandomUniformVariableMean
-          - TestBackendOps.testRandomUniformVariableDev
-          - TestBackendOps.testRandomNormalMean
-          - TestBackendOps.testVarSimple
-          - TestBackendOps.testWinograd
+        args: [backend_test.py]
```
tzerrell, line 149: This is purely style, but why is this `args` line changed to `[item]` style when the other lines were changed to `- item` style?

plaidml/op/op_test.cc
```diff
@@ -222,6 +222,43 @@ ConvolutionParams convParams[] = {
...
+    },
+    {
+        // 3D convolution
+        DType::FLOAT32,
+        {1, 16, 56, 56, 64},
+        {3, 3, 3, 64, 192},
+        [](op::convolution& conv) {
+          conv.strides({1, 1, 1}).autopad_mode(op::AutoPadMode::EXPLICIT).manual_padding({1, 1, 1, 1, 1, 1});
+        },
+    },
```
dgkutnic, line 261: I'm still seeing this test segfaulting in the PR build. Do we want to add this one as a todo?

flaub: I'm trying to repro this locally, it seems to pass fine for me, I'll dig into this one a bit.

plaidbench/plaidbench/frontend_keras.py
```diff
@@ -198,9 +198,9 @@ def setup(self):
             eval(code, mod)
         self.x = mod['scale_dataset'](self.x)
         self.model = mod['build_model'](**build_model_kwargs)
-        if self.enable_bn_folding:
-            click.echo('Model loaded, folding in batch_norm')
-            self.model = self.fold_batch_norm(self.model)
+        # if self.enable_bn_folding:
+        #     click.echo('Model loaded, folding in batch_norm')
+        #     self.model = self.fold_batch_norm(self.model)
```
tzerrell, line 203: Not certain, but perhaps it would be cleaner to disable this by changing line 341 to `enable_bn_folding = False` instead?

ci/plan.yml
flaub: I have some yaml formatter that runs when I hit save. It decided that anything that doesn't fit within a single line should be split. In this case tho, it does fit and the formatter doesn't change it on me.

plaidbench/plaidbench/frontend_keras.py
flaub: Actually I'd like to get @jbruestle feedback on this to decide if we should just remove it entirely from plaidbench or what the plan is for this. I think we should try to write an MLIR pass for it. So I'm thinking this comment would just be temporary until we remove it.

dgkutnic:
LGTM, I'll let Jeremy/Tim hit the approval when they see the changes.

plaidbench/plaidbench/frontend_keras.py
tzerrell: Do you want to wait for @jbruestle's feedback here for this PR, or shall we land this and have a followup PR if needed?

flaub: I guess it wouldn't hurt to land it, but I still have build/ci issues to fix it seems.

jbruestle: Honestly I'd leave the comment in for now, because in part I'm curious to try to track down while folding induced failure, and we may want to re-enable this if we fix the bug and haven't written a BN folding MLIR pass yet (which I agree is the long term solution).

flaub:
This is ready again for review.

PR #1231, opened 2020-07-28
Add AffineVectoreReduce instruction to the pxa layer.

LLITCHEV:
This is needed to allow us to use the data flow analysis mechanisms that the AffineReduceOp exposes/utilizes, so Fusion and other passes can operate on the AffineVectorReduceOp.

### DIFF
--- a/pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
+++ b/pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
@@ -19,6 +19,8 @@ using mlir::AffineIfOp;
 using mlir::AffineLoadOp;
 using mlir::AffineParallelOp;
 using mlir::AffineStoreOp;
+using mlir::AffineVectorLoadOp;
+using mlir::AffineVectorStoreOp;
 using mlir::AllocOp;
 using mlir::ArrayRef;
 using mlir::ConversionPattern;
@@ -38,6 +40,7 @@ using mlir::RankedTensorType;
 using mlir::ReturnOp;
 using mlir::Type;
 using mlir::Value;
+using mlir::VectorType;
 
 using util::AggregationKind;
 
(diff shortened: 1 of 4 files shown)


pmlc/dialect/pxa/tests/ops.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: pmlc-opt -pxa-dataflow-opt -canonicalize -convert-pxa-to-affine %s | FileCheck %s
+
+// CHECK-LABEL: func @pxa_vector_reduce
+func @pxa_vector_reduce(%arg0: memref<100x100xf32>, %arg1: memref<100x100xf32>, %arg2: memref<100x100xf32>) {
```
jbruestle, line 4: At the PXA level, outputs are typically returned rather than written into, i.e:

func @pxa_vector_reduce(%arg0: memref<100x100xf32>, %arg1: memref<100x100xf32>) -> (memref<100x100xf32>)

Then, inside the body, allocate the output (with alloc), add an affine.parallel that clears it via pxa.reduce assign, then do the vector reductions and then return the result of the vector reduction.

Something like

%a = alloc<...>
%b = affine.parallel ... {
  %cst = constant.zero
  %c = pxa.reduce assign %cst, %b[...] ...
  yield %c
}
%d = affine.parallel (existing AP) {
  ...
  %e = pxa.vector_reduce add %vec, %b : ...
  yield %e
}
return %d

Look at the EDSL tests IR for more complete examples, it's a bit weird.  This should lower to very similar code to what you already lower to.

Basically, the current test doesn't actually use the PXA memref-state semantics (and had you added a CSE pass, the vector_reduce would have vanished I think since it's NoSideEffect and it's return is unused)

pmlc/dialect/pxa/ir/ops.td
```diff
@@ -29,8 +29,7 @@ class PXA_OpWithPP<string mnemonic, list<OpTrait> traits = []> :
 
 def AnyStdScalar : AnyTypeOf<[AnyFloat, AnySignlessInteger]> {}
 
-// TODO: Add NoSideEffect
-def AffineReduceOp : PXA_OpWithPP<"reduce", []> {
+def AffineReduceOp : PXA_OpWithPP<"reduce", [NoSideEffect]> {
```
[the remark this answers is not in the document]
LLITCHEV: I just saw the TODO: and added the "no side effect" flag, but I can add them back and just deal with it later.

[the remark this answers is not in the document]
flaub: IIRC, the `Arg<AnyMemRef, "the memref to reduce", [MemRead, MemWrite]>:$mem,` supercedes `[NoSideEffect]` as far as the canonicalizer goes. I believe this is why I needed to write a separate canoncalizer pattern.

pmlc/dialect/pxa/ir/ops.cc
```diff
@@ -126,6 +136,22 @@ struct SimplifyDeadReduce : public OpRewritePattern<AffineReduceOp> {
   }
 };
 
+/// Fold vector_reduce operations with no uses. Vector_reduce has side effects
+/// on the heap, but can still be deleted if it has zero uses.
+struct SimplifyDeadVectorReduce
```
flaub, line 141: I think we can reuse the existing `SimplifyDeadReduce`, the only difference is the template parameter.

pmlc/dialect/pxa/tests/ops.mlir
LLITCHEV: Added more tests. and returning the value now.

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -224,6 +227,88 @@ struct AffineReduceOpConversion : public LoweringBase<pxa::AffineReduceOp> {
...
+  }
+
+  Value createVectorReduction(ConversionPatternRewriter &rewriter,
+                              pxa::AffineVectorReduceOp op,
+                              Value source) const {
+    // Must be a VectorType
+    assert(source.getType().isa<VectorType>());
+
+    switch (op.agg()) {
+    case AggregationKind::assign:
```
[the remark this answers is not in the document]
LLITCHEV: Added more tests.

pmlc/dialect/pxa/ir/ops.td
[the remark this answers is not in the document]
LLITCHEV: NoSideEffect flag removed (as it was before).

pmlc/dialect/pxa/ir/ops.cc
```diff
@@ -126,6 +136,22 @@ struct SimplifyDeadReduce : public OpRewritePattern<AffineReduceOp> {
...
+    : public OpRewritePattern<AffineVectorReduceOp> {
```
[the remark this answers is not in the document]
LLITCHEV: NoSideEffect flag removed (as it was before).

pmlc/dialect/pxa/ir/ops.cc
jbruestle: I agree, you should template SimplyDeadReduce on the operation type to and then just instantiate two versions in the patterns to reduce code duplication.

pmlc/dialect/pxa/tests/ops.mlir
```diff
@@ -0,0 +1,159 @@
...
+
+// CHECK-LABEL: func @pxa_reduce_mul
+func @pxa_reduce_mul(%arg0: memref<100x100xf32>, %arg1: memref<100x100xf32>) -> (memref<100x100xf32>) {
+  %a = alloc() : memref<100x100xf32>
+  %r = affine.parallel (%i, %j, %k) = (0, 0, 0) to (100, 100, 100) reduce ("assign") -> (memref<100x100xf32>) {
+    %0 = affine.load %arg1[%i, %k] : memref<100x100xf32>
+    %1 = affine.load %arg0[%k, %j] : memref<100x100xf32>
+    %2 = mulf %0, %1 : f32
+    %red = pxa.reduce mul %2, %a[%i, %j] :  memref<100x100xf32>
+    // CHECK: %{{.*}} = affine.load %{{.*}} : memref<100x100xf32>
```
jbruestle, line 73: The patterns are a bit needlessly permissive. You should probably do captures of the values.  You could verify that for example that the loaded value is passed into the mulf and that the output of the mulf goes into the store, and that the store and the load both reference the same buffer (which they should post pxa->affine)

pmlc/dialect/pxa/ir/ops.cc
LLITCHEV: It is done.

pmlc/dialect/pxa/tests/ops.mlir
LLITCHEV: Added more soecific checks.

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -236,9 +237,16 @@ struct AffineVectorReduceOpConversion
...
+    auto srcMapAttr = AffineMapAttr::get(op.map());
+    // Set the nap attribute
+    source.setAttr(AffineVectorLoadOp::getMapAttrName(), srcMapAttr);
     auto reduce = createVectorReduction(rewriter, op, source.getResult());
-    rewriter.create<AffineVectorStoreOp>(op.getLoc(), ArrayRef<Type>{}, reduce,
-                                         op.mem(), op.idxs());
+    auto dest = rewriter.create<AffineVectorStoreOp>(
+        op.getLoc(), ArrayRef<Type>{}, reduce, op.mem(), op.idxs());
+    // Set the nap attribute
+    auto destMapAttr = AffineMapAttr::get(op.map());
```
jbruestle, line 248: Both src + dest take the same map (and thus the same attribute), no need to call AffineMapAttr::get twice.

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -236,9 +237,16 @@ struct AffineVectorReduceOpConversion
                ConversionPatternRewriter &rewriter) const override {
     auto source = rewriter.create<AffineVectorLoadOp>(
         op.getLoc(), op.getVectorType(), op.mem(), op.idxs());
+    // Get an attribute form of the map
+    auto srcMapAttr = AffineMapAttr::get(op.map());
+    // Set the nap attribute
```
jbruestle, line 242: Typeo, although a nap attribute sounds really good right now...

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
LLITCHEV: Sure! I had it that way, but I have seen LLVM having issues with sharing state objects like this. I choose to stay on the safe side, but I'll change it.

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
LLITCHEV: Fixed

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
LLITCHEV: Changed.

jbruestle: So attributes + types are interned in MLIR (i.e. they are copied into a big buffer in the MLIR context that is never freed until context destruction) and so they are always safe (and cheap) to reuse aggressively, but constructing them (i.e. interning) is mildly expensive because it involves checking if an identical value exists, and if not, copying and compacting the memory representation across.

PR #1232, opened 2020-07-30
Added explicit_padding op

jd-bartlett96:
- Added explicit_padding op that corresponds to TensorFlow's & OpenVino's Pad functions (the existing spatial padding op does not work for these functions since all axes are padded and various filling modes for the paddings are allowable)
- Only implemented constant padding mode; Edge, Symmetric, and Reflect paddings remain to be implemented. Note: Tensorflow (which only accepts Constant, Symmetric, and Reflect modes) should still be able to compile any of these since it compiles Symmetric and Reflect paddings as a series of slice & concatenation ops; it only calls the "Pad" instruction for constant padding.
- Currently Constant padding with nonzero constants is not tested in OV; the infrastructure to test this parameter is not yet set up.


plaidml/op/op_test.cc
```diff
@@ -251,6 +251,22 @@ TEST_F(OpTest, Elu) {
...
+  auto program = makeProgram("explicit_padding", {O});
+
+  std::vector<float> I_input = {1, 2, 3,  //
+                                4, 5, 6};
+  std::vector<float> O_output = {-1, -1, -1, -1, -1,  //
+                                 -1, -1, -1, -1, -1,  //
+                                 -1, 1,  2,  3,  -1,  //
+                                 -1, 4,  5,  6,  -1,  //
+                                 -1, -1, -1, -1, -1,  //
+                                 -1, -1, -1, -1, -1};
```
dgkutnic, line 266: Minor nitpick: did you run `clang-format` here? Looks like the 3rd and 4th rows are offset from the rest of the matrix.

jd-bartlett96: This is actually what `clang-format` changed it to, I guess it tries to keep the fronts of the numbers aligned instead of the commas

dgkutnic: Interesting, thanks for clarifying. I see it in other places too.

plaidml/bridge/openvino/src/plaidml_plugin/ops/pad.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace PlaidMLPlugin {
+
+static OpRegistration reg("pad", [](const Context& ctx) {
+  auto* layer = dynamic_cast<ngraph::opset1::Pad*>(ctx.layer);
+  IE_ASSERT((ctx.operands.size() == 3) || (ctx.operands.size() == 4));
+
+  auto I = ctx.operands.at(0);
+  std::vector<int> lo_pads;
+  auto lopadset = get_axis_vector_from_constant_operand(1, ctx.layer);
```
tzerrell, line 24: This probably works, but it's a little awkward since these are padding amounts, not axes. Might be better to use an approach like StridedSlice's `cast_constant_operand`

tzerrell, line 24: Change the variable name to not use the word "set" since it's not a set

plaidml/bridge/openvino/src/plaidml_plugin/ops/pad.cpp
```diff
@@ -0,0 +1,44 @@
...
+  for (auto p : lopadset) {
+    lo_pads.push_back(p);
+  }
+  std::vector<int> hi_pads;
+  auto hipadset = get_axis_vector_from_constant_operand(2, ctx.layer);
```
tzerrell, line 29: Again, like `cast_constant_operand` would be better

plaidml/bridge/openvino/src/plaidml_plugin/ops/pad.cpp
```diff
@@ -0,0 +1,44 @@
...
+  for (auto p : hipadset) {
+    hi_pads.push_back(p);
+  }
+  float padval = 0.;  // OV tests currently only use default constant pad of 0
+  if (ctx.operands.size() == 4) {
+    auto padvalset = get_axis_set_from_constant_operand(3, ctx.layer);
```
tzerrell, line 35: This won't work in most cases (the get axis set function explicitly verifies you have an integer type); you'll need something like the `cast_constant_operand` mentioned above.

plaidml/bridge/openvino/tests/functional/plugin/plaidml/shared_tests_instances/single_layer_tests/pad.cpp
```diff
@@ -0,0 +1,37 @@
...
+    {std::vector<std::size_t>({0, 1}), std::vector<std::size_t>({10, 10})}};
+
+const std::vector<ngraph::op::PadMode> padModes = {{ngraph::op::PadMode::CONSTANT}};
+
+INSTANTIATE_TEST_CASE_P(CompareWithRefs, PadLayerTest,
+                        ::testing::Combine(::testing::ValuesIn(loPads), ::testing::ValuesIn(hiPads),
+                                           ::testing::ValuesIn(padModes), ::testing::ValuesIn(netPrecisions),
+                                           ::testing::ValuesIn(inputShapes),
+                                           ::testing::Values(CommonTestUtils::DEVICE_PLAIDML)),
+                        PadLayerTest::getTestCaseName);
```
tzerrell, line 35: Put each input to Combine on its own line with `//` at the end so linting doesn't reformat

plaidml/op/op.h
```diff
@@ -320,6 +328,16 @@ inline edsl::Tensor elu(const edsl::Tensor& I, double alpha) {
   return details::op("elu", args).as_tensor();
 }
 
+inline edsl::Tensor explicit_padding(  //
+    const edsl::Tensor& x,             //
+    const std::vector<int>& lo_pads,   //
+    const std::vector<int>& hi_pads,   //
+    PadMode mode,                      //
+    const float& padval) {
```
tzerrell, line 336: This should be `double` instead of `float`. (In the long run, we'll need to think about how to deal with `padval` since it's supposed to match the tensor's type, but `double` should be fine for now.)

flaub: Also, C++ wise, there's really no reason to pass a POD (Plain-Old-Datatype) by-const-ref (like `const float&`). Passing by value is usually easier to read and doesn't cause any performance hit.

plaidml/op/op.h
```diff
@@ -320,6 +328,16 @@ inline edsl::Tensor elu(const edsl::Tensor& I, double alpha) {
   return details::op("elu", args).as_tensor();
 }
 
+inline edsl::Tensor explicit_padding(  //
+    const edsl::Tensor& x,             //
+    const std::vector<int>& lo_pads,   //
+    const std::vector<int>& hi_pads,   //
+    PadMode mode,                      //
+    const double& padval) {
```
flaub, line 336: Like here, just pass `double`

plaidml/bridge/openvino/src/plaidml_plugin/ops/pad.cpp
```diff
@@ -13,27 +13,36 @@
...
+
+template <typename T>
+std::vector<T> cast_constant_operand(size_t operand_idx, ngraph::Node* layer) {
+  auto ngraph_const =
+      std::dynamic_pointer_cast<ngraph::op::Constant>(layer->input_value(operand_idx).get_node_shared_ptr());
+  if (ngraph_const) {
+    return ngraph_const->cast_vector<T>();
+  } else {
+    THROW_IE_EXCEPTION << "Dynamic slicing not currently supported by PlaidML plugin; all of begin, end, and stride "
+                          "must be Constants.";
```
tzerrell, line 26: This exception text was specific for StridedSlice, update it for Pad (e.g. Dynamic padding not currently supported by PlaidML plugin; all of pads_begin, pads_end, and pads_value must be Constants).

(We could also create a shared cast_constant_operand in plaidml_utils with some sort of joint error message; IMO the current approach is sufficient.)

PR #1236, opened 2020-07-31
Modify explicit_padding op

jd-bartlett96:
- Changed treatment of padval argument from double to Value for easier passing from TF & better handling of data type
- Implemented fluent API

### DIFF
--- a/plaidml/op/op.h
+++ b/plaidml/op/op.h
@@ -328,15 +328,44 @@ inline edsl::Tensor elu(const edsl::Tensor& I, double alpha) {
   return details::op("elu", args).as_tensor();
 }
 
-inline edsl::Tensor explicit_padding(  //
-    const edsl::Tensor& x,             //
-    const std::vector<int>& lo_pads,   //
-    const std::vector<int>& hi_pads,   //
-    PadMode mode,                      //
-    double padval) {
-  auto args = edsl::make_tuple(x, edsl::make_tuple(lo_pads), edsl::make_tuple(hi_pads), static_cast<int>(mode), padval);
-  return details::op("explicit_padding", edsl::Value(args)).as_tensor();
-}
+class explicit_padding {
+ public:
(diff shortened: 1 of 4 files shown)


plaidml/op/op.h
```diff
@@ -328,15 +328,44 @@ inline edsl::Tensor elu(const edsl::Tensor& I, double alpha) {
...
+
+  explicit_padding& mode(PadMode mode) {
+    mode_ = mode;
+    return *this;
+  }
+
+  explicit_padding& padval(const edsl::Value& padval) {
+    padval_ = padval;
+    return *this;
+  }
```
tzerrell, line 354: I don't know what @dgkutnic needs for the TF side to work well, but avoiding exposing a Value in this API is preferable where possible. (See https://github.com/plaidml/plaidml/pull/1165 for some discussion on this.)

Depending on what TF needs, it might be possible to have multiple `padval` functions taking e.g. `float`, `double`, `int` (or whatever exactly you want to support) as inputs and then setting `padval_ = edsl::Value(padval)` inside this function. Or perhaps even better, a single templated function to do that.

But I don't know what the TF-side requirements are, so this might not be viable.

dgkutnic: The TF-side requirements are that this would be passed in as a Tensor. We could definitely do templated functions if we think that is better.

tzerrell: I think a templated function is preferable for the same reasons @flaub gave in #1165.

@jd-bartlett96 that's only a rough outline, let me know if you want more details on what I have in mind.

tzerrell: Or honestly we could probably just have it take a tensor and have the OV bridge initialize a constant tensor with the value it was given

dgkutnic: Yeah, I like the `Tensor` idea.

PR #1237, opened 2020-08-03
Wip: Do not merge: Add OpenCL + SPIRV path

kdobros:
### How to run:
1. Currently branched from commit: f171a54279040e3a2d5fedb88508604a9ce1d55e, merge at this point
2. Manually apply LLVM [patch](https://github.com/kdobros/plaidml/blob/ocl_spirv/llvm_unified_diff.patch)
3. On Linux you will most likely need to fix libraries to link in vendor/opencl_icd_loader/opencl_icd_loader.BUILD

### Most interesting changes:
273d08c - Add comp dialect - host-device modeling
0377848 - LLVM patch
0d1343a - Fixes for existing gpu pipeline

### Current status:
- EdslCpp tests - 16/17 pass
- networks/oplib/resnet50 - able to execute with removed softmax layer, conformance not verified

### TODO:
- upstream existing llvm changes
- make sure linux build works correctly - especially opencl icd
- remove debug cout's
- add traces support
- add support for extended OCL/SPIRV operations, needed for resnet's softmax

### DIFF
--- a/bzl/workspace.bzl
+++ b/bzl/workspace.bzl
@@ -170,4 +170,28 @@ def plaidml_workspace():
         build_file = clean_dep("//bzl:zlib.BUILD"),
     )
 
+    http_archive(
+        name = "opencl_headers",
+        url = "https://github.com/KhronosGroup/OpenCL-Headers/archive/v2020.06.16.zip",
+        sha256 = "518703d3c3a6333bcf8e4f80758e4e98f7af30fbd72a09fe8c2673da1628d80c",
+        strip_prefix = "OpenCL-Headers-2020.06.16",
+        build_file = clean_dep("//vendor/opencl_headers:opencl_headers.BUILD"),
+    )
+
+    http_archive(
+        name = "opencl_hpp_headers",
+        url = "https://github.com/KhronosGroup/OpenCL-CLHPP/archive/v2.0.12.zip",
+        sha256 = "127936b3a5ef147f23b85fb043599d1480e9e57acabe2d2a67c5dac05aa4ad70",
+        strip_prefix = "OpenCL-CLHPP-2.0.12",
+        build_file = clean_dep("//vendor/opencl_hpp_headers:opencl_hpp_headers.BUILD"),
+    )
+
+    http_archive(
+        name = "opencl_icd_loader",
+        url = "https://github.com/KhronosGroup/OpenCL-ICD-Loader/archive/v2020.06.16.zip",
+        sha256 = "e4c27a5adcef4dbc0fee98864af203dc78dfc967ca7287c9bad9add030e7516e",
+        strip_prefix = "OpenCL-ICD-Loader-2020.06.16",
+        build_file = clean_dep("//vendor/opencl_icd_loader:opencl_icd_loader.BUILD"),
+    )
+
     openvino_workspace()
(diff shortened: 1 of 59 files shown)


flaub:
This is great work! Thanks, we'll do an in-depth review.

pmlc/conversion/comp/gpu_to_comp.cc
```diff
@@ -0,0 +1,324 @@
...
+  auto scheduleFuncOp = rewriter.create<comp::ScheduleFunc>(
+      loc, eventType, execEnv, mlir::Value());
+  auto postScheduleIP = rewriter.saveInsertionPoint();
+  auto newBlock =
+      rewriter.createBlock(&scheduleFuncOp.body(), scheduleFuncOp.body().end());
+  // TODO Is there a better way to move operation into block insted of
+  // clone->erase
+  newBlock->push_back(op.clone());
+  rewriter.create<comp::ScheduleEnd>(loc);
+  // TODO Isn't there RAII-idiomatic way to save/restore insertion point
```
flaub, line 136: Yes, see the `PatternRewriter::InsertionGuard`

pmlc/dialect/comp/ir/ops.td
```diff
@@ -0,0 +1,267 @@
...
+}
+
+def COMP_ScheduleSingleBlock : SingleBlockImplicitTerminator<"ScheduleEnd">;
+
+def COMP_ScheduleFunc : COMP_Op<"schedule_func",
+                                [COMP_ScheduleSingleBlock,
+                                 DeclareOpInterfaceMethods<COMP_ExecEnvOpInterface>,
+                                 DeclareOpInterfaceMethods<COMP_ScheduleOpInterface>]>
+                  , Arguments<(ins COMP_AnyExecEnv:$execEnv,
+                                   Optional<COMP_AnyEvent>:$depEvent)>
```
flaub, line 88: Perhaps we want to have a `Variadic` here to support multiple dependent events.

bzl/workspace.bzl
```diff
@@ -170,4 +170,28 @@ def plaidml_workspace():
...
+
+    http_archive(
+        name = "opencl_hpp_headers",
+        url = "https://github.com/KhronosGroup/OpenCL-CLHPP/archive/v2.0.12.zip",
+        sha256 = "127936b3a5ef147f23b85fb043599d1480e9e57acabe2d2a67c5dac05aa4ad70",
+        strip_prefix = "OpenCL-CLHPP-2.0.12",
+        build_file = clean_dep("//vendor/opencl_hpp_headers:opencl_hpp_headers.BUILD"),
+    )
+
+    http_archive(
```
earhart, line 189: FWIW -- we used to do this, since it got us an ICD loader with the most recent OpenCL API (so we could link, and then we could dynamically probe devices to see what they actually supported).  IIRC there was some problem with this approach, though, so we switched to using the system's OpenCL ICD, and using a shim layer that dynamically loaded the system OpenCL ICD, forwarding supported calls.

pmlc/dialect/comp/ir/ops.td
kdobros: Good point, previously I used solution with `group_events` and "empty" event in case I would need variadic for something else. I created more polished solution as #1282

pmlc/conversion/comp/gpu_to_comp.cc
kdobros: Thank you, I'll use that

PR #1251, opened 2020-08-05
Changes to StrideInfo and some fixes to tiling and resize

jbruestle:
Updates to StrideInfo, Tiling and ResizeTmps so that they work on non-canonical affine loops.

### DIFF
--- a/pmlc/dialect/pxa/analysis/strides.cc
+++ b/pmlc/dialect/pxa/analysis/strides.cc
@@ -30,7 +30,7 @@ static std::string getUniqueName(Block *ref, BlockArgument arg) {
 }
 
 // Generally useful helper function
-int64_t GetIVStep(BlockArgument arg) {
+int64_t getIVStep(BlockArgument arg) {
   // Check the kind of loop we are part of, and dispatch.
   Operation *baseOp = arg.getOwner()->getParentOp();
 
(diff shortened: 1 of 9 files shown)


jbruestle:
@shiyao00 Wanted to give you a heads up since I changed the semantics of tile for cases with non-1 step sizes (provided tile size is the total range not the relative range).  I'm not sure if any internal code relies on those semantics, but I noticed you added a test for that case.

pmlc/dialect/pxa/analysis/strides.cc
```diff
@@ -28,27 +29,44 @@ static std::string getUniqueName(Block *ref, BlockArgument arg) {
       .str();
 }
 
+// Generally useful helper function
+int64_t GetIVStep(BlockArgument arg) {
```
flaub, line 33: Rename `getIVStep`

pmlc/dialect/pxa/analysis/strides.cc
```diff
@@ -181,6 +215,14 @@ void StrideInfo::print(raw_ostream &os, Block *relative) const {
   os << ']';
 }
 
+AffineValueMap StridesToValueMap(MLIRContext *ctx, ArrayRef<StrideInfo> dims) {
```
flaub, line 218: Not sure if we should say `stridesToValueMap` or something else.

flaub: Like, `getStridesToValueMap` or `convertStridesToValueMap`

PR #1253, opened 2020-08-06
Testenv: Allow env overrides

earhart:


flaub:
Curious how is setDefault different from set?

earhart:
> Curious how is setDefault different from set?

setDefault doesn't apply the setting if it's already been set. The idea is that the test environment can set llvm as the default target, while allowing the user to override it.

flaub:
Doesn't this break reproducible builds? Like if a dev happens to have one of the env vars set, won't they get a different result?

earhart:
The builds are still reproducible.  They take an environment variable input (or a .plaidml2 settings file input), and you have to get that environment correct.  NB we're already explicitly plumbing the PLAIDML_DEVICE and PLAIDML_TARGET environment variables through Bazel (which otherwise would remove them), and we're explicitly setting them in our CI plan.yml.

(I'd definitely be in favor of having someone spend some time designing a better settings system, but I also think that's a longer project and this isn't really the time.)

earhart:
Update: Frank and I chatted about this in person, going over the pros and cons of various approaches.  We've settled on the idea of using explicit command-line flags to control the target and device, making what's going on extremely explicit and making a particular bazel invocation hermetic with regard to these environment variables.  I'm going to code it up and submit another PR.

Issue #1249, opened 2020-08-04
Fix various small lowering bugs in backend tests


jbruestle:
Fix all the test that do not involving a missing feature (i.e. gather).

jbruestle:
@mars1026 This is basically a placeholder ticket, feel free to expand on the details, etc.

mars1026:
testPow - fails to legalize eltwise.pow when one of the arguments is integer rather than float
testSparseCategoricalCrossentropy (& friends) - lowers to std.cmpi with condition as f32 instead of integer type
testSwitch - fails when eltwise.select condition is f32 instead of boolean

flaub:
Are these all done now?

mars1026:
indeed they are, looks like I just forgot to close it

PR #1257, opened 2020-08-06
Pick float vs int comparison for contractions based on type of comparands, not result

mars1026:
This was the issue blocking the sparse categorical crossentropy backend tests. Related to #1249.

### DIFF
--- a/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
+++ b/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
@@ -170,8 +170,17 @@ static Value createCastOp(OpBuilder &builder, Location loc, Value from,
       return builder.create<mlir::FPTruncOp>(loc, from, intoType).getResult();
     }
     if (auto fromIntType = fromType.dyn_cast<IntegerType>()) {
-      // SIToFPOp: IntegerType -> FloatType
-      return builder.create<mlir::SIToFPOp>(loc, from, intoType).getResult();
+      if (fromIntType.getWidth() == 1) {
+        // If UIToFPOp existed, we would use it, but it currently does not.
+        // Converting i1 as signed gives us [0,-1] rather than [0,1].
+        // We can hack our way around this by negating the result.
+        auto raw =
+            builder.create<mlir::SIToFPOp>(loc, from, intoType).getResult();
+        return builder.create<mlir::NegFOp>(loc, raw);
+      } else {
+        // SIToFPOp: IntegerType -> FloatType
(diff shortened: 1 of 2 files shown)


pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -1144,10 +1165,10 @@ struct LowerTileToPXAPass : public LowerTileToPXABase<LowerTileToPXAPass> {
                                 ResultIs<EltwiseInteger>>,
         ContractionOpConversion<CombinationKind::eq,
                                 CmpFloatOp<CmpFPredicate::OEQ>,
-                                ResultIs<EltwiseFloat>>,
+                                AnyComparandIs<EltwiseFloat>>,
         ContractionOpConversion<CombinationKind::eq,
                                 CmpIntOp<CmpIPredicate::eq>,
-                                ResultIs<EltwiseInteger>>,
+                                ComparandsAre<EltwiseInteger>>,
```
flaub, line 1171: Does this mean we no longer have a conversion for heterogenous operand types?

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -1144,10 +1165,10 @@ struct LowerTileToPXAPass : public LowerTileToPXABase<LowerTileToPXAPass> {
                                 ResultIs<EltwiseInteger>>,
         ContractionOpConversion<CombinationKind::eq,
                                 CmpFloatOp<CmpFPredicate::OEQ>,
-                                ResultIs<EltwiseFloat>>,
+                                AnyComparandIs<EltwiseFloat>>,
```
jbruestle, line 1168: I probably just don't understand some magic MLIR goo, but how is AnyComparandIs related to ComparandsAre?

jbruestle: Oh wait, I get it.  Basically float happens on *either* being float (which already exists), and int happens only when both are int....

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
mars1026: The previous ContractionOpConversion should cover that one, using CmpFloatOp. The issue was that when we have heterogenous operand types, they are promoted to float, not to int.

flaub: OK, sorry, took me a second to understand, LGTM.

PR #1258, opened 2020-08-06
Cache pass transform

jbruestle:
Initial version of cache pass.  Clearly needs tests, but my prototype subgroups pass which uses it produces correct code, sending up for early PR to improve coordination.


pmlc/util/ident.cc
```diff
@@ -0,0 +1,66 @@
...
+      agg = AtomicRMWKind::addi;
+    }
+  }
+
+  switch (agg) {
+  case AtomicRMWKind::assign:
+  case AtomicRMWKind::addf:
+    return typedFloat(0.0, type);
+  case AtomicRMWKind::addi:
+    return IntegerAttr::get(type, int64_t(0));
```
flaub, line 39: Use `static_cast<int64_t>(0)`

pmlc/util/BUILD
```diff
@@ -66,6 +66,7 @@ plaidml_cc_library(
         "@easylogging",
         "@llvm-project//llvm:Support",
         "@llvm-project//mlir:IR",
+	"@llvm-project//mlir:StandardOps",
```
flaub, line 69: whitespace, use buildifier

pmlc/dialect/pxa/transforms/cache.cc
```diff
@@ -0,0 +1,173 @@
...
+      return failure();
+    }
+    out = yieldOp.getParentOp()->getResult(use.getOperandNumber());
+  }
+  // Verify final output has a single use, TODO: THis probably isn't a hard
+  // requirement, but it's easier
+  if (!out.hasOneUse()) {
+    return failure();
+  }
+  auto &final_use = *out.use_begin();
```
flaub, line 134: nit: rename to `finalUse`

pmlc/dialect/pxa/transforms/cache.cc
```diff
@@ -0,0 +1,173 @@
...
+  while (out.getParentBlock() != par.getBody()) {
+    auto &use = *out.use_begin();
+    auto yieldOp = cast<AffineYieldOp>(use.getOwner());
+    out = yieldOp.getParentOp()->getResult(use.getOperandNumber());
+    out.setType(newType);
+  }
+  // Implement the copy loop
+  builder.setInsertionPoint(par.getBody(), std::prev(par.getBody()->end()));
+  SmallVector<StrideInfo, 4> zeroOffset(rap.innerCount.size());
+  auto copy_loop =
```
flaub, line 165: nit: rename `copyLoop`

pmlc/dialect/pxa/transforms/BUILD
```diff
@@ -23,6 +23,7 @@ plaidml_cc_library(
     name = "transforms",
     srcs = [
         "autotile.cc",
+	"cache.cc",
```
flaub, line 26: whitespace

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -122,6 +121,56 @@ struct AffineIfOpConversion : public OpConversionPattern<AffineIfOp> {
   }
 };
 
+static Value createReduction(ConversionPatternRewriter &rewriter,
+                             mlir::Location loc, AtomicRMWKind agg,
+                             Value source, Value val) {
```
adstraw, line 126: The variable name `val` is not very descriptive.  The calling convention is always:
```
createReduction(rewriter, op.getLoc(), op.agg(), source.getResult(), op.val());
```
Are we ever going to get location, aggregation, and value from anything but `op`?  If not the previous API makes more sense to me:

```
Value createReduction(ConversionPatternRewriter &rewriter, pxa::AffineReduceOp op, Value source)
```

pmlc/dialect/pxa/analysis/strides.cc
```diff
@@ -397,6 +398,59 @@ Optional<StrideInfo> computeStrideInfo(pmlc::dialect::pxa::AffineReduceOp op) {
...
+        maybeStrides =
+            computeStrideInfo(op.getAffineMap(), op.getMapOperands());
+        vecSize = op.getVectorType().getShape();
+      });
+  if (!maybeStrides) {
+    return llvm::None;
+  }
+  auto &strides = *maybeStrides;
+  RelativeAccessPattern ret;
+  for (size_t i = 0; i < strides.size(); i++) {
```
adstraw, line 429: Feels like some documentation is warranted to describe the math being done in this loop.  This may be just "learning curve" for me but I can't make sense of it.

pmlc/dialect/pxa/analysis/strides.h
```diff
@@ -24,6 +24,13 @@ struct StrideRange {
   explicit StrideRange(int64_t val)
       : valid(true), minVal(val), maxVal(val), stride(0) {}
 
+  explicit StrideRange(int64_t min, int64_t max, int64_t stride)
+      : valid(true), minVal(min), maxVal(max), stride(stride) {
+    if (min == max) {
```
adstraw, line 29: `if (min >= max)`?

pmlc/dialect/pxa/analysis/strides.h
```diff
@@ -113,6 +120,29 @@ Optional<StrideInfo> computeStrideInfo(AffineLoadOp op);
 Optional<StrideInfo> computeStrideInfo(AffineStoreOp op);
 Optional<StrideInfo> computeStrideInfo(pmlc::dialect::pxa::AffineReduceOp op);
 
+// For a given block + memory access:
```
adstraw, line 123: +1 for the comments!

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
flaub: We could fix this by using a template, but you might not have noticed that `op` in both usages are actually different types.

flaub: An interface is planned to cover all of these ops that have similar properties. Once the interface is done, this and many other parts of the code can be simplified.

adstraw: I did miss that `op` was a different type.

PR #1260, opened 2020-08-07
Minor cleanups

flaub:


networks/oplib/resnet50.cc
```diff
@@ -34,38 +34,19 @@ edsl::Tensor block(             //
...
   if (use_shortcut_conv) {
     auto conv_1 = op::convolution(I, W[3])
                       .name(base_name.str() + "_branch1")
                       .strides(strides.vec())
-                      .dilations({1, 1})
-                      .data_dilations({1, 1})
-                      .autopad_mode(op::AutoPadMode::VALID)
-                      .input_layout(op::TensorLayout::NXC)
-                      .filter_layout(op::TensorLayout::XCK);
+                      .autopad_mode(op::AutoPadMode::VALID);
```
mars1026, line 49: looks like these are all just removing declarations which restate the default values for these members?

PR #1264, opened 2020-08-10
Add support for dumping passses to a directory

flaub:
Passes can be dumped to a directory by using the `PLAIDML_DUMP` env var. The directory specified will be created if it does not yet exist. It does not clear the directory, so care must be taken to clear it before each run.


flaub:
@dgkutnic This might be useful for you.

PR #1265, opened 2020-08-10
AffineLoad to PXA

adstraw:


pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -17,8 +17,7 @@ namespace {
 // we're going to be processing with std::next_permutation() --
 // e.g. if we used pointers as comparison values, our order of
 // iteration could vary run-to-run, creating non-determinism.
-template <typename V>
-class Orderer {
+template <typename V> class Orderer {
```
flaub, line 20: What version of clang-format are you using? We need to use version 9.

adstraw: I will get version 9, thanks.

adstraw: I couldn't figure out how to run clang-format on *just* my changes so I did a `git reset` and recommitted all changes (with clang-format-9 formatting) as a single commit, force pushed.

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -178,8 +189,8 @@ struct AffineReduceOpConversion
   LogicalResult
   matchAndRewrite(pxa::AffineReduceOp op, ArrayRef<Value> operands,
                   ConversionPatternRewriter &rewriter) const final {
-    auto source = rewriter.create<AffineLoadOp>(op.getLoc(), op.mem(), op.map(),
-                                                op.idxs());
+    auto source = rewriter.create<pxa::AffineLoadOp>(op.getLoc(), op.mem(),
```
adstraw, line 192: It's not clear to me whether this should be a pxa.load or an affine.load.

pmlc/dialect/pxa/analysis/uses.cc
```diff
@@ -73,7 +73,7 @@ IndirectAccessUsesIterator &IndirectAccessUsesIterator::operator++() {
 
 void IndirectAccessUsesIterator::skipNonAccess() {
   while (inner != IndirectUsesIterator()) {
-    if (isa<AffineLoadOp>(inner->getOwner()) ||
+    if (isa<pxa::AffineLoadOp>(inner->getOwner()) ||
```
adstraw, line 76: This may be overkill, but I hit a bug with namespace so I decided to make pxa vs. mlir namespace explicit everywhere for AffineLoad.

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
jbruestle: In PXA -> Affine we want to move back out of PXA and into affine, so the one we are creating wants to be mlir::AffineLoad.

pmlc/dialect/pxa/analysis/uses.cc
jbruestle: I think explicit definitely better.

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
adstraw: The code above works, for what it's worth.  I believe the pxa.load created from the reduce op is later lowered to affine.load.  That said, I will change to the mlir:: version.

jbruestle: Yeah, the greedy rewritter keeps rewritting until there are only legal op, but no need to do two rewrites when one will work.

PR #1267, opened 2020-08-11
Tile accumulation

YangleiZouIntel:
Tile accumulation pass. Check if any reduce op inside a affine.parallelfor is an aggregation op. If there is one, then it will potentially cause race condition, so perfom tiling and place the nonparallelizable part inside, and will be running on single work item. 

Currently single for loop and any ops outside of for loops will be turned into llvm and running on host side, need discussion.


pmlc/conversion/SCFToGPU/SCFToGPUPass.cpp
```diff
@@ -29,6 +29,88 @@ using namespace mlir::scf; // NOLINT
...
+  LogicalResult matchAndRewrite(mlir::scf::ForOp forOp,
+                                PatternRewriter &rewriter) const override;
+};
+
+// From LowerToCFGPass, avoid conversion in gpu::LaunchOp contents.
+LogicalResult ForLowering::matchAndRewrite(scf::ForOp forOp,
+                                           PatternRewriter &rewriter) const {
+  Operation *nested = &forOp.getBody()->front();
+  auto parentOp = forOp.getParentOp();
+  if (dyn_cast<scf::ForOp>(nested) || isa<scf::ForOp>(*parentOp)) {
```
YangleiZouIntel, line 46: This whole thing is added to turn single/unnested for loops to standard and then run on host side. I think maybe a new pass need to be created to turn single for loops and ops outside of any for loops to 2 nested for loops , so they can be turned into a gpu module and run on GPU instead of running on host side.

flaub:
I think the goal is to eventually drop the local copy of SCFToGPU and instead use the upstream version. Can we split this PR so that we have the TileAccumulationPass as one PR and then have the SCFToGPU changes come in a separate PR?

Could you also add some `.mlir` tests for just the TileAccumulationPass? We should be able to test it without any EDSL.

pmlc/dialect/pxa/transforms/BUILD
```diff
@@ -33,6 +33,7 @@ plaidml_cc_library(
         "stencil_gemm.cc",
         "test_analysis.cc",
         "tile.cc",
+        "tileaccumlate.cc",
```
flaub, line 36: rename to: `tile_accumulate.cc`

pmlc/dialect/pxa/transforms/passes.td
```diff
@@ -8,6 +8,11 @@ def AutoTileExample : FunctionPass<"pxa-autotile-example"> {
   let constructor = "pmlc::dialect::pxa::createAutoTileExamplePass()";
 }
 
+def TileAccumulate : FunctionPass<"pxa-tile-accumulate"> {
+  let summary = "Tile accumulate";
```
flaub, line 12: Could you add a slightly better description?

pmlc/dialect/pxa/transforms/tileaccumlate.cc
```diff
@@ -0,0 +1,118 @@
...
+    } else if (dyn_cast<AffineIfOp>(defop)) {
+      defop->walk([&](AffineReduceOp op) {
+        opRes = op.getResult().cast<mlir::OpResult>();
+      });
+    }
+  }
+  return opRes.getOwner();
+}
+
+bool isAggregation(AffineParallelOp op) {
```
YangleiZouIntel, line 54: Currently a few test examples work as expected. Testing and checking with more complex tests.

YangleiZouIntel:
> I think the goal is to eventually drop the local copy of SCFToGPU and instead use the upstream version. Can we split this PR so that we have the TileAccumulationPass as one PR and then have the SCFToGPU changes come in a separate PR?
> 
> Could you also add some `.mlir` tests for just the TileAccumulationPass? We should be able to test it without any EDSL.

Thanks. Yes, I will split the SCFToGPU changes to a separate PR and add `.mlir` tests.

pmlc/dialect/pxa/transforms/BUILD
YangleiZouIntel: Sure

pmlc/dialect/pxa/transforms/passes.td
YangleiZouIntel: Yes, I will add more descriptive words.

pmlc/dialect/pxa/transforms/tile_accumulate.cc
```diff
@@ -0,0 +1,88 @@
...
+  auto srcDef = getOriginalDef(op.getResult(0));
+  auto red = mlir::cast<PxaReduceOp>(srcDef);
+  // Get strides for output
+  auto si = *computeStrideInfo(red);
+  // Find all the accumulation indexes (stride 0 with respect to output) and
+  // tile them into an inner block
+  auto ranges = *op.getConstantRanges();
+  SmallVector<int64_t, 6> accumTile;
+  auto steps = op.steps().cast<ArrayAttr>().getValue();
+  // Track if both inner + outer loops would bee used
```
flaub, line 47: typo: `be`

pmlc/dialect/pxa/transforms/tile_accumulate.h
```diff
@@ -0,0 +1,13 @@
...
+
+#pragma once
+
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "pmlc/dialect/pxa/ir/ops.h"
+
+namespace pmlc::dialect::pxa {
+
+mlir::AffineParallelOp tileAccumulations(mlir::AffineParallelOp op,
+                                         bool skipTrivial = true);
```
flaub, line 11: Maybe `performTilingForAccumulations`

pmlc/dialect/pxa/transforms/nest_loops.cc
```diff
@@ -0,0 +1,85 @@
...
+unsigned getNestedIVCount(AffineParallelOp op) {
+  unsigned count = 0;
+  while (op) {
+    count += op.getIVs().size();
+    op = dyn_cast<AffineParallelOp>(op.getBody()->front());
+  }
+  return count;
+}
+
+void buildNestLoop(AffineParallelOp op) {
```
flaub, line 42: Maybe `buildNestedLoop`

pmlc/dialect/pxa/transforms/tile_accumulate.cc
```diff
@@ -0,0 +1,88 @@
...
+    op = performTiling(op, accumTile);
+  }
+  return op;
+}
+
+struct TileAccumulatePass : public TileAccumulateBase<TileAccumulatePass> {
+  void runOnFunction() final {
+    auto func = getFunction();
+    // Tile only the outermost loops
+    for (auto op : func.getBody().getOps<AffineParallelOp>()) {
```
flaub, line 75: Not sure if you need `func.getBody()`, if `func` has a single region, then `func.getOps` should work

PR #1270, opened 2020-08-12
Add best cost in autotile cost model

shiyao00:
With the knowledge of the previous best cost in autotiling pass, we could skip heavy tests in the cost model if the cost of the current plan is already not the best.


flaub:
Seems fine to pass extra context to the costModel function, so LGTM.

@jbruestle WDYT?

PR #1271, opened 2020-08-13
Add intel_gen target

earhart:


plaidml/testenv.h
```diff
@@ -77,6 +79,57 @@ class TestFixture : public ::testing::Test {
...
+  //
+  //    if (setIsSkipped("some_target")) {
+  //      return;
+  //    }
+  //
+  bool setSkipOnTarget(const char* target) {
+    if (Settings::get("PLAIDML_TARGET") == target) {
+      [target]() -> void {
+        GTEST_SKIP() << "Test skipped on " << target;
+      }();
```
flaub, line 117: This seems strange, why make a lambda and then immediately call it?

plaidml/testenv.h
```diff
@@ -77,6 +79,57 @@ class TestFixture : public ::testing::Test {
...
+  //     // Code to run the program and check its output.
+  //
+  // On the other hand, if the test's output isn't being checked
+  // against its source by FileCheck, one can simply use:
+  //
+  //    if (setIsSkipped("some_target")) {
+  //      return;
+  //    }
+  //
+  bool setSkipOnTarget(const char* target) {
```
flaub, line 113: It's odd to me that a method with a prefix of `set` returns a boolean.

plaidml/testenv.h
earhart: GTEST_SKIP() includes a `return`.

plaidml/testenv.h
earhart: I'm open to better names... or maybe this should be two methods for the two use cases?

tzerrell:
For the build failures on `INSTANTIATE_TEST_CASE_P`, it looks like most of those are coming from OpenVINO tests. I'll try to get my changes to get onto OV's `master` branch into a PR as promptly as I can to help with this (it's a big change, but if all goes well that cold happen today), although it's not going to resolve it -- `INSTANTIATE_TEST_CASE_P` is very common in that code. We might need to make updates in our OV fork? :/ I was hoping we wouldn't have to keep that fork alive for too long.

Not quite sure why this is only a failure in Mac builds...

plaidml/BUILD
```diff
@@ -176,10 +176,24 @@ string_flag(
 
 string_flag(
     name = "target",
-    values = ["llvm_cpu", "foo"],
+    values = ["llvm_cpu", "intel_gen"],
```
flaub, line 179: Perhaps we should leave this as being an open-set. In fact this will be a requirement soon.

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -820,6 +825,7 @@ TEST_F(CppEdsl, CumSum) {
   // CHECK: %{{.*}} = tile.contract add, none, %[[cst]], %{{.*}} {cons = #set{{[0-9]+}}, sink = #map{{[0-9]+}}, srcs = [#map{{[0-9]+}}]} : tensor<f32>, tensor<10xf32> -> tensor<10xf32>
   // CHECK: return %{{.*}} : tensor<10xf32>
   // clang-format on
+
```
flaub, line 828: Can we remove these newlines?

plaidml/edsl/tests/BUILD
```diff
@@ -1,18 +1,31 @@
...
 
 plaidml_cc_test(
     name = "cc_test",
     srcs = ["edsl_test.cc"],
-    tags = ["manual"],  # the `edsl_test.cc.test` rule will run this test via lit
     deps = [
         "//plaidml:testenv",
         "//plaidml/edsl",
         "//plaidml/exec",
+        "//pmlc/rt/vulkan",
```
flaub, line 14: Can we make this conditional on being part of a GPU target?

bzl/workspace.bzl
```diff
@@ -61,12 +61,14 @@ def plaidml_workspace():
...
+    #       pull the license directly from the gmock release.
     http_archive(
         name = "gmock",
-        url = "https://github.com/google/googletest/archive/release-1.8.0.tar.gz",
-        sha256 = "58a6f4277ca2bc8565222b3bbd58a177609e9c488e8a72649359ba51450db7d8",
-        strip_prefix = "googletest-release-1.8.0",
-        build_file = clean_dep("//bzl:gmock.BUILD"),
+        url = "https://github.com/google/googletest/archive/release-1.10.0.tar.gz",
+        sha256 = "9dc9157a9a1551ec7a7e43daea9a694a0bb5fb8bec81235d8a1e6ef64c716dcb",
+        strip_prefix = "googletest-release-1.10.0",
```
flaub, line 71: Let's avoid making dependency changes until we're sure that OpenVINO tests are ready for the switch.

flaub:
Can we just rely on the PLAIDML_DEVICE and PLAIDML_TARGET settings? There are two sources for these settings:

- Environment variables (with the same name)
- `~/.plaidml2` settings file

Right now, all of the CI machines have the proper `~/.plaidml2` files setup. So, it'd be a lot easier if we just removed the command line arguments and simply relied on the `plaidml::Settings` API to handle this.

earhart:
N.B. If we rely on `~/.plaidml2`, then we're restricted to one set of settings per build account, and we won't respect the build configuration; the tests will be less hermetic and it won't be as obvious when they're failing in some particular developer's configuration.  I'm also not sure there's a good way to set environment variables for test programs from bazel -- although I might easily be missing something.

plaidml/edsl/tests/edsl_test.cc
earhart: Done

plaidml/BUILD
earhart: Done

plaidml/edsl/tests/BUILD
earhart: Done

bzl/workspace.bzl
earhart: Done

flaub:
OK, then the challenge with this approach is to figure out a way to hook up the command line arguments with the CI build machine. We need something that turns the environment variables defined as part of buildkite tags (metadata) into the command line arguments.  There seems to be a tension here between hermcity for developers vs for CI build machines.

bzl/workspace.bzl
tzerrell: OpenVINO tests will need 1.10 googletest when we advance our OV reference, I have a variation of this change on my OV branch.

That said it's possible that this change will break some OV tests until I get fixes in; I'm working on getting the OV advancement PR ready as my main priority right now. But if you aren't seeing CI issues from this, it's probably fine.

earhart:
It seems like we want a variant in `plan.yml` for each hardware type we want to test on; right now, we're setting `env/PLAIDML_DEVICE` and `env/PLAIDML_TARGET`, but I think we could change that pretty easily to setting top-level keys `plaidml_device` and `plaidml_target`, and then in `buildkite.py`, we could pull those settings out of the variant and pass them to the bazel command line.

(OTOH -- I know you wrote that code; I could well believe there's something more subtle required.  :-)

On the gripping hand, #1259 removed `PLAIDML_DEVICE` and `PLAIDML_TARGET` from the test environment, so we're already using the bazel flag defaults to specify the configuration.

flaub:
For specifying the target, it's reasonable to do this in the `plan.yml`. However, this won't work for the device. Consider the case where a build machine has multiple identical devices. In this scenario, we want the device to be specified as part of the build environment, not part of the plan. Which means, we need map the tags that are specified on a per buildkite agent basis.

earhart:
Sounds like something `buildkite.py` can handle...?  I mean, it doesn't matter where it gets its source of "what's going on", what matters is that it passes it to the bazel command line.

pmlc/testing/gtest_main.cc
```diff
@@ -17,6 +21,38 @@ int main(int argc, char *argv[]) {
...
+    for (std::string line; std::getline(fs >> std::ws, line);) {
+      // Erase trailing whitespace
+      line.erase(std::find_if_not(line.rbegin(), line.rend(), [](int c) -> bool { return std::isspace(c); }).base(), line.end());
+      ss << sep << line;
+      sep = ':';
+    }
+
+    testing::FLAGS_gtest_filter = ss.str();
+  }
+
```
flaub, line 55: How does this interact with existing filters? Let's say I want to try out a 'skipped' test, is there a way to override the filter so that the one from the bazel command line takes precedence?

flaub:
Yes but then now `buildkite.py` has more test-specific goo in it. If we just have all tags mapped to env vars and then those get picked up by random processes, then there's nothing specific in `buildkite.py` to do that. Consider what we're going to have to do for OpenVINO, we'll run into the same issue again, and it's not clear we can easily add command line arguments or have the same command line arguments.

pmlc/testing/gtest_main.cc
earhart: Omit the filter file and specify the skip filter explicitly (which is easy, because the test output contains the full setting after the filter file's been applied).  It's hard to do this any other way, since the filters are patterns.

FWIW, I also thought about allowing comments in the file format, so that we could describe why things were being omitted -- YAGNI for now, though.

flaub:
I was hoping we could use the INSTANCIATE test suites that we discussed in chat. It seems like a more explicit way for each target to specify which tests it wants to run and also provide the ability to tune the tests (like specifying different data types or shapes depending on what's supported by the target).

earhart:
This seemed simpler and more direct, but sure, I can go back to that approach.

flaub:
I'm OK with landing this as is for now, and then have a plan for moving to something more flexible in the future (like as soon as we need it, which appears to be soon). The only thing I'd block this on is not having the integration with buildkite in place. Or rather, we need a plan on what changes we need to make to our CI infrastructure to make this appraoch work (RE the command line vs env vars).

earhart:
Okay, sounds good -- so, do we just need a way to smuggle PLAIDML_TARGET and PLAIDML_DEVICE through tests we don't control?

PR #1279, opened 2020-08-15
Support for 3rd party compilers

flaub:
* Provide support for allowing external compilers to decide which targets to include.
* Decouple lib and test in openvino bridge.
* All dialects must be registered, until this lands: https://reviews.llvm.org/D85622
* A target is registered via a static initializer.
* OpenVINO tests no longer hard-code the target, instead use the PLAIDML_TARGET env var.

### DIFF
--- a/plaidml/bridge/openvino/src/plaidml_plugin/plaidml_executable_network.cpp
+++ b/plaidml/bridge/openvino/src/plaidml_plugin/plaidml_executable_network.cpp
@@ -33,7 +33,7 @@ InferRequestInternal::Ptr PlaidMLExecutableNetwork::CreateInferRequestImpl(Input
     IVLOG(2, "output: " << kvp.first);
     outputs.push_back(tensorIOMap_.at(kvp.first));
   }
-  auto program = edsl::ProgramBuilder("ie", outputs).target("llvm_cpu").compile();
+  auto program = edsl::ProgramBuilder("ie", outputs).compile();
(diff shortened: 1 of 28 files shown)


plaidml/bridge/openvino/src/plaidml_plugin/plaidml_executable_network.cpp
```diff
@@ -33,7 +33,7 @@ InferRequestInternal::Ptr PlaidMLExecutableNetwork::CreateInferRequestImpl(Input
     IVLOG(2, "output: " << kvp.first);
     outputs.push_back(tensorIOMap_.at(kvp.first));
   }
-  auto program = edsl::ProgramBuilder("ie", outputs).target("llvm_cpu").compile();
+  auto program = edsl::ProgramBuilder("ie", outputs).compile();
```
flaub, line 36: @tzerrell Note, you'll have to ensure `PLAIDML_TARGET` is set properly, one way to ensure this is to run `//plaidml:setup`.

plaidml/edsl/ffi.cc
```diff
@@ -114,8 +112,6 @@ void plaidml_edsl_init(  //
   ffi_wrap_void(err, [&] {
     std::call_once(is_initialized, []() {
       IVLOG(1, "plaidml_edsl_init");
-      registerAllDialects();
-      registerAllPasses();
```
flaub, line 118: This is the key bit, we can't expect all the dialects/passes will be known to the EDSL lib. Instead we make it a linker problem. This way 3rd parties can create their own version of `libplaidml.so` that contains the targets they wish to include.

PR #1282, opened 2020-08-17
Add "comp" dialect for device-host modeling

kdobros:
This change adds "comp" dialect that allows to model asynchronous
execution on device other than host.
It provides operations for managing memory transfers, operation ordering
with events and synchronizing from host with wait.


pmlc/dialect/comp/ir/interfaces.td
```diff
@@ -0,0 +1,80 @@
...
+
+  let methods = [
+    InterfaceMethod<"", "::mlir::Value", "getExecEnv", (ins), [{}], [{
+      ConcreteOp op = ::mlir::cast<ConcreteOp>(this->getOperation());
+      return op.execEnv();
+    }]>
+  ];
+}
+
+def COMP_ScheduleOpInterface : COMP_OpInterface<"ScheduleOpInterface"> {
```
earhart, line 22: I'd probably call this AsyncOp or SchedulableOp; this interface isn't being used to perform scheduling, it's the thing being scheduled.

pmlc/dialect/comp/ir/interfaces.cc
```diff
@@ -0,0 +1,75 @@
+#include "pmlc/dialect/comp/ir/interfaces.h"
```
earhart, line 1: No copyright on this one?  :-)

pmlc/dialect/comp/ir/interfaces.td
```diff
@@ -0,0 +1,80 @@
...
+  let description = [{
+    Interface for operations that are scheduled on execution environment.
+    Provides access for dependant event and produced event.
+    Operations implementing this interface are expected to also implement
+    ExecEnvOp interface.
+  }];
+
+  let methods = [
```
earhart, line 30: This seems to be defining a completely orthogonal mechanism for specifying dependencies between operations.  Is there any chance you could use MLIR's SSA arguments and results to do the same thing?  Or does that not work for some reason?  (If so, please add a comment.)

pmlc/dialect/comp/ir/ops.td
```diff
@@ -0,0 +1,238 @@
...
+                  , Results<(outs COMP_AnyEvent:$outEvent)> {
+  let summary = "Schedules function in region for execution on device.";
+  let description = [{
+    This is a wrapper around gpu.launch_func that schedules function
+    to execute on environment and provides synchronization with
+    event dependencies and output event.
+    The region is expected to have only this single gpu.launch_func operation
+    and implicit terminator - comp.schedule_end.
+    Example:
+    %ev = "comp.launch_func"(%execEnv, %depEvent) ( {
```
earhart, line 102: Need to update this example

pmlc/dialect/comp/ir/ops.td
```diff
@@ -0,0 +1,238 @@
...
+  }];
+}
+
+def COMP_ScheduleEnd : COMP_Op<"schedule_end", [Terminator]> {
+  let summary = "End of scheduling block.";
+}
+
+def COMP_ScheduleSingleBlock : SingleBlockImplicitTerminator<"ScheduleEnd">;
+
+def COMP_ScheduleFunc : COMP_Op<"schedule_func",
```
earhart, line 87: Maybe "ExecFunc"?  (Scheduling is a bit overloaded in this context, and the idea of "scheduling" is already being used for ScheduleOpInterface.)

pmlc/dialect/comp/ir/interfaces.td
earhart, line 22: I'm also not sure the concept of "do something asynchronously" should be represented at this level at all -- what's synchronous and what's asynchronous may vary from API to API.  IMHO, it might be better to simply represent the dependencies between operations at this level, and let the lowering to target API consider how to handle asynchronicity.

pmlc/dialect/comp/ir/ops.td
```diff
@@ -0,0 +1,238 @@
...
+                  , Arguments<(ins COMP_AnyExecEnv:$execEnv,
+                                   AnyMemRef:$deviceMem)> {
+  let summary = "Deallocates memory previously allocated for use in executon environment.";
+
+  let assemblyFormat = [{
+    $execEnv $deviceMem attr-dict `:` functional-type(operands, results)
+  }];
+}
+
+def COMP_ScheduleWrite : COMP_Op<"schedule_write",
```
earhart, line 161: How do you want to handle shared-memory devices, and devices that can expose their memory in the system physical address space (for write-combining transfers and such)?

pmlc/dialect/comp/ir/interfaces.cc
kdobros: Added

pmlc/dialect/comp/ir/interfaces.td
kdobros: I have added short description of advantages of event based approach over SSA dependencies to `EventType` description.

pmlc/dialect/comp/ir/interfaces.td
kdobros: I'm not sure if I understand this, all operations that implement this interface are expected to schedule work on device.
Also there is no differentiation between operations that are scheduled and that perform scheduling inside this dialect. Only operation I could see being "schedulable" is `gpu.launch_func`, but that's not a part of dialect.

pmlc/dialect/comp/ir/ops.td
kdobros: Updated

pmlc/dialect/comp/ir/ops.td
kdobros: ExecFunc to me suggests that operation executes and is finished after this operation. In general the idea is that operations named "schedule_XYZ" don't provide guarantee of finishing unless waited for, or even starting execution immediately.
Also I see no point in making a difference in naming for enqueuing/dispatching/scheduling kernel compared to any other task performed on device.

pmlc/dialect/comp/ir/interfaces.td
kdobros: I have added dialect description addressing this concern. I agree that there should be no requirement for asynchronous execution here and I have tried to avoid it. All the operations in `comp` dialect could be lowered to synchronous API in which case events and explicit dependencies can be ignored/removed.
This is ensured by the fact that "schedule" operations don't guarantee time after scheduling that operations will finish executing (it can be immediately)  and that all events are SSA values so operations can only depend on previously scheduled operations.

pmlc/dialect/comp/ir/ops.td
kdobros: In my original draft I have implemented this as `schedule_map : (execenv, memref, event...) -> (event, memref)` and matching `schedule_unmap`, but I have seen no practical use for this so far.
I'd like it to be implemented later, when it would be needed.

kdobros:
@earhart @jbruestle Sorry for the long turn around, I had a lot of other tasks outside of plaidml to finish.
I have added documentation for types and general overview of the dialect.

jbruestle (approved):

earhart (approved):

PR #1283, opened 2020-08-17
WIP: stdx subgroup lowering

abialas1:
Lowering vector broadcast pattern to stdx.subgroup_broadcast. And also stdx.subgroup_broadcast to spv.GroupBroadcast.


pmlc/dialect/stdx/transforms/subgroup_broadcast.cc
```diff
@@ -0,0 +1,207 @@
...
+///    store %1, %arg1[%idx] : memref<64xf32>
+///    scf.yield
+///  } { mapping = [
+///    {processor = 1, map = affine_map<(d0) -> (d0)>, bound = affine_map<(d0)
+///    -> (d0)>}, {processor = 0, map = affine_map<(d0) -> (d0)>, bound =
+///    affine_map<(d0) -> (d0)>}] }
+///
+/// This pass requires --convert-parallel-loops-to-gpu afterwards
+
+void runOnParallel(scf::ParallelOp loopOp) {
```
jbruestle, line 60: The pattern matching here is too specific.  Additionally, the 'devectorization' needs to happen universally thoughout the scf.parallel (i.e. all vector reads/writes need to be 'scalarized', and vector computational operations (i.e. mulf) need to be retyped to operate on scalars.  Specifically the transforms need to be roughly:

1) transfer_read X[i, j, k] -> load[i, j, k + sid]
2) transfer_write  a, X[i, j, k] -> store a, X[i, j, k + sid]
3) b = extractelement a[e] vec, broadcast b -> stdx.subgroup_broadcast a, e
4) Change the typing of any vector versions of std.mulf, etc, with scalar versions of the same.  You may want to rely on the op implementing VectorUnrollOpInterface to detect such ops as is done here: https://github.com/plaidml/plaidml/blob/plaidml-v1/pmlc/dialect/pxa/transforms/vectorize.cc#L84

Please let me know if this makes sense.  I can try to provide a more complete set of lit test by EOD Tuesday if that would help.  There is an open question in my mind regarding two points:

1) Should we consider converting adding some sort of attribute to the the load/store resulting from  the transfer_read/transfer_write so that it is marked for lowering to sub_group_block_read/sub_group_block_write or should we rely on the later lowering to detect this due to the stride 1 relative to SID + the alignment?

2) Is it reasonable long term 'devectorize' all computational ops greedily?  This means that any use of vectorization of ops on a per-thread basis would need to done after the subgroup lowering.  To reproduce the behavior of plaidml-v0, which had excellent performance, we can do things greedily, but I wonder if we are painting ourselves into a corner.

pmlc/dialect/stdx/transforms/subgroup_broadcast.cc
```diff
@@ -0,0 +1,207 @@
...
+///  %c1 = constant 1 : index
+///  %c8 = constant 8 : index
+///  %c64 = constant 64 : index
+///  scf.parallel (%i, %sid) = (%c0, %c0) to (%c64, %c8) step (%c8, %c1) {
+///    %idx = addi %i, %sid : index
+///    %0 = load %arg0[%idx] : memref<64xf32>
+///    %1 = stdx.subgroup_broadcast %0, %c1_i32 : f32
+///    store %1, %arg1[%idx] : memref<64xf32>
+///    scf.yield
+///  } { mapping = [
```
jbruestle, line 53: On further reflection, I think we should probably make this test not check the parallel loop mapping, since it seems independent of the transform we really care about and is a result of  --convert-parallel-loops-to-gpu  as you mention.

pmlc/dialect/stdx/transforms/BUILD
```diff
@@ -24,13 +24,19 @@ plaidml_cc_library(
...
     ],
     hdrs = [
         "passes.h",
     ],
     deps = [
         ":gen-pass",
+        "@llvm-project//mlir:SCFDialect",
+        "@llvm-project//mlir:GPUDialect",
+        "@llvm-project//mlir:GPUTransforms",
+        "@llvm-project//mlir:VectorOps",
```
flaub, line 38: Would be good to keep other dialects out of this one. If we need to perform a conversion from one dialect to another, we should put it in the `conversion` directory.

pmlc/dialect/stdx/ir/ops.td
```diff
@@ -110,4 +110,14 @@ def TanOp : StdX_UnaryFuncOp<"tan"> {
   let summary = "tangent";
 }
 
+def SubgroupBroadcastOp : StdX_Op<"subgroup_broadcast", [NoSideEffect, AllTypesMatch<["value", "result"]>]> {
+  let summary = "broadcast to all elements in subgroup";
+  let arguments = (ins AnyType:$value, AnyType:$localid);
```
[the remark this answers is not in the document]
abialas1: thanks, I will update those accordingly

pmlc/dialect/stdx/transforms/subgroup_broadcast.cc
abialas1: So do we want to remove adding of the mapping also from the transform logic as well? Will it be correctly mapped in the convert-parallel-loops-to-gpu?

pmlc/dialect/stdx/transforms/BUILD
abialas1: I can do that, will create vector_to_stdx directory and move the implementation there. I will do this at later step

abialas1:
Thanks Jeremy, already fixed the Linux warning.
There is also second problem with stdx->spv lowering, as the llvm upstream version does not have my changes for opGroupBroadcast yet. Now the question, do you guys plan to make update sometime soon, or should I provide additional PR for plaidml/llvm_project with OpGroupBroadcast?
For now I removed stdx->spv lowering to have the rest checkin. Please take a look and merge after approval.

PR #1285, opened 2020-08-19
Fix bazel buildkite farm

flaub:
Currently the OpenVINO tests use `plaidml::Settings` to determine which device/target to run on. Therefore, we need to have these env vars included so that CI will work. A PR that will use command line args instead to make things more hermetic is very welcome!

@tzerrell Is it possible to use `//plaidml:device` and `//plaidml:target` command line arguments?


tzerrell:
I'll investigate tomorrow with my larger OV PR. I hope so!

PR #1286, opened 2020-08-19
Add getByteSize in util

shiyao00:


pmlc/util/util.cc
```diff
@@ -32,4 +33,24 @@ void UpdateFuncOpType(Operation *op) {
...
 }
 
+uint64_t getByteSize(MemRefType type) {
+  int64_t offset;
+  llvm::SmallVector<int64_t, 8> strides;
+  if (failed(mlir::getStridesAndOffset(type, strides, offset))) {
+    throw std::runtime_error("Could not retrieve strides");
+  }
+  auto sizes = type.getShape();
+  unsigned total = 0;
```
flaub, line 43: IIUC, some platforms define `unsigned` to be 16-bit, so perhaps we should use `uint64_t` here to be safe.

shiyao00: done

PR #1287, opened 2020-08-19
Fix i1 memref in Tile function arguments on GPU

YangleiZouIntel:
* alloc and dealloc new memref to replace I1 buffer parameter in function.


pmlc/dialect/stdx/tests/i1_to_i32.mlir
```diff
@@ -20,3 +20,32 @@ func @simpleLoadStore(%i: index, %j: index) {
   // CHECK: store %{{.*}}, %{{.*}}[%{{.*}}, %{{.*}}] : memref<20x10xi32>
   return
 }
+
+// CHECK-LABLE: func @simpleI1Param
```
abialas1, line 24: typo: should be LABEL

pmlc/dialect/stdx/transforms/i1_to_i32.cc
```diff
@@ -93,11 +104,132 @@ void populateI1StorageToI32(MLIRContext *context,
...
 
+// Adaptor for building loop nests for the specific memRef shape
+void buildLoopForMemRef(
+    OpBuilder &builder, Location loc, Value memRef,
+    function_ref<void(OpBuilder &, Location, ValueRange)> bodyBuilder) {
+  auto memRefType = memRef.getType().dyn_cast_or_null<MemRefType>();
+  if (!memRefType)
+    return;
+
+  mlir::edsc::ScopedContext context(builder, loc);
```
abialas1, line 115: Is the context used anywhere?

YangleiZouIntel: It is from EDSC declarative builders API.  Please see https://github.com/llvm/llvm-project/blob/master/mlir/docs/EDSC.md
Here we implemented memory copy in standard dialect instead of scf.for to avoid converting it to a gpu module.

pmlc/dialect/stdx/tests/i1_to_i32.mlir
YangleiZouIntel: Fixed, thanks.

jbruestle:
Deprecated by #1311

PR #1289, opened 2020-08-19
Add per-target unit tests (outside bazel)

earhart:


ci/pipeline.yml
```diff
@@ -10,7 +10,7 @@ steps:
     command: '{{python}} ci/buildkite.py build {{name}}'
     key: '{{name}}'
     agents:
-      PLATFORM: '{{name}}'
+      PLAIDML_TARGET: '{{name}}'
```
flaub, line 13: This is an important abstraction because platform may encompass more than just a plaidml target. For instance, we want to be able to tag machines that are capable of running non-plaidml compilers (like tensorflow, or whatnot).

plaidml/BUILD
```diff
@@ -36,6 +36,7 @@ plaidml_cc_library(
         "//pmlc/target/demo",
         "//pmlc/target/intel_gen",
         "//pmlc/target/x86",
+        "//pmlc/rt/vulkan",
```
flaub, line 39: This shouldn't be added here, it should be part of the intel_gen target.

earhart:
FYI -- This is only an experiment; I wanted to see if it ran the way I expected.  I'll let you know as soon as it's ready for review.  :-)

ci/pipeline.yml
earhart: WAI -- experimental code

plaidml/BUILD
earhart: WAI -- experimental code

earhart:
@flaub I think this is ready.  PTAL?

bzl/plaidml.bzl
```diff
@@ -141,3 +154,178 @@ plaidml_settings = rule(
     },
     implementation = _plaidml_settings_impl,
 )
+
+
+def _plaidml_target(settings, attr):
+    return {"//plaidml:target": attr.plaidml_target}
+
+
+plaidml_target = transition(
```
flaub, line 163: Can you add some comments as to what this is and what it's doing?

bzl/plaidml.bzl
```diff
@@ -141,3 +154,178 @@ plaidml_settings = rule(
...
+
+    return [DefaultInfo(runfiles = runfiles, executable = ctx.outputs.executable)]
+
+
+_plaidml_device_test_builder = rule(
+    attrs = {
+        "plaidml_target": attr.string(mandatory = True),
+        "tests": attr.label_list(allow_empty = False, cfg = plaidml_target),
+        "_tpl": attr.label(default = "//bzl:device_test_builder.tpl.py", allow_single_file = True),
+        "_whitelist_function_transition": attr.label(
```
flaub, line 209: What is this all about?

ci/pipeline.yml
```diff
@@ -16,11 +16,23 @@ steps:
...
 
 # test phase
 
+  - label: ':linux:unit_tests/intel_gen'
+    agents:
+      PLAIDML_TARGET: intel_gen
+    command: |
+      python3 ci/buildkite.py unit linux_x86_64-opt intel_gen
+    env:
+      PLAIDML_TARGET: intel_gen
```
flaub, line 33: Is this still needed? I thought we wanted to use command line args instead? Also I see the environment being set from the test runner.

bzl/plaidml.bzl
earhart: Done

bzl/plaidml.bzl
earhart: Added comments in the code

ci/pipeline.yml
earhart: The environment variables set in the test runner come from plan.yml; I preserved that code in case there are environment variables that aren't set in pipeline.yml.

We still have tests that use the environment variables, and I don't think it hurts to set them.

bzl/plaidml.bzl
```diff
@@ -141,3 +154,220 @@ plaidml_settings = rule(
...
+
+    ctx.actions.write(ctx.outputs.executable, "\n".join(script), is_executable=True)
+
+    return [DefaultInfo(executable = ctx.outputs.executable)]
+
+
+# A rule to create a device-specific test runner script, which is
+# packaged into an out-of-bazel device-specific unit testing package.
+#
+# Note that the device-specific tests are built using a transition to
```
flaub, line 276: When you say "device-specific", do you actually mean "target-specific". I don't see anything in here that specifies the device, all I see is mentioning the target. So I guess I'm a bit confused.

earhart: Sorry -- yes, target-specific.  I'll reword for clarity.

bzl/plaidml.bzl
flaub: What are we transitioning from and to?

bzl/plaidml.bzl
flaub: Great! Thanks, that makes a ton more sense to me now.

ci/pipeline.yml
flaub: So how would the device be specified? Is that thru an environment variable? Let's say we have multiple devices on a machine and we want one agent to be responsible for one device.

bzl/plaidml.bzl
earhart: We're transitioning `//plaidml:target`.

So, for example:

* `bazel build --//plaidml:target=llvm_cpu //plaidml/edsl/tests:cc_test` -- builds `cc_test` for `llvm_cpu`

* `bazel build --//plaidml:target=intel_gen //plaidml/edsl/tests:cc_test` -- builds `cc_test` for `intel_gen`
* `bazel build --//plaidml:target=llvm_cpu //plaidml:intel_gen_test` -- builds `//plaidml/edsl/tests:cc_test` for `intel_gen` and builds `bazel-bin/plaidml/intel_gen.tar.gz` from it.  This happens regardless of what you specify for `--//plaidml:target` on the command line.

flaub:
I guess I'm OK with it, just seems like a lot of infrastructure for the skip list implementation. I might come along later and clean this all up.

ci/pipeline.yml
earhart: I think we'd tag the agent with the device, set the device constraint in pipeline.yml (possibly starting to construct the unit test block from plan.yml at that point), and then either pass it as an environment variable or pass it as a command-line parameter through buildkite.py.  (I prefer the latter, just to make things more explicit, but I don't have strong feelings about the device setting.)

We might want to do this as part of implementing device selection.

PR #1291, opened 2020-08-20
Prevent over-fusion via a memory acvitity threshold

flaub:


pmlc/dialect/pxa/transforms/fusion.cc
```diff
@@ -171,17 +179,69 @@ struct FusionInfo {
         bInfo.op.emitRemark("Mapping is not 1 to 1");
         return false;
       }
+      newAs.insert(argA);
```
jbruestle, line 182: I suspect that leaving things in aToB when the plan fails is actually incorrect behavior (i.e. each attempt should be from scratch, with empty mappings).  If you clear aToB and bToA at the top of considerPlan, I think it will have the same effect for less code than the newA's etc.  A better design would probably be to have an object per plan attempt so we don't need to clear (i.e. we using too monolithic of an object for fusion state tracking), but given the complexity of the pass, we should probably refactor only a bit at a time, so I'd leave that out of this PR.

PR #1294, opened 2020-08-20
Advance OpenVINO reference

tzerrell:
Advance our OpenVINO reference from 2020.2 to a recent(ish) master commit. This depends on https://github.com/plaidml/openvino/pull/27 landing (and will need to be slightly updated after that lands), so it is marked as a draft for now to prevent accidental landing.

I approached a few things in a somewhat hacky way and have not figured out how to clean them. I noted these with inline comments to call attention to them in case any reviewers have suggestions.


BUILD
```diff
@@ -67,3 +68,17 @@ toolchain(
     toolchain = ":py_runtime_pair",
     toolchain_type = "@rules_python//python:toolchain_type",
 )
+
+# TODO
+write_file(
+    name = "ov_plugins_xml",
+    out = "_solib_linux_x86_64/plugins.xml",
```
tzerrell, line 75: I don't like creating this file out of the root-directory BUILD, but I didn't see another way to get it in the right place. I'm also concerned about the use of the presumably machine-specific `_solib_linux_x86_64`, but I'm going to look for the output of the CI machines to see what happens with that.

plaidml/bridge/openvino/src/plaidml_plugin/plaidml_plugin.cpp
```diff
@@ -15,6 +15,10 @@
 #include "ie_plugin_config.hpp"
 #include "inference_engine.hpp"
 
+// TODO: The awkward include of a *.cpp is to get `ngraph_register_interpreter_backend`. See below where it's used for
+// more info.
+#include "ngraph/test/runtime/interpreter/int_backend.cpp"  // NOLINT[build/include]
```
tzerrell, line 20: I could avoid this by just writing out the contents of the function I'm including here: `ngraph_register_interpreter_backend`, as this is a short function. This is needed for the single_layer_tests to run. However, it really seems like that should be happening over on the testing side, rather than us registering the nGraph interpreter as part of the PlaidML plugin. I can't seem to find a place it's happening though, even for the other plugin tests; maybe I'm missing something?

plaidml/bridge/openvino/tests/functional/plaidml/single_layer_tests/activation_tests.cpp
```diff
@@ -1,328 +0,0 @@
...
-// SPDX-License-Identifier: Apache-2.0
-//
-
-#include "gtest/gtest.h"
-
-#include "ie_core.hpp"
-#include "ie_plugin_ptr.hpp"
-#include "ir_gen_helper.hpp"
-
-#include "cpp/ie_cnn_net_reader.h"
```
tzerrell, line 11: I'm removing these deprecated tests as this file no longer exists. It could probably be fixed, but since the tests are deprecated anyway, I'm thinking it might not be worth it. This does leave us with just one main path through the OV code, rather than 2, but I'm hoping that problem will be addressed soon by getting some full networks running.

vendor/openvino/workspace.bzl
```diff
@@ -23,10 +23,11 @@ def openvino_workspace():
         build_file = clean_dep("//vendor/openvino:ngraph.BUILD"),
     )
 
+    # TODO: Update the commit once the OV-repo PR lands
```
tzerrell, line 26: Will need to change this once https://github.com/plaidml/openvino/pull/27 lands

vendor/openvino/openvino.BUILD
```diff
@@ -207,9 +193,14 @@ cc_library(
             "inference-engine/src/inference_engine/os/lin/*.cpp",
         ]),
     }),
+    hdrs = glob([
+        # "inference-engine/src/readers/ir_reader/ie_ir_version.hpp",  # TODO: Ok to remove?
```
tzerrell, line 197: Here (and I think a few other places in this BUILD file) I seem to not be listing any *.h or *.hpp files from the directory `inference-engine/src/readers/ir_reader` (in this case) and everything builds fine; however, if I remove that directory from the `includes` list the build breaks. I'm confused as to how that is possible.

tzerrell:
Also this does not include the target/device work I mentioned I'd look into from https://github.com/plaidml/plaidml/pull/1285, I'll follow up on that either in this PR or a separate one depending on difficulty of that work & speed of review.

plaidml/bridge/openvino/src/plaidml_plugin/plaidml_executable_network.cpp
```diff
@@ -44,7 +44,7 @@ PlaidMLExecutableNetwork::PlaidMLExecutableNetwork(const ICNNNetwork& network, c
   IVLOG(2, "Layers:");
   for (auto& node : fcn->get_ordered_ops()) {
     IVLOG(2, "  " << node->description() << ": " << node->get_name() << "... " << node->get_friendly_name());
-    if (node->is_constant()) {
+    if (ngraph::op::is_constant(node)) {
```
earhart, line 47: Any chance ADL works here?

plaidml/bridge/openvino/tests/functional/plugin/plaidml/shared_tests_instances/single_layer_tests/BUILD
```diff
@@ -10,31 +10,64 @@ TAGS = [
...
 DEPS = [
-    "//pmlc/util:logging",
-    "@gflags",
-    "@gmock//:gtest_main",
-    "@openvino//:common_test_utils",
-    "@openvino//:helpers",
-    "@openvino//:inference_engine",
-    "@openvino//:shared_plugin_tests",
+    "//plaidml/bridge/openvino/tests/functional/plugin/plaidml/shared_tests_instances",
+    # "//pmlc/util:logging",
```
earhart, line 17: Can you remove this commented code?

flaub:
LGTM, all things considered. I don't know any better way to fix the xml generation, unless we can somehow patch upstream OpenVINO perhaps. It seems like we'll need to retain the tags to prevent macos/windows builds from failing, or spend time getting those builds to work (like we'd need a platform-specific select for the `solib` goo)

PR #1298, opened 2020-08-21
Enable tests on windows

flaub:


flaub:
@drgone This is blocked on upgrading our CI environment with bare-metal windows machines.

PR #1303, opened 2020-08-24
subgroups_broadcast lower to spirv

abialas1:
subgroups_broadcast lower to spirv after recent MLIR upstream update.


pmlc/conversion/gpu_to_spirv/passes.td
```diff
@@ -0,0 +1,11 @@
+#ifndef __PMLC_CONVERSION_GPU_TO_SPIRV_PASSES__
+#define __PMLC_CONVERSION_GPU_TO_SPIRV_PASSES__
+
+include "mlir/Pass/PassBase.td"
+
+def GPUToSPIRVCustom : Pass<"gpu-to-spirv-custom", "mlir::ModuleOp"> {
+  let summary = "Custom lowering from GPU to SPIRV including stdx dialect";
+  let constructor = "pmlc::conversion::gpu_to_spirv::createGPUToSPIRVCustomPass()";
+}
```
flaub, line 9: You'll need a `dependentDialects` field to tell the infrastructure that this pass generates other dialects.

abialas1: Thanks Frank, seems like this is something new that came with the last MLIR update? Can You write what is the actual purpose of this since it also works without these? I will add this change, but not sure if I should do it right now, or with my next change for lowering to subgroups_block_reads/writes that will be this week.

flaub: My understanding is that it is designed to allow dialects to be declared on-demand rather than having them be registered all up-front. Note that with `pmlc-opt`, all dialects are registered up front, so you'll likely not see an issue there. However, the new design of the pass pipeline system is to allow MLIR to only load dialects when they are needed. To accomplish this, first the input dialects are registered. In our case, we register `tile`, `eltwise`, and the `std` dialects. Next, as passes are executed, a pass needs to declare which dialects it might produce or emit as part of its transformation. So for instance, the `tile_to_pxa` conversion will declare `pxa` as a `dependantDialect`.

I don't think this is a critical issue; there isn't even any good documentation on this feature yet. Probably don't worry about it until it's convenient or it actually becomes a problem.

PR #1312, opened 2020-08-25
Get SinH etc to lower to GPU dialect.

jbruestle:
As of this commit, many of the remaining EDSL tests that fail are caused by not having lowerings for our various stdx ops in https://github.com/plaidml/plaidml/blob/plaidml-v1/pmlc/conversion/gpu_to_spirv/gpu_to_spirv.cc.  At this point it should be a straightforward (if tedious) process to add such lowerings, and in many cases, the needed additions to the SPIRV dialect.

### DIFF
--- a/pmlc/target/intel_gen/pass_detail.h
+++ b/pmlc/target/intel_gen/pass_detail.h
@@ -1,7 +1,10 @@
 #pragma once
 
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/GPU/GPUDialect.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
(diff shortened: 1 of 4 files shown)


pmlc/target/intel_gen/pass_detail.h
```diff
@@ -1,7 +1,10 @@
 #pragma once
 
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/GPU/GPUDialect.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/SCF/SCF.h"
+
```
flaub, line 7: Spurious newline

PR #1315, opened 2020-08-27
Generalized cache transformations

flaub:


pmlc/dialect/pxa/transforms/cache.cc
```diff
@@ -2,12 +2,20 @@
 
 #include "pmlc/dialect/pxa/transforms/cache.h"
 
+#include <memory>
+#include <utility>
+
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Affine/IR/AffineValueMap.h"
+#include "mlir/Support/DebugStringHelper.h"
```
jbruestle, line 10: Is this used?

pmlc/dialect/pxa/transforms/cache.cc
```diff
@@ -251,10 +269,191 @@ Optional<CacheInfo> cacheReduce(AffineParallelOp par, PxaReduceOp reduce) {
...
+          createCopyLoop(builder, loc, entry.rap.innerCount, finalValue, memref,
+                         zeroOffset, entry.rap.outer, AtomicRMWKind::assign);
+      copyLoop.getOperation()->setAttr("cache_out", builder.getUnitAttr());
+      finalUse.set(copyLoop.getResult(0));
+    }
+  }
+}
+
+struct CachePass : public CacheBase<CachePass> {
+  void runOnFunction() final {
```
jbruestle, line 425: Some of this is a bit more use case specific than it needs to be, but we can refactor it when we need to.

flaub: Yeah, and also this is meant for testing only at the moment.

PR #1316, opened 2020-08-28
Initial TensorFlow Bridge

dgkutnic:


bzl/workspace.bzl
```diff
@@ -136,6 +143,14 @@ def plaidml_workspace():
         build_file = clean_dep("//vendor/tbb:tbb.BUILD"),
     )
 
+    http_archive(
```
flaub, line 146: Let's keep the tensorflow stuff out of `workspace.bzl`. This file gets used by downstream repositories that may or may not want tensorflow included.

bzl/bazel.rc
```diff
@@ -25,6 +25,9 @@ build --action_env=HTTPS_PROXY
 build --action_env=no_proxy
 build --action_env=http_proxy
 build --action_env=https_proxy
+build --action_env=XLA_DUMPDIR="/tmp/hlo_module"
```
flaub, line 28: Could you keep at the TF-specific changes together in this file? We might do the integration with TF different in the future.

flaub:
Is the FileCheck integration based off of what TF is doing upstream?

bzl/bazel.rc
```diff
@@ -101,3 +101,13 @@ test:asan --test_env=ASAN_ENABLE=1
 test:asan --test_env=ASAN_OPTIONS
 test:asan --test_env=ASAN_SYMBOLIZER_PATH
 test:asan --test_env=DYLD_INSERT_LIBRARIES
+
+# For building tensorflow
+build --action_env=LD_PRELOAD # Needed on *nix to prevent malloc errors in tensorflow
+build --action_env=XLA_DUMPDIR="/tmp/hlo_module"
+build --action_env=XLA_FLAGS=--xla_dump_to="/tmp/hlo_module"
+build --action_env=TF_XLA_FLAGS="--tf_xla_auto_jit=2 --tf_xla_cpu_global_jit --tf_xla_min_cluster_size=0"
```
jd-bartlett96, line 109: I think having this here will generate & dump the hlo module whenever TF is called via python within the repo, instead of just when necessary for testing. Probably not a real issue but I believe this does make it unnecessary to specify these flags again in the BUILD file

plaidml/bridge/tensorflow/tests/BUILD
```diff
@@ -0,0 +1,109 @@
...
+    name = "conv_op_test_gen",
+    srcs = ["conv_op_test_gen.py"],
+    testonly = 1,
+)
+
+genrule(
+    name = "conv_op_test_genrule",
+    srcs = [],
+    outs = ["conv_op_test.h.inc"],
+    cmd = "XLA_FLAGS=\"--xla_dump_to=/tmp/hlo_module\" TF_XLA_FLAGS=\"--tf_xla_auto_jit=2 --tf_xla_cpu_global_jit --tf_xla_min_cluster_size=0\" $(location :conv_op_test_gen) --output $@",
```
jd-bartlett96, line 65: See other comment; this test runs fine even when these flags are removed here.

plaidml/bridge/tensorflow/service/compiler.cc
```diff
@@ -0,0 +1,762 @@
...
+          instr_map.insert(std::make_pair(cur_instr_id, op));
+          break;
+        }
+        case HloOpcode::kTranspose: {
+          // Tensor transpose operation
+          // TODO: test correctness, see if the axes operand is needed
+          auto op = plaidml_op::transpose(instr_map[operand_ids[0]]);
+          instr_map.insert(std::make_pair(cur_instr_id, op));
+          break;
+        }
```
jd-bartlett96, line 670: A fix to this was included in the shape op testing PR

plaidml/bridge/tensorflow/tests/BUILD
dgkutnic: I think I prefer specifying them in the `BUILD` file rather than the `bazelrc`, but I'm open to hearing any thoughts on why I should do it the other way around.

jd-bartlett96: No I agree; I think having them in the BUILD file makes more sense since those flags are only really needed for the testing (as far as I'm aware)

plaidml/testenv.h
```diff
@@ -36,7 +36,8 @@ class TestFixture : public ::testing::Test {
     EXPECT_EQ(a, b);
   }
 
-  void compareElements(float a, float b) { EXPECT_NEAR(a, b, (fabs(a) + fabs(b)) / 10000.0); }
+  // Higher tolerance for full network computations
+  void compareElements(float a, float b) { EXPECT_NEAR(a, b, (fabs(a) + fabs(b)) / 100.0); }
```
flaub, line 40: I'm not sure this is a good idea, @tzerrell ?

dgkutnic: To provide some insight here: `compareElements` is generally failing for values `a` and `b` that are sufficiently small - around `1e-05` and smaller. In those cases, I think that this test may be running into precision issues.

tzerrell: I think a higher tolerance for full network comparisons is reasonable, but this is IMO too high a tolerance and also it applies to anything using compareElements on floats, not just full networks. My suggestions:
* Rather than changing the global tolerance, add an optional tolerance parameter to this function. Then on the specific tests which need looser tolerance, provide what's needed for each specific test. Perhaps both an absolute and relative tolerance parameter (see below)
* It may be necessary to add an absolute tolerance component to the test; it's not uncommon to see outputs like 3.25e-33 vs. 8.12e-29 for different implementations computing "approximately zero", and that's fine for ML workloads that are OK with reordering floating point reductions (i.e. basically all of them), but the relative error is huge. A possible formula with both absolute and relative tolerance terms:
```
rtol * ((fabs(a) + fabs(b)) / 2.) + atol
```
This is about as close as you can get to the all_close formula we from numpy while being symmetric in `a` and `b`. See those docs for more: https://numpy.org/doc/stable/reference/generated/numpy.allclose.html

flaub (approved):

jd-bartlett96 (approved):

flaub:
I'll try to get the remaining tests revived, this is just a first pass to get something landed.

Also I'll try to figure out a way to deal with large weights that don't involve generating C++ header files which makes the compiler take a long time to process.

PR #1317, opened 2020-08-29
Reify the notion of a compilation target

earhart:


flaub:
What do you plan to add to this new `Target` type? In its current form it's not worth reifying.

earhart:
@flaub Device enumeration and opening.

flaub:
Can you describe how you see the device and target being intertwined? To me they are orthogonal things, albeit one is related to the other. To me, device enumeration should be part of the execution environment, and not related to compilation.

earhart:
Consolidating the two threads:
> It seems like we're confusing targets with devices. I think certain devices will have support for certain targets, but they should not be combined.

So, here's what I'm thinking.  To be clear -- I don't have a strong opinion about all this; if you'd like things to be done differently, just let me know.

To start with, we have device access _APIs_ that we're using to access the device: e.g. Metal, OpenCL, Vulkan, maybe some sort of Native for CPU, &c.  Let's also define a _Runtime_ as the part of the PlaidML codebase that's using a particular API to access devices.

I don't know that there's a general good way to associate physical devices across these APIs -- and the same physical device might show up as a Metal device and an OpenCL device.  So I think the identifier of the runtime that's using a particular API is an inextricable part of the device ID.

It seems reasonable to have a symbolic identifier for each runtime, enumerating them globally in the PlaidML sources.  So, `metal`, `ocl` or `opencl`, `vulkan`, &c -- these then are part of the device ID.

We also probably want the device class in the device ID, because otherwise things get confusing when devices are added and removed.  So perhaps something like `intel_gen-vulkan`.  The device class here is something specific to the runtime & API, though; OpenCL might have a different name than Metal for the same hardware device class.

There might also be multiple instances of a given device class, so we probably want an ordinal in there to distinguish them.  So we wind up with something like `intel_gen-vulkan.0`.

On to targets: currently, a target identifier identifies a compilation pipeline (or to be precise, a function that knows how to add a particular target-specific compilation pipeline to an MLIR pass manager).

The way we're currently doing compilation, we're lowering the entire kernel launch sequence into native code.  So the target is more than just the device class; lowering to run via OpenCL is fundamentally different from lowering to run via Vulkan, and lowering to a device accessed by a host running x86 is different from lowering to a device accessed by a host running x86_64.  Maybe that's ungood and should be changed at some point; I don't know.  But for now, that's what we've got.

So the runtime to use is implied by the target.  We've kind of papered over this so far, with proposals like `intel_gen` vs `intel_gen-ocl`.  To make the naming simpler and more symmetric, it seems reasonable to always include the symbolic runtime identifier in the target ID -- so `intel_gen-vulkan`, `intel_gen-ocl`, &c.

So the target identifier looks a lot like a device identifier without the device instance ordinal.

On top of all that: we've been linking the target code and the runtime code into binaries separately, but the PoR is to add the runtime as a target dependency, so that e.g. anything pulling in `//pmlc/target/intel_gen` would automatically link in `//pmlc/rt/vulkan`.

Given all this, we could still have separate registration mechanisms for the target and the runtime... but when they're this closely aligned, I honestly don't see the point; I think it's simpler for our users if we link them.  I think we should be able to take a target ID or a device ID, extract the `vulkan` component, look up the associated module (currently a `Target` in this change, but could easily be renamed), and let that module handle device enumeration, compilation, and execution.

For example: A user says, "What devices are on my machine?"  We look through the registry, and ask each thing there to enumerate its devices; we might return to the user, `intel_gen-vulkan.0` and `intel_gen-ocl.0`.  Or the user might query for the set of available compilation targets, and we might return `intel_gen-vulkan` or `intel_gen-ocl`.  When compiling, the user specifies the target; executing just adds a particular device index.  But the important bit is that the compilation target to use for a particular device is pretty obvious from the device ID.

An alternative might be to try to associate devices across APIs, so we could say something like, "Yeah, there's an `intel_gen` device; to use it, compile for `intel_gen` or `intel_gen-ocl`.  That seems harder (since we have to implement the association), more fragile (I don't know that we can always get it right), confusing to users, and unnecessary.

One interesting extension: for offline compilation, we should probably be thinking of including the OS and ABI, since we're lowering to native code.  And we might as well think of this as an extension to LLVM's target triples, since people are already familiar with those.  So we might want `<arch><subarch>-<vendor>-<sys>-<abi>-<api>.<device index>`, e.g. `gen9-intel-linux-x86_64-vulkan` for the target and `gen9-intel-linux-x86_64-vulkan.0` for the device.  The important bit here is to have a standard pattern that we can apply a regex to -- extract `vulkan`, look it up, and let that module figure out where to go from there.

Thoughts?

flaub:
I've been thinking of the problem like:

We start with a `Selector`, which is in the form of `<target>.<runtime>.<device>`.

Next, we have a few maps:
- `target (string) -> Pipeline`
- `runtime (string) -> Runtime`

Currently, we have:
- `Program`: consumes source code/EDSL, has a `compile()` method on it that uses a `target` to produce LLVM IR
- `Executable`: consumes a `Program`, `buffers`, and `device_id`, has an `invoke()` method for execution.

I propose then we add a new abstraction, called a `Runtime`, which looks like:

```
class Runtime {
public:
  virtual std::vector<std::string> listDevices() = 0;
  virtual Exectuable prepareProgram(Program, device_id) = 0;
};
```
There'd be a registry for runtimes, so the `<runtime>` portion of the `Selector` would give us a `Runtime` instance.
From there, you can get an `Exectuable` and off you go.

The other change would be for the map to the pipeline. Right now it's only `target` -> `Pipeline`. I suggest we extend this so that it's a map of `target.runtime` -> `Pipeline`. 

Maybe in the future we can have distinct pipelines, one that is runtime agnostic, so it would be `target -> pipeline`, and then another that goes `runtime -> pipeline`. The target pipeline might emit a runtime-agnostic dialect like `comp`, and then the runtime pipeline might convert `comp` into `vulkan` or whatnot.

WDYT?

earhart:
@flaub Just had a good convo w/Jeremy.  Sounds like we've changed our PoR to not include the runtime as a dep of the pipeline, and he'd strongly prefer for us to not try to pick a target ID for a given device; given that, I agree, we should separate the `Runtime` and the `Target`.  I think we might want to reify `Device`, and provide constant buffers to `prepare`, but otherwise that interface looks roughly right to me.

You don't mention what a `device` actually looks like as part of a `Selector` -- is an integer okay?  If we're not picking targets, then, devices would be enumerated as `vulkan.0`, `vulkan.1`, `opencl.0`, and we can parse that string to determine the runtime and device index.

One other idea we bandied about was that it'd be possible to combine the target and device into a single string -- e.g. `intel_gen9.vulkan.0`; we could then parse out the `intel_gen9.vulkan` target and the `vulkan.0` device separately.  It'd just give users one less thing to specify.  Thoughts?

And then, I'd still prefer something like LLVM target triples -- e.g. `gen9-intel-vulkan`. (Jeremy is okay with requiring an LLVM JIT on the far side of serialization, if we ever implement separate compilation, so we don't need system or ABI).  I think that's orthogonal, but what do you think?

earhart:
(Closing this PR -- will open a new one along the lines we've discussed.)

PR #1318, opened 2020-08-29
Expose target registry map 

earhart:


earhart:
FYI -- this includes changes from #1317; until that's landed, please make sure to only look at the most recent commit.

flaub:
What new functionality does this provide?

earhart:
I was thinking to build towards device access, going through the `Target` -- so each device is named `<target>.<device_index>`.

At some point, we might want to rename this `Target` class to something like `Api` or `Hal` or `Backend` -- something like Vulkan or OpenCL or Metal.  I don't see a need to do that just yet, though.

earhart:
(... the primary reason for keeping the current `Target` name is because targets are the things we're currently registering, and it's really easy to change this later.)

flaub:
It seems like we're confusing targets with devices. I think certain devices will have support for certain targets, but they should not be combined.

earhart:
(Closing this PR -- will open a new one along the lines we've discussed.)

PR #1322, opened 2020-08-31
add stdx to SPIRV GLSL lowering pattern

YangleiZouIntel:
* lower stdx::Round to spirv and enable CppEdsl.Round

* lower stdx::FloorOp to spv and enable CppEdsl.Floor

* lower stdx::TanOp to spirv and enable CppEdsl.Tan

* use stdx to spirv unary conversion template

* lower stdx::CosH spv and enable CppEdsl.CosH

* lower stdx::SinH spv and enable CppEdsl.SinH

* add mlir test for newly added stdx operations conversion

* use GLSL pattern for GLSL memory model ops

Co-authored-by: Ling, Liyang <liyang.ling@intel.com>


pmlc/conversion/gpu_to_spirv/gpu_to_spirv.cc
```diff
@@ -78,6 +160,16 @@ void populateStdxToSPIRVPatterns(MLIRContext *context,
   patterns.insert<StdxSubgroupBroadcastOpConversion>(context, typeConverter);
 }
 
+void populateStdxToSPIRVGLSLPatterns(MLIRContext *context,
+                                     SPIRVTypeConverter &typeConverter,
+                                     OwningRewritePatternList &patterns) {
+  patterns.insert<StdxRoundOpConversion,
```
abialas1, line 166: Changes look good to me, but I am having slight doubts regarding the implementations of the functions that are still missing in spv dialect in mlir upstream. If we are not using the functions from extended instruction sets (GLSL or OpenCL) we can have a gap in performance on those. I would recommend adding those to the upstream first and map them 1:1, later on rely on the gfx compiler during spirv consumption. Probably not needed to do this straightaway, but at least leave the TODO comments to mark those to replace in the future

YangleiZouIntel: I agree with you that we can leave these ops TODO and replace in the future. This patch is trying to fix these stdx lowing right now to help test networks that needs these ops.

jbruestle: I'm not sure if we are in such a hurry to add Round/SinH, etc that it is worth adding to technical debt.  I suspect if you made a PR for upstream to add support for these ops, it would be accepted fairly quickly (and there are all in the GLSL extension as far as I can tell https://www.khronos.org/registry/spir-v/specs/unified1/GLSL.std.450.html).  Let's keep the easy 1-to-1 conversions, but lets fix upstream support for the rest.

YangleiZouIntel: Thanks. I keeped the 1-to-1 convertions in this PR and will add those ops to upstream.

flaub:
FYI, the unit tests for intel_gen did fail in CI: https://buildkite.com/plaidml/plaidml-plaidml/builds/3696#691d5a40-e27a-46d1-a8da-8e7da5c6ac1f

The unit test is marked as soft_fail right now until we are sure that it's robust and non-flakey.

YangleiZouIntel:
> FYI, the unit tests for intel_gen did fail in CI: https://buildkite.com/plaidml/plaidml-plaidml/builds/3696#691d5a40-e27a-46d1-a8da-8e7da5c6ac1f
> 
> The unit test is marked as soft_fail right now until we are sure that it's robust and non-flakey.

Thanks for your note. Some tests passed on my local machine but failed on CI. Updated the test skip list.

PR #1327, opened 2020-09-01
Add support for vectorizations of 0 stride reductions

LLITCHEV:


pmlc/dialect/pxa/transforms/vectorize.cc
```diff
@@ -158,11 +157,25 @@ class Impl {
...
-        op.getLoc(), ArrayRef<Type>{op.getMemRefType()}, op.agg(), val,
-        op.memref(), op.map(), op.idxs());
-    op.replaceAllUsesWith(vecOp.getResult());
-    op.erase();
+    // Add vector_reduction only if the stride is 0
+    if (zeroStrideReductions.find(op.getOperation()) !=
+        zeroStrideReductions.end()) {
+      vector::ReductionOp reductionOp = builder.create<vector::ReductionOp>(
+          op.getLoc(), op.getMemRefType().getElementType(),
+          builder.getStringAttr("add"), val, ValueRange{});
```
jbruestle, line 165: The 'add' here needs to match the agg of the reduce (i.e. if we are doing an add reduce, it's an add, if we are doing a mul reduce, we need a mul).  Not all of our reductions are supported unfortunately, so we probably also need to preflight to make sure there is a translation.  The reductions supported only seems to appear in the vector dialect verifier:

https://github.com/llvm/llvm-project/blob/master/mlir/lib/Dialect/Vector/VectorOps.cpp#L134

pmlc/dialect/pxa/transforms/vectorize.cc
```diff
@@ -33,6 +33,35 @@ class Impl {
...
+    case AtomicRMWKind::mins:
+      return "min";
+    case AtomicRMWKind::minu:
+      return "min";
+    case AtomicRMWKind::mulf:
+      return "mul";
+    case AtomicRMWKind::muli:
+      return "mul";
+    }
+    return "";
```
flaub, line 62: use `llvm_unreachable` here

pmlc/dialect/pxa/transforms/vectorize.cc
```diff
@@ -33,6 +33,35 @@ class Impl {
   DenseSet<Operation *> vectorizedOps;
   llvm::DenseSet<Operation *> zeroStrideReductions;
 
+  ::llvm::StringRef
```
flaub, line 36: Return type should be `const char*`

pmlc/dialect/pxa/transforms/vectorize.cc
```diff
@@ -66,6 +95,17 @@ class Impl {
...
+            // vector::ReductionOp doesn't support pxa's assign reduction.
+            // Also, make sure we handle only the supported types -
+            // see the vector::ReductionOp verification code
+            // (https://github.com/llvm/llvm-project/blob/master/mlir/lib/Dialect/Vector/VectorOps.cpp#L134).
+            Type eltType = op.getMemRefType().getElementType();
+            if (op.agg() == AtomicRMWKind::assign ||
+                (!eltType.isF32() && !eltType.isF64() &&
+                 !eltType.isSignlessInteger(32) &&
+                 !eltType.isSignlessInteger(64))) {
+              return failure();
```
flaub, line 107: Use `op.emitError()` here.

LLITCHEV: Adding a remark emission, as we discussed. The vectorization just doesn't happen if the vector::ReductionOperation doesn't support the type of the operand or the reduction type.

pmlc/dialect/pxa/transforms/vectorize.cc
LLITCHEV: Fixed.

pmlc/dialect/pxa/transforms/vectorize.cc
LLITCHEV: Added llvm_unreachable.

pmlc/dialect/pxa/transforms/vectorize.cc
LLITCHEV: If the parameter types and the type of the reduction are not supported by the vector::ReductionOp, the vectorization doesn't happen.

pmlc/dialect/pxa/transforms/vectorize.cc
```diff
@@ -158,11 +198,27 @@ class Impl {
           builder.create<vector::BroadcastOp>(op.getLoc(), vecType, val);
       val = bcast.getResult();
     }
-    auto vecOp = builder.create<PxaVectorReduceOp>(
-        op.getLoc(), ArrayRef<Type>{op.getMemRefType()}, op.agg(), val,
-        op.memref(), op.map(), op.idxs());
-    op.replaceAllUsesWith(vecOp.getResult());
-    op.erase();
+    // Add vector_reduction only if the stride is 0
+    if (zeroStrideReductions.find(op.getOperation()) !=
```
flaub, line 202: I think you can use `zeroStrideReductions.count(op.getOperation())` here instead of find.

LLITCHEV: Yes, it could be... At least in some STL implementations the count() would do also the find internally (well, really a hash lookup, which find boils down to, and then get the count, which is probably a bit slower than the find and compare to a constant. I might be wrong, since this doesn't hold for all implementation of the STL interfaces, but I don't consider it being a big issue. If you insist, I can change it...

flaub: FYI, they have equal performance and so the decision to use one or the other should be based on semantics rather than runtime performance: https://stackoverflow.com/questions/14159682/unordered-map-which-one-is-faster-find-or-count?rq=1

PR #1328, opened 2020-09-01
Workaround to enable keras backend tests without breaking CI

YangleiZouIntel:
This patch add vulkan runtime to libplaidml.so and forces bazel native cc_test(plaidml_cc_test) to statically link (the default behavior of native cc_test is to link dependences dynamically) to volk and avoid breaking CI. 
Note that using statical linking could potentially increase bazel test run time. 

Not sure if this workaround is appropreate to be merged or we should keep it as a patch file, please check. Thanks.

### DIFF
--- a/pmlc/target/intel_gen/BUILD
+++ b/pmlc/target/intel_gen/BUILD
@@ -42,6 +42,9 @@ plaidml_cc_library(
         "@llvm-project//mlir:LLVMTransforms",
         "@llvm-project//mlir:SCFToGPUPass",
         "@llvm-project//mlir:StandardOpsTransforms",
-    ],
+    ] + select({
+        "//plaidml:gpu_target": ["//pmlc/rt/vulkan"],
+        "//conditions:default": [],
+    }),
     alwayslink = 1,
 )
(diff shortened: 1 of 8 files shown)


pmlc/target/intel_gen/BUILD
```diff
@@ -42,6 +42,9 @@ plaidml_cc_library(
         "@llvm-project//mlir:LLVMTransforms",
         "@llvm-project//mlir:SCFToGPUPass",
         "@llvm-project//mlir:StandardOpsTransforms",
-    ],
+    ] + select({
+        "//plaidml:gpu_target": ["//pmlc/rt/vulkan"],
+        "//conditions:default": [],
+    }),
```
flaub, line 48: Seems like we should always have `intel_gen` target depend on the vulkan runtime.

YangleiZouIntel: Thanks for your note. I think we can run keras backend tests with --//plaidml:target=intel_gen in plaidml-v1. This PR is not needed. I will close it. Thanks.

PR #1329, opened 2020-09-01
Add support for PRNG on intel_gen

YangleiZouIntel:
* Use the x86 implementation of PRNG to execute the op on the host.
* Lots of buildifier formatting changes (please remember to use buildifier for BUILD files).


pmlc/target/intel_gen/pipeline.cc
```diff
@@ -35,8 +35,30 @@ using namespace mlir; // NOLINT[build/namespaces]
...
+    conversion::pxa_to_affine::PXAToAffineConversionTarget target(ctx);
+
+    OwningRewritePatternList patterns;
+    populatePXAPrngToAffineConversionPatterns(patterns, &ctx);
+    conversion::pxa_to_affine::populatePXAToAffineConversionPatterns(patterns,
+                                                                     &ctx);
+
+    if (failed(applyPartialConversion(getOperation(), target, patterns,
+                                      nullptr))) {
+      getOperation().dump();
```
flaub, line 55: Drop this.

pmlc/target/intel_gen/pipeline.cc
```diff
@@ -35,8 +35,30 @@ using namespace mlir; // NOLINT[build/namespaces]
...
+      emitError(UnknownLoc::get(&ctx), "Error lowering pxa -> affine\n");
```
flaub, line 56: Use `getOperation()->emitError()`

pmlc/target/intel_gen/passes.td
```diff
@@ -28,6 +28,13 @@ def ConvertParallelLoopToGpu : Pass<"intel-gen-convert-parallel-loops-to-gpu"> {
   let dependentDialects = ["mlir::AffineDialect", "mlir::gpu::GPUDialect"];
 }
 
+def ConvertPXAToAffine : Pass<"intel_gen-convert-pxa-to-affine", "mlir::ModuleOp"> {
```
flaub, line 31: Use `intel-gen` instead of `intel_gen`

flaub:
This looks identical to the x86 target implementation. IIUC, the strategy here is to implement the PRNG on the host-side. If so, then I think we should use the x86 implementation directly, instead of making a copy.

YangleiZouIntel:
> This looks identical to the x86 target implementation. IIUC, the strategy here is to implement the PRNG on the host-side. If so, then I think we should use the x86 implementation directly, instead of making a copy.

Thanks. Yes it is to implement PRNG on the host side for now to run resnet50 on gpu. Do you think it is better to move prng lowering under pmlc/target/x86 to pmlc/conversion/pxa_to_affine/ ? Otherwise I think we will need to add x86 as a dependent to intel_gen which seems not good.

pmlc/target/intel_gen/passes.td
YangleiZouIntel: Fixed

pmlc/target/intel_gen/pipeline.cc
YangleiZouIntel: Fixed

flaub:
Yes, I think moving it to `pxa_to_affine` makes some sense. Or we could leave in `x86` but just make a library out of this one pattern so it can be reused by `intel_gen`. It's interesting to think about whether we should have `host` and `device` targets as a further distinction.

adstraw:
Are we thinking about this as a permanent fixture of the GPU pipeline or a temporary workaround until we figure out PRNG lowering on the GPU?  My understanding was that it was temporary.  If so, perhaps it is ok to let this through without further design review.  If permanent, or if we think we need host side compute for other things in the GPU pipeline then it's a different story.

flaub:
I'll fix this in my upcoming PR, stay tuned :)

YangleiZouIntel:
> I'll fix this in my upcoming PR, stay tuned :)

Thank you. I was running out of time yesterday. Looking forward to seeing the new PR.

flaub (approved):

adstraw:
Nice work @flaub !

PR #1330, opened 2020-09-02
Lowering of subgroups down to vulkan

jbruestle:
Initial reworking to connect all the passes end-to-end.  As of this commit, a 2d convolution lowers into what looks roughly like correct subgroup SPIRV code, and gets all the way to the runtime, at which point we segfault in VulkanRuntime::createComputePipeline.  Lots of rough code and I broke some tests that need fixing, but wanted to put up the current state for today.

To reproduce the failure run:

PLAIDML_VERBOSE=1 PLAIDML_DUMP=/tmp/ir/ PLAIDML_TARGET="intel_gen" bazelisk run -c dbg //plaidml/bridge/keras:backend_test -- TestBackendOps.res3a_branch2a


pmlc/conversion/gpu_to_spirv/gpu_to_spirv.cc
```diff
@@ -31,13 +32,31 @@ struct StdxSubgroupBroadcastOpConversion
...
+  using SPIRVOpLowering<AllocOp>::SPIRVOpLowering;
+
+  AllocOpPattern(MLIRContext *context, SPIRVTypeConverter &typeConverter)
+      : SPIRVOpLowering<AllocOp>(context, typeConverter, /*benefit=*/1000) {}
+
+  LogicalResult
+  matchAndRewrite(AllocOp operation, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    MemRefType allocType = operation.getType();
+    MemRefType allocType2 = MemRefType::Builder(allocType).setMemorySpace(6);
```
adstraw, line 52: Magic number 6?

pmlc/dialect/pxa/transforms/gpu_thread.cc
```diff
@@ -67,12 +67,22 @@ struct GPUThreadPass : public GPUThreadBase<GPUThreadPass> {
...
+      }
+      CostModel model(op, goalThreads);
       auto tileSize =
           findBestTileSize(EvenTilingGenerator(), model, *maybeRanges);
       auto inner = performTiling(op, tileSize);
       op.setAttr("hardware", StringAttr::get("gpu_block", &getContext()));
       inner.setAttr("hardware", StringAttr::get("gpu_thread", &getContext()));
+      if (subgroupAttr) {
+        op.removeAttr("subgroupSize");
+        inner.setAttr("subgroupSize", subgroupAttr);
```
adstraw, line 84: This seems superfluous as we got subgroupAttr from the op attribute "subgroupSize", did not modify it, and now resetting it.

pmlc/dialect/pxa/transforms/subgroups.cc
```diff
@@ -230,7 +232,9 @@ struct SubgroupsPass : public SubgroupsBase<SubgroupsPass> {
         6,       // Maximum register per thread
     };
     SubgroupCostModel cm(params, op);
-    if (cm.bestCost == std::numeric_limits<double>::infinity()) {
+    if (cm.bestCost == std::numeric_limits<double>::infinity() ||
+        op.getIVs().size() < 6) {
```
adstraw, line 236: Magic number 6, again.

adstraw (commented):
looking good so far, just a few minor comments

pmlc/conversion/gpu_to_spirv/gpu_to_spirv.cc
jbruestle: Yeah, this is just lack of cleanup, need to pull in the appropriate enum (6 is the function local spirv memory space).

pmlc/dialect/pxa/transforms/gpu_thread.cc
jbruestle: We tile the block (i.e. op changes to two block, op = outer, innner = inner).  We are 'moving' the attribute off of the outer block and placing it on the inner block.  In general, I feel like this use of attributes to squirrel information from one pass to another is a poor design, Frank and I have been talking over other options, but haven't arrived at a solution yet.

pmlc/dialect/pxa/transforms/subgroups.cc
jbruestle: This is a pure hack to make sure we only 'subgroup' the convolution and not the eltwise stuff so I only have one sub-grouped thing to test and will definitely not make it into the non WIP commit.

jbruestle:
As of the latest (even hackier) changes, we get correct results for some of the resnet convolutions.  Remaining issue seems to be subgroup size mismatch on the hardware side.

jbruestle:
Code is now tidy enough to land (hopefully).  Also, edsl test (except winograd) and many backend tests involving convolutions now work with subgroups, but due to the fact that 1) The CI machines dont have the latest driver version which is now required and 2) Resnet still doesn't quite work for subgroups, I've disabled it pmlc/target/intel_gen/pipeline.cc and replaced it with tileAccumulate for the official PR, but if you swap which pass is commented, subgroups work.

pmlc/dialect/pxa/transforms/gpu_thread.cc
```diff
@@ -61,18 +64,26 @@ struct GPUThreadPass : public GPUThreadBase<GPUThreadPass> {
...
+    // Nest outermost loops into 'blocks' and 'threads'
     for (auto op : func.getOps<AffineParallelOp>()) {
       auto maybeRanges = op.getConstantRanges();
       if (!maybeRanges) {
+        // Fail if we can't compute the ranges at compile time
         continue;
       }
-      CostModel model(op, maxThreads);
+      // We want 'logical threads' * 'threads per subgroup' (i.e. subgroupSize)
+      // to end up beging about 'maxThreads' threads, and the rest we put into
```
flaub, line 75: typo: 'beging'?

flaub (approved):
LGTM

PR #1331, opened 2020-09-02
Include assertion fix from MLIR

flaub:


flaub:
Once https://github.com/plaidml/llvm-project/pull/26 I'll bump the pointer again and mark this as ready for review.

PR #1334, opened 2020-09-02
Add Vulkan profiling support

adstraw:


pmlc/rt/vulkan/vulkan_runtime.cc
```diff
@@ -830,12 +839,46 @@ LogicalResult VulkanRuntime::createCommandPool() {
...
+    VkQueryPipelineStatisticFlags    pipelineStatistics;
+  } VkQueryPoolCreateInfo;
+  */
+
+  VkQueryPoolCreateInfo queryPoolCreateInfo = {};
+  queryPoolCreateInfo.sType = VK_STRUCTURE_TYPE_QUERY_POOL_CREATE_INFO;
+  queryPoolCreateInfo.pNext = nullptr;
+  queryPoolCreateInfo.flags = 0;
+  queryPoolCreateInfo.queryType = VK_QUERY_TYPE_TIMESTAMP;
+  queryPoolCreateInfo.queryCount = 1024; // TODO: hardcode?
```
YangleiZouIntel, line 866: I think is OK. We probably will need a larger number if we need to track every single gpu module.

pmlc/rt/vulkan/vulkan_runtime.cc
```diff
@@ -903,6 +946,16 @@ LogicalResult VulkanRuntime::createSchedule() {
...
                          "vkBeginCommandBuffer");
 
+  /*
+  void vkCmdWriteTimestamp(
+    VkCommandBuffer                             commandBuffer,
+    VkPipelineStageFlagBits                     pipelineStage,
+    VkQueryPool                                 queryPool,
+    uint32_t                                    query);
+  */
+  vkCmdWriteTimestamp(commandBuffer, VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT,
```
YangleiZouIntel, line 956: I think it is good. In compute pipeline, there are not many stages, basically top->draw indirect->compute shader->bottom. I think using VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT or compute or bottom does not make much difference.

pmlc/rt/vulkan/vulkan_runtime.cc
```diff
@@ -946,6 +999,16 @@ LogicalResult VulkanRuntime::createSchedule() {
...
   }
 
+  /*
+  void vkCmdWriteTimestamp(
+    VkCommandBuffer                             commandBuffer,
+    VkPipelineStageFlagBits                     pipelineStage,
+    VkQueryPool                                 queryPool,
+    uint32_t                                    query);
+  */
+  vkCmdWriteTimestamp(commandBuffer, VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT,
```
YangleiZouIntel, line 1009: I am thinking about adding timestamps for each action (run a shader or a buffer transfer), what do you think?

pmlc/target/intel_gen/BUILD
```diff
@@ -35,6 +35,7 @@ plaidml_cc_library(
         "//pmlc/dialect/pxa/transforms",
         "//pmlc/dialect/tile/transforms",
         "//pmlc/dialect/stdx/transforms",
+        "//pmlc/rt/vulkan",
```
YangleiZouIntel, line 38: The PRNG and linking issue are both fixed in Frank's PR https://github.com/plaidml/plaidml/pull/1329

pmlc/rt/vulkan/vulkan_runtime.cc
adstraw: Good idea.  Will add this.

pmlc/rt/vulkan/vulkan_runtime.cc
adstraw: I think you are right.  Doesn't seem to matter.  I went with VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT for consistency with the kernel actions.

pmlc/rt/vulkan/vulkan_runtime.cc
adstraw: Plan to add this in follow-up commit.

jbruestle (approved):

PR #1335, opened 2020-09-03
Use CI plan to run target-specific unit tests

flaub:


tzerrell (commented):
~LGTM~

Update: Accidentally saw the suggested reviewer list as an actual reviewer list -- I can formally approve if this is ready for a real review

flaub:
OK, I think this is finally ready for review.

ci/buildkite.py
```diff
@@ -151,9 +154,8 @@ def cmd_build(args, remainder):
 
     env = os.environ.copy()
     variant = plan['VARIANTS'][args.variant]
-    if 'env' in variant:
-        for key, value in variant['env'].items():
-            env[key] = str(value)
+    for key, value in variant.get('env', {}).items():
```
earhart, line 157: I think it'd be better to let buildkite set the environment; it's a bit more clear what's going on.

ci/plan.yml
```diff
@@ -124,28 +143,43 @@ SUITES:
...
+        cwd: tmp/input
         precision: untested
-        soft_fail: true
+        skip_platforms:
+          - pml-llvm-cpu
+        platform_overrides:
+          pml-vk-gen:
+            args:
+              - --plaidml_target=intel_gen
+              - --skip_test_file=edsl_test-skip.intel_gen.txt
```
earhart, line 165: DRY?

BUILD
```diff
@@ -39,6 +39,8 @@ pkg_tar(
         "//plaidbench:wheel",
         "//plaidml:wheel",
         "//plaidml/bridge/keras:wheel",
+        "//plaidml/edsl/tests:cc_test",
```
earhart, line 42: This approach requires that support for all runtimes is linked into `//plaidml/edsl/tests:cc_test` at all times; I was under the impression that we didn't want to do that, but I could be mistaken.  :-)

ci/plan.yml
flaub: Where is this being repeated?

ci/buildkite.py
flaub: Not sure you're reading this right, this is saying, try to get the `env` dictionary, but if it's not there, just make an empty one. This is functionally the same as before (and in fact was a change you made, adding the `if 'env' in variant`, but I think this way is a bit more pythonic.

BUILD
flaub: Eventually, I plan to have multiple `cc_test` targets, but for now this works just fine.

flaub: For tests I actually do want everything to be built and linked, so that even if someone is on a machine that can't run the tests, they can at least build them, so that people have less chance of breaking some other platform. I think the whole idea of transitions in bazel isn't a great one.

ci/buildkite.py
earhart: I mean, I don't know that you need to be manipulating env here at all; I just left the code here to be careful.

drgone (approved):

ci/plan.yml
earhart: The definition of `plaidml/edsl/tests:cc_test`

earhart: The definition of `//plaidml/edsl/tests:cc_test`

BUILD
earhart: SGTM

ci/plan.yml
flaub: Well the issue here is that we're running this test outside of bazel, and so we can't rely on any bazel rules to run anything.

earhart: Yes... but the previous code captured the arguments and propagated them to the out-of-bazel test, so that they'd only have to be defined in one place.

I don't know that it matters much, though.

PR #1337, opened 2020-09-07
Add device enumeration and selection

earhart:


plaidml/exec/BUILD
```diff
@@ -33,7 +33,9 @@ plaidml_cc_library(
     visibility = ["//visibility:public"],
     deps = [
         "//plaidml/core",
+        "//plaidml/edsl",
```
flaub, line 36: Execution should not depend on edsl

plaidml/edsl/tests/BUILD
```diff
@@ -23,14 +23,14 @@ plaidml_cc_test(
...
     }),
-    linkstatic = True,
     toolchains = ["//plaidml:settings"],
     visibility = ["//visibility:public"],
     deps = [
         "//plaidml:testenv",
         "//plaidml/edsl",
         "//plaidml/exec",
     ],
+    linkstatic = True,
```
flaub, line 33: Please use buildifer to format all bazel build files

pmlc/rt/device_id.h
```diff
@@ -0,0 +1,17 @@
...
+#pragma once
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "pmlc/rt/runtime.h"
+#include "llvm/ADT/StringRef.h"
+
+namespace pmlc::runtime {
```
flaub, line 12: The dir structure and namespace hierarchy should match

pmlc/rt/executable.cc
```diff
@@ -71,8 +78,7 @@ std::string makeCWrapperFunctionName(StringRef name) {
 // Define an interface function that wraps all the arguments of the original
 // function and all its results into an i8** pointer to provide a unified
 // invocation interface.
-static std::string packFunctionArguments(llvm::Module *module,
-                                         StringRef entry) {
+std::string packFunctionArguments(llvm::Module *module, StringRef entry) {
```
flaub, line 81: Please follow MLIR guidelines. Functions are static and types go into anonymous namespace

flaub (changes requested):
I haven't had a chance to review the design in detail but here's my first shallow pass.

plaidml/edsl/tests/BUILD
earhart: This was done, using the team standard pre-commit.  Not sure why buildifier is inconsistent here.

plaidml/exec/BUILD
earhart: Hmm.  Something needs to know about both `Program` and `Executable`.  How would you like that constructed?

pmlc/rt/device_id.h
earhart: That bothered me, as well; I figured it could go in a separate PR, but, done.

pmlc/rt/executable.cc
earhart: Done.

plaidml/exec/BUILD
flaub: Both of these are in `pmlc`, not part of the API surface.

networks/oplib/BUILD
```diff
@@ -21,7 +21,7 @@ plaidml_cc_binary(
     deps = [
         ":oplib",
         "//plaidml/exec",
-        "//pmlc/rt",
+        "//pmlc/runtime",
```
flaub, line 24: Let's call this `rt`, so that it's consistent with LLVM.

plaidml/exec/ffi.cc
```diff
@@ -122,7 +116,7 @@ plaidml_executable* plaidml_jit(  //
     } else if (jit == "MCJIT") {
       kind = EngineKind::MCJIT;
     }
-    exec->exec = std::make_unique<Executable>(program->program, bufptrs, kind);
+    exec->exec = std::make_unique<Executable>(program->program, std::move(device), bufptrs, kind);
```
flaub, line 119: Why can't we just pass the `device_id` here, what's the point of grabbing a `device` object at this stage?

pmlc/runtime/llvm/runtime.cc
```diff
@@ -0,0 +1,27 @@
+// Copyright 2020, Intel Corporation
+
+#include "pmlc/runtime/runtime_registry.h"
+
+#include <stdexcept>
+
+namespace {
+
+class Device final : public pmlc::runtime::Device {};
```
flaub, line 9: What's the point of this?

pmlc/runtime/runtime.h
```diff
@@ -0,0 +1,57 @@
...
+
+// Device represents a PlaidML device, capable of evaluating a PlaidML program.
+class Device {
+public:
+  // Returns the current Device, local to the current thread.
+  //
+  // If there is no current device, or the current device is of an unexpected
+  // type, an exception will be thrown.
+  //
+  // This is used to smuggle the current device through to runtimes that are
```
flaub, line 19: I think we should pass the `runtime` instance to the LLVM code as the first void* pointer to the program. The `runtime` instance should be instantiated with the `device_id`.

pmlc/runtime/runtime.h
```diff
@@ -0,0 +1,57 @@
...
+};
+
+// Runtime represents a particular runtime implementation.
+//
+// A runtime implementation is a component that understands how to use a
+// particular API (e.g. OpenCL, Metal, Vulkan, &c) to access a device capable of
+// evaluating a PlaidML program.  These are typically single-instanced within a
+// process, constructed once as needed (typically well after process
+// initialization) and then held onto by the global runtime registry.
+class Runtime {
```
flaub, line 48: Each runtime should list the set of devices it supports.

pmlc/runtime/runtime_registry.cc
```diff
@@ -0,0 +1,90 @@
...
+namespace pmlc::runtime {
+namespace {
+
+struct LoaderList {
+  std::mutex mutex;
+  std::list<std::pair<std::string, Loader>> loaders;
+
+  static LoaderList &instance() {
+    static LoaderList ll;
+    return ll;
```
flaub, line 22: This looks a lot like eleven.

pmlc/runtime/symbol_registry.cc
```diff
@@ -0,0 +1,33 @@
+// Copyright 2020 Intel Corporation
```
flaub, line 1: This doesn't seem to be runtime-specific, can we leave it on the compiler?

pmlc/runtime/vulkan/vulkan_runtime.cc
```diff
@@ -0,0 +1,45 @@
...
+
+    for (const auto &physicalDevice : physicalDevices) {
+      devices.emplace_back(
+          std::make_shared<VulkanDevice>(physicalDevice, state));
+    }
+  }
+
+  std::size_t deviceCount() const noexcept final { return devices.size(); }
+  std::shared_ptr<pmlc::runtime::Device> device(std::size_t idx) {
+    return devices.at(idx);
```
flaub, line 36: We should not force device construction just to do enumeration.

pmlc/runtime/executable.h
```diff
@@ -0,0 +1,35 @@
+// Copyright 2020 Intel Corporation
```
flaub, line 1: This doesn't seem to be runtime-specific. I think certain runtimes will have certain Executable implementations, but the interface should remain on the compiler.

pmlc/runtime/internal.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include "llvm/ADT/StringMap.h"
+
+#include "pmlc/runtime/runtime.h"
+
+namespace pmlc::runtime {
+
+// ScopedCurrentDevice sets the current thread's Device pointer.  This is read
+// by Device::current().
+class ScopedCurrentDevice {
```
flaub, line 17: I don't see why we need this if we pass the device/runtime to the program.

flaub:
I think we should hide the device inside of the runtime, I don't see any reason to expose it to the outside world. In fact, some runtimes probably don't even need a device abstraction (I'm thinking about the CPU, but I could imagine others).

flaub:
To me it goes like:

At startup:
1. all runtimes register themselves

When the user wants to enumerate devices:
1. Walk thru each registered runtime
2. Ask each runtime for the list of devices found
3. Merge the devices from each runtime and return a single array to the user. (Let's not have the user make a roundtrip for each device).

User selects a device by passing the `device_id` to `plaidml_jit()`:
1. The `Executable` knows how to parse a `device_id` so that it can figure out that it needs to dispatch execution to a specific runtime.

pmlc/rt/device_id.h
flaub: Use `pmlc::rt`

pmlc/runtime/runtime.h
```diff
@@ -0,0 +1,57 @@
...
+public:
+  virtual ~Runtime() {}
+
+  // Returns the devices supported by this Runtime.
+  virtual std::size_t deviceCount() const noexcept = 0;
+  virtual std::shared_ptr<Device> device(std::size_t idx) = 0;
```
flaub, line 54: Return a list of `device_id`s here, instead of making the user roundtrip for each one.

pmlc/runtime/runtime_registry.cc
```diff
@@ -0,0 +1,90 @@
...
+#include <stdexcept>
+#include <utility>
+
+#include "pmlc/util/logging.h"
+#include "llvm/Support/FormatVariadic.h"
+
+namespace pmlc::runtime {
+namespace {
+
+struct LoaderList {
```
flaub, line 16: Can you explain the design of the `Loader`? My thinking was that we could just statically initialize runtimes and have the problem of which runtimes are available be a linker issue.

pmlc/runtime/internal.h
```diff
@@ -0,0 +1,45 @@
...
+// getSymbolMap returns the map of symbols registered by components within the
+// current process.
+llvm::StringMap<void *> &getSymbolMap();
+
+// getDeviceIDRegex returns the regular expression used to match device
+// identifiers.
+//
+// When matching a device identifier, match[1] will be the runtime identifier,
+// and match[3] will be the device index within that runtime.
+const std::regex &getDeviceIDRegex();
```
flaub, line 33: Let's just return a struct instead of passing regex objects around.

pmlc/runtime/internal.h
```diff
@@ -0,0 +1,45 @@
...
+class ScopedCurrentDevice {
+public:
+  explicit ScopedCurrentDevice(Device *device);
+  ScopedCurrentDevice(const ScopedCurrentDevice &) = delete;
+  ~ScopedCurrentDevice();
+};
+
+// getSymbolMap returns the map of symbols registered by components within the
+// current process.
+llvm::StringMap<void *> &getSymbolMap();
```
flaub, line 26: What's wrong with `SymbolRegistry::getInstance()`?

pmlc/runtime/internal.h
```diff
@@ -0,0 +1,45 @@
...
+// and match[3] will be the device index within that runtime.
+const std::regex &getDeviceIDRegex();
+
+// getRuntimeMap returns the map of runtimes available to the current process.
+//
+// If the registered loaders have not yet been evaluated, this call will
+// evaluate them.  The caller may use the functions within the map to obtain a
+// pointer to the associated runtime; if the runtime has not yet been
+// instantiated, this will instantiate the associated runtime, which will be
+// memoized to make subsequent lookups less expensive.
```
flaub, line 41: It seems like we should just initialize the map at startup, why defer it?

pmlc/runtime/llvm/BUILD
```diff
@@ -3,20 +3,15 @@
...
-        "xsmm.cc",
-        "prng.cc",
     ],
-    visibility = ["//visibility:public"],
     deps = [
         "//pmlc/compiler",
-        "@half",
-        "@llvm-project//llvm:Support",
-        "@llvm-project//mlir:mlir_runner_utils",
-        "@xsmm",
```
flaub, line 19: Why isn't `@xsmm` specific to the `x86` runtime? It's not really an LLVM runtime, it's more like an `x86`-specific runtime.

pmlc/runtime/runtime_registry.h
```diff
@@ -0,0 +1,82 @@
...
+namespace pmlc::runtime {
+
+// Factory is used to instantiate a Runtime.
+//
+// Factory is invoked on-demand (typically well after process initialization) to
+// instantiate a runtime, either as a specific request by the PlaidML caller or
+// for general device enumeration.  The registry is responsible for arranging
+// runtimes to be instantiated exactly once; Factory implementations do not need
+// to memoize their results.
+using Factory = std::function<std::shared_ptr<Runtime>()>;
```
flaub, line 25: Do we really need a Factory, can we just register Runtimes directly? What's the point of having an extra level of indirection?

earhart: Added a top-level comment to address this.

pmlc/runtime/symbol_registry.cc
earhart: I'm not sure when the compiler would ever use this...?  Seems like the symbols are used at JIT time; I was thinking of that as a post-compilation step.

pmlc/runtime/vulkan/vulkan_runtime.cc
earhart: We don't -- enumeration depends solely on `deviceCount()`.

earhart (commented):
On devices vs. device IDs:

We need a way to pass devices around.  We can reify `Device`, or use device IDs.  But a device ID is effectively the same thing as a `Device`, in that they're both referencing some device-y entity; deviceID just has a string API instead of something we can attach additional info to.  So it makes sense to me to expose `Device` to components outside pmlc.  But on the gripping hand: eh, six of one, half a dozen of the other.  Done.

On loaders and runtimes:

So, we have a module of code, and one of the things that module provides is some number of `Runtime` instances (we might talk about "The Vulkan runtime", but the `Runtime` object is one very particular object defined by that module).

I freely use init-time to build global objects out of whatever modules we happen to have linked in, but I dislike doing anything complex there; it's a single-threaded environment that's typically blocking the program from starting up, so I want that code to execute as quickly as it can.  So I don't like having `pmlc/rt` assume that the various modules can figure out the `Runtime` objects that they're going to expose to the global runtime map in very small constant time.

So my usual way of writing things like this is to have the init-time registration code be as simple and small as possible.  In this case, that meant defining a `Loader` type, which regardless of how complicated the module is can always be a simple `std::function` wrapping a lazy-initialization lambda, performing whatever potentially-slow initialization is required by the module.

For simple builtin modules like Vulkan and LLVM, then, we have a trivial implementation that just does the right thing, so that they can be as simple as possible.

OTOH, while I can trivially come up with cases where the loader abstraction is useful, it's definitely not needed in today's code; I can remove it if that's your preference.

plaidml/core/internal.h
```diff
@@ -110,4 +110,8 @@ void ffi_wrap_void(plaidml_error* err, F fn) {
 plaidml_datatype convertIntoDataType(mlir::Type type);
 mlir::Type convertFromDataType(plaidml_datatype dtype, mlir::MLIRContext* context);
 
+// Converts the supplied RAII strings vector into a plaidml_strings*
+// for returning via FFI.
+plaidml_strings* toFFI(std::vector<std::string> strings);
```
flaub, line 115: Why is this by-value?

plaidml/core/ffi.cc
```diff
@@ -107,6 +108,28 @@ Type convertFromDataType(plaidml_datatype dtype, MLIRContext* context) {
   llvm_unreachable("Invalid plaidml_datatype");
 }
 
+plaidml_strings* toFFI(std::vector<std::string> strings) {
+  std::unique_ptr<plaidml_string* []> elts { new plaidml_string*[strings.size()] };
+  auto result = std::make_unique<plaidml_strings>();
+
+  std::size_t idx = 0;
+  try {
+    while (idx < strings.size()) {
```
flaub, line 117: Can we just use a for loop?

plaidml/core/internal.h
earhart: `plaidml_string` wraps a `std::string`; passing the vector of strings by value means we take ownership of the strings and can move them into the `plaidml_string`s we're creating instead of having to copy them.

plaidml/core/ffi.cc
earhart: Probably.  What's the goal?

The important bit is to keep track of where we are in `elts`, for the unwind.  I wrote this the first way that came to mind here and moved on to the next bit of coding, but there're probably ways to improve it.

Looking at it, I'm pretty sure the compiler can optimize it just fine (except that we should probably explicitly capture `strings.size()` if we really care), and I kinda like how clear it is about its use of `strings[idx]` as an rvalue in the move; I think using a ranged for loop over `strings` would work, but I don't know that it's as obviously correct, since then the cursor is working more as an lvalue except where we'd be putting it through `std::move()`.

A better approach might be to refactor `plaidml_strings` so that it works a bit more like a standard container, with proper RAII semantics, while preserving its FFI utility.  We might have bigger fish to fry, though...

flaub (dismissed):

pmlc/runtime/internal.h
flaub: Yes, we can do this later tho.

plaidml/core/ffi.cc
flaub: Code clarity, not a big deal.

ci/pipeline.yml
```diff
@@ -25,7 +25,7 @@ steps:
 {{#tests}}
   - label: '{{emoji}}{{engine}}{{shard_emoji}}{{suite}}/{{workload}}/{{platform}} BATCH_SIZE={{batch_size}}'
     command: |
-      {{python}} -m pip install pyyaml --user
+      {{python}} -m pip install pyyaml numpy --user
```
flaub, line 28: This is a smell, not sure why we need this.

pmlc/rt/device_id.cc
```diff
@@ -0,0 +1,54 @@
...
+
+#include "pmlc/rt/internal.h"
+#include "pmlc/rt/runtime_registry.h"
+#include "pmlc/util/logging.h"
+
+namespace pmlc::rt {
+
+const std::regex &getDeviceIDRegex() {
+  static const std::regex re(
+      "([[:alpha:]][_[:alnum:]]*(-[[:alpha:]][_[:alnum:]]*)*)[.]([[:digit:]]+)",
```
flaub, line 15: Can we just split on `.` to avoid having to use a regex?

flaub (dismissed):

earhart:
Okay, this needed a few not-so-minor changes to the Vulkan logic, but I think it's finally ready to land.  @flaub, PTAL?

pmlc/rt/device_id.cc
```diff
@@ -0,0 +1,54 @@
...
+}
+
+std::vector<std::string> getDeviceIDs() {
+  std::vector<std::string> result;
+  for (auto &[id, runtimeFunc] : getRuntimeMap()) {
+    Runtime *runtime;
+    try {
+      runtime = runtimeFunc();
+    } catch (const std::exception &e) {
+      IVLOG(1, "Runtime " << id << " initialization failed: " << e.what());
```
flaub, line 44: Does this imply that a runtime is initialized each time the user lists the devices?

pmlc/rt/internal.h
```diff
@@ -0,0 +1,46 @@
...
+// and match[3] will be the device index within that runtime.
+const std::regex &getDeviceIDRegex();
+
+// getRuntimeMap returns the map of runtimes available to the current process.
+//
+// If the registered loaders have not yet been evaluated, this call will
+// evaluate them.  The caller may use the functions within the map to obtain a
+// pointer to the associated runtime; if the runtime has not yet been
+// instantiated, this will instantiate the associated runtime, which will be
+// memoized to make subsequent lookups less expensive.
```
flaub, line 42: It seems like we should use `plaidml::exec::init()` to call all the registered runtimes to have them do their initialization. This gives the user control for when they'd like this to take place.

pmlc/rt/runtime.h
```diff
@@ -0,0 +1,58 @@
...
+  // Returns the current Device, local to the current thread.
+  //
+  // If there is no current device, or the current device is of an unexpected
+  // type, an exception will be thrown.
+  //
+  // This is used to smuggle the current device through to runtimes that are
+  // directly invoked by lowered code, when that lowered code doesn't support
+  // the use of a passthrough parameter.
+  template <typename T>
+  static std::shared_ptr<T> current() {
```
flaub, line 24: I think this will go away in the next set of PRs, because it's not really the current device that we need to remember, it's actually the whole initialized state of the vulkan runtime which along with this will have a device to be bound to.

pmlc/rt/runtime_registry.cc
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+private:
+  std::once_flag init_once;
+  Factory factory;
+  std::shared_ptr<Runtime> runtime;
+};
+
+const std::unordered_map<std::string, std::function<Runtime *()>>
+buildRuntimeMap() {
```
flaub, line 41: Can we just build this during `plaidml::exec::init()`?

pmlc/rt/runtime_registry.h
```diff
@@ -0,0 +1,82 @@
...
+
+// Loaders instantiate Factories.
+//
+// PlaidML Loaders are (typically) simple, stateless functions, automatically
+// registered via LoaderRegistration when their containing library is loaded
+// into the process.
+//
+// Loaders are invoked on-demand (typically well after process initialization)
+// to determine the runtime factories they're able to provide.
+using Loader = std::function<std::unordered_map<std::string, Factory>()>;
```
flaub, line 35: I'm still having trouble with the idea of having a Factory, a Loader, a Runtime, and a Device. To me all we need is a Runtime, and a runtime takes IR along with a device_id to produce an Executable. Runtimes are registered at startup using static initialization, making it a linker problem. Then for initialization, each registered runtime gets initialized during `plaidml::exec::init()`.

vendor/volk/volk.BUILD
```diff
@@ -9,5 +9,6 @@ cc_library(
         "@bazel_tools//src/conditions:windows": ["VK_USE_PLATFORM_WIN32_KHR"],
         "//conditions:default": [],
     }),
+    linkstatic = 1,
```
flaub, line 12: Does this actually fix the downstream problems where we need to use `linkstatic=1`? In my experiments it doesn't have any effect.

pmlc/rt/device_id.cc
earhart: Sure -- just curious, what's the rationale?  (I liked that this made the format for device IDs clear, and that it didn't require any subsequent validation on the ordinal.  NB the device ID -> device translation only occurs when the `Executable` is created; this isn't something that happens for every inference (although I doubt it'd noticeable if it did).

pmlc/rt/device_id.cc
earhart: Yes -- in this design, runtimes own device enumeration.  (E.g. there's no way to detect the set of OpenCL devices available without actually initializing an OpenCL runtime).

pmlc/rt/internal.h
earhart: I'd rather not initialize any runtime beyond what's necessary -- if you're using `llvm_cpu.0`, why load `vulkan` into your process?  But we could definitely build the map when `plaidml::exec::init()` is called.

pmlc/rt/runtime.h
earhart: I agree -- talking w/Jeremy, it sounds like we're all on board with moving the Vulkan calls out of the lowered code, making it trivial to know what the correct device is.

pmlc/rt/runtime_registry.cc
earhart: Definitely.

pmlc/rt/runtime_registry.h
earhart: We need to parse device IDs anyway in order to figure out the correct runtime.

But: if you'd like to have `Runtime` create an `Executable`, and pass in an ordinal to indicate the logical device entity to use, we can definitely do that.  It only becomes an anti-pattern when we try to do anything else with the logical device entity, e.g. querying its properties -- the set of things you can do with a number aren't the same set of things you can do with a `Device`, and putting additional `Device` methods into `Runtime` breaks the single-responsibility principle.

I think we could get rid of `Factory`.  What it's doing is wrapping up the concept of initializing a `Runtime`, so that the associated `Runtime` can use normal constructors and destructors to manage its state -- otherwise, every `Runtime` implementation picks up another bit of state information to indicate whether or not it's been initialized.  To get rid of it, we could instead implement a `Runtime` proxy template that manages the initialization state; I think the functional approach here is easier to maintain in the face of changes to `Runtime`, but on the other hand, some `Runtime` implementations (e.g. LLVM) wouldn't need to be wrapped at all.

I've already talked a bit about `Loader` -- want me to remove it?  I think it's a useful abstraction, but we can ditch it if needed.

vendor/volk/volk.BUILD
earhart: It does get downstream dependencies to pull in libvolk.a, which seemed to fix them just fine.
Not that it did much good, though, because the next thing that I ran into was that in order to use Vulkan, our Vulkan driver helpfully loaded LLVM from the system, which immediately collided with our built LLVM library's symbols and broke everything in subtle ways that were all sorts of fun to debug.  So we still need linkstatic=1 on our tests, and we should probably come up with a better story for managing LLVM DLL collisions at some point.

earhart (commented):
Overall, I'm happy to make various changes if they're really needed.

But I also think the current PR is good enough.  The underlying philosophy of the aggressive YAGNI approach we're taking is that you don't try to make the code perfect, and you accept that you're going to be refactoring continuously as requirements are discovered; the requested changes are not fit&finish sorts of things, they're actually requirements changes (e.g. "We don't need a Device abstraction" -- fine, but it's logically a removal of a low-level requirement to reify the `Device` abstraction; it's not like fixing a typo).

Given that, and the fact that the current code works as-is, and the time I spent today integrating the earlier PR that added Vulkan performance monitoring, and the likelihood of subsequent conflicting changes, I'd like to land the current PR as-is, turn the remaining refactorings into issues to be tracked and scheduled alongside everything else, and get on with the next bit of coding.  "Don't carry changes indefinitely; check in early and often".

Thoughts?

pmlc/rt/device_id.cc
flaub: Simplicity

flaub:
Generally, if I think a PR is blocked, I'll mark it as requiring changes. But if I make comments, it generally means, let's consider the design space and then decide how we want to get to that design over time. So I'm OK with approving this for now, provided we try to simplify the design as we go, and be will to let go of abstractions that we don't need.

flaub (approved):

Issue #1241, opened 2020-08-04
Add support for strided and offset based batch reduce for pxa.gemm and XSMM.


jbruestle:
Provide a mechanism for pxa.gemm to have an additional index/indexs which exist on A + B and which represent distinct matrixes to multiply together and sum/reduce into the output C.  Provide lowerings to XSMM.  Additionally, modify stenciling pass on CPU to use this support.

PR #1339, opened 2020-09-08
Brgemm polydl

brightlaboratory:
The commits implement strided batch reduce GEMM. Related issue: https://github.com/plaidml/plaidml/issues/1241 @alheinecke @jbruestle .


jbruestle (changes requested):
2 high level blockers, and then I will review further.  

First, something is wrong with your clang-format settings and there is a great deal of whitespace changes which need to be fixed.  Did you run ./configure and then use the conda environment?  That should install the pre-commit hooks and make sure all of the format/linter settings are correct.

Second, I don't think we should copy and modify the stencil_gemm pass as it leads to a lot of code duplication.  Instead, I'd like to just change the existing pass to support batch reduce, preferably with an option to prevent the use of batch reduce (which can be done via MLIR's option mechanism, see the vectorize pass for an example).  Similarly with the PxaGemm op.  I would imagine that since batch reduce gemm is a superset of regular gemm, we don't lose any representational power.

adstraw:
@jbruestle, all, you can ignore whitespace changes in the diff using GitHub.  Click on "Files changed" then in the top right corner there is a little gear icon next to "Changes from all commits * File filter... * Jump to... ".  When you click the gear icon select "Hide whitespace changes" which makes this PR much easier to review as-is.

jbruestle:
@adstraw certainly, however 1) Fixing the whitespace is a blocker to landing the commit anyway, 2) I also want to be able to see the diffs of stencil_gemm.cc vs stencil_brgemm.cc for example, but since the code was copy/pasted I would need to checkout and do a manual diff, and the code duplication induced is also a blocker IMO.

pmlc/dialect/pxa/transforms/stencil_gemm.cc
```diff
@@ -293,6 +293,7 @@ class StencilGEMM : public StencilBase {
                                            tileAttr, mapOperands);
 
     opC.result().replaceAllUsesWith(gemm);
+    opC.erase();
```
flaub, line 296: No need for this since follow on canonicalizations can clean it up.

brightlaboratory:
@jbruestle : I made the changes you suggested: there no longer is a new path for BRGemm in Pxa Transforms. Instead, I have modified the existing PxaGemmOp related code/files to generate BRGemms (using a user provided option to choose one or the other).

I did install clang-format hooks. There appear to be still some indentation related differences. (Did the .clang-format file change between the two pushes, perhaps?) Please let me know if I should run the clang-format command line manually in a particular way to make the files to be as formatted as before. In the meantime, https://github.com/plaidml/plaidml/pull/1339/files?diff=split&w=1 should show the actual differences between the two file versions sans whitespace changes.

pmlc/dialect/pxa/transforms/stencil_gemm.cc
brightlaboratory: Yeah, could be. If one doesn't run the canonicalization pass, it results in incorrect output (I have checked). Erasing opC here should be fine? (opC isn't used anywhere else, so I believe it should always produce the correct code?)

flaub:
@brightlaboratory  Are you on Windows? It appears that every line of every file has whitespace changes. This could be because you are saving the file with CRLF line endings instead of LF line endings.

brightlaboratory:
@flaub : I did change the files on Windows (in Visual Studio) and transferred them to a linux machine for compilation. That makes sense. I will look up how to get rid of those extra white spaces. Any tricks do it readily are appreciated. Thank you.

tzerrell:
In general, git's `core.autocrlf` option is helpful for avoiding unintentional EoL whitespace changes: https://docs.github.com/en/free-pro-team@latest/github/using-git/configuring-git-to-handle-line-endings

Visual Studio also has an option for this, although it might (?) only be on a per-file basis: https://docs.microsoft.com/en-us/visualstudio/ide/encodings-and-line-breaks?view=vs-2019

jbruestle (changes requested):
Still need to fix the newlines, but I can review now with whitespace set to ignore.  Definitely improved, and the 'strategy' mechanism seems reasonable, but variable names are too short and comments are too sparse, so I can't really tell what's happening in most places.  A few bits are definitely bugs (see inline comments).

pmlc/dialect/pxa/ir/ops.td
```diff
@@ -1,271 +1,272 @@
...
+    AffineMapAttr:$cAccessMap,
+    AffineMapAttr:$cTileMap,
+    Arg<AnyMemRef, "the A memref to load from", [MemRead]>:$a,
+    AffineMapAttr:$aAccessMap,
+    AffineMapAttr:$aTileMap,
+    Arg<AnyMemRef, "the B memref to load from", [MemRead]>:$b,
+    AffineMapAttr:$bAccessMap,
+    AffineMapAttr:$bTileMap,
+    I64ArrayAttr:$tile,
+    I64Attr:$lBr,
```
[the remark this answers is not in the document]
brightlaboratory: Sure. I renamed it to "numBatches" which I hope should be more descriptive. It's the number of "batches" of matmuls that are to be performed.

pmlc/dialect/pxa/transforms/stencil_gemm.cc
```diff
@@ -1,355 +1,450 @@
...
+    for (size_t i = 0; i < getBlockArgsAsSet().size(); i++) {
+      for (size_t j = 0; j < getTiledIdxCount(); j++) {
+        if (perm.indexes[j] == op.getBody()->getArgument(i)) {
+          steps[i] *= tileSize[j];
+
+          // K index
+          if (j == 2) {
+            IVLOG(3, "steps[" << i << "] = " << steps[i]);
+            if (kRange != -1) {
+              l_br = kRange / steps[i];
```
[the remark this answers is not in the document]
brightlaboratory: I renamed the variable to numBatches and have included a detailed comment as to what is being attempted here. I am pasting it in this comment also:

`              /*
              We want to transform "regular" pxa.gemm where numBatches is 1:
              affine.parallel (i, j, k) = (..., 0) to (..., kRange) step (.., kStep) {
                pxa.gemm C[i, j] = A[i, k], B[k, j]: [..., kStep], 1
              }

              to

              affine.parallel (i, j, k) = (..., 0) to (..., kRange) step (.., kRange) {
                pxa.gemm C[i, j] = A[i, k], B[k, j]: [..., kStep], (kRange/64)
              }

              where the number of batches of A and B matrices to multiply is
              the k loop's range divided by the original step size for the k loop.

              Subsequently, kStep is set to kRange. That is, in one step, a block of C
              is completely computed through reduction of batches of A and B matrix multiplies.
              */`

jbruestle:
I fixed the whitespace issues.  Remaining blockers: 1) Probably need to update to latest head 2) CI is currently failing (see the 'required' buildkite/plaidml-plaidml/pr link on 'checks').

pmlc/dialect/pxa/transforms/stencil_gemm.cc
```diff
@@ -260,12 +264,57 @@ class StencilGEMM : public StencilBase {
   }
 
   void transform(TensorAndIndexPermutation perm, ArrayRef<int64_t> tileSize) {
+    PxaGemmOp gemm;
```
flaub, line 267: What's this for?

pmlc/dialect/pxa/transforms/stencil_gemm.cc
```diff
@@ -58,7 +60,9 @@ struct GemmOperand {
 class StencilGEMM : public StencilBase {
 private:
   unsigned numThreads;
+  std::string strategy;
```
flaub, line 63: Let's use an `enum` instead of a `string`. Generating an enum with tblgen will give us the ability to convert to/from a string.

pmlc/dialect/pxa/transforms/stencil_gemm.cc
```diff
@@ -345,11 +448,17 @@ struct StencilGEMMPass : public PassWrapper<StencilGEMMPass, FunctionPass> {
   Option<unsigned> numThreads{
       *this, "threads",
       llvm::cl::desc("Specifies number of threads for the stencil pass")};
+
+  Option<std::string> strategy{
+      *this, "strategy", llvm::cl::desc("Specifies the stenciling strategy")};
 };
 
 std::unique_ptr<Pass> createStencilGEMMPass(unsigned numThreads,
+                                            std::string strategy,
```
flaub, line 457: Use `llvm::StringRef`

pmlc/target/x86/xsmm_lowering.cc
```diff
@@ -201,6 +215,138 @@ struct XSMMGemmInvokeLowering
...
+    auto int64Type = LLVM::LLVMType::getInt64Ty(op->getContext());
+    return rewriter.create<LLVM::LLVMFuncOp>(
+        rewriter.getUnknownLoc(), kBRGemmDispatchF32,
+        LLVM::LLVMType::getFunctionTy(int64Type,
+                                      ArrayRef<LLVM::LLVMType>{int32Type, // lda
+                                                               int32Type, // ldb
+                                                               int32Type, // ldc
+                                                               int32Type, // m
+                                                               int32Type, // n
+                                                               int32Type}, // k
```
flaub, line 277: Use `clang-format`

pmlc/dialect/pxa/transforms/stencil_gemm.cc
brightlaboratory: Hi @flaub : Did you mean to declare the strategy like it's done for "pxa-vectorize": https://github.com/plaidml/plaidml/blob/brgemm_polydl/pmlc/dialect/pxa/transforms/passes.td#L129 ? 

But, "x86-affine-stencil-xsmm" pass is not declared in pxa/transforms/passes.td. Instead it's declared in target/x86/passes.td. From target/x86/pipeline.cc, pxa/transforms/stencil_gemm.cc is invoked. I had tried adding options to pmlc/target/x86/passes.td here: https://github.com/plaidml/plaidml/blob/brgemm_polydl/pmlc/target/x86/passes.td#L25. Curiously, the options were not being read. I am not sure why.

pmlc/dialect/pxa/transforms/stencil_gemm.cc
brightlaboratory: Okay. Done: https://github.com/plaidml/plaidml/blob/brgemm_polydl/pmlc/dialect/pxa/transforms/stencil_gemm.cc#L457

pmlc/dialect/pxa/transforms/stencil_gemm.cc
flaub: No, I mean we should create an enum with tblgen and then use it to define the available strategies. The pass would still be configured with a string, it's just that the string can be converted into the enum and the rest of the code would use the enum.

pmlc/dialect/pxa/transforms/stencil_gemm.cc
```diff
@@ -260,12 +264,57 @@ class StencilGEMM : public StencilBase {
...
     // First, modify step size of all tiled indexes
     auto steps = op.getSteps();
     for (size_t i = 0; i < getBlockArgsAsSet().size(); i++) {
       for (size_t j = 0; j < getTiledIdxCount(); j++) {
         if (perm.indexes[j] == op.getBody()->getArgument(i)) {
           steps[i] *= tileSize[j];
+
+          // K index (reduction dimension)
+          if (j == 2) {
+            /*
```
flaub, line 292: Please use C++ style comments instead of C style ones.

pmlc/rt/xsmm.cc
```diff
@@ -34,6 +34,35 @@ extern "C" int64_t plaidml_rt_xsmm_gemm_dispatch_f32(int32_t lda, int32_t ldb,
   return reinterpret_cast<int64_t>(sgemm);
 }
 
+extern "C" void plaidml_rt_xsmm_brgemm_invoke_f32(int64_t funcAddr, float *a,
+                                                  float *b, float *c,
+                                                  int64_t lBr) {
```
flaub, line 39: What is `lBr`? Can we use a more descriptive variable name?

pmlc/dialect/pxa/transforms/stencil_gemm.cc
flaub: Why do we need a strategy in the first place? It seems like we should have a cost model that decided when to use a batched vs non-batched GEMM. I guess the question is, why would we ever want to use a non-batched GEMM? Is there some overhead/cost to doing so?

pmlc/rt/xsmm.cc
brightlaboratory: Done. I changed it to `numBatches`.

pmlc/dialect/pxa/transforms/stencil_gemm.cc
brightlaboratory: Yeah, in the future it would be nice to have a cost function which will use one or the other variant automatically. For now, I think that there is value to having both of them and switching between them manually. Which one will perform better would be a function of cache sizes and parallelization. Additionally, the quality of microkernels JITed, if there is a difference, would determine which one would be better.

pmlc/dialect/pxa/transforms/stencil_gemm.cc
```diff
@@ -279,24 +368,29 @@ class StencilGEMM : public StencilBase {
...
+        AffineMapAttr::get(b.tileMap), //
+        tileAttr, numBatchesAttr, mapOperands);
 
     opC.result().replaceAllUsesWith(gemm);
+    opC.erase();
   }
 
 public:
-  StencilGEMM(AffineParallelOp op, unsigned numThreads,
+  StencilGEMM(AffineParallelOp op, unsigned numThreads, std::string strategy,
```
flaub, line 393: Use `llvm::StringRef`

flaub: Generally, we shouldn't ever pass strings by-value because that causes an implicit copy.

pmlc/rt/xsmm.cc
```diff
@@ -34,6 +34,35 @@ extern "C" int64_t plaidml_rt_xsmm_gemm_dispatch_f32(int32_t lda, int32_t ldb,
   return reinterpret_cast<int64_t>(sgemm);
 }
 
+extern "C" void plaidml_rt_xsmm_brgemm_invoke_f32(int64_t funcAddr, float *a,
+                                                  float *b, float *c,
+                                                  int64_t numBatches) {
+  libxsmm_xmmfunction sgemm;
+  sgemm.xmm = reinterpret_cast<FunctionPtr>(funcAddr);
+  unsigned long long numBatchesVar = numBatches; // NOLINT
```
flaub, line 42: Why is this NOLINT?

brightlaboratory: No idea. It must have automatically gotten inserted by one of the commit hooks -- I didn't insert it myself. I am also surprised to see it.

pmlc/dialect/pxa/transforms/stencil_gemm.cc
brightlaboratory: I removed this stray variable declaration. Thanks for pointing it out.

pmlc/rt/xsmm.cc
flaub: If you remove it, does `cpplint` complain?

jbruestle: Lint doesn't like unsigned long long.

pmlc/dialect/pxa/transforms/stencil_gemm.cc
brightlaboratory: @flaub : Hope the `enum` thing is not a blocker. For my future reference, do you have an example you could point me to? Thanks

brightlaboratory:
@jbruestle : I trust the code looks good now and there are no blockers. If so, could you please merge this branch with plaidml-v1 when you get a chance?

pmlc/dialect/pxa/transforms/stencil_gemm.cc
jbruestle: I think not erasing here violated the PXA requirement of having only one 'write' user, but regardless the erase certainly doesn't hurt, so I think leaving it is good.

pmlc/rt/xsmm.cc
jbruestle: After poking at this more, basically XSMM need unsigned long long to avoid warnings, but uint64_t (which the linter likes) is technically unsigned long on a 64 bit machine.  Which also happens to be the exact same physical type as unsigned long long on 64 bit, but they still are different logical types.  NOLINT seems the least annoying way to make things happy here.

pmlc/target/x86/xsmm_lowering.cc
jbruestle: Done.

pmlc/target/x86/pipeline.cc
```diff
@@ -117,8 +115,7 @@ void pipelineBuilder(OpPassManager &pm) {
   pm.addPass(createCanonicalizerPass());
   pm.addPass(createCSEPass());
 
-  llvm::StringRef defaultStrategy = llvm::StringRef("simple");
-  pm.addPass(pxa::createStencilGEMMPass(/*numThreads=*/1, defaultStrategy,
+  pm.addPass(pxa::createStencilGEMMPass(/*numThreads=*/1, true,
```
flaub, line 118: Use `/*doBatch=*/` here

pmlc/target/x86/tests/xsmm_brgemm_stencil.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: pmlc-opt --pass-pipeline='x86-affine-stencil-xsmm{threads=4 strategy=strided_brgemm}' %s | FileCheck %s
+// RUN: pmlc-opt --pass-pipeline='x86-affine-stencil-xsmm{threads=4 do-batch=true}' %s | FileCheck %s
```
flaub, line 1: Perhaps rename to `batched=true` or `batch-mode=true`

pmlc/dialect/pxa/transforms/stencil_gemm.cc
```diff
@@ -260,12 +263,46 @@ class StencilGEMM : public StencilBase {
   }
 
   void transform(TensorAndIndexPermutation perm, ArrayRef<int64_t> tileSize) {
+    auto AStrideInfo = getStrideInfo(perm.ioOps[1]);
```
flaub, line 266: Rename to `aInfo` or `aStrideInfo`

flaub (dismissed):

jbruestle (approved):
Good to land!  I know the process can be a bit brutal, so thanks for sticking with it.  Hopefully next time will be easier.

PR #1340, opened 2020-09-08
subgroup block ops

abialas1:
Moved block ops to parametrized subgroup broadcast pass.
Added fix for vectorizeBuffers, where crash occurs on last iterator. Also changed types to vectors in this loop where appropriate. Still this does not work as expected due to issue in the returnOp memref mismatch.


jbruestle (approved):
Looks great!  Regarding alignment, it's an interesting question.  It would be easy to calculate at the PXA level (basically a one-liner) but we would somehow have to smuggle it down to the stdx-subgroup-broadcast pass.  At this level, verifying alignment is quite a chore.  I'll noodle on how to fix this, but it's not a blocker considering the block reads/writes are disabled by default for now anyway.

PR #1345, opened 2020-09-10
Enable ResNext Test

dgkutnic:
Need to resolve:
* ~`resnext_test_gen` (the py_binary) works correctly but `resnext_test_genrule` (the genrule that runs the binary) can't find the SavedModel that I am pulling in from `@models_resnext50`~
* `resnext_test.cc` is 526kb, slightly over the 500kb limit for precommit. It may be possible to pull out the module text from the test file and put it somewhere else.
* need to fix segfault happening in `resnext_test.cc`, currently debugging by turning up verbosity and seeing what I can figure out


plaidml/bridge/tensorflow/tests/resnext_test.cc
```diff
@@ -0,0 +1,3473 @@
...
+#include "tensorflow/compiler/xla/tests/test_utils.h"
+#include "tensorflow/compiler/xla/tests/verified_hlo_module.h"
+
+#include "plaidml/bridge/tensorflow/tests/archive_generated.h"
+#include "plaidml/bridge/tensorflow/tests/codegen_test.h"
+
+using plaidml::edsl::MultiBuffer;
+namespace zoo = plaidml::zoo;
+
+static MultiBuffer convertBuffer(const zoo::DataUnion& data) {
```
flaub, line 18: Would be great if we could refactor this so that it's in the `codegen_test.cc`, so that it's not duplicated between multiple networks.

dgkutnic: I agree, I'll refactor this once I have ResNext working.

plaidml/bridge/tensorflow/tests/resnext_test_gen.py
```diff
@@ -0,0 +1,59 @@
...
+    builder.Finish(packed)
+    args.output.write(builder.Output())
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Generate archive for resnext')
+    parser.add_argument('output',
+                        type=argparse.FileType('wb'),
+                        help='location to write the generated archive')
+    parser.add_argument('saved_model_path', help='location where the saved_model is located')
```
flaub, line 57: If you use `type=pathlib.Path`, then later when you use it, you can say: `args.saved_model_path / model_name`

plaidml/bridge/tensorflow/tests/BUILD
```diff
@@ -200,6 +201,7 @@ tf_cc_test(
     deps = [
         ":archive_cc_fbs",
         ":codegen_test",
+	":utils",
```
flaub, line 204: whitespace, use buildifier

plaidml/bridge/tensorflow/tests/utils.cc
```diff
@@ -0,0 +1,55 @@
+#include "plaidml/bridge/tensorflow/tests/utils.h"
```
flaub, line 1: Could we merge all this into `codegen_test.cc`? Right now we have a `py_library` called `util` and a `cc_library` called `utils`.

flaub:
There seems to be changes to I3D in this PR, could you split that out?

dgkutnic:
Sorry about that! It should be fixed now. The only I3D changes that are in here are related to the refactoring which pulled `convertBuffer` and `ReadFile` into `codegen_test.cc`

flaub:
Since it takes about 3 minutes to run resnext right now, we need to avoid making it be a bazel test rule unless we mark it as a 'large' test. So I think we should instead consider running this like we run networks, outside of bazel.

dgkutnic:
That sounds reasonable to me. I'll go ahead and make those changes.

dgkutnic:
Closing this and migrating to #1472

PR #1346, opened 2020-09-10
implement tile.gather operation and enable backend tests

mars1026:


plaidml/bridge/keras/backend_test.py
```diff
@@ -1515,30 +1515,33 @@ def testReverse(self, b, x, ax):
          ]],
         skip_theano=False,
         skip_tensorflow=True)
-    @unittest.skip("gather is not yet implemented")
     def testGather(self, b, v):
         I = b.variable(np.array([0, 2, 1, 0], dtype='int32'), dtype='int32')
-        I2 = b.variable(np.array([[2, 1], [0, 1], [1, 0], [2, 1], [0, 0]], dtype='int32'),
-                        dtype='int32')
+        #I2 = b.variable(np.array([[2, 1], [0, 1], [1, 0], [2, 1], [0, 0]], dtype='int32'),
```
jbruestle, line 1520: If I2 is unused (and apparently was previously as well), let's delete it rather than just commenting it out.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -879,6 +884,97 @@ struct ContractionOpConversion : public OpConversionPattern<ContractionOp> {
...
+
+    TypeConverter typeConverter;
+    auto resultType = typeConverter.convertType(op.result().getType());
+    auto memrefType = resultType.cast<MemRefType>();
+
+    // Make an allocation for the output
+    auto resultMemRef = rewriter.create<AllocOp>(loc, memrefType).getResult();
+
+    // affine loop nest
+    // symbol representing load of index from index tensor
```
jbruestle, line 915: Out of date comment?

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -879,6 +884,97 @@ struct ContractionOpConversion : public OpConversionPattern<ContractionOp> {
...
+    // we need an array of int64_t representing the results tensor's dims
+    ArrayRef<int64_t> size = memrefType.getShape();
+
+    auto loop = rewriter.create<AffineParallelOp>(
+        loc, ArrayRef<Type>{memrefType},
+        ArrayRef<AtomicRMWKind>{AtomicRMWKind::assign}, size);
+
+    auto txBuilder = loop.getBodyBuilder();
+
+    // create an affine map for loading the index, using the leading counters
```
jbruestle, line 926: This block of code is correct, but it would be easier to do:
auto idxLoadMap = AffineMap::getMultiDimIdentityMap(idxDims.size(), ctx)
auto idxLoadOps = loop.getIVs().slice(0, idxDims.size())

ArrayRef and MutableArrayRef are awesome BTW (slice, etc).

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -879,6 +884,97 @@ struct ContractionOpConversion : public OpConversionPattern<ContractionOp> {
...
+    SmallVector<Value, 4> srcOps;
+    srcOps.push_back(indexVal);
+    for (size_t i = idxDims; i < dstDims; ++i) {
+      srcOps.push_back(loop.getIVs()[i]);
+    }
+
+    // load the specified value from the source tensor
+    auto loaded = txBuilder.create<mlir::LoadOp>(loc, tensor, srcOps);
+
+    // create a destination map using all of the dimensions
```
jbruestle, line 959: Similarly to the above, no need for a loop here.

jbruestle (changes requested):
Looks great!  Glad to have this working.  Some minor comments + MLIRisms inline

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
flaub: Actually you could use `loop.getIVs().take_front(idxDims.size())` I think?

flaub:
It'd be really useful to add a `.mlir` test to show what we expect the tile code to look like on input and what we expect the pxa code to look like on output of the conversion.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
jbruestle: Yeah, even better.

mars1026: That's cool!

flaub:
LGTM

jbruestle (approved):
Awesome!

PR #1347, opened 2020-09-10
Abstract executable creation

earhart:


plaidml/exec/ffi.cc
```diff
@@ -114,7 +114,7 @@ plaidml_executable* plaidml_jit(  //
     } else if (jit == "MCJIT") {
       kind = EngineKind::MCJIT;
     }
-    exec->exec = std::make_unique<Executable>(program->program, deviceID, bufptrs, kind);
+    exec->exec = Executable::fromProgram(program->program, deviceID, bufptrs, kind);
```
flaub, line 117: Seems to be like we'd want to create an `Executable` from a `Runtime`?

flaub: Like:

```
Runtime* runtime = Runtime::getRuntime(deviceId);
Executable* executable = runtime->createExecutable(program, bufptrs); // kind can be wrapped up in the Runtime/deviceID it seems
```

earhart: Yes -- I think that'll be in a future PR, though.  (And I think we want the deviceID->Runtime lookup to happen over in pmlc.)

flaub: OK, so what does this PR buy us then?

earhart: Before this PR, ffi.cc et al think that Executable is some concrete class that they can go instantiate.  After this PR, they go ask pmlc to build them some object with the Executable interface.  (It's a small change, for sure, but I prefer small changes, even if I do sometimes wind up producing large-ish PRs; I think small PRs are easier to review and understand.)

I think the next step after this will be to move the concrete Executable implementations into the runtimes, so that what's left in `pmlc/rt/executable.cc` is just deviceID parsing and dispatch.

flaub: Agreed on small changes, I just don't see how this gets us closer to where we want to be...

flaub (approved):

PR #1348, opened 2020-09-10
Load runtimes in plaidml::exec::init

earhart:


earhart:
Chatted w/@flaub -- we'll do this a different way; stay tuned.

PR #1350, opened 2020-09-10
Simplify runtime creation

earhart:


pmlc/rt/runtime_registry.cc
```diff
@@ -2,89 +2,31 @@
...
 
 } // namespace details
 
-const std::unordered_map<std::string, std::function<Runtime *()>> &
-getRuntimeMap() {
-  static std::unordered_map<std::string, std::function<Runtime *()>>
-      runtimeMap = buildRuntimeMap();
-  return runtimeMap;
+void initRuntimes() {
+  [[maybe_unused]] static bool inited = []() {
```
flaub, line 24: I've been using 
```
    static std::once_flag is_initialized;
    std::call_once(is_initialized, []() {
      // do init here
    });
```

Any reason why we shouldn't use that here?

To me, I can look at the `std::call_once` docs to understand how it works, while looking at a lambda in c++ with a strange attribute makes me pause and wonder if it does the right thing or not (I think it does...).

earhart: They probably compile to the same code.

The static bool is guaranteed to be initialized exactly once, the first time the block is entered, with a barrier to synchronize parallel callers, and re-trying if there's an exception -- this is why Meyers singletons work.

With `call_once`: a naive compiler might generate code that checks that `is_initialized` has been initialized every single time the function is entered, and then do another atomic check + barrier in `call_once` to see whether it should run the lambda. But a smart compiler can use `once_flag`'s constexpr constructor to do the initialization at compilation time and elide the first runtime check, so I think it's fine.

Reviewing it myself, though, I think there's a bigger problem: I'm not sure what to do when a `Runtime::init()` throws an exception; the caller might call `initRuntimes()` again, so we either need to make `Runtime::init()` idempotent or we need to remember that we've called a Runtime's `init()` method somewhere.  And we should probably continue on to initialize all of the `Runtime`s, or declare that you can't use the system unless all `Runtime`s successfully initialize and `initRuntimes()` completes normally.  I'll work on this a bit more and send an update.

jbruestle: I think Franks point is the call_once is easier to read (and I agree), and even if the compiler is dumb (hint: it's not), we save < 1us per call to initRuntimes, which is probably called exactly once anyway, so use the easier to read one.

earhart: Sorry -- forgot to add that I'm happy to change it if you guys prefer `call_once`.
Should we also do the same for other cases where we use Meyers singletons, e.g. `getRuntimeMap()`?

flaub: Meyers singletons are reasonable, it's the usage of a lambda with the non-standard attribute + being a Meyers singleton that gives me pause.

I'd prefer if we used `std::call_once` in the places that it makes sense (to me the impl of `getRuntimeMap` does not qualify). If we can encapsulate the initialization in the construction of an object, then I'd say a Meyers singleton is great. But in this case, we're calling a sequence of code, that seems like better suited for `std::call_once`

earhart: SGTM.  (FWIW, `maybe_unused` is a standard attribute.  :-)

pmlc/rt/internal.h
```diff
@@ -34,13 +34,8 @@ llvm::StringMap<void *> &getSymbolMap();
...
-// If the registered loaders have not yet been evaluated, this call will
-// evaluate them.  The caller may use the functions within the map to obtain a
-// pointer to the associated runtime; if the runtime has not yet been
-// instantiated, this will instantiate the associated runtime, which will be
-// memoized to make subsequent lookups less expensive.
-const std::unordered_map<std::string, std::function<Runtime *()>> &
-getRuntimeMap();
+// N.B. This is mutable at static initialization time, but should be treated as
+// immutable after main() has been called; there is no associated lock.
+std::unordered_map<std::string, std::unique_ptr<Runtime>> &getRuntimeMap();
```
flaub, line 39: I think if we just exposed `registerRuntime`, `getRuntime` and `enumerateRuntimes`, then we can hide the impl details and make enumerate return a const thing.

flaub: Or better yet,

```
struct Runtime
  static void registerRuntime();
  static Runtime* getRuntime(StringRef name);
  static ArrayRef<Runtime*> getAllRuntimes();

  virtual void init() = 0;
  // etc
};
```

earhart: NB `getRuntimeMap()` is declared in `internal.h`; it's not public, and everything in this module should be using it correctly already.

pmlc/rt/runtime_registry.cc
earhart: Done

pmlc/rt/runtime.h
```diff
@@ -46,13 +47,32 @@ class Device {
...
 // process, constructed once as needed (typically well after process
 // initialization) and then held onto by the global runtime registry.
+//
+// N.B. The Runtime lifecycle is a little unusual.
+//
+//      Typically, Runtime will be instantiated (via RuntimeRegistration) at
+//      static initialization time, i.e. before main() is invoked.  Since this
+//      is an unusual process environment in which various global objects may or
+//      may not be initialized, Runtime implementations may defer complicated
+//      initialiation to their init() method; the contract with the caller is
```
flaub, line 57: I don't think we should say they 'may' , but they 'must'. All initialization should happen in `init()`. Could we drop the 'unusual' business?

pmlc/rt/runtime_registry.h
```diff
@@ -2,80 +2,36 @@
...
-    details::registerLoader(id, std::move(loader));
-  }
-};
-
-// RuntimeRegistration is used to register a PlaidML Runtime class at image load
-// time.  This is used by runtimes that are linked into the image (vs. being
-// dynamically discovered and loaded).
+// N.B. This requires that an instance of the runtime must be instantiated at
+// static initialization time.
+//      Implementations should delegate non-trivial initialization to init().
```
flaub, line 25:
```
Implementations must perform initializations within `init()`.
```

pmlc/tools/pmlc-jit/pmlc-jit.cc
```diff
@@ -96,6 +97,7 @@ int main(int argc, char **argv) {
   llvm::InitializeNativeTarget();
   llvm::InitializeNativeTargetAsmPrinter();
   mlir::initializeLLVMPasses();
+  pmlc::rt::Runtime::initRegisteredRuntimes();
```
flaub, line 100: Would be nice if we had `pmlc::InitializeRuntimes()` so that we are a little more consistent with LLVM.

pmlc/rt/runtime_registry.h
```diff
@@ -2,80 +2,36 @@
...
-//
-// Loaders are invoked on-demand (typically well after process initialization)
-// to determine the runtime factories they're able to provide.
-using Loader = std::function<std::unordered_map<std::string, Factory>()>;
-
 namespace details {
 
-// Creates a global registration for a loader.
-void registerLoader(llvm::StringRef id, Loader loader);
+void registerRuntime(llvm::StringRef id, std::unique_ptr<Runtime> runtime);
```
flaub, line 16: Why is this a 'detail', why not just have a public API here.

pmlc/rt/runtime_registry.cc
```diff
@@ -2,89 +2,22 @@
...
-          });
-      if (!inserted) {
-        throw std::runtime_error{
-            llvm::formatv("Multiple runtime implementations found for {}", id)};
-      }
-    }
-  }
-
-  return result;
+std::unordered_map<std::string, std::unique_ptr<Runtime>> &getRuntimeMap() {
```
flaub, line 10: It took me awhile to find this function, because it's in `internal.h`, it wasn't clear to me that I should look in `runtime_registry.cc`.

pmlc/rt/runtime_registry.cc
flaub: TIL

pmlc/rt/runtime.h
earhart: Done

pmlc/rt/runtime_registry.h
earhart: Done

pmlc/tools/pmlc-jit/pmlc-jit.cc
earhart: Done

pmlc/rt/runtime_registry.h
earhart: Done

pmlc/rt/runtime_registry.cc
earhart: Maybe we should dump `internal.h` and put everything private into the `details` namespace?

pmlc/rt/runtime_registry.h
earhart: ... it's a little dangerous to use, though.  If `plaidml::exec::init()` might have been called, the new runtime will not have its `init()` method called by the system, so the caller needs to be very careful.

flaub: I'm going to make it a requirement that we make this function public, because in other contexts, I will want to call this directly without using static initializers.

pmlc/rt/runtime_registry.cc
flaub: What if we just hung all this stuff off of `Runtime` itself (like static methods on the base)?

pmlc/rt/runtime_registry.h
earhart: It remains Done.  :-)

pmlc/rt/runtime.h
```diff
@@ -46,13 +47,40 @@ class Device {
 // evaluating a PlaidML program.  These are typically single-instanced within a
 // process, constructed once as needed (typically well after process
 // initialization) and then held onto by the global runtime registry.
+//
+// N.B. The Runtime lifecycle is a little unusual.
```
flaub, line 51: Can we just say:

```
The Runtime Lifecycle:
```

And then let the reader decide if they think it's unusual or not.

pmlc/rt/runtime_registry.h
```diff
@@ -2,80 +2,33 @@
...
-struct LoaderRegistration {
-  LoaderRegistration(llvm::StringRef id, Loader loader) {
-    details::registerLoader(id, std::move(loader));
-  }
-};
-
-// RuntimeRegistration is used to register a PlaidML Runtime class at image load
-// time.  This is used by runtimes that are linked into the image (vs. being
-// dynamically discovered and loaded).
+// N.B. This requires that an instance of the runtime must be instantiated at
```
flaub, line 20: What requires that it happens at static initialization time? I can imagine scenarios, in fact I'm planning on scenarios, where this will happen in `main()`, not at static initialization time.

pmlc/rt/runtime.cc
```diff
@@ -19,4 +19,14 @@ ScopedCurrentDevice::~ScopedCurrentDevice() { currentDevice = nullptr; }
 
 std::shared_ptr<Device> Device::currentUntyped() { return currentDevice; }
 
+void Runtime::initRegisteredRuntimes() {
+  static std::once_flag inited;
+  std::call_once(inited, []() {
+    for (auto &[id, runtime] : getRuntimeMap()) {
+      std::call_once(runtime->initialized,
+                     [runtime = runtime.get()]() { runtime->init(); });
+    }
```
flaub, line 28: Seems like we need to skip runtimes that fail to initialize. Or we could remove them from the map if they fail. Thoughts?

pmlc/rt/runtime.h
earhart: Okay.  But it _is_ unusual: in this design, `Runtime` is not a normal object, initialized to a usable state in its constructor; there's this extra step you need to know to perform, and that is definitely worth calling out in big bold flashing marquee text.

flaub: That's subjective, please leave that out of code comments :)

earhart:
Okay, thinking about this one overnight: this was meant to be a small cleanup PR, but it's become a bigger refactoring; I want to take a step back and think some more about where it could/should go.  I'm going to pull it for now, and replace it with a smaller, more targeted change that directly addresses a couple of requirements that were elicited by this discussion -- things we know we need sooner rather than later.  Stay tuned.

PR #1352, opened 2020-09-11
clean up IVLOG1 logs

adstraw:


pmlc/dialect/pxa/transforms/vectorize.cc
```diff
@@ -305,15 +305,16 @@ LogicalResult vectorizeOverOutputs(AffineParallelOp op, unsigned vectorWidth) {
   }
   auto reduce = dyn_cast<PxaReduceOp>(getPrevWriter(op.getResult(0)));
   if (!reduce) {
-    return op.emitRemark(
-        "vectorizeOverOutputs: Failed, missing previous PxaReduceOp");
+    return failure();
```
adstraw, line 308: We are hitting this case a LOT on ResNet50 which bulks up the logs quite a bit as we print the entire op IR when we do a remark.  Not sure if this is the right thing to do here, or not.

jbruestle (approved):

pmlc/dialect/pxa/transforms/vectorize.cc
```diff
@@ -305,15 +305,16 @@ LogicalResult vectorizeOverOutputs(AffineParallelOp op, unsigned vectorWidth) {
...
+    // return op.emitRemark(
+    //    "vectorizeOverOutputs: Failed, missing previous PxaReduceOp");
```
flaub, line 310: Remove comments

plaidml/op/lib/ops.cc
```diff
@@ -431,7 +431,7 @@ std::vector<int64_t>* extend_manual_padding(std::vector<int64_t>* pads, size_t r
 }  // namespace
 
 Value abs(const Value& value) {
-  IVLOG(1, "abs");
+  IVLOG(2, "abs");
```
flaub, line 434: Hmm, I think these are good to have at this level.

pmlc/dialect/tile/builder.cc
```diff
@@ -578,7 +578,7 @@ TileBuilder::MakeProgram(StringRef name, const ProgramMutations &mutations,
   if (name.empty()) {
     name = "noname";
   }
-  IVLOG(1, "TileBuilder::MakeProgram> " << name.str());
+  IVLOG(2, "TileBuilder::MakeProgram> " << name.str());
```
flaub, line 581: I was hoping to keep this one at 1

plaidml/op/lib/ops.cc
jbruestle: Why specifically?  I'd like to keep log level 1 to basic status, basically something with O(1) lines per program execution or thereabouts.  Basically if you run resnet50 at log level 1, it spews > 2k lines prior to this commit.

flaub: Yeah, this should be happening at the O(1) level.

pmlc/dialect/tile/builder.cc
jbruestle: Does this get called every time Keras does shape inference?  I think that might be the issue.  If not, I agree that it's high level enough to keep.

adstraw: As an example, this is called 342 times during a ResNet50 run on Intel Gen9 GPU.  Each time we print quite a lot of IR.  Log size takes a big hit:

```
$ ls -la ~/temp/logMakeProgram 
-rw-rw-r-- 1 adstraw adstraw 37026716 Sep 14 13:03 <path>/logMakeProgram

$ ls -la ~/temp/noLogMakeProgram 
-rw-rw-r-- 1 adstraw adstraw 196395 Sep 14 13:09 <path>/noLogMakeProgram
```

pmlc/dialect/pxa/transforms/vectorize.cc
adstraw: Will do.

plaidml/op/lib/ops.cc
adstraw: What's the verdict?  O(1) or O(2)?  These logs aren't very illuminating, in my opinion.  How does logging the operation name help?

pmlc/dialect/tile/builder.cc
flaub: OK, yes, I could see this causing lots of logs for keras. Level 2 is fine here.

plaidml/op/lib/ops.cc
flaub: Usually, this comes up like something crashed and it'd be interesting to see which operation caused it. The `O(1)` means that we expect the log to be printed once (not like `O(n)` or `O(n*n)`). Overall it's not a big deal, I just didn't think the oplib logs were excessive so far.

adstraw: OK, I think it is reasonable to put these back.  Slight increase in log size for ResNet50.  235535 vs. 196395.  Same order of magnitude.

adstraw:
Follow up PR:
https://github.com/plaidml/plaidml/pull/1357

PR #1353, opened 2020-09-12
rm Runtime loader, add post-main binding

earhart:


pmlc/rt/runtime_registry.cc
```diff
@@ -13,20 +14,17 @@
...
 
-struct LoaderList {
-  std::mutex mutex;
-  std::list<std::pair<std::string, Loader>> loaders;
-
-  static LoaderList &instance() {
-    static LoaderList ll;
-    return ll;
-  }
+class DuplicateRuntimeError final : public std::runtime_error {
```
flaub, line 17: So far we haven't been using custom exceptions, is `std::runtime_error` insufficient?

pmlc/rt/runtime_registry.cc
```diff
@@ -37,54 +35,86 @@ class MemoizedFactory final {
...
-  for (auto &[loaderId, loader] : ll.loaders) {
-    std::unordered_map<std::string, Factory> factories;
-    try {
-      factories = loader();
-    } catch (const std::exception &e) {
-      IVLOG(1, "Loader " << loaderId << " initialization failed: " << e.what());
-      continue;
+  void registerFactory(llvm::StringRef id, Factory factory) {
+    std::lock_guard<std::mutex> lock{mutex};
+    auto memoizer = std::make_shared<Memoizer>(std::move(factory));
```
flaub, line 49: What do we need the `Memoizer` for?

pmlc/rt/runtime_registry.cc
```diff
@@ -37,54 +35,86 @@ class MemoizedFactory final {
...
+            .emplace(id,
+                     [runtime = std::move(runtime)]() { return runtime.get(); })
+            .second;
+    if (!inserted) {
+      throw DuplicateRuntimeError{id};
+    }
+  }
+
+  Runtime *getRuntime(llvm::StringRef id) {
+    std::lock_guard<std::mutex> lock{mutex};
```
flaub, line 74: I'm happy with marking these functions as not-thread-safe so we don't have to worry about locks.

pmlc/rt/runtime_registry.cc
```diff
@@ -37,54 +35,86 @@ class MemoizedFactory final {
...
-  return runtimeMap;
+Runtime *getRuntime(llvm::StringRef id) {
+  return RuntimeRegistry::instance()->getRuntime(id);
+}
+
+std::unordered_map<std::string, Runtime *> getRuntimeMap() {
+  return RuntimeRegistry::instance()->getRuntimeMap();
+}
+
+void bindRuntime(llvm::StringRef id, std::shared_ptr<Runtime> runtime) {
```
flaub, line 116: Can we call this `registerRuntime`?

pmlc/rt/runtime_registry.cc
earhart: Done.

pmlc/rt/runtime_registry.cc
earhart: Done.

pmlc/rt/runtime_registry.cc
earhart: Done.

pmlc/rt/runtime_registry.cc
earhart: Done.

pmlc/rt/runtime_registry.h
```diff
@@ -75,8 +72,27 @@ struct LoaderRegistration {
 template <class R>
 struct RuntimeRegistration {
```
flaub, line 73: Please call this `FactoryRegistration`, since it's registering a factory. I'd prefer that we had a `RuntimeRegistration` that just registered a runtime directly.

flaub: To clarify, I'd expect to see:

```
template <class R>
struct FactoryRegistration {
  explicit RuntimeRegistration(llvm::StringRef id) {
    registerFactory(id, makeStaticFactory<R>());
  }
};

struct RuntimeRegistration {
  explicit RuntimeRegistration(llvm::StringRef id, std::shared_ptr<Runtime> runtime) {
    registerRuntime(id, runtime);
  }
};
```

pmlc/rt/runtime_registry.h
```diff
@@ -17,56 +17,53 @@ namespace pmlc::rt {
...
-// Loaders are invoked on-demand (typically well after process initialization)
-// to determine the runtime factories they're able to provide.
-using Loader = std::function<std::unordered_map<std::string, Factory>()>;
-
 namespace details {
 
-// Creates a global registration for a loader.
-void registerLoader(llvm::StringRef id, Loader loader);
+// Creates a global registration for a factory.
+void registerFactory(llvm::StringRef id, Factory factory);
```
flaub, line 43: Can we make this part of the public interface? What's the rationale for `details` namespace?

earhart: No problem; done.  (Details: because it's tricky to use correctly, because it requires that `initRuntimes()` is called after it, and because I don't think anything ever needs to use it; if you're post-main(), just instantiate your Runtime and call `registerRuntime()`.)

pmlc/rt/runtime_registry.h
earhart: I'd rather hide the whole factory thing from components that use this if possible.  Also, we can't actually call registerRuntime() at static initialization time, because we want the exception to be caught by `initRuntimes()`.  So, how about this?

flaub: OK, LGTM

flaub (approved):

PR #1354, opened 2020-09-14
Fix fusion

flaub:


pmlc/dialect/pxa/analysis/strides.cc
```diff
@@ -606,4 +684,45 @@ Optional<StrideArray> computeStrideArray(AffineMap map) {
   return ret;
 }
 
+bool hasPerfectAliasing(const RelativeAccessPattern &aRap,
+                        const RelativeAccessPattern &bRapOrig,
```
flaub, line 688: Pass by-value if you want to make a copy of this one.

jbruestle (approved):

PR #1356, opened 2020-09-15
Add buffer copy mode to only copy buffers from/to device when necessary and improve inference latency

YangleiZouIntel:
Only map the user buffers from&to host memory.


pmlc/rt/vulkan/vulkan_invocation.cc
```diff
@@ -789,29 +789,32 @@ void VulkanInvocation::submitCommandBuffersToQueue() {
...
-                                           reinterpret_cast<void **>(&payload)),
-                               "vkMapMemory");
-            std::memcpy(hostMemoryBuffer.ptr, payload, hostMemoryBuffer.size);
-            vkUnmapMemory(device->getDevice(), deviceMemoryBuffer.deviceMemory);
-          }
-        }
+void VulkanInvocation::updateLastShaderHostMemoryBuffers() {
+  const auto &action = schedule.back();
+  auto kernel = std::dynamic_pointer_cast<LaunchKernelAction>(action);
+  if (!kernel) {
```
adstraw, line 795: Is there a different way to handle this case?  Throwing here seems pretty severe.  Could we search from back-toward-front until we find a LaunchKernelAction?

adstraw (commented):
Wondering about the `throw` but otherwise looks good.  Will leave final approval to @jbruestle based on the conversation about multiple-output nets, etc.

pmlc/rt/vulkan/vulkan_invocation.cc
YangleiZouIntel: Yes, I agree we can search back to front and avoid throws exception. This patch is sort of a workaround I think we should check and map all buffers needed not only the last a few buffers as Jeremy mentioned in the meeting.

adstraw:
Good ResNet50 results on my local machine (below).  Overall inference time, measured by stopwatch, down from about 1.5s to 1.1s with this change.

```
2020-09-15 11:08:40,953 VERBOSE-1 [default] Overall Vulkan time: 243.792ms
2020-09-15 11:08:40,953 VERBOSE-1 [default] Total Vulkan kernels: 1367
2020-09-15 11:08:40,953 VERBOSE-1 [default] Total Vulkan kernel exec time: 163.864ms
2020-09-15 11:08:40,953 VERBOSE-1 [default] Percentage Vulkan kernel exec time: 67.2148%
2020-09-15 11:08:40,953 VERBOSE-1 [default] Total Vulkan memory transfers: 1644
2020-09-15 11:08:40,953 VERBOSE-1 [default] Total (esimtated) Vulkan memory transfer time: 79.9276ms
2020-09-15 11:08:40,953 VERBOSE-1 [default] Percentage (estimated) Vulkan memory transfer time: 32.7852%
Example finished, elapsed: 12.281s (compile), 1.148s (execution)

-----------------------------------------------------------------------------------------
Network Name         Inference Latency         Time / FPS          
-----------------------------------------------------------------------------------------
resnet50             1148.32 ms                1148.32 ms / 0.87 fps
Correctness: PASS, max_error: 7.918380106275436e-06, max_abs_error: 1.4603137969970703e-06, fail_ratio: 0.0
```

pmlc/rt/vulkan/vulkan_invocation.cc
jbruestle: Yes.  We do know which buffers are user buffers (passed in as arguments) vs locals (alloced within the original function), so we should base the decision on that information rather than kernel #.  I'd rather not commit this change as as.

pmlc/rt/vulkan/wrappers.cc
```diff
@@ -34,12 +34,12 @@ namespace {
 template <typename T>
 void bindBuffer(void *vkInvocation, DescriptorSetIndex setIndex,
                 BindingIndex bindIndex, uint32_t bufferByteSize,
-                ::UnrankedMemRefType<T> *unrankedMemRef) {
+                uint32_t bufferType, ::UnrankedMemRefType<T> *unrankedMemRef) {
```
adstraw, line 37: Use BufferType instead of uint32_t

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -265,11 +265,21 @@ ConvertGpuLaunchFuncToVulkanCalls::bindBuffers(Location loc, OpBuilder &builder,
           builder.getI32IntegerAttr(numElement * elementTypeSize));
       Value unrankedBuffer = builder.create<mlir::MemRefCastOp>(
           loc, buffer, getUnrankedMemRefType(elementType));
+
+      Value bufferType;
+      if (auto arg = buffer.dyn_cast<mlir::BlockArgument>()) {
+        bufferType = builder.create<LLVM::ConstantOp>(
```
adstraw, line 271: wondering if we should use something like `isBlockArgument` and set to `true` here instead of the 1/0 uint32_t type system you have now.  Or, `isHostMemory` ?

adstraw (commented):
Just a few small things about types and names.  I think it's OK to make this PR "ready for review"

pmlc/rt/vulkan/wrappers.cc
```diff
@@ -82,8 +82,9 @@ void submitCommandBuffers(void *vkInvocation) {
 #define BIND_BUFFER_IMPL(_name_, _type_)                                       \
   void _mlir_ciface_bindBuffer##_name_(                                        \
       void *vkInvocation, DescriptorSetIndex setIndex, BindingIndex bindIndex, \
-      uint32_t bufferByteSize, ::UnrankedMemRefType<_type_> *unrankedMemRef) { \
-    bindBuffer(vkInvocation, setIndex, bindIndex, bufferByteSize,              \
+      uint32_t bufferByteSize, uint32_t buffer_type,                           \
```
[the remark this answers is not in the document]
earhart: Actually -- to keep the ABI really well-specified here, please do use uint32_t (but cast to `BufferType` immediately).

pmlc/rt/vulkan/vulkan_invocation.cc
YangleiZouIntel: Thanks, right now map only user buffers. Resnet50 inference latency improved from about 650ms to 300ms. However, some keras backend tests failed,  checking now.

pmlc/conversion/gpu/gpu_to_vulkan.cc
YangleiZouIntel: Yes, that is better. Using isBlockArgument and bool now.

pmlc/rt/vulkan/wrappers.cc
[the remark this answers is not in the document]
YangleiZouIntel: Updated with bool type isBlockArgument.

pmlc/rt/vulkan/vulkan_invocation.cc
YangleiZouIntel: Some local buffers involved in calculations on the host function, I am working on adding logic to detect these buffers and copy them from/to device.

Example 1:
testSwitch: need to copy buffer %0 from host to device.
```
   func @get_value(%arg0: memref<2x3xf32> {tile.name = "anon"}, %arg1: memref<2x3xf32> {tile.name = "anon_0"}, %arg2: memref<f32> {tile.name = "anon_1"}, %arg3: memref<2x3xf32>) {
   %c1_i32 = constant 1 : i32
   %c0_i32 = constant 0 : i32
    %c1 = constant 1 : index
    %c2 = constant 2 : index
    %c3 = constant 3 : index
    %0 = alloc() : memref<i32>
    %1 = load %arg2[] : memref<f32>
    %2 = fptoui %1 : f32 to i1
    %3 = select %2, %c1_i32, %c0_i32 : i32
    store %3, %0[] : memref<i32>
    "gpu.launch_func"(%c1, %c1, %c1, %c2, %c3, %c1, %0, %arg1, %arg0, %arg3) {kernel = @get_value_kernel::@get_value_kernel} : (index, index, index, index, index, index, memref<i32>, memref<2x3xf32>, memref<2x3xf32>, memref<2x3xf32>) -> ()
    return
  }
```

Example 2:
testAvgPool: need to copy buffer %1 and %2 from device to host.
```
  func @get_value(%arg0: memref<1x4x4x1xf32> {tile.name = "anon"}, %arg1: memref<1x1x1x1xf32>) {
    %cst = constant 0.000000e+00 : f32
    %c0_i32 = constant 0 : i32
    %c0 = constant 0 : index
    %c4 = constant 4 : index
    %c1 = constant 1 : index
    %0 = alloc() : memref<1x4x4x1xi32>
    "gpu.launch_func"(%c1, %c1, %c1, %c4, %c4, %c1, %0) {kernel = @get_value_kernel::@get_value_kernel} : (index, index, index, index, index, index, memref<1x4x4x1xi32>) -> ()
    "gpu.launch_func"(%c1, %c1, %c1, %c4, %c4, %c1, %0) {kernel = @get_value_kernel_0::@get_value_kernel} : (index, index, index, index, index, index, memref<1x4x4x1xi32>) -> ()
    %1 = alloc() : memref<1x1x1x1xi32>
    store %c0_i32, %1[%c0, %c0, %c0, %c0] : memref<1x1x1x1xi32>
    "gpu.launch_func"(%c1, %c1, %c1, %c1, %c1, %c1, %0, %1) {kernel = @get_value_kernel_1::@get_value_kernel} : (index, index, index, index, index, index, memref<1x4x4x1xi32>, memref<1x1x1x1xi32>) -> ()
    %2 = alloc() : memref<1x1x1x1xf32>
    store %cst, %2[%c0, %c0, %c0, %c0] : memref<1x1x1x1xf32>
    "gpu.launch_func"(%c1, %c1, %c1, %c1, %c1, %c1, %arg0, %2) {kernel = @get_value_kernel_2::@get_value_kernel} : (index, index, index, index, index, index, memref<1x4x4x1xf32>, memref<1x1x1x1xf32>) -> ()
    %3 = load %2[%c0, %c0, %c0, %c0] : memref<1x1x1x1xf32>
    %4 = load %1[%c0, %c0, %c0, %c0] : memref<1x1x1x1xi32>
    %5 = sitofp %4 : i32 to f32
    %6 = divf %3, %5 : f32
    store %6, %arg1[%c0, %c0, %c0, %c0] : memref<1x1x1x1xf32>
    return
  }
```

adstraw:
@earhart does this latest commit work for you in terms of the ABI?

earhart:
@adstraw No -- how wide is a bool?  I've used systems where it's a single byte, and where it's a full machine word, and where it could be either depending on the API (Windows NT, and yes, this caused bugs).

I think you should pick a fixed integer size for the ABI, and then use an enum internally (because bool parameters to APIs are also a source of bugs).

earhart:
@adstraw @YangleiZouIntel If you do use bool, please make sure to intention comments on callees -- https://llvm.org/docs/CodingStandards.html#comment-formatting

YangleiZouIntel:
> @adstraw @YangleiZouIntel If you do use bool, please make sure to intention comments on callers -- https://llvm.org/docs/CodingStandards.html#comment-formatting

Thanks @adstraw and @earhart. I checked the backend tests, it turned out we need at least 4 different buffer copy modes to get minimum number of buffer copies and achieve good inference latency. So I changed back to use uint32_t. Most backend tests are fixed now, still checking the last a few (<10) tests (which turned from pass to fail after this patch).

pmlc/rt/vulkan/vulkan_invocation.cc
YangleiZouIntel: Thanks @jbruestle, the user buffers and the issues above are resolved.

YangleiZouIntel:
Ready for review. Changing host buffer after shader submitted to device and before device returns is not allowed (generate warning)

pmlc/rt/vulkan/vulkan_invocation.h
```diff
@@ -22,6 +22,7 @@ namespace pmlc::rt::vulkan {
 
 using DescriptorSetIndex = uint32_t;
 using BindingIndex = uint32_t;
+using BufferCopyMode = uint32_t;
```
earhart, line 25: Use an enum here.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -71,6 +80,9 @@ static constexpr const char *kBindBufferInteger32 = "bindBufferInteger32";
 static constexpr const char *kBindBufferInteger64 = "bindBufferInteger64";
 
 static constexpr const int kByteBits = 8;
+static constexpr const int kBufferCopyModeInit = 0;
```
earhart, line 83: I think these should be an enum attribute (generated by tblgen), but @jbruestle and @flaub might have other thoughts.  (And this could be done later if we decide to do it; just add a TODO.)

earhart (commented):
Overall looks great!  Just a couple very minor comments.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -237,10 +262,86 @@ Value ConvertGpuLaunchFuncToVulkanCalls::createEntryPointNameConstant(
...
+      if (op->isBeforeInBlock(callOp.getOperation())) {
+        copyMode |= kBufferCopyModeHostToDevice;
+      } else if (deinitVulkan.getOperation()->isBeforeInBlock(op)) {
+        copyMode |= kBufferCopyModeDeviceToHost;
+      } else {
+        callOp.emitWarning("A host side buffer is used after copied to "
+                           "device and before device returns.");
+      }
+    } else {
+      copyMode |= (kBufferCopyModeHostToDevice | kBufferCopyModeDeviceToHost);
```
adstraw, line 333: You can `return copyMode` here as well.  Once we know there is a dependency outside the current block we are done.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -158,19 +175,18 @@ class ConvertGpuLaunchFuncToVulkanCalls
...
   LLVM::LLVMType llvmInt32Type;
   LLVM::LLVMType llvmInt64Type;
 
   mlir::Type mlirFloat32Type;
-  mlir::Type mlirIndexType;
 
   uint64_t numKernel = 0;
   uint64_t lauchFuncIndex = 0;
   mlir::Value vulkanRuntime;
+  LLVM::CallOp deinitVulkan;
```
adstraw, line 189: It seems like there may be a design issue here if we need an (uninitialized) member variable to communicate between two member methods --- `convertGpuLaunchFunc` which sets `deinitVulkan` and `getBufferCopyMode` which consumes it.  I see that both methods are private and that ordering of calls to these methods is preserved by `runOnOperation` so it's not a bug, technically.  Again, just wondering if the need for this sort of thing indicates some design issue.  Could we reorganize code, somehow, to avoid this problem?

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -237,10 +262,86 @@ Value ConvertGpuLaunchFuncToVulkanCalls::createEntryPointNameConstant(
...
+    return copyMode;
+  }
+
+  auto operationDeps = getExternalDependentOperations(buffer);
+  auto currentBlock = callOp.getOperation()->getBlock();
+  for (auto op : operationDeps) {
+    if (op->getBlock() == currentBlock) {
+      if (op->isBeforeInBlock(callOp.getOperation())) {
+        copyMode |= kBufferCopyModeHostToDevice;
+      } else if (deinitVulkan.getOperation()->isBeforeInBlock(op)) {
```
adstraw, line 326: I had some trouble reading this code given that the state we are conditioning on changes from line 324 (if dependency is before the *call op*) and line 326 (if *de initialization* is before the dependency).

I think it would read quite a lot more readable if you did this:
```
if dependency is before the call op then 
  set the copy-host-to-device bit
else // dependency is after the call op
  if de initialization is before the dependency then
    set the copy device-to-host bit
  else
    warning
```

I am wondering about this warning, though.  Under what circumstances would we hit this case?  Does it have to do with whether or not we are printing the buffers?  And, if it is just a warning is it worth checking?  Goes back to my design question from earlier --- maybe makes sense just to drop this check.  In which case you have...

```
if dependency is before the call op then 
  set the copy-host-to-device bit
else // dependency is after the call op
  set the copy device-to-host bit
```

And, no need to track the deinit across two member functions.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -237,10 +262,86 @@ Value ConvertGpuLaunchFuncToVulkanCalls::createEntryPointNameConstant(
                                   shaderName, LLVM::Linkage::Internal);
 }
 
+llvm::SmallSet<Operation *, 4>
+ConvertGpuLaunchFuncToVulkanCalls::getExternalDependentOperations(
```
adstraw, line 266: Just noticing that there is a lot of overlap between this function and the next function `isInternalOperation` and the analysis code in pmlc/dialect/pxa/analysis/uses.cc.  Wondering if we could / should find a more general place for this code so that it can be reused later.

adstraw (changes requested):

pmlc/rt/vulkan/vulkan_invocation.cc
```diff
@@ -22,6 +22,9 @@
 
 namespace pmlc::rt::vulkan {
 
+static constexpr const int kBufferCopyModeHostToDevice = 1 << 0;
+static constexpr const int kBufferCopyModeDeviceToHost = 1 << 1;
```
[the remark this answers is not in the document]
earhart: There's actually a tblgen thing to make it easy to generate enum attrs.  But IIRC there was some objection to defining new attributes; definitely check w/@flaub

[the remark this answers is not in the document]
flaub: No objections from my end. I think we should use tblgen for generating enums as much as we can, since it provides lots of nice extra features (stringifying/symbolizing/documentation generation, etc)

pmlc/conversion/gpu/gpu_to_vulkan.cc
YangleiZouIntel: Fixed

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -237,10 +262,86 @@ Value ConvertGpuLaunchFuncToVulkanCalls::createEntryPointNameConstant(
...
+                                                     Value &buffer) {
+  uint32_t copyMode = kBufferCopyModeInit;
+  if (buffer.isa<BlockArgument>()) {
+    copyMode |= (kBufferCopyModeHostToDevice | kBufferCopyModeDeviceToHost);
+    return copyMode;
+  }
+
+  auto operationDeps = getExternalDependentOperations(buffer);
+  auto currentBlock = callOp.getOperation()->getBlock();
+  for (auto op : operationDeps) {
```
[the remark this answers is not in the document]
YangleiZouIntel: Fixed

pmlc/rt/vulkan/vulkan_invocation.cc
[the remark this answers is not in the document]
YangleiZouIntel: Fixed. Thanks all.

pmlc/conversion/gpu/gpu_to_vulkan.cc
YangleiZouIntel: Fixed and replaced "deinitVulkan" by "run", which is clearer.

pmlc/rt/vulkan/vulkan_invocation.h
YangleiZouIntel: Fixed

pmlc/conversion/gpu/gpu_to_vulkan.cc
YangleiZouIntel: Because currently we are doing calculations on both host and device side, thus some backend tests run into a situation where the host buffers are used after submit to vulkan and before returns, such as in testStd,
 
Below is a section of the IR for testStd
```
 func @get_value(%arg0: memref<3x4xf32> {tile.name = "anon"}, %arg1: memref<f32>) {
    %cst = constant 0.000000e+00 : f32
    %c12_i32 = constant 12 : i32
    %c0 = constant 0 : index
    %c3 = constant 3 : index
    %c4 = constant 4 : index
    %c1 = constant 1 : index
    %0 = alloc() : memref<1x1xf32>
    store %cst, %0[%c0, %c0] : memref<1x1xf32>
    "gpu.launch_func"(%c1, %c1, %c1, %c1, %c1, %c1, %arg0, %0) {kernel = @get_value_kernel::@get_value_kernel} : (index, index, index, index, index, index, memref<3x4xf32>, memref<1x1xf32>) -> ()
    %1 = alloc() : memref<1x1xf32>
    %2 = load %0[%c0, %c0] : memref<1x1xf32>
    %3 = sitofp %c12_i32 : i32 to f32
    %4 = divf %2, %3 : f32
    store %4, %1[%c0, %c0] : memref<1x1xf32>
    %5 = alloc() : memref<3x4xf32>
    "gpu.launch_func"(%c1, %c1, %c1, %c1, %c3, %c4, %arg0, %1, %5) {kernel = @get_value_kernel_0::@get_value_kernel} : (index, index, index, index, index, index, memref<3x4xf32>, memref<1x1xf32>, memref<3x4xf32>) -> ()
    %6 = alloc() : memref<3x4xf32>
    "gpu.launch_func"(%c1, %c1, %c1, %c1, %c3, %c4, %5, %6) {kernel = @get_value_kernel_1::@get_value_kernel} : (index, index, index, index, index, index, memref<3x4xf32>, memref<3x4xf32>) -> ()
    %7 = alloc() : memref<f32>
    store %cst, %7[] : memref<f32>
    "gpu.launch_func"(%c1, %c1, %c1, %c1, %c1, %c1, %6, %7) {kernel = @get_value_kernel_2::@get_value_kernel} : (index, index, index, index, index, index, memref<3x4xf32>, memref<f32>) -> ()
    %8 = alloc() : memref<f32>
    %9 = load %7[] : memref<f32>
    %10 = divf %9, %3 : f32
    store %10, %8[] : memref<f32>
    %11 = load %8[] : memref<f32>
    %12 = sqrt %11 : f32
    store %12, %arg1[] : memref<f32>
    return
  }
```
the code section below between the first and second gpu.launch_func is changing host side buffers after submit to vulkan, so it will end up with an incorrect result. But if the operation is stx.reshape, the result may still be correct.
```
    %1 = alloc() : memref<1x1xf32>
    %2 = load %0[%c0, %c0] : memref<1x1xf32>
    %3 = sitofp %c12_i32 : i32 to f32
    %4 = divf %2, %3 : f32
    store %4, %1[%c0, %c0] : memref<1x1xf32>
    %5 = alloc() : memref<3x4xf32>
```

I have detected a list of backend tests that currently has this issue, as follows,
FAIL: testBranchWithSoftmax (__main__.TestBackendOps) . 
FAIL: testRandomNormalDev (__main__.TestBackendOps) 
FAIL: testRandomUniformDev (__main__.TestBackendOps)  
FAIL: testRandomUniformVariableDev (__main__.TestBackendOps)  
FAIL: testStd (__main__.TestBackendOps) 
FAIL: testTruncatedNormalDev (__main__.TestBackendOps)  
FAIL: testVarSimple (__main__.TestBackendOps)  

Do you think put this message in IVLOG is better than warning?

To solve this issue we can 
1. Put everything between gpu.launch_func into a gpu module so they will be send to device
2. Rearrange the algorithm/pipeline passes to avoid changing host side buffers after copied to device
3. Make several vulkan command buffers and run the gpu.launch_funcs(and copy buffers back) before these host side ops.

pmlc/conversion/gpu/gpu_to_vulkan.cc
YangleiZouIntel: I think right now for debugging purposes, we have to check all dependencies to see if some of them changed host side buffers after copied to device. 
If we fixed this issue or we do not want to check here, we can early return copyMode.

pmlc/conversion/gpu/gpu_to_vulkan.cc
adstraw: Addressing the larger design / testing issue in follow-on PRs works for me.  I see that you rewrote and commented the conditional logic in `getBufferCopyMode` and that looks good, thanks.

adstraw:
Waiting for CI to pass before I do one final review but overall this looks good to me.  I would like if @earhart also reviewed and gave a 👍 to this PR when he's ready.

pmlc/conversion/gpu/enums.td
```diff
@@ -0,0 +1,21 @@
+#ifndef __PML_VULKAN_ENUMS__
+#define __PML_VULKAN_ENUMS__
+
+#ifndef OP_BASE
+include "mlir/IR/OpBase.td"
+#endif
+
+def BUFFER_COPY_MODE : BitEnumAttr<
```
flaub, line 8: Use `CapitalCasing` instead of all caps.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -109,14 +121,20 @@ class ConvertGpuLaunchFuncToVulkanCalls
   /// Declares all needed runtime functions.
   void declareVulkanFunctions(Location loc);
 
+  /// Check buffer type and only copy from/to device when necessary
+  BUFFER_COPY_MODE getBufferCopyMode(mlir::CallOp &callOp, Value &buffer);
```
flaub, line 125: Don't pass `Value` can `CallOp` by reference, always use by-value semantics.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -109,14 +121,20 @@ class ConvertGpuLaunchFuncToVulkanCalls
...
+
+  bool isInternalOperation(Operation *op);
+
+  llvm::SmallSet<Operation *, 4> getExternalDependentOperations(Value &value);
```
flaub, line 129: Pass `Value` by-value

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -93,8 +106,7 @@ class ConvertGpuLaunchFuncToVulkanCalls
                                      Location loc, OpBuilder &builder);
 
   /// bind gpu.launchOp buffers to Vulkan runtime.
-  LogicalResult bindBuffers(Location loc, OpBuilder &builder,
-                            gpu::LaunchFuncOp launchOp);
+  LogicalResult bindBuffers(mlir::CallOp &callOp);
```
flaub, line 109: Pass `CallOp` by value.

flaub:
I feel like this entire functionality should be done at the `comp` dialect level.

adstraw:
Changes LGTM 👍.  I have reset my review status so it no longer says "change requested".  Will leave it to either @flaub or @earhart to provide a second 👍 so this PR can land.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -181,7 +197,7 @@ class ConvertGpuLaunchFuncToVulkanCalls
   };
 
   llvm::SmallSet<mlir::Type, 4, mlirTypeComparator> bufferElementTypes;
-  llvm::SmallSet<const char *, 4> optionalSymbols;
+  llvm::SmallSet<StringRef, 4> optionalSymbols;
```
earhart, line 200: As long as you're changing this -- maybe use `llvm::StringSet`?  (It's not completely clear to me what the ownership rules are, here.)

pmlc/rt/vulkan/wrappers.cc
```diff
@@ -83,9 +84,10 @@ void run(void *vkInvocation) {
 #define BIND_BUFFER_IMPL(_name_, _type_)                                       \
   void _mlir_ciface_bindBuffer##_name_(                                        \
       void *vkInvocation, DescriptorSetIndex setIndex, BindingIndex bindIndex, \
-      uint32_t bufferByteSize, ::UnrankedMemRefType<_type_> *unrankedMemRef) { \
+      uint32_t bufferByteSize, BUFFER_COPY_MODE bufferCopyMode,                \
```
earhart, line 87: Please make bufferCopyMode an int32_t (matching the call op), and then cast it before passing it to bindBuffer, so that the ABI is declared the same way on both the caller and callee.

earhart (commented):
Overall looks good; just a few minor additional comments.

YangleiZouIntel:
> I feel like this entire functionality should be done at the `comp` dialect level.

That makes a lot of sense. I am thinking about replace this pass with comp_to_vulkan pass.

pmlc/rt/vulkan/wrappers.cc
YangleiZouIntel: Fixed, please check.

pmlc/conversion/gpu/gpu_to_vulkan.cc
YangleiZouIntel: Fixed

pmlc/conversion/gpu/enums.td
YangleiZouIntel: Fixed

pmlc/conversion/gpu/gpu_to_vulkan.cc
YangleiZouIntel: Fixed

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -237,10 +263,115 @@ Value ConvertGpuLaunchFuncToVulkanCalls::createEntryPointNameConstant(
...
+  if (buffer.isa<BlockArgument>()) {
+    copyMode = copyMode |
+               (BufferCopyMode::HostToDevice | BufferCopyMode::DeviceToHost);
+    return copyMode;
+  }
+
+  auto currentBlock = callOp.getOperation()->getBlock();
+
+  // Get llvm.call @run in current block
+  LLVM::CallOp *pRunOp = nullptr;
```
flaub, line 325: Don't create pointers to `CallOp`, instead use it by-value.

Like:
```
LLVM::CallOp runOp = nullptr;
```

And later like:
```
runOp = llvmCallOp;
```

flaub: Even better to extract this to a new function like:

```
Optional<LLVM::CallOp> findCallOp(llvm::StringRef name) {
}
```

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -314,16 +448,16 @@ LogicalResult ConvertGpuLaunchFuncToVulkanCalls::transferBuffers(
...
 LogicalResult ConvertGpuLaunchFuncToVulkanCalls::printBuffer(Location loc,
                                                              OpBuilder &builder,
-                                                             Value &buffer) {
+                                                             Value buffer) {
   auto type = buffer.getType();
   if (auto memRefType = type.dyn_cast_or_null<MemRefType>()) {
     auto elementType = memRefType.getElementType();
     if (elementType.isF32()) {
-      auto unrankedBuffer = builder.create<mlir::MemRefCastOp>(
+      auto unrankedBuffer = builder.create<MemRefCastOp>(
```
flaub, line 456: We should avoid using `MemRefCastOp`, the internal lowering of a MemRefCast is to use `alloca` memory (when lowered to LLVM). If this happens inside of a loop, it's easy for the stack to be exhausted.

pmlc/conversion/gpu/gpu_to_vulkan.cc
YangleiZouIntel: Fixed, added a findLLVMCallOp, since here only checks LLVM::CallOp

pmlc/conversion/gpu/gpu_to_vulkan.cc
YangleiZouIntel: I see, should we add a custom conversion pattern for MemRefCastOp to llvm and avoid use `alloca` memory?

YangleiZouIntel: I see, this is a very good point, I will think more about it.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -237,10 +267,122 @@ Value ConvertGpuLaunchFuncToVulkanCalls::createEntryPointNameConstant(
...
+llvm::Optional<LLVM::CallOp>
+ConvertGpuLaunchFuncToVulkanCalls::findLLVMCallOp(Block *block,
+                                                  llvm::StringRef name) {
+  LLVM::CallOp op = nullptr;
+  auto llvmCallOps = block->getOps<LLVM::CallOp>();
+  for (auto llvmCallOp : llvmCallOps) {
+    if (llvmCallOp.callee().getValue().equals(name)) {
+      if (op) {
+        llvmCallOp.emitError("Does not support multiple llvm.call @run.");
+        signalPassFailure();
```
flaub, line 324: Need to return `llvm::None` here.

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -237,10 +267,122 @@ Value ConvertGpuLaunchFuncToVulkanCalls::createEntryPointNameConstant(
...
+      }
+      op = llvmCallOp;
```
flaub, line 326: Are you looking for the latest callOp or can you return early here?

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -237,10 +267,122 @@ Value ConvertGpuLaunchFuncToVulkanCalls::createEntryPointNameConstant(
...
+  if (buffer.isa<BlockArgument>()) {
+    copyMode = copyMode |
+               (BufferCopyMode::HostToDevice | BufferCopyMode::DeviceToHost);
+    return copyMode;
+  }
+
+  auto currentBlock = callOp.getOperation()->getBlock();
+
+  // Get llvm.call @run in current block
+  auto runOp = findLLVMCallOp(currentBlock, kRun);
```
flaub, line 346: Need to check the result here.

flaub:
Is this PR still valid?

YangleiZouIntel:
> Is this PR still valid?

No, this changes will be moved to comp dialect level. I will close this PR.

PR #1358, opened 2020-09-15
Refactor device/executable plumbing

earhart:


plaidml/BUILD
```diff
@@ -44,7 +44,7 @@ plaidml_cc_library(
     name = "runtimes",
     visibility = ["//visibility:public"],
     deps = [
-        "//pmlc/rt/llvm",
+        "//pmlc/rt/llvm_cpu",
```
flaub, line 47: This doesn't seem like a good name change to me. Why is this runtime CPU specific?

flaub:
Not sure this PR has enough meat on its bones, I don't know where it's headed so hard for me to decide if this is a good move or not.

plaidml/BUILD
earhart: We want namespaces to follow the directory hierarchy, and `llvm::` is already used a few times in this code.

earhart:
NP -- I'll mark it as WIP and add a few more commits as they become ready.

plaidml/BUILD
flaub: Please use `pmlc::rt:llvm`, to disambiguate, use `::llvm`.

pmlc/rt/device.h
```diff
@@ -0,0 +1,47 @@
...
+  // Returns the current Device, local to the current thread.
+  //
+  // If there is no current device, or the current device is of an unexpected
+  // type, an exception will be thrown.
+  //
+  // This is used to smuggle the current device through to runtimes that are
+  // directly invoked by lowered code, when that lowered code doesn't support
+  // the use of a passthrough parameter.
+  template <typename T>
+  static std::shared_ptr<T> current() {
```
flaub, line 25: I thought we were going to pass the device pointer as a void* into the JIT'd code. If so, why do we need this?

pmlc/rt/device.h
```diff
@@ -0,0 +1,47 @@
...
+    if (!result) {
+      throw std::runtime_error{"Incompatible target/device combination"};
+    }
+    return result;
+  }
+
+  virtual ~Device() = default;
+
+  virtual std::unique_ptr<Executable>
+  compile(const std::shared_ptr<pmlc::compiler::Program> &program,
```
flaub, line 40: It seems like only a `Program*` is needed here? Why would an `Executable` need to hold onto a `program`?

pmlc/rt/device.h
earhart: (Actually -- not the device pointer, but an invocation pointer.)

pmlc/rt/device.h
earhart: This is just moving around & refactoring existing code... but I could go either way on this -- the counterargument is that since we're passing around Program with shared_ptr in other places, we should avoid passing it via a raw pointer, since it's entirely too easy to build a shared_ptr out of a raw pointer and get a double-delete.  (When do we switch to Rust, again?  :-)

plaidml/BUILD
earhart: Done.  If you have the time, I'd like to better understand your rationale.  Note that this affects all of the headers in `pmlc/rt`, which now need to explicitly use `::llvm` everywhere, since they might be included in a context in which `pmlc::rt::llvm` exists.  (C++ namespace resolution being what it is, I find it useful to try to keep my sub-namespace names globally unique where possible.)

flaub: I have a few intuitions:
* Maybe we should do away with nested namespaces, perhaps `pmlc` is sufficient (`mlir` fine for MLIR so far).
* Maybe we could call the namespace `llvm_rt`? It doesn't seem CPU-specific to me, but maybe I'm misunderstanding the purpose of this library.


Separately, it'd be really nice if we could somehow retain something in the spirit of the old `rt` library, which contained only runtime functions that were linked into the resulting binary. It's possible that we might need to dynamically link in the runtime functions, and so having a library dedicated to these functions only seems good. I feel like we're starting to mix in the concept of the runtime itself from the original `rt` of just functions that are linked at runtime. I can see how it gets blurry with things like vulkan.

pmlc/rt/device.h
flaub: I guess what's needed here is a description of the intended lifetime and usage policy for these objects. My thinking was that the user would be in control over when a program, device, executable, or runtime would be destroyed. But perhaps it's better to use shared_ptr everywhere so that users don't have to be as careful (and we don't have to work out every single workflow). Overall it's cheap, so ya, let's stick with shared_ptr.

bzl/bazel.rc
```diff
@@ -32,11 +32,13 @@ test --test_env=HOME
 test --test_env=PLAIDML_DEVICE
 test --test_env=PLAIDML_TARGET
 test --test_env=PLAIDML_SETTINGS
+test --test_env=PLAIDML_VERBOSE
 test --test_sharding_strategy=disabled
 run --test_env=HOME
 run --test_env=PLAIDML_DEVICE
 run --test_env=PLAIDML_TARGET
 run --test_env=PLAIDML_SETTINGS
+run --test_env=PLAIDML_VERBOSE
```
flaub, line 41: Why do we need this? I've been using PLAIDML_VERBOSE without a need to add this to the .bazelrc for a long time now.

pmlc/dialect/tile/builder.cc
```diff
@@ -612,6 +612,7 @@ TileBuilder::MakeProgram(StringRef name, const ProgramMutations &mutations,
   // Construct a function to represent the entire program
   auto initialFuncType = FunctionType::get(inputTypes, {}, getContext());
   auto funcOp = FuncOp::create(loc, name, initialFuncType, {});
+  funcOp.setAttr("entrypoint", mlir::UnitAttr::get(getContext()));
```
flaub, line 615: Actually, there's an `entry` member on the `Program` which was meant to contain this information. If we don't need that anymore, we should remove it from `Program`.

flaub:
Could we make a separate PR for moving/reorganizing the directory structure? It's hard to review the diff to see what has changed in `jit_executable.cc` for instance.

pmlc/rt/runtime_registry.h
```diff
@@ -47,7 +47,7 @@ using Factory = std::function<std::shared_ptr<Runtime>()>;
 // N.B. This function is NOT synchronized.  It is the caller's responsiblity to
 // ensure that other components are not concurrently accessing the system global
 // runtime map or registering other Factory instances.
-void registerFactory(llvm::StringRef id, Factory factory);
+void registerFactory(::llvm::StringRef id, Factory factory);
```
flaub, line 50: Actually you could use `mlir::StringRef` here

plaidml/BUILD
earhart: Interesting idea.  So basically: the output of target compilation is essentially native code, and we can either JIT it or write it to a platform-specific image format and dynamically load it, and then it'll just do the right thing?

Assuming I've got it right -- I think we could get there.  LMK if I've got the idea; in the meantime, I'll mark this as a draft again in case I want to change the approach a bit before landing it.

bzl/bazel.rc
earhart: Interesting; I'd expect bazel to have cleaned it from your test environment.  IIRC it doesn't work without something like this, but I can remove it and add it to my .bazelrc.user (assuming it's needed) too.

pmlc/dialect/tile/builder.cc
earhart: > Could we make a separate PR for moving/reorganizing the directory structure? It's hard to review the diff to see what has changed in `jit_executable.cc` for instance.

I could use some more guidance on when a PR has "too little meat on its bones" vs. "hard to review".  :-)   In the meantime, could you look at the individual commits?

bzl/bazel.rc
flaub: It's not needed AFAICT, was it not working for you in some context? I'd like to understand what prompted you to add this.

flaub: I think we have the sandbox disabled BTW

pmlc/dialect/tile/builder.cc
flaub: The issue is that when I look at `jit_executable.cc`, it all looks like new code, but it's actually just a renamed file. It'd be good if we could make a NFC commit for renames/reorgs so that I can see the delta to the executable code. Like temporarily define the contents of `jit_exectuable.cc` in its original location (which now I'm not sure where it was, it used to be in `compiler/executable.cc`.

pmlc/rt/jit_executable.h
```diff
@@ -15,6 +15,7 @@ namespace pmlc::rt {
 std::unique_ptr<Executable>
 makeJitExecutable(const std::shared_ptr<pmlc::compiler::Program> &program,
                   std::shared_ptr<Device> device,
+                  llvm::ArrayRef<void *> preParams,
```
flaub: I would have thought that the `Device` we are passing in here would be sufficient? Can we pass the `device*` to the jitted program?

plaidml/BUILD
flaub: It'd be nice is my way of saying, let's do this in post, no need to draft it, let's get something landed :)

bzl/bazel.rc
earhart: I actually added it in my local codebase ages ago, for some reason; I've just been carrying it forward.  I'll try removing it along with all the other environment passthroughs.

flaub: AFAIK, no changes are needed to the `.bazelrc`, or at least no changes that are related to this PR :)

pmlc/rt/jit_executable.h
earhart: The thinking was that the JIT logic should not know or care about any extra parameters that're needed by a target.  But I'm no longer convinced this is necessary at all, if we're going to a more opaque (from the PoV of the caller) model.

plaidml/BUILD
earhart: The thing is, if we're going in that direction, I want to think more about this entire chunk of code before landing it; there're a few things that I think I might want to undo.  :-)

pmlc/rt/runtime_registry.h
earhart: Turns out that doesn't work for `::llvm::StringMap`, but otherwise, it's great - thanks!  Done.

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -386,7 +386,7 @@ TEST_F(CppEdsl, DoubleDot) {
   // clang-format off
   // CHECK-LABEL: CppEdsl.DoubleDot
   // CHECK: func @double_dot
-  // CHECK: tensor<10x20xf32>) -> tensor<10x40xf32> {
+  // CHECK: tensor<10x20xf32>) -> tensor<10x40xf32> attributes {entrypoint} {
```
flaub, line 389: Can you explain why we need a new attribute? There should already be an entry member on the `Program` that we can use to identify and lookup the entrypoint.

pmlc/rt/timed_executable.cc
```diff
@@ -0,0 +1,62 @@
...
+#include <utility>
+
+#include "pmlc/util/logging.h"
+
+namespace pmlc::rt {
+namespace {
+
+// A trivial stopwatch class.
+//
+// TODO: Consider using std::chrono::high_resolution_clock.  Also consider
```
flaub, line 15: Steady clock is better for this sort of thing.

pmlc/rt/timed_executable.h
```diff
@@ -0,0 +1,16 @@
...
+#include <memory>
+
+#include "pmlc/rt/executable.h"
+
+namespace pmlc::rt {
+
+// Returns an Executable implementation that times how long it takes to perform
+// an inference.
+std::unique_ptr<Executable>
+makeTimedExecutable(std::unique_ptr<Executable> base);
```
flaub, line 14: I  think we should build profiling into every Executable, and have it be reported back via the Runtime interface. Does anything actually use this?

plaidml/edsl/tests/edsl_test.cc
earhart: During lowering, I needed to figure out the name of the function to attach the `Device` parameter to.  OTOH, I didn't try too hard to find an alternative mechanism; I think it's better to have this information in the MLIR than in a side channel.

pmlc/rt/timed_executable.cc
earhart: IIRC, on Windows, they're the same.  :-)

pmlc/rt/timed_executable.h
earhart: I have no idea.  This is a refactoring of existing code to preserve existing functionality, not adding anything.

I'm pretty sure this logic doesn't belong in the JIT executable; it should be applied to all sorts of executables.  So I think a decorator is a fine place for it.

The broader question of building profiling into every Executable is interesting, and more complicated than I want to address in this PR.  It depends a bit on what the purpose is; `steady_clock` (and `high_resolution_clock`) is pretty good for end-user perceived latency, which is certainly something we care about; we also might want more per-device statistics to guide development, and that might require running the network (or slices of the network) multiple times via a benchmarking harness (e.g. on x64 CPU, you can count all kinds of interesting events, but you only get a few performance counters to use at a time; PEBS info is even more useful, but also requires multiple runs to get out everything you might care about).

I think we could use a design spec on this.  But in the meantime, I'm happy to stick with preserving the existing code.

pmlc/rt/timed_executable.cc
flaub: Can we drop this TODO then? I don't think there's anything else to consider.

pmlc/rt/timed_executable.h
flaub: Could we make it part of normal jit executable creation for now? I'm depending on this behavior for some of my testing. Eventually, we should roll profiling up into the Runtime API.

pmlc/rt/vulkan/tests/addf.mlir
```diff
@@ -20,7 +20,7 @@ module attributes {
     }
   }
 
-  func @main() {
+  func @main() attributes {entrypoint} {
```
flaub, line 23: We need to resolve this design, which is right now we have two sources of information for the entrypoint. We should decide on having one.

flaub:
I think we should defer adding OpenCL to the mix until a follow up PR, WDYT?

pmlc/rt/timed_executable.cc
earhart: Done

pmlc/rt/timed_executable.h
earhart: I'm not sure what you mean -- what changes if we move it into jit_executable?

earhart:
> I think we should defer adding OpenCL to the mix until a follow up PR, WDYT?

What do you mean by "adding OpenCL to the mix"?  (I just wanted to get the code compiling.  :-)

pmlc/rt/vulkan/tests/addf.mlir
earhart: +1 -- can we do it in a followup PR, or do you want it resolved in this one?

pmlc/dialect/comp/ir/types.h
```diff
@@ -63,6 +63,47 @@ class RuntimeType : public mlir::Type {
...
+// ============================================================================
+// DeviceType;
+// ============================================================================
+class ExecEnvType;
+
+struct DeviceStorage : public RuntimeTypeStorage {
+  explicit DeviceStorage(ExecEnvRuntime runtime)
+      : RuntimeTypeStorage(runtime) {}
+
+  using KeyTy = ExecEnvRuntime;
```
flaub, line 75: I don't understand this. Are we saying that the Device and the ExecEnvRuntime are 1-to-1?

pmlc/conversion/gpu_to_comp/gpu_to_comp.cc
```diff
@@ -80,8 +80,32 @@ mlir::LogicalResult
...
+  // Look up the containing function.
+  auto func = op.getParentOfType<mlir::FuncOp>();
+  if (!func) {
+    return mlir::failure();
+  }
+  // Add the execution device parameter to the containing function, unless it
+  // already has one.
+  auto funcTy = func.getType();
+  mlir::Value device;
+  if (funcTy.getNumInputs() && funcTy.getInput(0).isa<comp::DeviceType>()) {
```
flaub, line 92: I don't like the idea that we are saying the the first argument to the function has to be a device. Let's chat about different design choices here.

pmlc/rt/timed_executable.h
flaub: I mean, let's just have every run thru the Executable measure the time, instead of having a separate TimedExecutable.

earhart: That's what this enables...?  We talked about having Executable implementations that aren't JitExecutable, e.g. something interpreted; using a decorator lets us compose this functionality with every Executable.

flaub: I'm trying to say:

Let's move the stopwatch into the Executable itself, where it was originally. And also, let's not have a separate `makeTimedExecutuable` since it doesn't look like anythign is calling it. Instead, I'm saying, let's have every default Executable do the timings.

pmlc/conversion/gpu_to_comp/gpu_to_comp.cc
```diff
@@ -80,8 +80,32 @@ mlir::LogicalResult
...
+    device = func.getArgument(0);
+  } else {
+    auto deviceTy =
+        rewriter.getType<comp::DeviceType>(execEnvType.getRuntime());
+    std::vector<mlir::Type> inputs{deviceTy};
+    inputs.insert(inputs.end(), funcTy.getInputs().begin(),
+                  funcTy.getInputs().end());
+    func.setType(
```
flaub, line 100: This is actually illegal

earhart: Fixed; thanks!

pmlc/dialect/comp/ir/types.h
earhart: As discussed -- removed runtime from device type until/unless we actually need it.

earhart:
@flaub -- assuming tests pass, I think this is ready for re-review.

flaub (approved):
LGTM, thanks!

PR #1359, opened 2020-09-15
WIP: Adding PlaidML Linear Algebra Solvers

jd-bartlett96:
Using PlaidML v0 to build Linear Algebra Solvers
(Did not build from source; runs with the v0 wheel on BuildKite)

`conjgrad.py` performs full solution using conjugate gradient method. This version does not take sparsity into account in its computation (currently working on a version that does). An iterative solve method.

`cholesky.py` performs a Cholesky decomposition (likewise, on a dense matrix), which decomposes a matrix into the product of two triangular matrices (allowing a system to be solved with triangular solvers). The "hard" step of a direct solve method.

Both scripts also have a corresponding script with a numpy implementation of the same operations.

Currently, both produce issues when run on the CPU but run fine on nvidia GPUs (a previous version of the conjgrad script used to work on the CPU; I'm investigating this further now).


jd-bartlett96:
Added actual FEM test cases for more useful validation. 

Conjugate gradient method quickly outpaces numpy black-box solver on the gpu (still assuming dense matrices). To try it out run with
`python conjgrad.py 10x10`
(where '10x10` can be replaced with any of the other test names in `test_data`)

However I'm running into an odd issue when I try to run on the cpu (which I'd like to do to get a idea of serial performance); it doesn't seem to be compiling correctly. I can provide more detail but the gist is that in this section:
https://github.com/plaidml/plaidml/blob/a91ad00e99a54c74abcfe186328d6681d4dabf9c/networks/scitile/LAS/conjgrad.py#L38-L42
for some reason the line 
`OP = OR + RSQN/RSQ * P`
is writing new data into the `OX` Tensor. If I even just remove the `OR` term from this expression everything works as expected. @jbruestle any ideas what might be causing that?

flaub:
What's the plan for this PR? I'd like to close it if it's no longer going to receive any more work or maintenance.

jd-bartlett96:
As these examples fall into a similar category as those in #1212 (interesting examples that someone may want to look at in the future but aren't currently being worked on/maintained), it would make sense to me to do with this whatever we do with that PR. Whether that's closing the branch and restoring later if needed, porting these to cmake & landing, etc. is up to you.

tzerrell:
I think if this is currently correct & tested, or could be made correct & tested with brief work, we should review & land it. If it is broken or missing key tests that would let us tell if it's broken, and cannot be quickly fixed, we should close it. This applies both here and to #1416, and it would not surprise me if this question has different answers for the 2 PRs. Feel free to reach out to me directly if you want my help assessing that.

jd-bartlett96:
Deciding to close this and land #1416 as some work would be needed to get ci passing on this and it is fairly redundant to #1416, which should be able to be landed with minimal effort.

PR #1361, opened 2020-09-15
WIP: add subgroup size control

adstraw:
Thanks to @abialas1 for providing the source code for this change:
https://github.com/plaidml/plaidml/compare/plaidml-v1...abialas-vulkan-subgroup-size-control

### DIFF
--- a/pmlc/conversion/gpu/gpu_to_vulkan.cc
+++ b/pmlc/conversion/gpu/gpu_to_vulkan.cc
@@ -348,7 +348,8 @@ void ConvertGpuLaunchFuncToVulkanCalls::declareVulkanFunctions(Location loc) {
 
   builder.create<LLVM::LLVMFuncOp>(
       loc, kSetVulkanLaunchKernelAction,
-      LLVM::LLVMType::getFunctionTy(getLLVMVoidType(), {getLLVMPointerType()},
+      LLVM::LLVMType::getFunctionTy(getLLVMVoidType(),
+                                    {getLLVMPointerType(), getLLVMInt32Type()},
                                     /*isVarArg=*/false));
 
   builder.create<LLVM::LLVMFuncOp>(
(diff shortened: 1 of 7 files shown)


pmlc/rt/vulkan/vulkan_invocation.cc
```diff
@@ -565,13 +565,22 @@ void VulkanInvocation::createPipelineLayout() {
...
 }
 
-void VulkanInvocation::createComputePipeline() {
+void VulkanInvocation::createComputePipeline(uint32_t subgroupSize) {
   if (!curr) {
     throw std::runtime_error{"createComputePipeline: curr is nullptr!"};
   }
 
+  VkPipelineShaderStageRequiredSubgroupSizeCreateInfoEXT subgroupSizeInfo;
+  subgroupSizeInfo.sType = VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_CREATE_INFO;
```
adstraw, line 574: Changed intentionally from VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SUBGROUP_PROPERTIES to VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_CREATE_INFO based on debug results from @jbruestle

pmlc/conversion/gpu/gpu_to_vulkan.cc
```diff
@@ -448,11 +449,16 @@ void ConvertGpuLaunchFuncToVulkanCalls::convertGpuLaunchFunc(
     return signalPassFailure();
   }
 
+  // TODO: select subgroup size based on the previous passes and some special
```
jbruestle, line 452: So technically we cannot set subgroupSize for non-subgroup kernels to 8 if the number of threads is not divisible by 8.  I think we need to grab the attribute on the kernel launch and pass that down.  non-subgrouped kernels should either have no attribute or have an attribute set to 1 depending on lowering stage.

pmlc/rt/vulkan/vulkan_invocation.cc
```diff
@@ -565,13 +565,22 @@ void VulkanInvocation::createPipelineLayout() {
...
+  subgroupSizeInfo.requiredSubgroupSize = subgroupSize;
+  subgroupSizeInfo.pNext = NULL;
+
   VkPipelineShaderStageCreateInfo stageInfo = {};
   stageInfo.sType = VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_CREATE_INFO;
+  if (device->isExtensionSupported("VK_EXT_subgroup_size_control"))
```
jbruestle, line 580: Similar to above, we probably need to skip specifying subgroup size if the value is 1 (or the extension is not supported).

jbruestle (changes requested):

abialas1:
Checked in some lowerings, however still missing the SCF->GPU part that is within llvm upstream. Working on the best way to have it added there.

PR #1362, opened 2020-09-16
implement tile.scatter

mars1026:
Lowering to affine loop

### DIFF
--- a/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
+++ b/pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
@@ -45,6 +45,8 @@ using dialect::tile::PaddingInfo;
 using dialect::tile::PrngOp;
 using dialect::tile::ReshapeOp;
 using dialect::tile::ReshapeOpAdaptor;
+using dialect::tile::ScatterOp;
+using dialect::tile::ScatterOpAdaptor;
 using dialect::tile::ShapeOp;
 using dialect::tile::ShapeOpAdaptor;
 using dialect::tile::TraceOp;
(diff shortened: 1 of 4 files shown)


jbruestle (approved):

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -1070,6 +1072,92 @@ struct ShapeOpConversion : public OpConversionPattern<ShapeOp> {
...
+    auto loc = op.getLoc();
+    auto ctx = rewriter.getContext();
+    TypeConverter typeConverter;
+
+    // Create an adaptor, to interpret the operands
+    ScatterOpAdaptor adaptor(operands);
+    // 'tensor' provides update values
+    // 'dims' contains the destination indices
+    // 'other' is the shape of the output
+    // this is redundant because the result type also specifies output shape
```
flaub, line 1094: If `other` is redundant, can we remove it as an operand for this op?

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -1070,6 +1072,92 @@ struct ShapeOpConversion : public OpConversionPattern<ShapeOp> {
...
+    // Make an allocation for the output
+    auto resultType = typeConverter.convertType(op.result().getType());
+    auto resultMemRefType = resultType.cast<MemRefType>();
+    auto resultMemRef =
+        rewriter.create<AllocOp>(loc, resultMemRefType).getResult();
+
+    // Get the shape of the update tensor and create a parallel loop over its
+    // indexes; we will load each value from the updates, load its destination
+    // from the indexes, and store the value to the result.
+    auto updatesType = typeConverter.convertType(updates.getType());
```
flaub, line 1107: I would expect the type of the adaptor operands to already have been converted, so why do we need to convert it manually here?

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -1070,6 +1072,92 @@ struct ShapeOpConversion : public OpConversionPattern<ShapeOp> {
...
+    auto updatesMemRefType = updatesType.cast<MemRefType>();
+    ArrayRef<int64_t> updatesShape = updatesMemRefType.getShape();
+
+    auto loop = rewriter.create<AffineParallelOp>(
+        loc, ArrayRef<Type>{resultMemRefType},
+        ArrayRef<AtomicRMWKind>{AtomicRMWKind::assign}, updatesShape);
+    auto txBuilder = loop.getBodyBuilder();
```
flaub, line 1114: It's actually illegal to create ops or mutate anything at all outside of the `rewriter`. So to do this properly, we need to move the insertion point of the rewriter and use that instead of creating and using a new builder.

PR #1389, opened 2020-09-24
Fix some warnings that were causing errors building with the latest X…

LLITCHEV:
It appears the clang in the latest Xcode has better logic to detect copying caused by assigning a non reference return to a reference var.
    See below the exact fixes.
    Also, disabled the newly added opencl tests/build on MacOS and Windows - it was not building on MacOS, pulling linux header files.

    ERROR: /Users/llitchev/local/plaidml-v1/pmlc/dialect/pxa/transforms/BUILD:23:19: C++ compilation of rule '//pmlc/dialect/pxa/transforms:transforms' failed (Exit 1)
    pmlc/dialect/pxa/transforms/stencil.cc:145:22: error: loop variable 'blockArg' is always a copy because the range of type 'pmlc::dialect::pxa::BlockArgumentSet' (aka 'SmallPtrSet<mlir::BlockArgument, 8>') does not return a reference [-Werror,-Wrange-loop-analysis]
        for (const auto &blockArg : getBlockArgsAsSet()) {
                         ^
    pmlc/dialect/pxa/transforms/stencil.cc:145:10: note: use non-reference type 'mlir::BlockArgument'

    plaidml/op/lib/ops.cc:2908:19: error: loop variable 'dim' of type 'const plaidml::edsl::TensorDim' creates a copy from type 'const plaidml::edsl::TensorDim' [-Werror,-Wrange-loop-analysis]
      for (const auto dim : O_dims) {

    ERROR: /Users/llitchev/local/plaidml-v1/pmlc/dialect/tile/BUILD:7:19: C++ compilation of rule '//pmlc/dialect/tile:tile' failed (Exit 1)
    pmlc/dialect/tile/gradient.cc:87:22: error: loop variable 'operand' is always a copy because the range of type 'mlir::Operation::operand_range' (aka 'mlir::OperandRange') does not return a reference [-Werror,-Wrange-loop-analysis]
        for (const auto &operand : op->getOperands()) {
                         ^
    pmlc/dialect/tile/gradient.cc:87:10: note: use non-reference type 'mlir::Value'
        for (const auto &operand : op->getOperands()) {
             ^~~~~~~~~~~~~~~~~~~~~
    pmlc/dialect/tile/gradient.cc:134:22: error: loop variable 'operand' is always a copy because the range of type 'mlir::Operation::operand_range' (aka 'mlir::OperandRange') does not return a reference [-Werror,-Wrange-loop-analysis]
        for (const auto &operand : op->getOperands()) {
                         ^
    pmlc/dialect/tile/gradient.cc:134:10: note: use non-reference type 'mlir::Value'
        for (const auto &operand : op->getOperands()) {
             ^~~~~~~~~~~~~~~~~~~~~
    pmlc/dialect/tile/gradient.cc:216:26: error: loop variable 'dim' is always a copy because the range of type '::mlir::Operation::operand_range' (aka 'mlir::OperandRange') does not return a reference [-Werror,-Wrange-loop-analysis]
            for (const auto &dim :
                             ^
    pmlc/dialect/tile/gradient.cc:216:14: note: use non-reference type 'mlir::Value'
            for (const auto &dim :
                 ^~~~~~~~~~~~~~~~~
    pmlc/dialect/tile/gradient.cc:239:26: error: loop variable 'dim' is always a copy because the range of type '::mlir::Operation::operand_range' (aka 'mlir::OperandRange') does not return a reference [-Werror,-Wrange-loop-analysis]
            for (const auto &dim :
                             ^
    pmlc/dialect/tile/gradient.cc:239:14: note: use non-reference type 'mlir::Value'
            for (const auto &dim :
                 ^~~~~~~~~~~~~~~~~
    pmlc/dialect/tile/gradient.cc:307:20: error: loop variable 'dim' is always a copy because the range of type '::mlir::Operation::operand_range' (aka 'mlir::OperandRange') does not return a reference [-Werror,-Wrange-loop-analysis]
      for (const auto &dim : target_src_op.dims()) {
                       ^
    pmlc/dialect/tile/gradient.cc:307:8: note: use non-reference type 'mlir::Value'
      for (const auto &dim : target_src_op.dims()) {
           ^~~~~~~~~~~~~~~~~
    pmlc/dialect/tile/gradient.cc:328:20: error: loop variable 'pair' is always a copy because the range of type '::mlir::Operation::operand_range' (aka 'mlir::OperandRange') does not return a reference [-Werror,-Wrange-loop-analysis]
      for (const auto &pair : src_cons.pairs()) {
                       ^
    pmlc/dialect/tile/gradient.cc:328:8: note: use non-reference type 'mlir::Value'
      for (const auto &pair : src_cons.pairs()) {
           ^~~~~~~~~~~~~~~~~~

    plaidml/edsl/tests/edsl_test.cc:255:44: error: result of '2 ^ 11' is 9; did you mean '1 << 11' (2048)? [-Werror,-Wxor-used-as-pow]
      std::vector<uint64_t> C_output{1 ^ 10, 2 ^ 11, 3 ^ 12,  //
                                             ~~^~~~
                                             1 << 11

### DIFF
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -2905,7 +2905,7 @@ Value squeeze(const Value& value) {
     }
   }
   std::vector<Value> O_dims_values;
-  for (const auto& dim : O_dims) {
+  for (const TensorDim& dim : O_dims) {
     O_dims_values.push_back(Value{dim});
   }
   return reshape(make_tuple(Value{I}, Value{O_dims_values}));
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -2905,7 +2905,7 @@ Value squeeze(const Value& value) {
     }
   }
   std::vector<Value> O_dims_values;
-  for (const auto dim : O_dims) {
+  for (const auto& dim : O_dims) {
     O_dims_values.push_back(Value{dim});
   }
   return reshape(make_tuple(Value{I}, Value{O_dims_values}));
--- a/pmlc/rt/opencl/BUILD
+++ b/pmlc/rt/opencl/BUILD
@@ -9,4 +9,12 @@ cc_test(
     srcs = ["clprobe.cc"],
     deps = ["@opencl_icd_loader", "@opencl_hpp_headers"],
     size = "small",
+    copts = [
+      "-Werror",
+      "-Wimplicit-function-declaration",
+    ],
+    tags = [
+      "skip_macos",
+      "skip_windows",
+    ],
 )
--- a/pmlc/rt/opencl/BUILD
+++ b/pmlc/rt/opencl/BUILD
@@ -9,11 +9,8 @@ cc_test(
     srcs = ["clprobe.cc"],
     deps = ["@opencl_icd_loader", "@opencl_hpp_headers"],
     size = "small",
-    copts = [
-      "-Werror",
-      "-Wimplicit-function-declaration",
-    ],
     tags = [
+      "manual",
       "skip_macos",
       "skip_windows",
     ],
--- a/pmlc/rt/opencl/BUILD
+++ b/pmlc/rt/opencl/BUILD
@@ -9,9 +9,4 @@ cc_test(
     srcs = ["clprobe.cc"],
     deps = ["@opencl_icd_loader", "@opencl_hpp_headers"],
     size = "small",
-    tags = [
-      "manual",
-      "skip_macos",
-      "skip_windows",
-    ],
 )
--- a/plaidml/edsl/tests/edsl_test.cc
+++ b/plaidml/edsl/tests/edsl_test.cc
@@ -252,9 +252,9 @@ TEST_F(CppEdsl, BitXor) {
   std::vector<uint64_t> B_input{10, 11, 12,  //
                                 13, 14, 15,  //
                                 16, 17, 18};
-  std::vector<uint64_t> C_output{1 ^ 10, 2 ^ 11, 3 ^ 12,  //
-                                 4 ^ 13, 5 ^ 14, 6 ^ 15,  //
-                                 7 ^ 16, 8 ^ 17, 9 ^ 18};
+  std::vector<uint64_t> C_output{0x1 ^ 10, 0x2 ^ 11, 0x3 ^ 12,  //
+                                 0x4 ^ 13, 0x5 ^ 14, 0x6 ^ 15,  //
+                                 0x7 ^ 16, 0x8 ^ 17, 0x9 ^ 18};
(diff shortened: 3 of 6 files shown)


plaidml/op/lib/ops.cc
```diff
@@ -2905,7 +2905,7 @@ Value squeeze(const Value& value) {
     }
   }
   std::vector<Value> O_dims_values;
-  for (const auto dim : O_dims) {
+  for (const auto& dim : O_dims) {
```
flaub, line 2908: This should not be by reference, `Value` is a smart-pointer but it has value-like semantics.

flaub (changes requested):
Most of these changes are good, there's just one that is going in the wrong direction. We should always use value-like semantics instead of reference-like semantics for `Value`, `Attribute`, and `Type`.

pmlc/rt/opencl/BUILD
```diff
@@ -9,4 +9,9 @@ cc_test(
     srcs = ["clprobe.cc"],
     deps = ["@opencl_icd_loader", "@opencl_hpp_headers"],
     size = "small",
+    tags = [
+      "manual",
+      "skip_macos",
+      "skip_windows",
```
flaub, line 15: How are CI builds passing without these tags?

plaidml/op/lib/ops.cc
flaub: Sorry, I misread this code. I thought we were looping over `Value`, but we are looping over `TensorDim`, sorry. Avoiding the use of `auto` here would be useful in this case for making it easier to review.

pmlc/rt/opencl/BUILD
LLITCHEV: Maybe the MacOS machines don't have the latest Xcode installed.

flaub: And windows?

plaidml/op/lib/ops.cc
LLITCHEV: The type is TensorDim. Here is the error from the compiler if it is not ra ref...
plaidml/op/lib/ops.cc:2908:19: error: loop variable 'dim' of type 'const plaidml::edsl::TensorDim' creates a copy from type 'const plaidml::edsl::TensorDim' [-Werror,-Wrange-loop-analysis]
  for (const auto dim : O_dims) {
                  ^
plaidml/op/lib/ops.cc:2908:8: note: use reference type 'const plaidml::edsl::TensorDim &' to prevent copying
  for (const auto dim : O_dims) {
       ^~~~~~~~~~~~~~~~
                  &

flaub (approved):

pmlc/rt/opencl/BUILD
LLITCHEV: Well, to me it looked it was not tested on Windows. The changes now look differently... No skipping.

PR #1390, opened 2020-09-24
Range padding

jbruestle:
Pad ranges to maximize opportunities to subgroup, etc.  TODO: lit tests and code to add additional constraints during PadRanges when needed (which I want to write a lit test to verify).  Overall, this code does seem to do the correct thing on resnet layer 1 at least, and doesn't break correctness or perf.  It doesn't improve perf however, likely due to broken subgroup cost model.


pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+    // Skip some cases where the padding pass can't operate.
+    if (op.getNumSymbols()) {
+      op.emitRemark("padding cannot be run on symbolic contractions");
+      return;
+    }
+    if (!op.lowerBounds() || !op.upperBounds()) {
+      op.emitRemark("contraction bounds must be computed");
+      return;
+    }
+    // Extract which dimensions of the contraction are which output dimension
```
adstraw, line 48: This function seems like a good candidate for a utility function as with `util::computeStrideArray` below.

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+          op.upperBounds()->getResult(i).dyn_cast<AffineConstantExpr>();
+      if (!lowerConst || lowerConst.getValue() != 0) {
+        op.emitRemark("Invalid lower bound");
+        return;
+      }
+      if (!upperConst) {
+        op.emitRemark("Invalid lower bound");
+        return;
+      }
+      ranges.push_back(upperConst.getValue() + 1);
```
adstraw, line 75: Why not push upper - lower + 1 so we can handle the lower != 0 case?  Given `ranges` is signed we can handle this case?  Do we only ever expect 0 for a lower bound?

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+    auto outStrides = util::computeStrideArray(op.getResultType(), op.sink());
+    IVLOG(1, "outStrides = " << *outStrides);
+    // Decide wthat to round up
+    bool didChange = false;
+    SmallVector<int64_t, 6> newRanges = ranges;
+    for (unsigned i = 0; i < indexCount; i++) {
+      if (outStrides->strides[i] == 0) {
+        continue;
+      }
+      for (int64_t po2 = maxPowerOfTwo; po2 >= minPowerOfTwo; po2 /= 2) {
```
adstraw, line 87: Probably want to enforce that `minPowerOfTwo` and `maxPowerOfTwo` are indeed powers of two and > 0 so this code doesn't do wonky things.  Forcing powers is not a huge deal, but this loop does not terminate if `minPowerOfTwo` is less than zero.  You probably want to `uint64_t` for `po2` as well as the `minPowerOfTwo` and `maxPowerOfTwo` parameters.  There is quite a lot of use of `int64_t` in this PR for things that will be / must be unsigned.

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+      auto outDim = op.sink().getResult(i).dyn_cast<AffineDimExpr>();
+      if (!outDim) {
+        op.emitRemark("Invalid output access");
+        return;
+      }
+      outIdx[i] = outDim.getPosition();
+    }
+    // Verify bounds are simple
+    unsigned indexCount = op.lowerBounds()->getNumResults();
+    SmallVector<int64_t, 6> ranges;
```
adstraw, line 61: Interesting use of signed `int64_t` when constraints below force unsigned ranges.

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+      if (outStrides->strides[i] == 0) {
+        continue;
+      }
+      for (int64_t po2 = maxPowerOfTwo; po2 >= minPowerOfTwo; po2 /= 2) {
+        int64_t roundedRange = llvm::divideCeil(ranges[i], po2) * po2;
+        if (roundedRange == ranges[i]) {
+          break;
+        }
+        int64_t increase = roundedRange - ranges[i];
+        int64_t precIncrease = increase * 100 / ranges[i];
```
adstraw, line 93: Nit:  Do you want `percIncrease` as in *perc*ent.

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+        IVLOG(1, "Orig: " << ranges[i] << ", rounded: " << roundedRange
+                          << ", precIncrease = " << precIncrease);
+        if (precIncrease < maxIncreasePercent) {
+          newRanges[i] = roundedRange;
+          didChange = true;
+          break;
```
adstraw, line 99: This `break` implies that the cost of being closer to the "max" is always lower than being closer to the "min".  Do we know that will always be the case?

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+    }
+    // Exit early if nothing is changing
+    if (!didChange) {
+      return;
+    }
+    IVLOG(1, "Making a brave new op");
+    // Start building a new contraction to 'shrink'
+    OpBuilder builder(op.getContext());
+    builder.setInsertionPointAfter(op.getOperation());
+    // Make shinking contraction
```
adstraw, line 111: Typo:  "shrinking"

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+    auto ident = createIdentity(builder, op.getLoc(), AtomicRMWKind::assign,
```
adstraw, line 112: Is  `AtomicRMWKind::assign` something that is fixed for life?  OK to hardcode?

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+    auto idMap = AffineMap::getMultiDimIdentityMap(outRank, op.getContext());
+    auto newOp = builder.create<ContractionOp>(
+        op.getLoc(), op.getResultType(), ident, ArrayRef<Value>{op},
+        AggregationKind::assign, CombinationKind::none, idMap,
+        ArrayRef<AffineMap>{idMap},
+        IntegerSet::getEmptySet(outRank, 0, op.getContext()),
+        /*noReduce=*/true, "shrink");
+    newOp.setLowerBounds(SmallVector<int64_t, 4>(outRank, 0));
+    SmallVector<int64_t, 4> newOutSize;
+    SmallVector<int64_t, 4> shinkUpperBounds;
```
adstraw, line 123: Typo: "shrink"

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+    for (unsigned i = 0; i < outRank; i++) {
+      newOutSize.push_back(newRanges[outIdx[i]]);
```
adstraw, line 125: `newOutSize` is redundant here

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+    }
+    newOp.setUpperBounds(shinkUpperBounds);
+    // Switch all uses to the new contracton
+    op.getResult().replaceAllUsesExcept(
+        newOp.getResult(), SmallPtrSet<Operation *, 1>{newOp.getOperation()});
+    // Resize + rerange the output of the original op
+    op.getResult().setType(
+        RankedTensorType::get(newOutSize, op.getResultType().getElementType()));
+    for (unsigned i = 0; i < indexCount; i++) {
+      newRanges[i] -= 1;
```
adstraw, line 136: Might use another variable with "bounds" in the name to make it clear you are converting from a range to bounds.

pmlc/util/strides.cc
```diff
@@ -0,0 +1,91 @@
...
+#include "pmlc/util/strides.h"
+
+#include <vector>
+
+#include "mlir/Analysis/AffineStructures.h"
+#include "pmlc/util/logging.h"
+
+namespace pmlc::util {
+
+StrideArray::StrideArray(unsigned numDims, int64_t offset)
```
adstraw, line 12: Do you want to support offset < 0?

pmlc/util/strides.cc
```diff
@@ -0,0 +1,91 @@
...
+    : offset(offset), strides(numDims) {}
+
+StrideArray &StrideArray::operator*=(int64_t factor) {
```
adstraw, line 15: Do you want to support factor < 0?

pmlc/util/strides.cc
```diff
@@ -0,0 +1,91 @@
...
+  offset *= factor;
+  for (auto &dim : strides)
+    dim *= factor;
+  return *this;
+}
+
+StrideArray &StrideArray::operator+=(const StrideArray &rhs) {
+  assert(strides.size() == rhs.strides.size() && "strides sizes much match");
+  offset += rhs.offset;
+  for (unsigned i = 0, e = strides.size(); i < e; ++i) {
```
adstraw, line 25: Maybe just a style thing, but why not just i < strides.size()?

pmlc/util/strides.cc
```diff
@@ -0,0 +1,91 @@
...
+
+mlir::Optional<StrideArray> computeStrideArray(mlir::MemRefType memRefType,
+                                               mlir::AffineMap map) {
+  assert(map.getNumResults() == memref.getRank());
+  int64_t offset;
+  llvm::SmallVector<int64_t, 4> strides;
+  if (failed(getStridesAndOffset(memRefType, strides, offset)))
+    return llvm::None;
+
+  IVLOG(1, "WOOT");
```
adstraw, line 64: Woot indeed!

pmlc/util/strides.cc
flaub: This is an LLVM-ism, their style guide prefers to compute the end once per loop instead of once per iteration.

https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop

pmlc/util/strides.cc
flaub: Are you wondering why we use `int64_t` instead of `uint64_t`?

pmlc/dialect/tile/transforms/pad_constraints.cc
```diff
@@ -0,0 +1,266 @@
...
+#include "pmlc/dialect/tile/transforms/pass_detail.h"
+
+namespace pmlc::dialect::tile {
+
+using mlir::AffineExpr;
+using mlir::AffineMap;
+
+namespace {
+
+bool validForPadding(AggregationKind agg, CombinationKind comb) {
```
flaub, line 25: nit: call this `isValidForPadding` since it returns a `bool`

pmlc/dialect/tile/transforms/pad_constraints.cc
```diff
@@ -0,0 +1,266 @@
...
+      op.setConstraints(makeConstraintSet(numDims, savedConstraints));
+    }
+  });
+}
+
+} // namespace
+
+llvm::Optional<PaddingInfo> getPaddingInfo(Operation *op) {
+  if (!op)
+    return llvm::None;
```
flaub, line 236: Would it be useful to have error messages with these failure cases? Or is it OK/expected to fail in some cases?

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+      if (!upperConst) {
+        op.emitRemark("Invalid lower bound");
+        return;
+      }
+      ranges.push_back(upperConst.getValue() + 1);
+    }
+    // Get strides for output
+    auto outStrides = util::computeStrideArray(op.getResultType(), op.sink());
+    IVLOG(1, "outStrides = " << *outStrides);
+    // Decide wthat to round up
```
flaub, line 80: typo "wthat"

pmlc/dialect/tile/transforms/pad_constraints.cc
```diff
@@ -0,0 +1,266 @@
+// Copyright 2019, Intel Corporation
```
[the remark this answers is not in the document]
jbruestle: Yeah, not sure why git doesn't realize it's a move, I did a git mv.

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+namespace pmlc::dialect::tile {
+
+using namespace mlir; // NOLINT
+
+namespace {
+
+struct PadRangesPass : public PadRangesBase<PadRangesPass> {
+  int64_t maxPowerOfTwo = 16;
+  int64_t minPowerOfTwo = 8;
+  int64_t maxIncreasePercent = 5;
```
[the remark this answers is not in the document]
jbruestle: Working down is intentional here in that if we can pad to 8 for 1% increase, and 16 for 3% increase, we want to pad to 16 I think.  If we work up, we will needlessly update twice (and we will have to track update % from original), so working down is better.

pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -0,0 +1,148 @@
...
+
+void PadRangesPass::runOnFunction() {
+  auto func = getFunction();
+  func.walk([&](ContractionOp op) {
+    // Skip some cases where the padding pass can't operate.
+    if (op.getNumSymbols()) {
+      op.emitRemark("padding cannot be run on symbolic contractions");
+      return;
+    }
+    if (!op.lowerBounds() || !op.upperBounds()) {
```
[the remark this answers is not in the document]
jbruestle: Yeah, I think that is a seperate issue. Right now we have lots of places where we fail early for dynamic sizes.  I think we need to consider a plan for doing what we can sometime soon so we don't go too far without considering it. but we will need to phase it in.

pmlc/dialect/tile/transforms/pad_ranges.cc
jbruestle: I think this is a somewhat unusual case as we only handle simple output dimensions (1 to 1), which is not normally the case.  I'd like to hold of extracting this to a utility until at leas one other use arises.

pmlc/dialect/tile/transforms/pad_ranges.cc
jbruestle: Done.

pmlc/util/strides.cc
jbruestle: 1) Yes, nothing in this code should preclude offset < 0, and in fact we often have accesses like A[x + i - 1], which would result in offset < 0.
2) Also, even when stride + offsets are expected to be >= 0, the MLIR upstream idiom is to always use int64_t.  I can't honestly remember the precise justification.

pmlc/util/strides.cc
jbruestle: Yes, technically negative stride memrefs would cause such an effect, and I see no reason to disallow them (and upstream MLIR does, although given that we currently only use 'natural' layout, they will never occur).

pmlc/dialect/tile/transforms/pad_constraints.cc
jbruestle: Fixex.

pmlc/dialect/tile/transforms/pad_constraints.cc
jbruestle: Yes, in general the goal of this code is to extract the padding attributes back into a c++ structure if they exist, but in most cases (non-padded operations) they will not.  To be honest, we probably should use a custom attribute here, but given that this is really just a moved copy of the existing padding pass, I don't think it's worth mucking with right now.

pmlc/dialect/tile/transforms/pad_ranges.cc
jbruestle: Fixed.

pmlc/util/strides.cc
adstraw: I would be interested in the justification for using int64_t when we expect a value > 0.

jbruestle: So as I mentioned, in this case we do expect values < 0, however some arguments for always using int64_t are:

1) Arithmetic: It's is very common to perform arithmatic on two values, one that is always >= 0, and one that may be < 0.  In this case, default casting behavior is the make the result unsigned (even for something like int64_t - uint64_t), which is very often incorrect and results in confusion or requires explicit casting
2) Comparison: comparing two integer of different types results in warnings, requiring explicit casts, which don't really improve clarity at all
3) Counting down: uint64_t range = ...; for (auto i = range; i >= 0; i--) is an infinite loop (very easy mistake to make).
4) Upstream uses negative values as 'invalid' in many cases (personally I think this is a poor choice BTW), so for example many functions return -1 for a 'range' when the range is dynamic.
5) Upstream MLIR follows the convention of always using int64_t for ranges, strides, sizes, etc both positive and non-positive (argument by authority, again not my favorite but relevant it we want to someday upstream some of these things)
6) If you really want to state that a number is non-negative, assert(x >= 0) on int64_t x is arguably better.  It may seem that having the compiler check for you would be nicer than having the runtime check it, but the compiler doesn't really check.  Consider:
uint64_t x, y = ...;
uint64_t r = x - y;
If y is > x, the subtract just underflows and instead of getting x < 0 (which a later assert might catch), you just get r somewhere near maxint.  Now clearly one should have probably asserted (x >= y) prior to the subtract, but my point here is that the compiler isn't really validating r >= 0 via making it unsigned, since underflow isn't a failure in C, so the promise of using unsigned numbers as type checking is a lie.

Generally, I do think that using uint vs int as sort of documentation on a function parameters is nice and an argument in favor of unsigned, but I think it outweighed by the points above and whenever these slightly bike-shedding sorts of things come up I like to just use upstream (although I will never understand their love of leaving braces off of trivial if's I even try to follow that one).

flaub:
LGTM

pmlc/dialect/tile/transforms/pad_ranges.cc
adstraw: Some of my comments got hidden by GitHub and therefore not seen by @jbruestle.  Commenting here in hopes that they will be made visible.

jbruestle: Handling non-zero lower bound requires applying the proper offset and some additional analysis.  However, the prior passes/lowering should have already normalized lower bound to zero, so it will never fail in the existing pipeline.  I still check to be sure, since technically the IR allows non-zero lower bound.

pmlc/dialect/tile/transforms/pad_ranges.cc
jbruestle: Fixed.

pmlc/dialect/tile/transforms/pad_ranges.cc
jbruestle: Yes, in that if we can round to 16 for less than 5% and we can round to 8 for less than 5%, I would argue we should always round to 16.  I think that is a reasonable assumption.

pmlc/dialect/tile/transforms/pad_ranges.cc
jbruestle: Fixed.

pmlc/dialect/tile/transforms/pad_ranges.cc
jbruestle: Yes.  Basically we are doing a 1-to-1 copy, so the 'reduction' is an 'assign' (i.e. no reduction actually).

pmlc/dialect/tile/transforms/pad_ranges.cc
jbruestle: Fixed.

adstraw (approved):

PR #1391, opened 2020-09-24
WIP: fix subgroup cost model

adstraw:


plaidml/bridge/keras/backend_test.py
```diff
@@ -1738,7 +1738,7 @@ def testBigRollup(self, b, x):
         return [b.sum(x, axis=1)]
 
     # Resnet sized tests
-    @opTest([[m(1, 224, 224, 3), m(7, 7, 3, 64)]], do_grads=False)
+    @opTest([[m(1, 229, 229, 3), m(7, 7, 3, 64)]], do_grads=False)
```
flaub, line 1741: Do we want to keep this change? @tzerrell

adstraw: No, this is for testing.   Still debugging.  Changes are not ready for review.  I was hoping the "WIP" tag would signify that this was not yet ready for prime time.

adstraw: Please also ignore logging changes --- as is this PR creates WAY to many IVLOG1 logs.  Etc.  Will remove "WIP" when it's in a reviewable state.  Pushed it to PR to get @jbruestle eyeballs on it as my calculateMemInfo changes are causing long runtimes for ResNet50.

flaub: Yep, no problem.

pmlc/dialect/pxa/transforms/subgroups.cc
```diff
@@ -165,16 +166,28 @@ struct SubgroupCostModel {
...
-    for (unsigned i = 0; i < ranges.size(); i++) {
-      auto iv = op.getIVs()[i];
-      if (si.strides.count(iv)) {
-        if (plan.subgroupTile[i] == plan.subgroupSize && si.strides[iv] == 1) {
+    for (auto si : blah) {
+      IVLOG(1, "stride info = " << si);
+      int64_t pos = 0;
+      // TODO: add neg
+      for (auto kvp : si.strides) {
+        auto index = kvp.first.getArgNumber();
```
jbruestle, line 176: So technically this is a bug, since it's valid to have something like this:

affine.parallel (%i) = (0) to (10)
  affine.parallel (%j) = (0) to (10)
    %0 = pxa.load %A[%i + %j]

Now, both %i and %j will have getArgNumber = 0, but they are on different blocks, and perhaps %j is being considered for tiling but %i is not.  Basically this code presumes a single block nest only.

The better way to do this is to make a tile map: DenseMap<BlockArgument, int64_t>for all tiled indexes.  If an index in si.strides is not in the tile map, we should consider it a tile size of 1 (or more simply, don't add anything to 'pos', which is equivalent).

That said, I don't think this bug is the issue with the slow runtimes, since when we call the subgroup pass in the existing pipeline we have a block depth of exactly one since it's basically fresh from the tile lowering.

adstraw:
rebooted this effort on another branch.  will open separate PR.

Issue #1381, opened 2020-09-23
Fusion pass for intel_gen


abialas1:
Currently fusion pass works naively for intel_gen, and does not fuse ops that are using the subgroup ops. The following points should be added to this pass:
o	Needs integration into pipeline.cc (including probably adding tagging to results of TileAccumulation fallback in the sugroups pass to allow promote inner to be turned back on)
o	Needs support for tiling one of the two fused loops for compatability to fuse eltwise into subgroups post subgrouping
o	Needs support for more control over block level and inner vectorization

abialas1:
Quick update on fusions without going deeper into details.
- The overall boost in current implementation is 10%. The analyzed conv1x1s + bias + relu, have up to 50% perf boost.
- No change or worse results on conv 3x3.
- Bad perf on the residual block eltwise.
- Had to remove localize pass, need to add more logic on that since too big buffers are getting promoted to in kernel memory
- Debug in progress on the cases where no boost or perf drop is seen

abialas1:
More updates:
- localize pass fails on the branches with skip layer on kernel with fused branch1 + branch2c + eltwise add + relu
- fusing on branches with skip layer needs additional logic to not fuse them (these should be executed with OOQ, start upon branch2a and have dependency on res2a only)
- perf boost on residual blocks (res2) without skip layers is ~+25%.  More than half of the time takes conv3x3 (which has the lower boost after fusing, few % only), rest are two conv1x1 + eltwise (boost +~50% after fusing)

PR #1396, opened 2020-09-25
WIP: adding tilings to outer loops fusions

abialas1:
Do not merge. Works only on simplest cases

### DIFF
--- a/pmlc/conversion/gpu/kernel_outlining.cc
+++ b/pmlc/conversion/gpu/kernel_outlining.cc
@@ -157,13 +157,17 @@ public:
         spirv::getTargetEnvAttrName());
     if (!target_env) {
       auto triple = spirv::VerCapExtAttr::get(
-          spirv::Version::V_1_0,
-          {spirv::Capability::Shader, spirv::Capability::Groups,
+          spirv::Version::V_1_5,
+          {spirv::Capability::Shader, spirv::Capability::GroupNonUniformBallot,
+           spirv::Capability::SubgroupBufferBlockIOINTEL,
            spirv::Capability::Int64, spirv::Capability::Int16,
            spirv::Capability::Int8, spirv::Capability::Float64,
-           spirv::Capability::Float16},
+           spirv::Capability::Float16,
+           spirv::Capability::StorageBuffer16BitAccess},
           ArrayRef<spirv::Extension>(
-              spirv::Extension::SPV_KHR_storage_buffer_storage_class),
+              {spirv::Extension::SPV_KHR_storage_buffer_storage_class,
+               spirv::Extension::SPV_KHR_16bit_storage,
+               spirv::Extension::SPV_INTEL_subgroups}),
           &getContext());
       getOperation().setAttr(
           spirv::getTargetEnvAttrName(),
--- a/pmlc/dialect/pxa/transforms/fusion.cc
+++ b/pmlc/dialect/pxa/transforms/fusion.cc
@@ -8,6 +8,7 @@
 #include "pmlc/dialect/pxa/ir/ops.h"
 #include "pmlc/dialect/pxa/transforms/pass_detail.h"
 #include "pmlc/util/logging.h"
+#include "pmlc/util/tags.h"
 #include "pmlc/util/util.h"
 
 using namespace mlir; // NOLINT[build/namespaces]
@@ -379,6 +380,15 @@ struct FusionInfo {
         /*lbMap=*/lowerC, /*lbArgs=*/aInfo.op.getLowerBoundsOperands(),
         /*ubMap=*/upperC, /*ubArgs=*/aInfo.op.getUpperBoundsOperands(),
         /*steps=*/stepsC);
+
+    // Copy across any tags, prefer A's.
+    if (hasTags(aInfo.op)) {
+      copyTags(apC, aInfo.op);
+    } else if (hasTags(bInfo.op)) {
+      copyTags(apC, bInfo.op);
+    }
+    clearTags(aInfo.op);
+    clearTags(bInfo.op);
 
     // Move the two parallel for's inside the new op
     aInfo.op.getOperation()->moveBefore(apC.getBody(), apC.getBody()->end());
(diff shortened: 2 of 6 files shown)


pmlc/dialect/pxa/transforms/fusion.cc
```diff
@@ -309,10 +338,14 @@ struct FusionInfo {
     }
     // For each raw & waw, consider the plan
     for (auto &raw : readAfterWrites) {
-      considerPlan(raw.first, raw.second);
+      auto write = cast<AffineWriteOpInterface>(raw.first);
+      auto read = cast<AffineReadOpInterface>(raw.second);
```
flaub, line 342: PXA doesn't actually implement these interfaces. We should probably create our own PXA based interfaces, and we've talked about doing so, but it hasn't been implemented yet.

abialas1: I saw that Affine interface was used for load and vector_load ops, so added the same to reduce. Let me separate these two based on the Affine ones.

pmlc/dialect/pxa/transforms/fusion.cc
```diff
@@ -16,8 +18,8 @@ namespace pmlc::dialect::pxa {
 
 namespace {
 
-using WriteRead = std::pair<PxaReduceOp, PxaLoadOp>;
-using WriteWrite = std::pair<PxaReduceOp, PxaReduceOp>;
+using WriteRead = std::pair<Operation *, Operation *>;
```
jbruestle, line 21: maybe use PxaWriteOpInterface, PxaReadOpInterface here?

pmlc/dialect/pxa/ir/interfaces.td
```diff
@@ -0,0 +1,169 @@
...
+      /*args=*/(ins),
+      /*methodBody=*/[{}],
+      /*defaultImplementation=*/[{
+        return cast<ConcreteOp>(this->getOperation());
+      }]
+    >,
+  ];
+}
+
+def PxaWriteOpInterface : PxaOpInterface<"PxaWriteOpInterface"> {
```
jbruestle, line 89: Should we call this PxaReduceOpInterface?  Write doesn't imply the fact that a reduction is happening (unless type is assign).

pmlc/dialect/pxa/transforms/tile.cc
```diff
@@ -64,4 +64,55 @@ AffineParallelOp performTiling(AffineParallelOp op,
   return inner;
 }
 
+AffineParallelOp undoTiling(AffineParallelOp op,
```
jbruestle, line 67: This code looks correct to me, but a small test of tile / untile would be good.

pmlc/dialect/pxa/transforms/fusion.cc
```diff
@@ -189,6 +210,14 @@ struct FusionInfo {
       return false;
     }
 
+    // Add tiling if needed
+    if (needsTiling) {
+      if (reverseFusion)
+        performTiling(aInfo.op, tileSizes);
```
jbruestle, line 216: I presume that the code for untiling isn't yet written?  I realize this is a WIP.

jbruestle (commented):
Directionally, looks very good, I like the interface refactors and the untile method seems like it was straightforward.  Some minor thoughts inline.

pmlc/conversion/gpu/kernel_outlining.cc
```diff
@@ -157,13 +157,17 @@ class GpuKernelOutliningPass
         spirv::getTargetEnvAttrName());
     if (!target_env) {
       auto triple = spirv::VerCapExtAttr::get(
-          spirv::Version::V_1_0,
-          {spirv::Capability::Shader, spirv::Capability::Groups,
+          spirv::Version::V_1_5,
```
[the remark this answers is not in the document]
abialas1: yes, actually they are in the separate PR already, I need these for local testing, will clean it up before moving from WIP to production

pmlc/dialect/pxa/transforms/fusion.cc
abialas1: yes, please see the latest checkin which has the tiling\untiling implemented

abialas1:
The PR got huge and would like to push it at this point to not struggle with the code rebases. The functionality should be transparent to what we have on plaidml-v1 branch right now, and added more options for fusing, added vector support for dataflow_opt, I1StorageToI32 and other changes needed for fusings. I will put detailed status for fusing in https://github.com/plaidml/plaidml/issues/1381

jbruestle:
Totally agree, In general I'm happy land code even if it's not complete as long as it doesn't break anything and has gotten code review.  However right now this code is failing CI hard, so that would need to be fixed first.  Also, it would be slightly nicer to land this in 2-3 chunks since there are a few unrelated changes, but it would also be nice to land it soon, so your call on whether or not to split it up.

abialas1:
Hi Jeremy, dividing this code into several chunks is doable, however will need some work on my side to have it separate, as some chunks are dependent on others to have the full pipeline working. Unless it is really needed I would like to focus on perf analysis instead. Let me know if that would be ok.
Basically there are 4 parts to this checkin:
1. Interfaces for PXA Read and Reduce
2. Vector support for I1ToI32Storage
3. Vector support for dataFlowOpt
4. Parametrized tiled fusions + undoTiling

jbruestle:
I think in the interest of time, it's OK to land this basically as is.  However in CI it seems to be breaking one test.  You should be able to repro via:

bazelisk run --//plaidml:device=vulkan.0 --//plaidml:target=intel_gen //plaidml/edsl/tests:cc_test

It's probably an actual edge case that needs proper handling, so once you root cause it, maybe add a simplified lit test if it's not too much work.  Also, I'm a bit worried about architecture specific stuff (subgroups) leaking too much into general purpose stuff (fusion), however I think I have a way to fix it in this case (basically compute the tiling to use based on the the relations of the loop indexes rather than the attribute), and I can refactor after this lands, so that's not a blocker.

PR #1400, opened 2020-09-29
Add OpenCL+SPIRV target

kdobros:


plaidml/BUILD
```diff
@@ -36,6 +36,7 @@ plaidml_cc_library(
         "//pmlc:all_passes_and_dialects",
         "//pmlc/target/demo",
         "//pmlc/target/intel_gen",
+        "//pmlc/target/intel_gen_ocl_spirv",
```
earhart, line 39: This shouldn't block this PR, but I think we need a better organization scheme for our targets.  :-)

earhart (approved):

PR #1404, opened 2020-09-29
Update OpenVINO Reference

tzerrell:
Advance our reference to OpenVINO to a more recent version of their `master` branch.

If we wanted we could instead change our reference to be OpenVINO 2020.4.


flaub (approved):

tzerrell:
There was a bug in this, see #1405

PR #1406, opened 2020-09-30
PlaidML API update

flaub:
Breaking Changes:
* Contraction construction in C++ and Python is done with a fluent builder. `TensorOutput` is removed and replaced with a `Contraction` builder class.
  Example:
```
  Tensor O = Contraction()
	.outShape(I, J)
	.outAccess(i, j)
	.sum(X(i, k) * Y(k, j));
```
  The Contraction gets built into a Tensor in one of two ways:
  1. Explicitly, via the `build()` method.
  2. Implicitly, via the `operator Tensor()` method.

* `plaidml::edsl::ProgramBuilder` is replaced by `plaidml::edsl::buildProgram()`
  Both inputs and outputs are now specified during program construction. The order of inputs specified will correspond to the order of buffer arguments to any executables created later.
* Moved `plaidml::edsl::Program` into `plaidml::Program`.
* The `Program` returned from `buildProgram()` starts off as target-independent Tile dialect. Use the `compile()` method to transform it into a target-dependent module.
* Dropped `plaidml::exec::Binder`, this is replaced by specifying buffers by position when constructing a `plaidml::exec::Executable`.
* Dropped `plaidml::edsl::LogicalShape`, this is replaced by `plaidml::TensorShape`. A `TensorShape` may optionally specify strides to provide layout information.
* Dropped the `plaidml::edsl::Tensor` constructors that allow buliding of placeholders and constants. Use `plaidml::edsl::Placeholder()` and `plaidml::edsl::Constant()` instead.
* `floatx` and `intx` are now DataTypes, they represent arbitrary precision floating-point and integer values. During type/shape inference, these arbitrary precision types always get promoted to concrete precision types.
* `plaidml::edsl::select()` now requires that the then/else values must have the same data type and they must not be arbitrary precision. This means that casts must be used beforing calling `select()`.
* The assignment contraction must use a concrete precision type, use a cast when necessary.
* `plaidml::Buffer` must now be constructed with a `plaidml::TensorShape`.
* 'Updates' are now handled outside of the PlaidML program builder and execution APIs.
* Dropped `plaidml_view`, the view methods are now directly available on the `Buffer` object.
* Dropped support for Gradients. This means PlaidML can no longer be used for training.
* Dropped being able to bind concrete shapes to existing Tensor objects. Keras is the only backend that requires this and it can now be done at Program construction time; an optional `shapes` argument is provided for this purpose.
* Renamed `Contraction::no_reduce()` to `Contraction::simplify(bool)`
* Renamed `Contraction::use_default()` to `Contraction::init()`
* Dropped `plaidml_dim_expr_get_int` and corresponding C++ and python wrappers.

Other changes:
* Update Keras backend to use new API.
  * Updates are handled by keras backend.
  * Variables must be traced and processed before using the PlaidML APIs.
  * To support variable tracing, each `_KerasNode` now contains a list of `operands`, representing dependent nodes.
  * Re-enable some backend tests that weren't comparing results to any other backend.
* Update OpenVINO plugin to use new API.
  * Disabled ReshapeCheckDynBatch/ReshapeLayerTest for now as they cause crashes.
* Update Tensorflow bridge to use new API.
  * Disabled `i3d_test` for now, need to figure out how to freeze the graph so that we can use `Constant()` instead of `Placeholder()` for variables.
* Remove the 'builder' sub-dialect from the Tile dialect. A separate AST implemention is used as the backing for the eDSL. This AST gets converted into tile dialect at program construction time.
* Add `plaidml_buffer_adopt` C API to allow for user owned buffers to be used.
* Remove lit testing from `edsl_test.py`
* Verify that input/output buffers specified to an `Executable` match the shapes specified in the corresponding `Program`.
* During lowering of Tile to PXA, inject `eltwise.ident` ops for each return operand that is a direct block argument.
* Drop lots of unused code and cruft.


plaidml/bridge/keras/backend_test.py
```diff
@@ -1330,17 +1330,17 @@ def testRandomChanges(self, b):
...
 
-    @compareForwardClose(skip_tensorflow=True, atol=1e-6)
+    @compareForwardClose(skip_tensorflow=True, skip_theano=False, atol=1e-6)
     def testBatchNormAndUpdate(self, b):
         b.set_learning_phase(1)
         x = b.variable(n(4, 7))
-        moving_mean = b.variable(m(4, 1))
-        moving_var = b.variable(m(4, 1))
+        moving_mean = b.variable(m(4, 4))
+        moving_var = b.variable(m(4, 4))
```
flaub, line 1343: @tzerrell This test was failing once I re-enabled Theano comparisons. The `(4, 1)` shape for these variables weren't matching the resulting shape of `(4, 4)` for the 'target' of the updates (i.e. `moving_mean` and `moving_var`). Does this seem alright to you?

plaidml/bridge/openvino/tests/functional/plugin/plaidml/shared_tests_instances/single_layer_tests/reshape.cpp
```diff
@@ -15,15 +15,15 @@ const std::vector<InferenceEngine::Precision> netPrecisions = {
...
-                         ReshapeLayerTest::getTestCaseName);
+// INSTANTIATE_TEST_SUITE_P(ReshapeCheckDynBatch, ReshapeLayerTest,
+//                          ::testing::Combine(::testing::Values(true),                                      //
+//                                             ::testing::ValuesIn(netPrecisions),                           //
+//                                             ::testing::Values(std::vector<size_t>({1, 16, 16, 16})),      //
+//                                             ::testing::Values(std::vector<size_t>({1, 0, 256})),          //
+//                                             ::testing::Values(CommonTestUtils::DEVICE_PLAIDML),           //
+//                                             ::testing::Values(std::map<std::string, std::string>(         //
+//                                                 {{CONFIG_KEY(DYN_BATCH_ENABLED), CONFIG_VALUE(YES)}}))),  //
+//                          ReshapeLayerTest::getTestCaseName);
```
flaub, line 26: @tzerrell @cnamrata15 @brianretford These tests now appear to crash with the latest changes. I wasn't able to determine exactly why, but it seems similar to the full network issues that were reported recently. Is it OK to drop these for now?

plaidml/bridge/tensorflow/tests/BUILD
```diff
@@ -143,7 +146,10 @@ tf_cc_test(
     data = [":i3d_test_genrule"],
     extra_copts = ["-fexceptions"],
     linkstatic = 1,
-    tags = ["skip_windows"],
+    tags = [
+        "manual",
```
flaub, line 150: @dgkutnic I made this test `manual` for now until we can either:
1. Get the model frozen before we convert it to HLO, so that all the existing variables come in as `Constant()` rather than `Placeholder()`.
2. Update the arguments being passed to the Executable so that the weights are in the correct order. This seems like lots of tedious work, and it's actually pretty brittle; we shouldn't require users to have to know the correct order.

I've made some progress on figuring out how to freeze the graph, we can chat later about it.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -1301,6 +1301,18 @@ struct TraceOpConversion : public OpConversionPattern<TraceOp> {
...
+    getOperation().walk([&](ReturnOp op) {
+      OpBuilder builder(op);
+      for (OpOperand &operand : op.getOperation()->getOpOperands()) {
+        if (operand.get().isa<BlockArgument>()) {
+          Value value = builder.create<ew::IdentOp>(op.getLoc(), operand.get());
+          operand.set(value);
+        }
+      }
+    });
+
```
flaub, line 1315: @jbruestle  This is the injection of `ident` ops that I mentioned.

pmlc/rt/device.h
```diff
@@ -17,7 +18,8 @@ class Device {
 
   virtual std::unique_ptr<Executable>
   compile(const std::shared_ptr<pmlc::compiler::Program> &program,
-          mlir::ArrayRef<void *> bufptrs) = 0;
+          mlir::ArrayRef<util::BufferPtr> inputBuffers,
+          mlir::ArrayRef<util::BufferPtr> outputBuffers) = 0;
```
flaub, line 22: @earhart Passing `BufferPtr` and splitting into inputs vs outputs makes some of the logic in other places a little easier. WDYT?

vendor/tensorflow/workspace.bzl
```diff
@@ -22,6 +22,13 @@ def plaidml_tf_workspace():
         },
     )
 
+    http_archive(
+        name = "tfhub_i3d_kinetics_400",
+        url = "https://storage.googleapis.com/tfhub-modules/deepmind/i3d-kinetics-400/1.tar.gz",
+        sha256 = "bafe29bb4528badad428207d8fe86ca2a998b5b1386b82e42175339f12ea2ff5",
+        build_file = clean_dep("//vendor/tensorflow:tfhub_i3d_kinetics_400.BUILD"),
+    )
+
```
flaub, line 31: @dgkutnic This approach obviates the need to manage the TF hub cache; we let bazel manage it instead.

vendor/opencl_icd_loader/opencl_icd_loader.BUILD
```diff
@@ -61,8 +61,8 @@ cc_library(
     linkopts = opencl_icd_loader_lnks,
     linkstatic = 1,
     local_defines = select({
-        "@bazel_tools//src/conditions:windows": [],
-        "//conditions:default": ["HAVE___SECURE_GETENV"],
+        "@bazel_tools//src/conditions:linux_x86_64": ["HAVE_SECURE_GETENV"],
+        "//conditions:default": [],
```
flaub, line 65: @earhart @LLITCHEV @kdobros I had to make this change to make things work correctly in my environment. WDYT?

LLITCHEV: Yeah... This is very sensitive on the environment, but Linux has the SEC_ENV, so this should be fine.

tzerrell (commented):
Awesome!

I have one thing in Keras backend_test I'll be following up on (details inline on that file). Other than that, I have a bunch of comments but I don't think any of them preclude landing this if everyone else things it's good to go.

If you haven't done so already you may want to try running Keras backend_tests with --fp16 and see what breaks; most of those tests are passing.

vendor/opencl_icd_loader/opencl_icd_loader.BUILD
earhart: I think you should delete the entire `local_defines` block.

At least when I tried it: this condition wasn't working on our Linux builds.  I suspect it's because we use "linux_x86_64" for our CPU, instead of "k8", which is what the condition uses.

Also, when I tried using `HAVE_SECURE_GETENV`, it didn't work on our CI; we needed the double-underscore version.

pmlc/rt/device.h
earhart: Completely fine; we just need a convention for ordering.  It might also be good to talk somewhere about using the same buffer as an input and an output (explicitly noting that it's legal, if it's legal).  It'd also be useful to have another input for "things we expect to not change" -- I'm not quite sure what the plan is for constant folding.

earhart (commented):
This is maybe a bit large for a thorough review.  :-). It mostly looks fine to me, though.

plaidml/bridge/keras/__init__.py
```diff
@@ -104,73 +101,120 @@ def wrapper(*args, **kwargs):
...
+        var = variable(x)
+        return var, var.tensor
+    if isinstance(x, _KerasNode):
+        return x, x.tensor
+    return x, x
+
+
+class _Runner(object):
+
+    def __init__(self, name, shapes, inputs, outputs, updates, vars):
```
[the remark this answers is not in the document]
flaub: They actually do get added, although indirectly. There's a separate process which traces down all the variables in the program. It turns out the 'source' of an update with always be a variable. These get collected into `vars`.

pmlc/rt/jit_executable.cc
```diff
@@ -386,10 +389,18 @@ class JitExecutable final : public Executable {
...
 
-    descriptors.reserve(bufptrs.size());
-    for (unsigned i = 0; i < bufptrs.size(); i++) {
-      descriptors.emplace_back(bufptrs[i], program->arguments[i].shape);
-      ptrs.push_back(descriptors[i].ptr());
+    for (auto [type, buffer] : llvm::zip(program->inputs, inputBuffers)) {
+      descriptors.emplace_back(buffer->data(), type.cast<RankedTensorType>());
+      ptrs.push_back(descriptors.back().ptr());
+    }
+    for (const compiler::ConstantArgument &arg : program->constants) {
```
flaub, line 396: @earhart Hopefully this answers your question on how constant buffers are currently handled. They come from the `Program` and are injected here, sandwiched between inputs and outputs.

pmlc/rt/device.h
flaub: Ordering is determined by the `Program` (and actually by the end-user).

plaidml/edsl/__init__.py
```diff
@@ -355,112 +295,171 @@ def __rfloordiv__(self, lhs):
...
-            >>> R = TensorOutput()
-            >>> R[()] <= A[i, j]
+            >>> R = Contraction().sum(A[i, j]).build()
         """
-        self._tensor._set_contraction(self._make_contraction(lib.PLAIDML_AGG_OP_MIN, rhs))
+        self.__agg_op = lib.PLAIDML_AGG_OP_SUM
+        self.__rhs = rhs
+        return self
+
+    def simplify(self, flag):
```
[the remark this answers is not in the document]
flaub: To be removed in future PR

plaidml/edsl/__init__.py
```diff
@@ -247,22 +196,19 @@ def __lt__(self, rhs):
         Example:
             >>> i, j = TensorIndexes(2)
             >>> A = Placeholder(DType.FLOAT32, [3, 3])
-            >>> R = TensorOutput()
-            >>> R[()] += A[i, j]
-            >>> R.add_constraint(i < 5)
+            >>> R = Contraction().sum(A[i, j]).add_constraint(i < 5).build()
```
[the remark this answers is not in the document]
flaub: I'll add a `Reduction()` builder later for these cases.

plaidml/bridge/keras/__init__.py
```diff
@@ -104,73 +101,120 @@ def wrapper(*args, **kwargs):
...
-        exe = self._cache.get(input_shapes)
-        if not exe:
-            exe = self._compile(inputs)
-            self._cache[input_shapes] = exe
-        return exe(inputs)
+        runner = self._cache.get(input_shapes)
+        if not runner:
+            runner = self._compile(inputs)
+            self._cache[input_shapes] = runner
+        # logger.debug('run({})'.format(inputs))
```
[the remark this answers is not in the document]
flaub: Agreed

plaidml/bridge/keras/__init__.py
```diff
@@ -1407,15 +1464,15 @@ def time_expand(val, ii, t, prev):
...
-        O = edsl.TensorOutput(*O_dims)
         O_idxs = [batch_idx] + [ii] + idxs
-        O[O_idxs] = I[I_idxs]
+        OC = edsl.Contraction().outShape(*O_dims).outAccess(*O_idxs).assign(I[I_idxs])
         if prev is None:
             if ii != 0:
                 raise RuntimeError(
                     'Generating RNN at time step {} with no previous time step'.format(ii))
         else:
-            O.use_default(prev.tensor)
```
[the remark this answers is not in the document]
flaub: Oops, good catch!

plaidml/bridge/keras/__init__.py
```diff
@@ -1407,15 +1464,15 @@ def time_expand(val, ii, t, prev):
...
+            OC.simplify(False)
```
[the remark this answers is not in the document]
flaub: OK, we'll remove this in the future as discussed

jbruestle (approved):

plaidml/bridge/keras/backend_test.py
tzerrell, line 1343: This is not a fix for this PR, but:

It looks like the weird shape (4, 4) is being introduced by the interaction of the lines
```
    # squeeze the mean and variance in all cases, that's what TF does
    m = squeeze(m)
    v = squeeze(v)
```
with the unsqueezed sizes (4, 1) for the moving mean & var; this ends up adding a (4, 1) tensor to a (4) tensor, which broadcasts to (4, 4). Probably we should switch to using size (4) for the moving mean & var

flaub: OK, thanks for looking into this

pmlc/rt/device.h
earhart: I meant for the network function call -- "inputs before outputs".  And now I see the constants -- awesome!

PR #1424, opened 2020-10-02
Move -Wall to PlaidML code only

earhart:
Hey, @flaub -- thoughts?  I was doing a build and watching the build warnings from the parts of the codebase we don't control; I'm thinking that maybe we should keep being strict about our own code, but maybe remove some of the noise coming from other components.


flaub (changes requested):
Please hold off on this until I can verify that downstream uses of this repository won't break.

earhart:
No problem

PR #1425, opened 2020-10-02
Add exactly-match option for fusion pass

shiyao00:
If exactly-match is true, fuse the loops with only exactly matched indices. The default value is false.
In GRN case, apply one fusion pass with exactly-match==true and another pass with exactly-match==false. It would fuse GRN as a single big loop.

### DIFF
--- a/pmlc/dialect/pxa/tests/fusion-match.mlir
+++ b/pmlc/dialect/pxa/tests/fusion-match.mlir
@@ -1,0 +1,34 @@
+// RUN: pmlc-opt -pxa-fusion="exactly-match=true" -pxa-normalize -canonicalize %s | FileCheck %s
+
+func @fusion_different_idxs(%A: memref<2x3xf32>, %B: memref<2x3xf32>, %C: memref<2x3x4xf32>, %D: memref<2x3x4xf32>) -> memref<2x3x4xf32> {
+  %T = alloc() : memref<2x3xf32>
+  %4 = affine.parallel (%i, %j) = (0, 0) to (2, 3) reduce ("assign") -> (memref<2x3xf32>) {
+    %0 = pxa.load %A[%i, %j] : memref<2x3xf32>
+    %1 = pxa.load %B[%i, %j] : memref<2x3xf32>
+    %2 = addf %0, %1 : f32
+    %3 = pxa.reduce assign %2, %T[%i, %j] : memref<2x3xf32>
+    affine.yield %3 : memref<2x3xf32>
+  }
+  %5 = affine.parallel (%i, %j, %k) = (0, 0, 0) to (2, 3, 4) reduce ("assign") -> (memref<2x3x4xf32>) {
+    %0 = pxa.load %4[%i, %j] : memref<2x3xf32>
+    %1 = pxa.load %C[%i, %j, %k] : memref<2x3x4xf32>
+    %2 = mulf %0, %1 : f32
+    %3 = pxa.reduce assign %2, %D[%i, %j, %k] : memref<2x3x4xf32>
+    affine.yield %3 : memref<2x3x4xf32>
+  }
+  return %5 : memref<2x3x4xf32>
+}
+
+// CHECK-LABEL: func @fusion_different_idxs
+// CHECK:       affine.parallel (%{{.*}}, %{{.*}}) = (0, 0) to (2, 3)
+// CHECK:         pxa.load
+// CHECK:         pxa.load
+// CHECK:         addf
+// CHECK:         pxa.reduce
+// CHECK:         affine.yield
+// CHECK:       affine.parallel (%{{.*}}, %{{.*}}, %{{.*}}) = (0, 0, 0) to (2, 3, 4)
+// CHECK:         pxa.load
+// CHECK:         pxa.load
+// CHECK:         mulf
+// CHECK:         pxa.reduce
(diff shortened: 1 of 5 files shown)


flaub:
IIUC, instead of requiring two passes, we could have the fusion pass consider an additional point to perform fusion. Right now, it skips over the 'yield'. I think the fix is to modify the iteration so that it stops on 'yield' ops. @jbruestle?

shiyao00:
> IIUC, instead of requiring two passes, we could have the fusion pass consider an additional point to perform fusion. Right now, it skips over the 'yield'. I think the fix is to modify the iteration so that it stops on 'yield' ops. @jbruestle?

The current pass performs fusion only at the top-level. I'm not confident if it can recursively work for inner level. If it's not good for inner loops, we should use multiple passes.

flaub:
I was under the impression that GRN only needed top-level fusions.

shiyao00:
> I was under the impression that GRN only needed top-level fusions.

Yes, it needs only top-level fusion. Currently, for the first pass, the last loop can't be fused when the last second loop is fused.  This is another issue. When I let fused the last loop, the last and second last loops are both inner loop. As the current pass can't perform inner loop fusion, they can't be fused then.

So we have to avoid the fusion of the second last loop and its previous loos in order to fuse the last and second last loop first. Otherwise, they become inner and we can't fuse them.

shiyao00:
> I was under the impression that GRN only needed top-level fusions.

If we perform only top-level fusion from the beginning to the end, we'll get a whole loop (when fixing a minor problem). However, it looks like

for (0,0) to (128,256)
    ......
    # last second
    for (0) to (24)
          ....
    # last
    for (0) to (24)
          ....

jbruestle:
So two things:  
1) I do think we want to support 'inner' fusion at some point.  Specifically, I've talked with Artur and Frank about possibly adding a 'recurse' option to fusion as some point which would try another fusion scan of all the ops inside any loops it fused: i.e. if we do a fusion on A + B, reconsider the interior of the new merged loop and fuse any inner blocks further.  That said, I don't believe anyone is actively working on that at this moment.
2) On the other hand an option to fusion to only do 'exact' fusion has multiple uses (for example, perhaps a platform only really support eltwise non-broadcast fusions for some reason, or we want to avoid complex loops), and it doesn't add much code or complexity.  So I think this is a useful PR.
3) I'm not sure if two passes with the first set to 'exact' is the way to go for GRN.  Seems like it might work for this case but not in general, whereas handing yields between the write and the read properly might be more 'general'.  However I haven't considered it in much detail, and I don't have a strong feeling either way.  And like I said, I think the actual content of the change is arguably useful even if we decide to go with a different solution to this particular case.

Conclusion:  Personally, I'm OK with this landing, but I'd like to hear if there are any additional thoughts before landing it.

shiyao00:
> So two things:
> 
> 1. I do think we want to support 'inner' fusion at some point.  Specifically, I've talked with Artur and Frank about possibly adding a 'recurse' option to fusion as some point which would try another fusion scan of all the ops inside any loops it fused: i.e. if we do a fusion on A + B, reconsider the interior of the new merged loop and fuse any inner blocks further.  That said, I don't believe anyone is actively working on that at this moment.
> 2. On the other hand an option to fusion to only do 'exact' fusion has multiple uses (for example, perhaps a platform only really support eltwise non-broadcast fusions for some reason, or we want to avoid complex loops), and it doesn't add much code or complexity.  So I think this is a useful PR.
> 3. I'm not sure if two passes with the first set to 'exact' is the way to go for GRN.  Seems like it might work for this case but not in general, whereas handing yields between the write and the read properly might be more 'general'.  However I haven't considered it in much detail, and I don't have a strong feeling either way.  And like I said, I think the actual content of the change is arguably useful even if we decide to go with a different solution to this particular case.
> 
> Conclusion: Personally, I'm OK with this landing, but I'd like to hear if there are any additional thoughts before landing it.

I agree. This PR adds the "exact" option which is a trivial walkaround for GRN, which isn't harmful to anything. I've verified that this works for GRN. In essential, we may need the recursive fusion.

shiyao00:
> So two things:
> 
> 1. I do think we want to support 'inner' fusion at some point.  Specifically, I've talked with Artur and Frank about possibly adding a 'recurse' option to fusion as some point which would try another fusion scan of all the ops inside any loops it fused: i.e. if we do a fusion on A + B, reconsider the interior of the new merged loop and fuse any inner blocks further.  That said, I don't believe anyone is actively working on that at this moment.
> 2. On the other hand an option to fusion to only do 'exact' fusion has multiple uses (for example, perhaps a platform only really support eltwise non-broadcast fusions for some reason, or we want to avoid complex loops), and it doesn't add much code or complexity.  So I think this is a useful PR.
> 3. I'm not sure if two passes with the first set to 'exact' is the way to go for GRN.  Seems like it might work for this case but not in general, whereas handing yields between the write and the read properly might be more 'general'.  However I haven't considered it in much detail, and I don't have a strong feeling either way.  And like I said, I think the actual content of the change is arguably useful even if we decide to go with a different solution to this particular case.
> 
> Conclusion: Personally, I'm OK with this landing, but I'd like to hear if there are any additional thoughts before landing it.

However, I think recursive fusion may not be always good in some cases. So we shouldn't just implement it. We have to consider a cost model probably.

jbruestle (approved):

PR #1428, opened 2020-10-02
Fix comp to llvm memref conversion

kdobros:
Default function signature conversion also converts memrefs in function
signature which results in lowering not expected by runtime.
Proposed solution is to convert only `comp` dialect types here.


pmlc/conversion/comp_to_llvm/common.cc
```diff
@@ -112,9 +114,57 @@ getManglingTypes(mlir::MLIRContext *context) {
       mlir::FloatType::getF32(context),    mlir::FloatType::getF64(context)};
 }
 
+namespace {
+struct ConvertFuncSignature : mlir::OpConversionPattern<mlir::FuncOp> {
+  ConvertFuncSignature(mlir::TypeConverter &typeConverter,
+                       mlir::MLIRContext *context)
+      : mlir::OpConversionPattern<mlir::FuncOp>(typeConverter, context) {}
```
earhart, line 121: You can use `using mlir::OpConversionPattern<mlir::FuncOp>::OpConversionPattern;` to pick up the constructors directly.

pmlc/conversion/comp_to_llvm/common.cc
```diff
@@ -112,9 +114,57 @@ getManglingTypes(mlir::MLIRContext *context) {
       mlir::FloatType::getF32(context),    mlir::FloatType::getF64(context)};
 }
 
+namespace {
+struct ConvertFuncSignature : mlir::OpConversionPattern<mlir::FuncOp> {
```
earhart, line 118: This struct should be `final`

pmlc/conversion/comp_to_llvm/common.cc
```diff
@@ -112,9 +114,57 @@ getManglingTypes(mlir::MLIRContext *context) {
...
+  ConvertFuncSignature(mlir::TypeConverter &typeConverter,
+                       mlir::MLIRContext *context)
+      : mlir::OpConversionPattern<mlir::FuncOp>(typeConverter, context) {}
+
+  mlir::LogicalResult
+  matchAndRewrite(mlir::FuncOp op, mlir::ArrayRef<mlir::Value> operands,
+                  mlir::ConversionPatternRewriter &rewriter) const override {
```
earhart, line 125: This method should be `final`

earhart (changes requested):
Is there any reason not to use `mlir::populateFuncOpTypeConversionPattern(patterns, context, typeConverter);` for this, using a type converter that returns its input for any non-`comp` type?  (And then you can use that for the dynamic legality check, too -- something like,

      target.addDynamicallyLegalOp<mlir::FuncOp>([&](mlir::FuncOp op) -> bool {
        return typeConverter.isSignatureLegal(op.getType());
      });

)

pmlc/conversion/comp_to_llvm/common.cc
kdobros: Done

pmlc/conversion/comp_to_llvm/common.cc
kdobros: Done

pmlc/conversion/comp_to_llvm/common.cc
kdobros: This pattern requires type converter, so I don't want to also pick up constructor that sets it to nullptr.

kdobros:
> Is there any reason not to use `mlir::populateFuncOpTypeConversionPattern(patterns, context, typeConverter);` for this, using a type converter that returns its input for any non-`comp` type? (And then you can use that for the dynamic legality check, too -- something like,
> 
> ```
>   target.addDynamicallyLegalOp<mlir::FuncOp>([&](mlir::FuncOp op) -> bool {
>     return typeConverter.isSignatureLegal(op.getType());
>   });
> ```
> 
> )

That's a good point, I somehow rejected it after seeing two type converters. I changed it and now it looks much cleaner, there is no duplication of such simple functionality as converting function signature.

earhart (approved):

PR #1429, opened 2020-10-02
Fix potential bug if outStrides is empty in pad_range

shiyao00:


pmlc/dialect/tile/transforms/pad_ranges.cc
```diff
@@ -79,7 +79,7 @@ void PadRangesPass::runOnFunction() {
     }
     // Get strides for output
     auto outStrides = util::computeStrideArray(op.getResultType(), op.sink());
-    if (!outStrides) {
+    if (!outStrides || outStrides->strides.size() == 0) {
```
flaub, line 82: Use `strides.empty()`. I feel like maybe we should make `util::computeStrideArray` consistently return either nothing or an empty array but not both.  @jbruestle ?

shiyao00:
For normalize, the output is zero dimensional. In this case, it accesses outStrides->strides[i] later which is a bug.

jbruestle (changes requested):
This is a bug, but this is the wrong bugfix.  The bug is here: https://github.com/plaidml/plaidml/blob/plaidml-v1/pmlc/util/strides.cc#L62.  
StrideArray(0) should be StrideArray(map.getNumDims(), 0).

flaub (approved):

PR #1432, opened 2020-10-05
Fix explicit_padding infinite padding value bug

dgkutnic:


plaidml/op/lib/ops.cc
```diff
@@ -1659,9 +1659,8 @@ Value explicit_padding(const Value& value) {
       IVLOG(2, "Constant padding requested");
 
       auto padval = args[4].as_tensor();
-      I = I - padval;
       O(O_idxs) = I(I_idxs);
-      O = O + padval;
+      O.use_default(padval);
```
tzerrell, line 1663: FYI, this works here because it's an assignment aggregation. But as a warning for elsewhere, `use_default` sets the initial value -- so if it's aggregated with e.g. sum and initialized to neg_inf, in the non-padding region this would just add a bunch of stuff to neg_inf, which isn't going to work. Just an FYI though, this is 100% OK here b/c of the assignment agg_op.

tzerrell (commented):
LGTM

plaidml/op/op.h
```diff
@@ -331,7 +331,7 @@ inline edsl::Tensor elu(const edsl::Tensor& I, double alpha) {
 class explicit_padding {
  public:
   explicit explicit_padding(const edsl::Tensor& I, const std::vector<int>& lo_pads, const std::vector<int>& hi_pads)
-      : I_(I), lo_pads_(lo_pads), hi_pads_(hi_pads), mode_(PadMode::CONSTANT), padval_(0) {}
+      : I_(I), lo_pads_(lo_pads), hi_pads_(hi_pads), mode_(PadMode::CONSTANT), padval_(edsl::Constant(0)) {}
```
flaub, line 334: Is it nicer to say `Tensor(0)` or to say `Constant(0)`? I've been internally debating whether I should drop the `Constant()` for scalars.

dgkutnic: I don't have a strong opinion here, but I thought we were trying to avoid exposing the `Tensor` API based on this comment. Perhaps I am misunderstanding what you were trying to say?

> Dropped the plaidml::edsl::Tensor constructors that allow buliding of placeholders and constants. Use plaidml::edsl::Placeholder() and plaidml::edsl::Constant() instead.

flaub: That comment was referring to Tensors/Buffers, not scalars.

tzerrell (approved):

plaidml/op/op.h
dgkutnic: @tzerrell that is a good point. Using `Constant` for scalars would also prevent a situation where someone has to think "Oh, I can just use `Constant` to store my weights, except for scalars, where I'd have to use `Tensor`."

PR #1434, opened 2020-10-06
Use uniform ABI for all runtimes

earhart:


earhart:
Updated to simplify things a bit (there's some scaffolding, but it can be removed in the next PR).  PTAL?

earhart:
After simplifying things, this PR was looking a little light; here's the next phase (which also removes the scaffolding).

pmlc/transforms/make_entrypoints.cc
```diff
@@ -0,0 +1,160 @@
...
+    signalPassFailure();
+  }
+
+  return entryFunc;
+}
+
+void MakeEntrypointsPass::makeSetup(mlir::OpBuilder &builder) {
+  // This method builds:
+  //
+  //   Device* setup(Device* device) {
```
flaub, line 86: What's the return-value for?

pmlc/util/ids.h
```diff
@@ -0,0 +1,13 @@
+// Copyright 2020 Intel Corporation
+#pragma once
+
+// A collection of well-known identifiers used in codegen.
+
+namespace pmlc {
```
flaub, line 6: `pmlc::util`

pmlc/transforms/passes.td
```diff
@@ -0,0 +1,19 @@
+#ifndef __PMLC_TRANSFORM_PASSES__
```
flaub, line 1: Not sure this belongs in a top-level `transforms` directory. Seems like it's `rt`-specific

pmlc/rt/jit_executable.cc
```diff
@@ -428,7 +458,8 @@ class JitExecutable final : public Executable {
     descriptors.clear();
     ptrs.clear();
 
-    std::copy(preParams.begin(), preParams.end(), std::back_inserter(ptrs));
+    network = abi.setupFunc(this->device.get());
```
flaub, line 461: We need to pass all the user arguments to the setup function also.

flaub: Actually, just the constant buffers need to be passed

pmlc/transforms/make_entrypoints.cc
earhart: Added a comment.

pmlc/util/ids.h
earhart: Done.  (Seems unnecessary to me, tho -- lots of `util::` sprinkled everywhere for no obvious value.)

pmlc/transforms/passes.td
earhart: Subdirectories of `rt` are runtimes, and the pass is used by the target pipelines, not`rt` itself -- it's a way to implement the expected `rt` ABI, and not even the only way.

pmlc/rt/jit_executable.cc
earhart: Definitely.  But I'd rather do that in a followup PR.

pmlc/util/ids.h
```diff
@@ -3,11 +3,11 @@
...
 // A collection of well-known identifiers used in codegen.
 
-namespace pmlc {
+namespace pmlc::util {
 
 static constexpr const char kEntrypoint[] = "main";
 static constexpr const char kPlaidmlInit[] = "plaidml_init";
-static constexpr const char kPlaidmlExecute[] = "plaidml_exec";
+static constexpr const char kPlaidmlExec[] = "plaidml_exec";
 static constexpr const char kPlaidmlFini[] = "plaidml_fini";
```
flaub: I feel like these don't belong in `util`? Maybe these are `rt` specific?

pmlc/util/ids.h
flaub: It's just consistency with the directory structure. I'm fine with flattening the namespace so it's not so deep, but I think we should do that with buy-in from others and that should be its own NFC PR.

pmlc/ast/builder.cc
```diff
@@ -38,8 +39,6 @@ namespace tile = pmlc::dialect::tile;
 
 namespace {
 
-static constexpr const char *kEntrypoint = "main";
```
flaub, line 41: Not sure we should move this, seems like it's only local to the builder.

pmlc/rt/jit_executable.cc
flaub: I think we should call `setupFunc` in the constructor, not in `bindArguments`, but I suppose that can wait in a follow up.

pmlc/transforms/make_entrypoints.cc
```diff
@@ -0,0 +1,160 @@
...
+  auto ctx = &getContext();
+  llvmInt32Type = LLVMType::getInt32Ty(ctx);
+  llvmPtrTy = LLVMType::getInt8PtrTy(ctx);
+  llvmVoidTy = LLVMType::getVoidTy(ctx);
+}
+
+mlir::FuncOp MakeEntrypointsPass::getNetworkEntry() {
+  mlir::FuncOp entryFunc;
+
+  getOperation().walk([&](mlir::FuncOp func) {
```
flaub, line 58: Don't use walk, use `getOps<FuncOp>` so that we don't look for recursive functions.

pmlc/transforms/make_entrypoints.cc
```diff
@@ -0,0 +1,160 @@
...
+  mlir::FuncOp entryFunc = getNetworkEntry();
+  if (!entryFunc) {
+    return;
+  }
+
+  initMembers();
+  mlir::OpBuilder builder(getOperation().getBody()->getTerminator());
+  makePlaidmlInit(builder);
+  makePlaidmlExecute(builder, entryFunc);
+  makePlaidmlFini(builder);
```
flaub, line 45: We should check to see if these functions are already defined or at least fail if they're already defined.

pmlc/transforms/make_entrypoints.cc
```diff
@@ -0,0 +1,160 @@
...
+  //
+  //   Network* plaidml_init(Device* device) {
+  //     return device;  // Network == Device for now.
+  //   }
+  //
+  // This is a rather trivial function; the plan is to expand it later.
+  // In particular, the Network parameter gives networks a mechanism to carry
+  // additional information from plaidml_init to plaidml_exec and
+  // plaidml_fini.
+  auto func = builder.create<mlir::FuncOp>(
```
flaub, line 91: I'd expect to be operating in the LLVM dialect and to create `llvm::FuncOp` instead of `mlir::FuncOp`

pmlc/transforms/make_entrypoints.cc
```diff
@@ -0,0 +1,160 @@
...
+    entryFunc = func;
+    return mlir::WalkResult::advance();
+  });
+  if (!entryFunc) {
+    getOperation().emitError(
+        "Expected a single entrypoint function definition");
+    signalPassFailure();
+  }
+
+  return entryFunc;
```
flaub, line 77: Why don't we just use a symbol lookup?

pmlc/transforms/passes.td
```diff
@@ -0,0 +1,19 @@
...
+#define __PMLC_TRANSFORM_PASSES__
+
+include "mlir/Pass/PassBase.td"
+
+def MakeEntrypointsPass : Pass<"pmlc-make-entrypoints", "mlir::ModuleOp"> {
```
flaub, line 6: It seems like this pass should operate at the llvm dialect level, so I'd call it `llvm-make-entrypoints` or perhaps some better name (can't think of one right now).

pmlc/transforms/make_entrypoints.cc
flaub: Alternatively, I could see this split happening at PXA dialect or comp dialect level, and then having a straightforward legalization to convert high-level `mlir::FuncOp` into low level `llvm::FuncOp` (converting a `Network` data type in to a `void*` for instance).

pmlc/rt/jit_executable.cc
```diff
@@ -388,26 +418,28 @@ class JitExecutable final : public Executable {
...
     }
 
-    jitEntry = impl->compile(std::move(llvmModule), std::move(ctx), entryPoint);
-    if (!jitEntry) {
-      throw std::runtime_error("jitEntry function is null");
+    abi = impl->compile(std::move(llvmModule), std::move(ctx));
+    if (!abi) {
+      throw std::runtime_error("Entrypoint functions not found");
     }
   }
```
flaub, line 431: Do reduce duplication, let's build up the ABI here, after `impl->compile` so it's part of the common code path.

plaidml/edsl/tests/BUILD
```diff
@@ -29,7 +29,7 @@ plaidml_cc_test(
         "//plaidml:intel_gen_ocl_spirv": [":edsl_test-skip.intel_gen_ocl_spirv.txt"],
         "//conditions:default": [],
     }),
-    linkstatic = True,
+    # linkstatic = False,
```
flaub, line 32: Probably need to undo this

pmlc/conversion/abi_to_llvm/abi_to_llvm.cc
```diff
@@ -0,0 +1,429 @@
...
+#include "mlir/IR/BlockAndValueMapping.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/Pass/Pass.h"
+
+#include "pmlc/conversion/abi_to_llvm/passes.h"
+#include "pmlc/dialect/abi/ir/dialect.h"
+#include "pmlc/util/ids.h"
+
+// #pragma GCC diagnostic ignored "-Wunused-function"
+// #pragma GCC diagnostic ignored "-Wunused-variable"
```
flaub, line 17: I don't think this is supported on all compilers, why do we need these?

pmlc/conversion/abi_to_llvm/abi_to_llvm.cc
```diff
@@ -0,0 +1,429 @@
...
+    builder.setInsertionPointToStart(moduleOp.getBody());
+    func =
+        builder.create<LLVM::LLVMFuncOp>(builder.getUnknownLoc(), name, funcTy);
+  }
+  return func;
+}
+
+namespace {
+
+constexpr char emptyNetworkSingletonName[] = "plaidml_empty_network";
```
flaub, line 40: Rename to `kNetworkSingletonName`

pmlc/conversion/abi_to_llvm/input.mlir
```diff
@@ -0,0 +1,55 @@
+Converting fn: llvm.func @main(%arg0: !llvm.ptr<float>, %arg1: !llvm.ptr<float>, %arg2: !llvm.i64, %arg3: !llvm.i64, %arg4: !llvm.i64, %arg5: !llvm.i64, %arg6: !llvm.i64, %arg7: !llvm.ptr<float>, %arg8: !llvm.ptr<float>, %arg9: !llvm.i64, %arg10: !llvm.i64, %arg11: !llvm.i64, %arg12: !llvm.i64, %arg13: !llvm.i64, %arg14: !llvm.ptr<float>, %arg15: !llvm.ptr<float>, %arg16: !llvm.i64, %arg17: !llvm.i64, %arg18: !llvm.i64, %arg19: !llvm.i64, %arg20: !llvm.i64) {
```
flaub, line 1: Can we make this a lit test?

pmlc/dialect/abi/ir/dialect.h
```diff
@@ -0,0 +1,84 @@
...
+using mlir::StringAttr;
+using mlir::StringRef;
+using mlir::Type;
+using mlir::TypeAttr;
+using mlir::ValueRange;
+using mlir::VectorType;
+
+namespace MemoryEffects = mlir::MemoryEffects;
+namespace SideEffects = mlir::SideEffects;
+*/
```
flaub, line 68: Drop comments

pmlc/dialect/abi/ir/loop_op.cc
```diff
@@ -0,0 +1,118 @@
...
+      auto definingOp = val.getDefiningOp();
+      if (definingOp && definingOp->hasTrait<mlir::OpTrait::ConstantLike>()) {
+        // N.B. Constant-Like operations have no side-effects, exactly one
+        //      result, and no operands.
+        auto *denormOp = definingOp->clone();
+        rewriter.insert(denormOp);
+        auto arg = loopOp.bodyEntryBlock()->getArgument(idx);
+        arg.replaceAllUsesWith(denormOp->getResult(0));
+        loopOp.bodyEntryBlock()->eraseArgument(idx);
+        networkOp.getOperation()->eraseOperand(idx);
```
flaub, line 37: Use `m_Constant` rather than looking for `ConstantLike`

tools/conda_run/conda_unix.cc
```diff
@@ -53,9 +53,9 @@ int main(int argc, char* argv[]) {
 
 #ifdef DEBUG
     // Useful debugging code
-    std::cout << "argc: " << argc << std::endl;
+    std::cerr << "argc: " << argc << std::endl;
```
flaub, line 56: Probably doesn't belong in this PR

pmlc/rt/jit_executable.cc
```diff
@@ -411,82 +360,105 @@ class JitExecutable final : public Executable {
...
+      throw std::runtime_error(
+          "Program outputs arguments and buffers mismatch");
+    }
     StopWatch stopWatch;
     if (VLOG_IS_ON(1)) {
       stopWatch.start();
     }
-    bindArguments(inputBuffers, outputBuffers);
-    abi.executeFunc(ptrs.data());
+    auto memrefIt = memrefs.begin();
```
flaub, line 413: Rename to `itMemref`

pmlc/rt/jit_executable.cc
```diff
@@ -73,89 +73,41 @@ static void setupTargetTriple(llvm::Module *llvmModule) {
...
-}
-
 namespace {
 
-class MemRefDescriptor {
-private:
-  struct Base {
+class MemRef {
+public:
+  struct Descriptor {
```
flaub, line 80: What do we gain by refactoring as such? This seems confusing to me, we want to manipulate a `MemRefDescriptor` not a `MemRef` itself (a memref is just a logical concept).

earhart:
FWIW -- this is currently just a draft PR.  Please feel free to ignore it; if you do want to review, please focus on major architectural issues, since there are many, many small things to be cleaned up before it's ready to submit.

flaub:
SGTM, however it's hard for me to grok what the pass is doing without a lit test. That's probably the first thing I would try to understand when doing an architectural review.

earhart:
+1 -- as things get a little closer to being nailed down, I'll definitely be adding some lit tests.  :-)

earhart:
(I'm actually not all that concerned about the overall architecture -- it's basically the same as what we discussed in the doc review.)

pmlc/dialect/abi/ir/loop_op.cc
earhart: Just wondering -- what's wrong with `ConstantLike`?

flaub: Nothing, it's just that `m_Constant` is probably a much more concise way of matching a constant op. But if it's unsuitable, obviously this way is fine :)

flaub:
Is it possible to split up this PR so that the hoisting is done in a follow up? We're going to need to switch to a new task soonish.

earhart:
> Is it possible to split up this PR so that the hoisting is done in a follow up? We're going to need to switch to a new task soonish.

Definitely possible.

pmlc/dialect/abi/ir/loop_op.cc
```diff
@@ -0,0 +1,184 @@
...
+
+void LoopOp::getCanonicalizationPatterns(
+    mlir::OwningRewritePatternList &patterns, mlir::MLIRContext *ctx) {
+  patterns.insert<DenormalizeConstantsPattern>(ctx);
+  patterns.insert<RemoveUnusedNetworkFieldsPattern>(ctx);
+}
+
+mlir::Region &LoopOp::getLoopBody() { return bodyRegion(); }
+
+bool LoopOp::isDefinedOutsideOfLoop(mlir::Value value) {
```
flaub, line 103: You can probably use `mlir/Transforms/RegionUtils.h`, specifically the `getUsedValuesDefinedAbove` or `getUsedValuesDefinedAbove` helper functions.

earhart: Awesome -- thanks!

flaub:
Instead of writing our own hoisting pass, I think we can use the existing LICM pass that is defined upstream. I realize that your pass also is splitting allocs and deallocs. But I think this could be done in a different pass. I believe we already have a pass that injects deallocs for each alloc. So ideally, this hoisting would occur before we get to the point of adding the deallocs.

earhart:
> Instead of writing our own hoisting pass, I think we can use the existing LICM pass that is defined upstream. I realize that your pass also is splitting allocs and deallocs. But I think this could be done in a different pass. I believe we already have a pass that injects deallocs for each alloc. So ideally, this hoisting would occur before we get to the point of adding the deallocs.

@flaub Initially, this code did use the existing LICM pass.  That pass wound up having two limitations: it wasn't able to move operations to "after the loop" (LoopLike just doesn't define a mechanism for it), and it doesn't move operations with memory effects.

Any post-processing to inject de-allocs would have to know to inject them into the correct location of the fini region of the loop; it's not as trivial as one might like.

Another solution might be to require the execution environment to implicitly destroy kernels and allocs.  That was pretty roundly shot down in the design review, but maybe now that the complexity's a little more obvious, we could go back to that approach?  OTOH, we'd still need to do our own thing for operations with memory effects (or just pretend that the relevant operations don't have effects), and it'd be strictly less general than the existing code.

My own thinking is that we should probably aim to contribute this logic upstream; code review there may suggest additional simplifications.

flaub:
I basically think we should make this a later problem to solve, because we might come up with different solutions over time with varying degrees of simplicity. To me, hoisting is an optimization, not a requirement for legalization/lowering. So I think we should just drop all the hoisting stuff for now and leave that as a follow up PR (or series of PRs).

earhart:
We should definitely split this PR up.

I agree that hoisting is an optimization, not a requirement for legalization and lowering; OTOH, performance is its own requirement.

flaub:

PR #1446, opened 2020-10-10
Adjusted the loop unroll factor

LLITCHEV:
Experimentally this is much better than the factor of 6 by about 27%


flaub:
Where does the number `213` come from?

LLITCHEV:
@flaub Bunch of experimentation with resnet50 network. For Gen-9 that seems to be the point where the code size starts t deminish the benefit of the loop unrolling.

jbruestle:
213 seems like the right order of magnitude based on prior experience, but I imagine a 'rounder' number might create less complex trailing loops.  Maybe that's not true?  How does 224=7*2^5 do?

LLITCHEV:
@jbruestle The 213 had a little better results than numbers above - on average. On the 100s of rtuns, it was very close, but a tiny little bit (in general the numbers very quite a bit). I adjusted it to be 256, since the difference is so small, and potentially on other networks the multiple of power of 2 can be beneficial.

jbruestle (approved):

PR #1447, opened 2020-10-10
Specify IO buffers at invocation time

flaub:
This can potentially eliminate an extra buffer copy per invocation.


jbruestle (commented):
Looks good to me, but seems to be failing CI...

jbruestle (approved):
Looks good.

PR #1449, opened 2020-10-12
Drop simplify/no_reduce from contractions

flaub:


pmlc/dialect/tile/transforms/contraction.cc
```diff
@@ -662,7 +661,7 @@ BoundsAndConstraints Contraction::ComputeBounds(ArrayRef<Shape> shapes,
   GatherConstraints(shapes);
   IVLOG(3, "Constraints:" << to_string(range_constraints.constraints));
   // Reduce if needed
-  if (NeedReduce() && !no_reduce) {
+  if (NeedReduce()) {
```
jbruestle, line 664: I think we need to also skip reduce for assignment contractions, or else we will damage perf for some cases.  Specifically, you may want to look at the code for:
O[i * 10 + j] = A[i, j], i < 10, j < 10, with O = tensor<100xfp> and A = tensor<10x10xfp>
If we don't skip reduce on assignment, I think this turns into a 10x10x10 inner loop or something of the sort, whereas it should only be 10x10.  Probably easy to just add a check for assignment reduction in NeedReduce, but might be good to verify it's broken as is and add a lit test (maybe verifying two affine parallels after reduction to PXA).

jbruestle (changes requested):
See inline comment.

jbruestle (approved):
Sweet.

PR #1452, opened 2020-10-12
[comp] Fix execenv_coalescing access after erase

kdobros:


jbruestle (approved):
Yup, see the bug and the fix.

PR #1453, opened 2020-10-12
[rfc] Setup/Teardown Plan

earhart:
Here's the first draft of what Frank and I discussed over the long weekend.  (NB This is still a work-in-progress; I need to go back through and make sure I've captured everything from the original discussion.)


flaub (changes requested):
I don't think we should commit RFCs

designs/RFC - Network Setup Teardown Optimization.md
```diff
@@ -0,0 +1,396 @@
...
+
+### Wait Cleanup
+
+Although it's not strictly necessary for setup/teardown optimization, the code's a bit more optimal (and easier to grok)
+if we coalesce `comp.wait` operations into subsequent operations that depend on the values touched by the `comp.wait`;
+we can also optimize a bit by eliminating redundant memory transfers.
+
+(Note that since we don't have information about which values
+are kernel inputs and which are kernel outputs, kernels reading the same data will be forced to serialize; fixing this
+is an important optimization, but not one we're considering in this document).
```
flaub, line 256: We can determine which are inputs vs outputs by doing analysis. I don't think we need a special marker from the user.

designs/RFC - Network Setup Teardown Optimization.md
```diff
@@ -0,0 +1,396 @@
...
+  %1 = comp.alloc %0 : (!comp.execenv<ocl:0,(11)>) -> memref<8x32xf32, 11>
+  comp.execenv_set_alloc %0, %1 {midx = 0} : !comp.execenv<ocl:0,(11)>, memref<8x32xf32, 11>
+  %4 = comp.alloc %0 : (!comp.execenv<ocl:0,(11)>) -> memref<16x32xf32, 11>
+  comp.execenv_set_alloc %0, %4 {midx = 1} : !comp.execenv<ocl:0,(11)>, memref<16x32xf32, 11>
+  %5 = comp.alloc %0 : (!comp.execenv<ocl:0,(11)>) -> memref<8x16xf32, 11>
+  comp.execenv_set_alloc %0, %5 {midx = 2} : !comp.execenv<ocl:0,(11)>, memref<8x16xf32, 11>
+  %k1 = comp.create_kernel %0 {kernel = @main_kernel::@main_kernel} : (!comp.execenv<ocl:0,(11)>) -> !comp.kernel<ocl, 11, 8x32xf32>
+  comp.execenv_set_kernel %0, %k1 {kidx = 0} : !comp.execenv<ocl:0,(11)>, !comp.kernel<ocl, 11, 8x32xf32>
+  %k2 = comp.create_kernel %0 {kernel = @main_kernel_0::@main_kernel} : (!comp.execenv<ocl:0,(11)>) -> !comp.kernel<ocl, 11, 16x32xf32, 8x16xf32, 8x32xf32>
+  comp.execenv_set_kernel %0, %k2 {kidx = 1} : !comp.execenv<ocl:0,(11)>, !comp.kernel<ocl, 11, 16x32xf32, 8x16xf32, 8x32xf32>
```
flaub, line 315: I think we should return these values instead of 'setting' them onto the environment. We'll then need to inject these values into the parameters of `plaidml_exec`

designs/RFC - Network Setup Teardown Optimization.md
```diff
@@ -0,0 +1,396 @@
...
+  return %0
+}
+
+func @plaidml_exec(%arg0: !comp.execenv<ocl:0,11>, %arg1: memref<8x16xf32>, %arg2: memref<16x32xf32>, %arg3: memref<8x32xf32>) {
+  %c2 = constant 2 : index
+  %c32 = constant 32 : index
+  %c1 = constant 1 : index
+  %c8 = constant 8 : index
+  %c4 = constant 4 : index
+  %1 = comp.execenv_get_alloc %arg0 {midx = 0}: (!comp.execenv<ocl:0,(11)>) -> memref<8x32xf32, 11>
```
flaub, line 325: I thought we were going to use values instead of using `midx` and `kidx`?

designs/RFC - Network Setup Teardown Optimization.md
earhart: Oh, good point -- just to clarify, do you mean by analyzing the SPIR-V code?

designs/RFC - Network Setup Teardown Optimization.md
earhart: Hmm... that's tricky; what do you lower to?  (I mean, what's the ABI even look like?)  I have another idea for how we could do this, though, by creating another type we can lower to an LLVMIR struct -- just a sec, I'll write it up.

designs/RFC - Network Setup Teardown Optimization.md
earhart: Misunderstand on my end; I thought it was just the way the code was using those to smuggle values.  Again, I think I can eliminate it; it'll be in the next update.

earhart:
> I don't think we should commit RFCs

NP -- what do you think the process for this should look like?  It does seem useful to be able to use standard code review tools for doc reviews... maybe we could use another repository?

designs/RFC - Network Setup Teardown Optimization.md
flaub: Best to probably do it during PXA, because we have interfaces for determining read/write and being able to follow the chain.

designs/RFC - Network Setup Teardown Optimization.md
flaub: I think we can pass allocations as `memref` directly, and I think we can add a new type for kernels.

flaub:
Yeah, I was hoping we could stand up a discourse forum for this purpose. Using another repo would be an alternative.

designs/RFC - Network Setup Teardown Optimization.md
flaub: I think the lowering for kernels could actually turn into an `int32` or whatnot, but that should only be at the LLVM dialect level I would think.

flaub: I think we should avoid using LLVM structs unless we have a real good reason to use one.

designs/RFC - Network Setup Teardown Optimization.md
earhart: Done

designs/RFC - Network Setup Teardown Optimization.md
earhart: PTAL

kdobros (commented):
Overall concept looks good to me, only some comments to implementation details.

designs/RFC - Network Setup Teardown Optimization.md
```diff
@@ -0,0 +1,381 @@
...
+  %c4 = constant 4 : index
+  %0 = comp.create_execenv %arg0 : (!comp.device) -> !comp.execenv<ocl:0,(11)>
+  %1 = comp.alloc %0 : (!comp.execenv<ocl:0,(11)>) -> memref<8x32xf32, 11>
+  %4 = comp.alloc %0 : (!comp.execenv<ocl:0,(11)>) -> memref<16x32xf32, 11>
+  %5 = comp.alloc %0 : (!comp.execenv<ocl:0,(11)>) -> memref<8x16xf32, 11>
+  %k1 = comp.create_kernel %0 {kernel = @main_kernel::@main_kernel} : (!comp.execenv<ocl:0,(11)>) -> !comp.kernel<ocl, 11, 8x32xf32>
+  %k2 = comp.create_kernel %0 {kernel = @main_kernel_0::@main_kernel} : (!comp.execenv<ocl:0,(11)>) -> !comp.kernel<ocl, 11, 16x32xf32, 8x16xf32, 8x32xf32>
+  comp.loop(%arg1: memref<8x16xf32>, %arg2: memref<16x32xf32>, %arg3: memref<8x32xf32>) {
+    %w1 = comp.schedule_write %arg3 to %1 on %0 : (!comp.execenv<ocl:0,(11)>, memref<8x32xf32>, memref<8x32xf32, 11>) -> !comp.event<ocl>
+    %2 = comp.launch_kernel %k1, %c4, %c1, %c1, %c1, %c2, %c32, %1 wait for %w1: (!comp.kernel<ocl, 11, 8x32xf32>, index, index, index, index, index, index, memref<8x32xf32, 11>, !comp.event<ocl>) -> !comp.event<ocl>
```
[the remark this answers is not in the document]
flaub: Actually, multiple variadic operands are fine, you can use this trait: `AttrSizedOperandSegments`, but the custom assembly format should probably be split like you suggest (but can still be done in a single op)

designs/RFC - Network Setup Teardown Optimization.md
```diff
@@ -0,0 +1,381 @@
...
+a bidirectional request/response pipe, and pass it to `comp.loop`.  Our sense is that it's more obvious to split the
+current `main()` function into multiple functions invoked individually to set up the network, run it, and tear it down.
+
+To do this, we add a new `!comp.network` to serve as a resource container, and add construction and deconstruction
+operations for it; this allows us to return a single value to the caller to represent the constructed network, and in
+the lowering to LLVMIR, we can trivially replace it with a pointer to a heap-allocated well-typed struct.  For our
+example code, this transformation looks something like this:
+
+```mlir
+func @plaidml_init(%arg0: !comp.device) -> !comp.network {
```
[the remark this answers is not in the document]
flaub: The intent is to eventually have the init function do constant buffer folding and perhaps other things like doing re-layouts of weights.

designs/RFC - Network Setup Teardown Optimization.md
```diff
@@ -0,0 +1,381 @@
...
+### Final Lowering
+
+With a minor tweak, we could run the above code as a coroutine &mdash; we could pass in an extra parameter to represent
+a bidirectional request/response pipe, and pass it to `comp.loop`.  Our sense is that it's more obvious to split the
+current `main()` function into multiple functions invoked individually to set up the network, run it, and tear it down.
+
+To do this, we add a new `!comp.network` to serve as a resource container, and add construction and deconstruction
+operations for it; this allows us to return a single value to the caller to represent the constructed network, and in
+the lowering to LLVMIR, we can trivially replace it with a pointer to a heap-allocated well-typed struct.  For our
+example code, this transformation looks something like this:
```
flaub, line 300: What's the advantage of `!comp.network`? I think we'll find that it's often easier to pass each value standalone, not contained within a 'struct'.

designs/RFC - Network Setup Teardown Optimization.md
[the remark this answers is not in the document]
flaub: Like:
```
%2 = comp.launch_kernel %env, args(%kernel_set, %c1, %c1, %c1, %c8, %c4, %c1), events(%w1)
    : (!comp.execenv<ocl:0,(11)>, !comp.kernel<ocl>, index, index, index, index, index, index,
       !comp.event<ocl>) -> !comp.event<ocl>
```

designs/RFC - Network Setup Teardown Optimization.md
earhart: I'm thinking the thing doing the passing is `JitExecutable`; in its constructor, it's running `plaidml_init`, getting some value back to hold the network, and then using that value for its `plaidml_exec` calls and then passing it to `plaidml_fini` in its destructor.  It can't directly handle a variable number of return values and arguments.  So something, somewhere, on the JITted code side, needs to be boxing the variables being passed from init to exec and fini.  Given that, I think it's simplest to have an explicit box at the comp dialect, and lower it to a heap-allocated LLVMIR struct.

I'm definitely open to other ideas, though -- maybe there's some MLIR facility for handling this sort of thing?  Some other good pattern people are using?

designs/RFC - Network Setup Teardown Optimization.md
```diff
@@ -0,0 +1,381 @@
...
+    %2 = comp.launch_kernel %k1, %c4, %c1, %c1, %c1, %c2, %c32, %1 : (!comp.kernel<ocl, 11, 8x32xf32>, index, index, index, index, index, index, memref<8x32xf32, 11>) -> !comp.event<ocl>
+    %3 = comp.schedule_read %arg3 from %1 on %0 wait for %2 : (memref<8x32xf32>, memref<8x32xf32, 11>, !comp.execenv<ocl:0,(11)>, !comp.event<ocl>) -> !comp.event<ocl>
+    comp.wait %3 : !comp.event<ocl>
+    %w4 = comp.schedule_write %arg2 to %4 on %0 : (!comp.execenv<ocl:0,(11)>, memref<16x32xf32>, memref<16x32xf32, 11>) -> !comp.event<ocl>
+    comp.wait %w4 : !comp.event<ocl>
+    %w5 = comp.schedule_write %arg1 to %5 on %0 : (!comp.execenv<ocl:0,(11)>, memref<8x16xf32>, memref<8x16xf32, 11>) -> !comp.event<ocl>
+    comp.wait %w5 : !comp.event<ocl>
+    %6 = comp.schedule_write %arg3 to %1 on %0 : (memref<8x32xf32>, memref<8x32xf32, 11>, !comp.execenv<ocl:0,(11)>) -> !comp.event<ocl>
+    comp.wait %6 : !comp.event<ocl>
+    %7 = comp.launch_kernel %k2, %c1, %c1, %c1, %c8, %c4, %c1, %4, %5, %1 : (!comp.kernel<ocl, 11, 16x32xf32, 8x16xf32, 8x32xf32>, index, index, index, index, index, index, memref<16x32xf32, 11>, memref<8x16xf32, 11>, memref<8x32xf32, 11>) -> !comp.event<ocl>
```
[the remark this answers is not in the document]
earhart: SGTM

designs/RFC - Network Setup Teardown Optimization.md
```diff
@@ -0,0 +1,381 @@
...
+  %c2 = constant 2 : index
+  %c32 = constant 32 : index
+  %c1 = constant 1 : index
+  %c8 = constant 8 : index
+  %c4 = constant 4 : index
+  %0 = comp.create_execenv %arg0 : (!comp.device) -> !comp.execenv<ocl:0,(11)>
+  %1 = comp.alloc %0 : (!comp.execenv<ocl:0,(11)>) -> memref<8x32xf32, 11>
+  %4 = comp.alloc %0 : (!comp.execenv<ocl:0,(11)>) -> memref<16x32xf32, 11>
+  %5 = comp.alloc %0 : (!comp.execenv<ocl:0,(11)>) -> memref<8x16xf32, 11>
+  %k1 = comp.create_kernel %0 {kernel = @main_kernel::@main_kernel} : (!comp.execenv<ocl:0,(11)>) -> !comp.kernel<ocl, 11, 8x32xf32>
```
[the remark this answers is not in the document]
earhart: I think I'll just remove memory space for now, then.

designs/RFC - Network Setup Teardown Optimization.md
earhart: @jbruestle and I just had a long chat about ways to improve this: we think we've found a way to get rid of the !comp.network type for passing values between the init and exec/fini functions, without running into all the complexity I was seeing in the lowering and execution.  I'm working on a revision.

earhart: PTAL?

earhart:
FYI -- just had a chat w/@flaub to get some feedback; we came up with a minor tweak that makes this approach work for non-comp-dialect code as well.  More changes incoming.  :-)

earhart:
FYI: This turned out to have problems.  Most notably, when you return memrefs as values, the compiled code literally returns memrefs as values (unsurprisingly), which makes it hard to modify them afterwards to fill in the actual desired contents.

I'm currently experimenting with some ideas to fix this (while keeping in mind the goals of simplicity and being able to handwrite MLIR in various dialects and get it executed); I'll update the doc once I have something that seems to work.

jbruestle:
Re the return memref vs take outbuf bit, see the PXA -> affine (https://github.com/plaidml/plaidml/blob/plaidml-v1/pmlc/conversion/pxa_to_affine/pxa_to_affine.cc#L218) which basically does a transform to changes returned alloced memrefs into 'output buffers' which are passed in as memrefs.  It's actually a pretty easy transform (and we could actually extract it from that conversion into a reusable pattern).

flaub:

PR #1455, opened 2020-10-13
Add sort intrinsic to eDSL

flaub:
Future work: implement a lowering from the tile dialect.


tzerrell (commented):
I'm not sure whether it should go on this op or a separate op, but we are going to want a "partial sort" at some point for perf reasons (i.e. for `topk`), where only the `k` largest (or smallest) elements are placed in their sorted positions and the other elements are left in arbitrary positions.

This code looks good except for that omission, and if we'd rather deal with that as a separate PR I'll happily approve.

jbruestle (commented):
I have changes, so hold off for a bit, but I need to take care of a few other things first.

flaub:
Looks like #1517 will obsolete this one.

PR #1461, opened 2020-10-14
Fix float16 support

flaub:


plaidml/edsl/tests/edsl_test.cc
```diff
@@ -358,7 +360,48 @@ TEST_F(CppEdsl, Dot) {
...
+  }
+  std::vector<half> expected(M * N);
+  for (int i = 0; i < M; i++) {
+    for (int j = 0; j < N; j++) {
+      for (int k = 0; k < K; k++) {
+        expected[i * N + j] += in1[i * K + k] * in2[k * N + j];
+      }
+    }
+  }
+  // checkClose(program, {in1, in2}, {expected}, /*tolerance=*/1e-2);
```
flaub, line 404: @tzerrell This is failing even with tolerance of `1e-2` which seems bad. @jbruestle Does this mean we need to add a cast for the accumulation?

jbruestle: Hmm... so we are only accumulating 16 (K) values, and fp16 should have about 3 digits precision (i.e. 1e-3).  For such a small number, I wouldn't expect so much precision loss, especially on normally distributed data (more high variance data might have more loss  during exponent matching).  Of course we are accumulating 32*8 = 256 times, so maybe we get unlucky over so many accumulations?  But also, given that we are doing the 'comparison' in halfs as well, and I don't know why tile would be reordering the additions, I would expect we should probably get exact results here anyway, since the C++ code and the PXA code should be 1-to-1 (at least per output).  I would try to reduce the problem size and/or look at the numbers to try to better understand what exactly is up.  If it is just one or two outputs with bad precision out of 256, we might need to add a 'mostly close' funcition, which is like 95% are good to tolerance of 1e-2, and the remaining 5% are good to tolerance 1e-1 or some such.  Like checkMostlyClose(...., .95, 1e-2, 1e-1).

pmlc/dialect/pxa/transforms/stencil_gemm.cc
```diff
@@ -62,95 +61,50 @@ class StencilGEMM : public StencilBase {
...
+                     m_Op<AffineYieldOp>( //
+                         m_Capture(&reduce, m_PxaReduceOp(
+                                                AtomicRMWKind::addf,
+                                                m_Op<MulFOp>(m_Capture(&load1),
+                                                             m_Capture(&load2)),
+                                                matchers::m_Any())))) ||
+        matchPattern(yield,
+                     m_Op<AffineYieldOp>( //
+                         m_Capture(&reduce, m_PxaReduceOp(
+                                                AtomicRMWKind::addf,
```
jbruestle, line 76: I think this needs to be addi?

pmlc/dialect/xsmm/ir/ops.cc
```diff
@@ -21,29 +21,29 @@ void XSMMDialect::initialize() {
 }
 
 //
-// ---- GemmInvokeOp ----
+// ---- SGemmInvokeOp ----
```
jbruestle, line 24: What is SGemm here?  Also, maybe we should put renames (this/indicies/etc) in as separate NFC changes?

flaub: Sure. It's following the naming convention of xsmm (s = single, d = double, etc)

plaidml/edsl/tests/edsl_test.cc
tzerrell: So I've taken a look at the numbers in one of the failing elements, and the results are weird: Specifically, in the `i = 7, j = 18` element, we see PlaidML computing `-0.4755859375` and the raw loop computing `-0.45849609375` according to gtest's failure. (Incidentally, if I do the sum manually in f32, I get `-0.470967` suggesting PlaidML is more accurate in this case.) This is within the range I can get by manually reordering the summands: if I try my best to minimize the final result, I get `-0.481445` and if I try my best to maximize, I get `-0.453125`.

So far so good, but, if I 

1. compute and print the partial sums at each stage of the loop, or
2. manually accumulate values into a `half` with the same values in the same order as the loop

I get different partial sums and a different final answer! However, most of the partial sums _are_ the same, and where there are differences, the jumps are large. I would hypothesize loss of precision from printing the `half`s and then using the printed decimal form to initialize the `half`s for the manual test -- however, the errors are larger than should be possible from that (I'm seeing errors >20 units in the lowest-order printed decimal digit of the least precise input used, that shouldn't be possible in a single jump from a print-decimal-then-read-back-to-binary-of-same-precision error)... so I'm rather mystified as to what's happening in even this no-PlaidML case.

Raw partial sums from the two same-order cases if that's helpful:
Printing each partial sum from the loop:

```
2020-10-14 18:44:01,309 VERBOSE-1 [default] Value at i=7 j=18 k=0 is -1.07183
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -1.07129
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=1 is 0.715884
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -0.355225
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=2 is -0.985329
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -1.33984
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=3 is -2.23091
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -3.57031
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=4 is 0.379913
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -3.18945
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=5 is 0.403094
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -2.78516
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=6 is -0.047905
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -2.83203
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=7 is 0.0024303
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -2.82812
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=8 is 0.110359
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -2.7168
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=9 is -0.122547
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -2.83789
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=10 is -0.599677
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -3.4375
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=11 is 1.12196
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -2.31445
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=12 is 0.286131
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -2.02734
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=13 is 0.483816
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -1.54297
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=14 is 0.0442439
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -1.49805
2020-10-14 18:44:01,310 VERBOSE-1 [default] Value at i=7 j=18 k=15 is 1.0394
2020-10-14 18:44:01,310 VERBOSE-1 [default] ... partial sum is -0.458496
2020-10-14 18:44:01,310 VERBOSE-1 [default] Result at i=7 j=18 as computed by naive loops is -0.458496
```

Partial sums performed directly:
```
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -1.07129
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -0.355469
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -1.33984
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -3.57031
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -3.18945
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -2.78516
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -2.83203
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -2.82812
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -2.7168
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -2.83789
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -3.4375
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -2.31641
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -2.0293
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -1.54492
2020-10-14 18:55:13,786 VERBOSE-1 [default] Partial naive: -1.5
2020-10-14 18:55:13,786 VERBOSE-1 [default] Final naive: -0.460938
```

pmlc/dialect/xsmm/ir/ops.cc
flaub: Actually, this rename is important, because it includes the data type, which is actually one of the main parts of this PR.

.gdbinit
```diff
@@ -1,3 +1,2 @@
-target remote localhost:2345
```
flaub, line 1: If we leave this in, the initial connection doesn't always load symbols.

plaidml/edsl/tests/edsl_test-skip.intel_gen.txt
```diff
@@ -4,6 +4,7 @@ CppEdsl.BitNot
 CppEdsl.BitRightScalar
 CppEdsl.BitRightTensor
 CppEdsl.ConvI8
+CppEdsl.DotF16_AccF32
```
flaub, line 7: @jbruestle This is broken on `pml-vk-gen`, need someone to investigate further.

jbruestle (approved):

PR #1462, opened 2020-10-14
enable tiled fusions

abialas1:


abialas1:
I added some more checks and enabled tiled fusions by default. Unfortunately there are some issues with subgroup size 16 kernels that I would need to investigate further, however this does not affect resnet50 at this moment. If subgroup size 16 will be chosen the we will fallback to naive fusion, in other words there should be no negative perf impact of adding this in current shape.

jbruestle (approved):

PR #1463, opened 2020-10-14
Add a basic 'threading' pass for the CPU.

jbruestle:
Tiles the outermost loops (presumed to be thread safe via a prior tile-accumulate) for threading, and attaches the a tag for lowering.  Currently we always produce a single index loop to make lowering to OpenMP easier.  Once we having working integration and multicore support running, we may want to increase the sophistication of the OpenMP lowering and consider a more complex cost model.  Additionally, we should consider doing the tile accumulate during threading to prevent pass dependency, however, currently the autotile system does not correctly handle non-normalized affine.parallels, so we need to intersperse a normalize pass.


pmlc/dialect/pxa/transforms/cpu_thread.cc
```diff
@@ -0,0 +1,92 @@
...
+namespace {
+using llvm::DenseMap;
+using llvm::DenseSet;
+using llvm::SmallVector;
+
+// Pick the tiling that:
+// 1) Has at most one index
+// 2) Is as large as possible without going over maxThreads
+struct CostModel {
+  AffineParallelOp op;
```
flaub, line 38: This appears to be unused

pmlc/dialect/pxa/transforms/cpu_thread.cc
```diff
@@ -0,0 +1,92 @@
...
+        innerSize = tile[i];
+      }
+    }
+    return 1.0 / innerSize;
+  }
+};
+
+struct CPUThreadPass : public CPUThreadBase<CPUThreadPass> {
+  CPUThreadPass() = default;
+  explicit CPUThreadPass(unsigned threads) { this->threads = threads; }
```
flaub, line 58: I'm thinking that if `threads` is not specified, the default should be the number of cores on the current machine?

jbruestle: How do I do that with the passes.td thing?

jbruestle: I.e. I think the base constructor is the one that parses the values from the command line (and sets defaults if they are not there).  So if I overload the default constructor it to get number of threads, won't that break the ability to set it from arguments?

mars1026: perhaps something like the createXSMMStencilPass() function in target/x86/pipeline.cc:98?

pmlc/util/tags.h
```diff
@@ -63,5 +63,6 @@ int64_t getIntegerTag(T op, llvm::StringRef name, int64_t defaultVal) {
 inline llvm::StringRef subgroupSizeTag() { return "subgroupSize"; }
 inline llvm::StringRef gpuThreadTag() { return "gpuThread"; }
 inline llvm::StringRef gpuBlockTag() { return "gpuBlock"; }
+inline llvm::StringRef cpuThreadTag() { return "cpuThread"; }
```
flaub, line 66: Any reason these aren't just 
```
static constexpr const char *kTagCpuThread = "cpuThread";
```

pmlc/dialect/pxa/transforms/cpu_thread.cc
jbruestle: Right, but XSMMStencilPass doesn't use passes.td to do it's parsing (or registration or anything).  It seems like we would be adding a lot of boilerplate already implemented by passes.td to make this change, considering that pipeline.cc etc will be constructing it in c++ and can pass it the number of threads there, so it's really only pmlc-opt command line usage that we would be fixing.  It's unclear to me if we even want the 'command line' version to have a system dependent default for the number of threads.

jbruestle: Realized that the 'default' is just a string that gets pasted into the C++, so it can just as happily be a function.  Implemented.

pmlc/dialect/pxa/transforms/cpu_thread.cc
jbruestle: Fixed.

pmlc/util/tags.h
jbruestle: Unsure, but I'd rather not add cleanup of all the other uses (gpuBlockTag, etc) to this commit.

mars1026 (approved):

pmlc/dialect/pxa/transforms/cpu_thread.cc
```diff
@@ -0,0 +1,90 @@
...
+  void threadOp(AffineParallelOp op) {}
```
flaub, line 57: Is this unused?

PR #1467, opened 2020-10-15
[OpenCL] Fix incorrect builtins type

kdobros:
This change fixes problem with a lot of incorrect results using OpenCL
backed. The issue was that declared builtins have i32 type, where
hardware uses 64-bit values. Current solution is to redeclare builtins
with correct type and insert conversions just after loading the values.
Complete lowering of index to i64 causes IGC to crash.
Validated on resnet layer 1 and res2a_branch2b keras backed tests.


jbruestle (approved):
Interesting issue.  The solution feels a bit like a workaround, but it's unclear to me what a better solution would be.  Great work tracking this down, it seems very non-obvious!

pmlc/target/intel_gen_ocl_spirv/legalize_spirv.cc
```diff
@@ -0,0 +1,161 @@
...
+
+#include "pmlc/target/intel_gen_ocl_spirv/pass_detail.h"
+#include "pmlc/target/intel_gen_ocl_spirv/passes.h"
+
+#include "mlir/Dialect/SPIRV/SPIRVOps.h"
+#include "mlir/Dialect/SPIRV/TargetAndABI.h"
+#include "mlir/IR/StandardTypes.h"
+#include "mlir/Support/LLVM.h"
+#include "mlir/Transforms/DialectConversion.h"
+
```
flaub, line 13: Adding a `using namespace mlir; // NOLINT` I think is a good way to reduce lots of noise.

kdobros: I assumed we are trying to be as close as possible to llvm's code style in pmlc (minus ccplint's weirdness with include order).
In this case such `using namespace` would be against it:
> In implementation files (e.g.` .cpp` files), the rule is more of a stylistic rule, but is still important. Basically, using explicit namespace prefixes makes the code **clearer**, because it is immediately obvious what facilities are being used and where they are coming from. [...]
> The general form of this rule is that any `.cpp` file that implements code in any namespace may use that namespace (and its parents’), but should not use any others.
~ LLVM Coding Standards - "Do Not Use `using namespace std`"

Also as someone who just recently started working with MLIR and PlaidML I know how painful it can be to see some function or class without `mlir::` and spend a good amount of time trying to find it inside wrong project.

flaub: So far, we've been using `using namespace mlir;` liberally in our code base. But, you bring up good points. I think we should decided one or the other, I'm not partial actually, but I do want it to be consistent.

flaub: I'd also strongly support the idea of moving all our code into the `mlir` namespace.

flaub: Or, we could do what IREE is doing:
https://github.com/google/iree/blob/main/iree/compiler/Dialect/IREE/IR/IREEDialect.cpp#L28

PR #1472, opened 2020-10-16
Frozen Graph Functionality for TF Backend

dgkutnic:
This is a WIP PR tracking the frozen graph functionality for the TF backend. I'm creating this now because I want feedback on the general direction of this work -- there are still some bugs I need to fix which I have triaged below and am working on.

A recap:
* Frozen graphs store weights _within_ the graph. They also perform folding/other optimizations to the weights, so they are different than the weights in tfhub or other archives. This means we need the entire context of the graph/module to be able to compile it down through the stack.
* Due to this, I've decided to remove `weights` and `module_text` from the test archives, as they are redundant (info is stored in the graph, which takes precedence). The archive now contains test cases (known correct inputs/outputs that we can use as a baseline for correctness tests.) 

Where we're at:
* We are generating frozen graphdefs for I3D and ResNext. I have visually inspected these for correctness.
* The function `ImportFrozenGraph` takes a frozen graphdef and lowers to XLA HLO via `tf2xla`. 
  * Known bug (ResNext): Dynamic shapes aren't supported in frozen graphs. The full input shape (including batch size) has to be known _before_ the graph is frozen. The input placeholder shape for ResNext is `None, 224, 224, 3` which causes a "missing dim" error in the `tf2xla` stack. I am planning on defining the batch size for this and testing it out.
  * Known bug (I3D): Lowering bug - this is occurring on a 3D convolution while it is being lowered from TF to XLA. It's an intermediate layer, but the input is not getting propagated correctly. I am looking into this.
* The tests for I3D and ResNext are going to lower all the way from TF Graph through to PlaidML. They do this by calling `ImportFrozenGraph` to get from TF Graph -> XLA HLO, then `CompileAndCheck` to get from XLA HLO -> PlaidML.


flaub:
I really love this approach. I looked very briefly and seems good. I'll do a more thorough review later 👍

plaidml/bridge/tensorflow/service/graph_util.h
```diff
@@ -0,0 +1,21 @@
...
+#include "tensorflow/compiler/xla/service/hlo_module.h"
+#include "tensorflow/compiler/xla/status.h"
+#include "tensorflow/compiler/xla/statusor.h"
+
+namespace xla {
+namespace plaidml {
+
+StatusOr<std::unique_ptr<HloModule>> ImportFrozenGraph(std::string frozen_graph_file_path,
+                                                       std::vector<std::string> input_names,
+                                                       std::vector<std::string> output_names);
```
flaub, line 18: Either use `llvm::StringRef` and `llvm::ArrayRef` or pass by reference instead of by value.

plaidml/bridge/tensorflow/tests/resnext_test.cc
```diff
@@ -0,0 +1,82 @@
...
+    outputs.emplace_back(convertBuffer(buffer->data));
+  }
+
+  auto lookup = [&](const char* key) {
+    auto it = weights.find(key);
+    if (it == weights.end()) {
+      throw std::runtime_error(std::string("Key not found: ") + key);
+    }
+    return it->second;
+  };
```
flaub, line 60: This seems to be unused, drop it.

plaidml/bridge/tensorflow/tests/i3d_test.cc
```diff
@@ -108,1595 +59,13 @@ TEST_P(PlaidMLI3DOperationTest, SimpleI3D) {
     return it->second;
   };
```
flaub, line 60: This seems unused, drop it.

plaidml/bridge/tensorflow/tests/resnext_test.cc
```diff
@@ -0,0 +1,82 @@
...
+class PlaidMLResNextOperationTest : public PlaidMLCodegenTest, public ::testing::WithParamInterface<ResNextTestSpec> {};
+
+TEST_P(PlaidMLResNextOperationTest, SimpleResNext) {
+  auto data = ReadFile("plaidml/bridge/tensorflow/tests/resnext.pml");
+  zoo::ArchiveT archive;
+  zoo::GetArchive(data.data())->UnPackTo(&archive);
+
+  std::vector<MultiBuffer> inputs;
+  std::vector<MultiBuffer> outputs;
+  std::unordered_map<std::string, MultiBuffer> weights;
```
flaub, line 39: Drop this

plaidml/bridge/tensorflow/tests/i3d_test.cc
```diff
@@ -81,19 +39,12 @@ TEST_P(PlaidMLI3DOperationTest, SimpleI3D) {
   std::unordered_map<std::string, MultiBuffer> weights;
```
flaub, line 39: Drop this.

flaub (commented):
LGTM. But I wonder if there's a way to use C++ to freeze the graph so that we can use the same version of TF (the one we are building from source) for everything.

dgkutnic:
> I wonder if there's a way to use C++ to freeze the graph so that we can use the same version of TF (the one we are building from source) for everything.

Yes, there is a way to do this. The same API exists in both Python and C++. I agree with your thought process, and will work on this.

flaub:
What's the status of this?

dgkutnic:
Closing this, the relevant pieces are in #1541

PR #1473, opened 2020-10-16
[edsl] Add support for layout-agnostic ops

flaub:
Adding the concept of a `TensorLens`. A lens is used to provide a translation map between some `source` logical layout into some `target` logical layout. Op authors can use a lens so that they can focus on writing the op in some canonical logical layout.

Note that these logical layouts have no relation to the physical layout that a tensor will have in memory. This should be specified by users for inputs/outputs/constants and decided on by the compiler for intermediates.

See the `reorg_yolo` op for an example of its usage.

### DIFF
--- a/plaidml/core/ffi.cc
+++ b/plaidml/core/ffi.cc
@@ -228,30 +228,16 @@ plaidml_datatype plaidml_shape_get_dtype(  //
 plaidml_integers* plaidml_shape_get_sizes(  //
     plaidml_error* err,                     //
     plaidml_shape* shape) {
-  return ffi_wrap<plaidml_integers*>(err, 0, [&] {
-    const auto& sizes = shape->shape.sizes;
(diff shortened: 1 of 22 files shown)


tzerrell:
Interesting! I'm going to think a bit on the implications of this design, but at a first look I like it!

tzerrell:
Having thought more: I really like this design. It's a concise way of expressing both the input and output layout "lenses" which naturally supports them being the same (the common case) while retaining the expressivity allowing them to be different.

I think we should make sure this is very carefully tested: At least for me it is extremely easy to accidentally write the inverse layout permutation of what was intended (which is presumably the forward lens permutation on the input and the inverse lens permutation on the output?)

flaub:
Will add python impl and better checks to this PR.

jbruestle (approved):

PR #1477, opened 2020-10-16
[OpenCL] Add lowering for std.exp to spv.OCL.exp

kdobros:
ResNet-50 running in full with conformance - checked with `plaidbench keras`.


kdobros:
FYI as a follow-up I'll upstream this to mlir, but for now GLSL ops are hardcoded in multiple places so custom conversions are faster to add.

jbruestle (approved):

PR #1479, opened 2020-10-19
[RFC][pxa] Layout reorder pass

kdobros:
Basic flow and concept of a pass performing optimization of memory by
changing layout.
Main aim of this change is to show general concept, some details of
the implementation may change.
As of now the plan is to perform this optimization right after
subgroup/tile pass. In future it probably should be tightly integrated
with those passes.

Future roadmap, not included in this draft, would encompass:
- adding support for memory with multiple users
- taking into account parallelization when sorting dimensions
- adding heuristic for cache misses and determining if separate reorder
  is optimal when cannot fuse into producer
- integrating with tiling pass - estimate tiling of loops together with
  tiling memory
- taking into account constness of data (ie. weights) when selecting
  which memory to reorder


kdobros:
@jbruestle @abialas1 @flaub @earhart 
Please take a look

pmlc/dialect/pxa/transforms/reorder_layout.cc
```diff
@@ -0,0 +1,264 @@
...
+///   2,
+///                              (d0 * 16 + d1 * 8) floordiv 8 % 2, 0)
+///   constraints = {0 <= d1 < 2, 0 <= d0 < 6}
+///
+/// Output:
+///   (d0, d1) -> (d0 + d1, d0, d1, 0)
+mlir::AffineMap simplifyConstrainedAffineMap(mlir::AffineMap map,
+                                             mlir::IntegerSet constraints) {
+  // TODO: Implement.
+  return map;
```
jbruestle, line 45: What is the strategy for implementation on this?  A set of rewrites or something more general (such as flattening to a system of pure affines with slack variables or the like)?  Have you examined the MLIR affine code to see if there is existing code for any of this?  Looking at the above, I can see how this might be done via rewrites, but it seems potentially brittle.

pmlc/dialect/pxa/transforms/reorder_layout.cc
```diff
@@ -0,0 +1,264 @@
...
+///   8 % 2, 0) reordered shape  = <7, 6, 2, 8> reordered vector = <1, 1, 1, 8>
+///
+/// Note: to obtain affine map from input space to expanded space composition
+///       A o B can be used (with simplification).
+/// A o B = (d0, d1) -> (d0 + d1, d0, d1, 0)
+ReorderDesc expandAffineMap(mlir::AffineMap map, mlir::ArrayRef<uint64_t> shape,
+                            mlir::ArrayRef<uint64_t> vector,
+                            mlir::IntegerSet constraints) {
+  // TODO: Implement.
+  auto reorderMap = mlir::AffineMap::getMultiDimIdentityMap(map.getNumResults(),
```
jbruestle, line 66: I can see that what the example is doing, in that it is converting a the second dimension into components, and providing a mapping and new size/vector data.  However I don't understand what the overall goal in general is.  Is it to split all dimensions into the maximal number of 'independently strideable' bits or something like that?  I mean it's clear the the first dimension can't be 'tiled' in any useful way, and that the second dimension can be tiled into 3 parts, so maybe that makes sense?  But more description would be helpful.  Also, what would the algorithm for this be?

pmlc/dialect/pxa/transforms/reorder_layout.cc
```diff
@@ -0,0 +1,264 @@
...
+/// Output:
+///   reorder map (B)  = (d0, d1, d2, d3) -> (d2, d0, d1, d3)
+///   reordered shape  = <2, 7, 6, 8>
+///   reordered vector = <1, 1, 1, 8>
+///
+/// Note: to obtain affine map from input space to sorted space composition
+///       A o B can be used.
+///   A o B = (d0, d1) -> (d1, d0 + d1, d0)
+ReorderDesc sortAffineMap(mlir::AffineMap map, mlir::ArrayRef<uint64_t> shape,
+                          mlir::ArrayRef<uint64_t> vector,
```
jbruestle, line 90: Also a bit unclear in intent. Seems to be doing a permutation where the dimensions in the 'split' version of the tensor are sorted by whether or not the use a certain input dimension (i.e. dim 0 and 2 seems to be moved to the front, presumable since they used d1, which is 0 in the schedule?). Also, the composition doesn't make sense since it should have a 4d output I think?

pmlc/dialect/pxa/transforms/reorder_layout.cc
```diff
@@ -0,0 +1,264 @@
...
+///   schedule    = <1, 0>
+/// Output:
+///   reorder map = (d0, d1) -> (d1 % 16, d0, d1 floordiv 16)
+///   shape map = (d0, d1) -> (16, d0, d1 ceildiv 16)
+///   vector map = (d0, d1) -> (1, 1)
+///
+/// Note: to obtain affine map from input space to tiled space composition
+///       A o B can be used (with simplification).
+///   A o B = (d0, d1) -> (d1, d0 + d1, d0)
+mlir::Optional<ReorderDesc> tileAffineMap(mlir::AffineMap map,
```
jbruestle, line 117: The example seems to have errors in it or perhaps I just don't follow (why is vector map's output 2d?). Also based on the building blocks above, it seems like we would split the lower dimension into 3 bits, and then reorder, so we should have a 4d output?

jbruestle (changes requested):
Clearly a lot of good thinking going on in terms of generating tiled layouts, however the presentation was quite confusing, and went directly into details too fast without any lead up, so I felt like I was reverse engineering something rather than reading a design document.  An overview would be good, as well as some examples of desired before/after MLIR.

Also, I'm a bit worried as to whether some of the transforms are implementable in a way that works in the general case, as it's not clear how some of the algorithms would even be written (but again that may be my lack of understanding).

Finally, conceptually, I wonder if we are putting too much in one pass, and the result will be code which is hard to understand and maintain.  Perhaps it is inevitable, but I wonder if we couldn't split to work into two passes: First a 'dimension splitting' pass that would attempt to increase the dimensionality of tensors by splitting given dimensions maximally (given constraints on their usage, possibly what 'expand' is doing), and second, a reordering pass that just changes striding of the new dimensions (which would effectively be tiling).  Interesting the second pass would be useful even on it's own, just to change simple layout.

pmlc/dialect/pxa/transforms/reorder_layout.cc
kdobros: I was thinking about rewrites with matching some patterns, as that seems to be enough for all maps I saw generated in our code. MLIR already has simplification without constraints via flattening, so that could be used as a cleanup after our rewrites.
I looked briefly at what is directly exposed in MLIR and there doesn't seem to be anything like this. Some implementation details used in MLIR look useful though, so I think it may be good to upstream such simplification in future.

pmlc/dialect/pxa/transforms/reorder_layout.cc
kdobros: I have added some additional description. General aim is to transform map of read operation to reduce number of input dimensions used for each output, ie (d0 * 16 + d1) uses two input dimensions, but in (d0, d1) both use one.
This is done in order to provide stronger hint for sorting where each dimension should be placed.

As for implementation again I  was thinking about rewrites with pattern matching, only one I see really needed after subgroups/tiling being e*i* * N + e*j*, where e*j* < N -> e*i*, e*j*

pmlc/dialect/pxa/transforms/reorder_layout.cc
kdobros: Yes, it should be 4d, I missed that.
I have added some loose set of rules that hint at order for which we want to sort, but I'm not really sure how to convey this most clearly without explicitly defining strict weak ordering here.
I feel that best way to think about this is in reverse order, since more frequently changing dimension appears later in affine schedule, so they should also appear later in map.
So to sort for given schedule in basic case we are interested in dimension that appears latest in schedule.

In this case, d3 (0) is last as that's vector dimension, next is d1 (A: d0) as it uses only d0 in input which is last in schedule.
Then d0 (A: d0 + d1), because it uses d0, but addition of d1 makes it before only d0. Finally d2 (A: d1) as previous in schedule.

pmlc/dialect/pxa/transforms/reorder_layout.cc
kdobros: Sorry, I forgot to update after some changes, fixed. The result of this operation should be just composition of expand and sort.

pmlc/dialect/pxa/transforms/reorder_layout.cc
kdobros: Also worth to note that what sorting aims to do is to order dimensions in terms of how often they will repeat same value (least frequently to most), but I have no idea how to measure this exactly with non-trivial maps, hence this approximation.

jbruestle (commented):
Thanks for the additional description.  It makes much more sense now.  I think some of the map transforms may be tricky to do in the general case, but that can be worked out during implementation.  Overall, I think the plan is sound, and I like the design.  I'd click approve, but we probably don't need to land this. Let's give everyone else a change weight in and then we can close this.

kdobros:
Thanks @jbruestle for review.
Unfortunately I have already started doing some implementation work, sorry for the noise it adds.

> the presentation was quite confusing, and went directly into details too fast without any lead up, so I felt like I was reverse engineering something rather than reading a design document. An overview would be good, as well as some examples of desired before/after MLIR.

I changed order of declarations to make reading easier and split it away from implementation details.
Added some overview and simple example of what this pass is expected to do, based on CppEdsl.Convolution test.

> Also, I'm a bit worried as to whether some of the transforms are implementable in a way that works in the general case, as it's not clear how some of the algorithms would even be written (but again that may be my lack of understanding).

I'm not sure what exactly "generality" would mean in this case. Even if algorithms would be implemented in terms of rewrite patterns they can still be distilled to most broad case. It's also the way majority of operations on affine maps are implemented.

> Finally, conceptually, I wonder if we are putting too much in one pass, and the result will be code which is hard to understand and maintain. Perhaps it is inevitable, but I wonder if we couldn't split to work into two passes: First a 'dimension splitting' pass that would attempt to increase the dimensionality of tensors by splitting given dimensions maximally (given constraints on their usage, possibly what 'expand' is doing), and second, a reordering pass that just changes striding of the new dimensions (which would effectively be tiling). Interesting the second pass would be useful even on it's own, just to change simple layout.

I think separate passes sound good, but if we will want to do both in sequence having single pass allows to avoid analyzing memory, gathering information twice and mutating IR without certain benefit.
I think that some functions may be general enough to be extracted outside of this pass (ie. `simplifyConstrainedAffineMap`, `foldOrCreateReorder`, maybe even gathering information about memory usage). This should reduce amount of code directly tied to this pass.
Additionally I think that more performant direction may be to reuse memory layout selection in loop tiling pass and calculate more accurate heuristics of cache behavior.

jbruestle:
Well, we can leave this PR open for doing the implementation.  As I said, I think I understand the plan here better now, and I no longer think we need two passes.  Regarding 'generality', I simply mean: Can we guarantee all semi-affine operations (floor-div, mod) introduced by any splitting can be fully reduced by simplification?  Otherwise we will need to add better support for semi-affine operations throughout the stack (and upsteam), which may itself be useful but adds complexity.

abialas1:
Hi Jeremy, this is the last part of Konrad's changes for layout reorders. As we talked yesterday please take a look, and accept for merging.

jbruestle (approved):
Very nice work.

PR #1480, opened 2020-10-19
Merge eltwise dialect into tile dialect

flaub:


jbruestle:
Is this basically (mostly) NFC code movement?  I presume this is really just cleaning up.  I would tend to agree that conceptually eltwise is really a subset of tile at this point since we lower to std typing during conversion to PXA at this point.

flaub:
Yes, this is mostly NFC, but there are a few changes for cleanup:

* Merge the `eltwise::ScalarConstantOp` with the `tile::ConstantOp`.
* Due to above, all types are always `RankedTensorType`, no more scalars anywhere.
* Custom assembly formats for all ops

jbruestle (approved):

PR #1483, opened 2020-10-20
resizeTmpPass improvements

abialas1:
Added vector ops support.
Added option to apply only to allocOp in nested ops (right now resizing global mem causes output verification fails)


jbruestle (approved):
Looks good!  I think long term there are some way to generalize vector support, but those involve reworking the stride analysis a bit, and this is great for now.

PR #1485, opened 2020-10-20
Add resnet50 workload to pml-ocl-gen platform

drgone:


jbruestle (dismissed):

flaub (approved):

adstraw (changes requested):
please adjust thresholds in plan.yml so that we are gating performance for both vk and ocl gen resnet50 runs.  golden value is 21ms.  my thought is to set thresholds to 0.33 for vk and ocl.  this would mean we gate vk and ocl commits for 3x the golden.

adstraw (approved):
LGTM

flaub (approved):

PR #1488, opened 2020-10-20
route OpenCL and Vulkan inference time to plaidbench

adstraw:
Gen9 / ResNet50 plaidbench output for OpenCL device:
```
2020-10-20 16:30:38,450 VERBOSE-1 [default] Total OpenCL time: 11595.6ms
2020-10-20 16:30:38,450 VERBOSE-1 [default] Total OpenCL execute time: 68.074ms
                                                                       ^^^^^^^^
2020-10-20 16:30:38,450 VERBOSE-1 [default] Total OpenCL kernels: 97
2020-10-20 16:30:38,450 VERBOSE-1 [default] Total OpenCL kernel execute time: 30.4867ms
2020-10-20 16:30:38,450 VERBOSE-1 [default] Total OpenCL memory transfers: 606
2020-10-20 16:30:38,450 VERBOSE-1 [default] Total OpenCL memory transfer time: 37.5873ms
2020-10-20 16:30:38,450 VERBOSE-1 [default] Execution time: 11682.2ms
Example finished, elapsed: 39.463s (compile), 0.068s (execution)

-----------------------------------------------------------------------------------------
Network Name         Inference Latency         Time / FPS          
-----------------------------------------------------------------------------------------
resnet50             68.07 ms                  68.07 ms / 14.69 fps
                                               ^^^^^^^^
```


plaidbench/plaidbench/frontend_keras.py
```diff
@@ -242,12 +242,9 @@ def run(self, once=False, warmup=False):
...
         out = self.model.predict(x=self.x[:epoch_size], batch_size=self.params.batch_size)
-        # Hack to ensure eventlog gets written
-        if not once and not warmup:
-            import keras.backend as b
-            if b.backend() == 'plaidml':
-                b._ctx.shutdown()
-        return (out, {})
+
+        import plaidml.bridge.keras as keras_bridge
+        return (out, {"time": keras_bridge.lastExecuteTime / 1000})
```
flaub, line 247: We need to check if the backend is plaidml before assuming that we can import it. There are cases where we want to run against a different backend and plaidml may not be there.

flaub: Also, this seems wrong, wouldn't it be `keras_bridge.lastExecuteTimeInMS`?

plaidml/exec/exec.h
```diff
@@ -90,7 +90,7 @@ class Executable {
     for (size_t i = 0; i < raw_outputs.size(); i++) {
       raw_outputs[i] = outputs[i].as_ptr();
     }
-    ffi::call_void(              //
+    return ffi::call_double(     //
```
flaub, line 93: `ffi::call_double` is unnecessary, just use `ffi::call<double>`.

plaidml/exec/ffi.cc
```diff
@@ -113,8 +114,10 @@ void plaidml_executable_run(   //
       }
       outputBuffers.push_back(outputs[i]->buffer);
     }
-    exec->exec->invoke(inputBuffers, outputBuffers);
+    execTimeInMS = exec->exec->invoke(inputBuffers, outputBuffers);
```
flaub, line 117: Just return it directly.

plaidml/exec/ffi.cc
```diff
@@ -75,13 +75,14 @@ void plaidml_executable_free(  //
...
-    size_t noutputs,           //
+double plaidml_executable_run(  //
+    plaidml_error* err,         //
+    plaidml_executable* exec,   //
+    size_t ninputs,             //
+    plaidml_buffer** inputs,    //
+    size_t noutputs,            //
     plaidml_buffer** outputs) {
+  double execTimeInMS = 0.0;
   ffi_wrap_void(err, [&] {  //
```
flaub, line 86: Use `ffi_wrap<double>(...`

pmlc/rt/jit_executable.cc
```diff
@@ -400,18 +400,20 @@ class JitExecutable final : public Executable {
     }
   }
 
-  void invoke(mlir::ArrayRef<util::BufferPtr> inputBuffers,
-              mlir::ArrayRef<util::BufferPtr> outputBuffers) final {
+  double invoke(mlir::ArrayRef<util::BufferPtr> inputBuffers,
+                mlir::ArrayRef<util::BufferPtr> outputBuffers) final {
+    bindArguments(inputBuffers, outputBuffers);
```
flaub, line 405: This should be included in the stopwatch time.

plaidml/core/core.h
```diff
@@ -40,6 +40,16 @@ void call_void(F fn, Args... args) {
...
+template <typename F, typename... Args>
+double call_double(F fn, Args... args) {
+  plaidml_error err;
+  double ret = fn(&err, args...);
+  if (err.code) {
+    throw std::runtime_error(str(err.msg));
+  }
+  return ret;
+}
+
```
flaub, line 52: No need for this, there's a templated version already.

flaub (changes requested):
Some minor changes needed, but otherwise looks great

plaidbench/plaidbench/frontend_keras.py
adstraw: Fixed the name bug.  lastExecuteTime -> lastExecuteTimeInMS.

plaidml/core/core.h
adstraw: Done.

plaidml/exec/exec.h
adstraw: Done.

plaidml/exec/ffi.cc
adstraw: Done.

plaidml/exec/ffi.cc
adstraw: Done.

pmlc/rt/jit_executable.cc
adstraw: Done.

plaidbench/plaidbench/frontend_keras.py
adstraw: Done.

flaub (approved):

flaub (changes requested):
Blocking for now

pmlc/rt/jit_executable.cc
```diff
@@ -400,18 +400,21 @@ class JitExecutable final : public Executable {
...
+    stopWatch.start();
     bindArguments(inputBuffers, outputBuffers);
     jitEntry(ptrs.data());
-    if (VLOG_IS_ON(1)) {
-      stopWatch.stop();
-      IVLOG(1, "Execution time: " << stopWatch.delta_ms() << "ms");
+    stopWatch.stop();
+
+    double execTime = device->execTimeInMS;
+    if (execTime == 0.0) {
```
adstraw, line 412: @flaub please check my logic here.  by doing this, I *always* return a non-zero execution time.  which means that even for pipelines such as CPU which do not set device->execTimeInMS we will use this "inner" stopwatch time to override `exec_per_example` in plaidbench.  not sure if that's what we want.  alternatively, I can move this check-for-zero to frontend_keras.py and only set the override in the non-zero case.  in that case I would simply return device->execTimeInMS from this function.

adstraw: the latest commit moves check-for-zero to frontend_keras.py and sets override accordingly.  I think this is what we want.

flaub (approved):

PR #1502, opened 2020-10-23
Some lowering code improvements

abialas1:
- LoopInvariantCodeMotionPass to reduce redundant mem reads in the generated code
- llvm's data flow opt to remove redundant buffers after loop unrolling
- changed condition to enable block ops for more buffers
These changes give few % boost on my machine.

### DIFF
--- a/pmlc/target/intel_gen/subgroup_broadcast.cc
+++ b/pmlc/target/intel_gen/subgroup_broadcast.cc
@@ -89,8 +89,8 @@ public:
     // TODO: Current HW supports only block read\write on global mem scope.
     // This can change in the future so probably need to be better handled with
     // HW specific parameters
-    auto invalidMemScope =
-        dyn_cast_or_null<AllocOp>(op.memref().getDefiningOp());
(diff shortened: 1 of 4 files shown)


jbruestle (approved):
Looks good

jbruestle:
Hmm, looks like there are test issues, and at least one of them looks like an actual issue.

abialas1:
Hi Jeremy,
Yes turns out that there were 2 problems, none of which affect resnet50, but still break other things.
1. Added additional checks for gpuBlock parallel op (it is present in resnet but not in tests)
2. Had to remove llvm's dataflow opt. Turns out that for some cases (not resnet) due to mixture of affine.for and affine.parallel there is an assert. The llvm's version works correctly only when ForOps are present. In some places I saw that there are redundant load->store traffic in resnet50 kernels. Right now I am working on a pass that vectorizes reads and writes to fully leverage on the block ops. Hopefully this would kind of replace what is achieved with affine's dataflow optimization.

jbruestle (approved):

PR #1504, opened 2020-10-24
Mark constant program args with `tile.const` attr

flaub:


earhart (approved):
Seems fine, but it'd be nice to have a dialect-independent marker -- like an interface, but for values.

earhart:
On further review -- why not make the attribute a unit attribute that's only present on constant arguments?  (Or at least a boolean?)

PR #1505, opened 2020-10-24
Add plaidml_program_save

flaub:


plaidml/core/core.h
```diff
@@ -339,6 +295,52 @@ class Buffer {
...
+  std::vector<TensorShape> outputs() const {
+    auto shapes = details::make_ptr(ffi::call<plaidml_shapes*>(plaidml_program_get_outputs, as_ptr()));
+    std::vector<TensorShape> ret(shapes->size);
+    for (size_t i = 0; i < shapes->size; i++) {
+      ret[i] = TensorShape(shapes->elts[i]);
+    }
+    return ret;
+  }
+
+  Buffer save() const { return Buffer(ffi::call<plaidml_buffer*>(plaidml_program_save, as_ptr())); }
```
szilveszteri, line 336: like the idea of save(). Could we also have a load() function that could consume a previously saved program? 
Would also vote for import/export naming instead of save/load, but that's just a nit

flaub: Export is good. Could also pass in a 'format' for the export.

flaub: Downside with `export` is that this is a reserved keyword in c++.

jbruestle (approved):

PR #1506, opened 2020-10-26
Fix keras backend to support additional networks

flaub:


drgone (approved):
LGTM

plaidml/core/__init__.py
```diff
@@ -287,7 +287,7 @@ def copy_into_ndarray(self, dst):
     def copy_from_ndarray(self, src):
         dst = np.frombuffer(self.data, dtype=self.shape.dtype.into_numpy())
         dst = dst.reshape(self.shape.sizes)
-        np.copyto(dst, src)
+        np.copyto(dst, src, casting='unsafe')
```
flaub, line 290: @tzerrell Needed to add this to allow casting from `f32` to bool`, WDYT?

tzerrell: Seems fine in that context. Perhaps it would be nice to be a little more paranoid and check that the data types are one of the expected unsafe casts, but I'm not sure we even know what all of the expected unsafe casts are at this point (e.g. presumably f64 to bool is also a thing that we'll be expected to do sometimes?). So I think this is ok as is

PR #1511, opened 2020-10-29
[pxa] Add simplification with constraints pass

kdobros:
This change adds pass to simplify affine maps using constraints on input
dimensions, and several related utilities.


pmlc/dialect/pxa/transforms/simplify_with_constraints.cc
```diff
@@ -0,0 +1,64 @@
...
+  if (newMap.getAffineMap() == oldMap.getAffineMap())
+    return;
+  auto newMapAttr = mlir::AffineMapAttr::get(newMap.getAffineMap());
+  op.setAttr(op.getMapAttrName(), newMapAttr);
+  op.idxsMutable().assign(newMap.getOperands());
+}
+
+} // namespace
+
+void simplifyAffineMapsWithConstraints(mlir::Operation *op) {
```
jbruestle, line 52: Technically, we could use the interfaces (PxaReadOpInterface, PxaReduceOpInterface), and actually, for this use and other similar ones, we should probably add a PxaMemOpInterface that both derive from so we could just cast as the interface and be done, but that can wait till another commit.

pmlc/dialect/pxa/transforms/simplify_with_constraints.cc
```diff
@@ -0,0 +1,64 @@
...
+  mlir::AffineValueMap newMap = simplifyMapWithConstraints(oldMap);
+  if (newMap.getAffineMap() == oldMap.getAffineMap())
+    return;
+  auto newMapAttr = mlir::AffineMapAttr::get(newMap.getAffineMap());
+  op.setAttr(op.getMapAttrName(), newMapAttr);
+  op.indicesMutable().assign(newMap.getOperands());
+}
+
+template <typename OpTy>
+void simplifyReduceMap(OpTy op) {
```
jbruestle, line 38: I believe this is identical to SimplifyloadMap, so maybe just SimplifyMap?  See also comment below about interfaces.

pmlc/dialect/pxa/analysis/affine_constraints.cc
```diff
@@ -0,0 +1,210 @@
...
+
+template <typename T>
+struct maximum {
+  T operator()(const T &lhs, const T &rhs) { return std::max(lhs, rhs); }
+};
+
+} // namespace
+
+std::pair<mlir::Optional<int64_t>, mlir::Optional<int64_t>>
+getLowerUpperBounds(mlir::AffineExpr expr,
```
jbruestle, line 71: This code is heavily overlapping in purpose with StrideRange in pxa/analysis, but this code is better I think.  Specifically, it more properly supports semi-affine operations like mod.  It is probably worth considers how to dedup on some later commit.

jbruestle (approved):
Great work, a very nice design for this pass.  One minor nit, but I'm approving anyway since it's a trivial fix later and I don't want to block things needlessly given your time frame.

pmlc/dialect/pxa/transforms/simplify_with_constraints.cc
kdobros: Completely agree, one interface for all operations having affine map would be nice. 
Unfortunately current interfaces don't have methods to update affine map and operands, which requires parametrization over concrete type.

pmlc/dialect/pxa/transforms/simplify_with_constraints.cc
kdobros: Unfortunately `op.getMapAttrName()` exists only in concrete operation type, same for `op.idxsMutable()`/`op.indicesMutable()`. Having single function would require incorporating these into interfaces.

PR #1512, opened 2020-10-29
[pxa] Add separate GPU thread function

kdobros:
This change adds separate function for threading parallel op for GPU -
tiling into global and local loops.
Such function is useful when new operation needs to be added after
normal threading pass, such as when inserting data reorders.
It then needs to be manually threaded to comply with expected semantics.


jbruestle (approved):
In general, I think making more 'transforms' composable independently of 'passes' is a good idea, so I approve of this commit, although there is something interesting going on here with pass ordering issues between threading and layout reorders that deserves some thinking.

PR #1513, opened 2020-10-29
[pxa] Add reordering utilities

kdobros:
This change adds several utility functions for converting data layouts,
creating and inserting reorders.


jbruestle (commented):
High level concern:  What is the rationale for changing load/store ops rather than using the memory layout maps supported by memref (via 'setAffineMaps' https://github.com/llvm/llvm-project/blob/master/mlir/include/mlir/IR/StandardTypes.h#L505)?  I suppose if we are intending to split dimensions, this may be the path of least resistance (vs adding support for semi-affine ops upstream), but I just want to understand the choice in more detail.

Minor nit, more use of interfaces to reduce type switches.

kdobros:
> High level concern: What is the rationale for changing load/store ops rather than using the memory layout maps supported by memref (via 'setAffineMaps' https://github.com/llvm/llvm-project/blob/master/mlir/include/mlir/IR/StandardTypes.h#L505)? I suppose if we are intending to split dimensions, this may be the path of least resistance (vs adding support for semi-affine ops upstream), but I just want to understand the choice in more detail.
> 
> Minor nit, more use of interfaces to reduce type switches.

Main rationale is ,as you mentioned, poor support for semi-affine both upstream and in some of our passes.
I'm also not entirely sure if using layout maps will simplify anything, as they are tied to type, same as shape.
Additionally using layout maps may limit some more advanced layout optimizations like padding, as it cannot change actual size of memory.

As for using interfaces, I'm not sure how it could be done. Currently the concept is to create same operation with modified affine map and updated type. Using interfaces would require adding methods to do this in place, which is significantly more effort.

jbruestle (approved):
I'm still uncertain about reordering ops vs generalizing memref layouts long term, that said, this approach doesn't require upstream code changes.  Additionally, we need better PXA interfaces, but I'm planning to add that in a near future PR, so this code will be another example use case to test the design.  Overall, looks good!  Sorry for the delay on approving.

abialas1 (approved):
Approving after mac test rerun and passed.

PR #1514, opened 2020-10-30
CMake build improvements

flaub:
Add targets:
  * check
  * check-test
  * clang-format
  * check-clang-format
  * fix-clang-format

Add support for running edsl lit test

Add CPM support and use it to download/use boost


tzerrell (commented):
LGTM although I'd like it if someone with more cmake experience could review as well

jbruestle (approved):

PR #1515, opened 2020-10-30
[comp] Memory synchronization passes and utilities

kdobros:
This change:
- adds utilities to track memory synchronization status (i.e. that have same
  content under comp dialect)
- pass that removes trivially redundant memory transfers using above
  utils
- modifies minimize allocations pass to prioritize reusing memory that
  is already synchronized, so not needing extra write

This lowers number of memory transfers in ResNet + OCL from 525 to 411.
Further reduction of that number will require tracking of which memory is
actually modified by kernel and extending synchronization tracking
utilities with this information.


jbruestle (approved):
Looks good.

PR #1516, opened 2020-10-30
WIP: Vectorize memory ops

abialas1:
Show general direction of the memory ops vectorization pass, e.g. use subgroup_block_read8 instead of 8x subgroup_block_read1.

Things that are missing in llvm upstream (locally implemented, will move these to llvm next week):
1. New SPIRV Op OpVectorExtractDynamic
2. Lowering of vector.extract_element to SPIRV
3. Updates to Vector size verification. For OpenCL with Vector16 capability we are allowed to use vector size 8 and 16, which we want to leverage on. Per spirv spec, Vulkan does not allow it.

Things that are missing and not there yet:
1. Simple mlir tests for all new features (worked on real case conv1x1 scenario)
2. Proper logic and functional tests with the ops that we can actualy block read using these vectors, will check with Konrad's layouts reorders
3. Currently only vector_load supported, support vector write as well
4. Lowerings of vector_loads\stores to spirv in the llvm upstream

### DIFF
--- a/pmlc/conversion/gpu_to_spirv/BUILD
+++ b/pmlc/conversion/gpu_to_spirv/BUILD
@@ -35,5 +35,6 @@ plaidml_cc_library(
         "@llvm-project//mlir:GPUToSPIRVTransforms",
         "@llvm-project//mlir:GPUTransforms",
         "@llvm-project//mlir:SPIRVDialect",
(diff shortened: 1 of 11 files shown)


abialas1:
moving to different branch

PR #1518, opened 2020-10-30
Allow TensorDims to be used in Repeat

tzerrell:
Accept `TensorDim`s for the `repeats` parameter of the oplib function `repeat`. This is intended to help us avoid `compute_shape` in the OV Bucketize code in https://github.com/plaidml/openvino/pull/90

### DIFF
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -2234,8 +2234,16 @@ Value repeat(const Value& value) {
     throw std::runtime_error(llvm::formatv("PlaidML repeat op expects 3 arguments (received {0})", args.size()));
   }
   auto I = args[0].as_tensor();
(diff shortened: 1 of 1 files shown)


dgkutnic:
LGTM - seems like a reasonable way to avoid compute_shape/explicit broadcasting.

For what it's worth, this wouldn't really have any impact on the TF frontend because XLA implements `tf.tile`/`tf.repeat` as a bunch of low-level broadcasts/reshapes.

flaub:
FYI with the recent rewrite of the eDSL implementation based on the AST, `compute_shape` is no longer as expensive as it once was. However, I think we should continue to follow the advice of using `bind_dims` or `TensorDim` over using `compute_shape`. And while it's not as expensive as it was before, it's still something that can usually be avoided.

flaub (approved):

PR #1519, opened 2020-10-31
Add support for subprojects

flaub:


pmlc/dialect/pxa/ir/matchers.h
```diff
@@ -27,7 +27,7 @@ struct PxaReduceOpMatcher {
 
 template <typename A, typename B>
 inline auto m_PxaReduceOp(mlir::AtomicRMWKind agg, A aMatcher, B bMatcher) {
-  return PxaReduceOpMatcher(agg, aMatcher, bMatcher);
+  return PxaReduceOpMatcher<A, B>(agg, aMatcher, bMatcher);
```
jbruestle, line 30: Why was this required?  I presume there was some reason.

jbruestle (approved):
Looks good, one question, but not a blocker, just curiosity.

pmlc/dialect/pxa/ir/matchers.h
flaub: It seems like there's an issue when compiling in a subproject with different compiler options or perhaps a different version of the compiler. This was needed to prevent a compiler error; for some reason the template types were not being deduced. I suspect this is related to `-std=1z` being used to emulate C++17. Given issues like this, I think we should use C++14 for the time being to ensure a wider range of compiler support.

PR #1521, opened 2020-11-02
Fixes for CMake build and downstream users

flaub:


pmlc/dialect/pxa/transforms/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@
     nest_loops.cc
     normalize.cc
     resize_tmps.cc
+    simplify_with_constraints.cc
```
jbruestle, line 18: Hmm, looks like we need at least basic CI that calls cmake to prevent this kind of omission.

plaidml/core/CMakeLists.txt
```diff
@@ -1,13 +1,31 @@
...
+  HDRS ${_HDRS}
+  SRCS ${_SRCS}
+  DEPS
+    pmlc::ast
+    pmlc::compiler
+    pmlc::util
+    LLVMSupport
+)
+
+pml_cc_library(
```
jbruestle, line 24: I wonder if there is some way to refactor this into a 'pml_cc_with_obj' or something to avoid the duplication, but probably not a huge deal given there are only 4 places this pattern seems to be used.

jbruestle (approved):
Looks good.

plaidml/core/CMakeLists.txt
flaub: Good idea, I'll give it some thought

PR #1522, opened 2020-11-02
Add repeat function declaration of TensorDim version.

XinWangIntel:


XinWangIntel:
In order to use TendorDim for repeatOp, need to change the declaration of repeat,  but not sure add a overload version or keep one version with edsl::Value argument, which is better.

PR #1523, opened 2020-11-02
Add repeat function declaration of TensorDim version.

YangleiZouIntel:

### DIFF
--- a/plaidml/op/op.h
+++ b/plaidml/op/op.h
@@ -604,6 +604,11 @@ inline edsl::Tensor repeat(const edsl::Tensor& I, int repeats, int axis) {
   return details::op("repeat", args).as_tensor();
 }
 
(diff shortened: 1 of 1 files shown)


XinWangIntel:
In order to use TendorDim for repeatOp, need to change the declaration of repeat, but not sure add a overload version or keep one version with edsl::Value argument, which is better.

tzerrell:
Thanks for correcting my oversight here!

Regarding overload vs. `edsl::Value`: We actually don't want to use either. Instead, as we upgrade the oplib we've been updating the header to use a Fluent API (as seen in e.g. `convolution` and `relu`). I'll put together a PR with that change so you can see what I mean.

tzerrell:
Closing this in favor of #1524. See my previous comment for details.

PR #1524, opened 2020-11-02
Update Repeat to Fluent API

tzerrell:
Upgrade the API for repeat to enable the use of the new TensorDim repeats parameter, which I overlooked in #1518.

### DIFF
--- a/plaidml/op/op.h
+++ b/plaidml/op/op.h
@@ -599,10 +599,35 @@ inline edsl::Tensor reorg_yolo(const edsl::Tensor& I, int stride, bool decrease,
   return details::op("reorg_yolo", args).as_tensor();
 }
 
-inline edsl::Tensor repeat(const edsl::Tensor& I, int repeats, int axis) {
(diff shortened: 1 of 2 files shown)


plaidml/op/op.h
```diff
@@ -599,10 +599,35 @@ inline edsl::Tensor reorg_yolo(const edsl::Tensor& I, int stride, bool decrease,
...
 
-inline edsl::Tensor repeat(const edsl::Tensor& I, int repeats, int axis) {
-  auto args = edsl::make_tuple(I, repeats, axis);
-  return details::op("repeat", args).as_tensor();
-}
+class repeat {
+ public:
+  explicit repeat(const edsl::Tensor& I) : I_(I) {}
+
+  repeat& repeats(int repeats) {
```
flaub, line 606: Maybe call this `count` to avoid confusion?

tzerrell: Sure. The `repeats` matches numpy, but I agree that `count` is a less confusing name, so I made the switch.

flaub (approved):

PR #1527, opened 2020-11-03
Add support for user-specified data layout to ReorderLayoutsPass

brightlaboratory:
Placeholder for future data layout transforms work.


brightlaboratory:
@jbruestle @flaub : I was wondering if for memrefs, user layouts are allowed in the pxa dialect such as the ones here: https://github.com/plaidml/plaidml/blob/private/sanket/data_layout_opts/pmlc/target/x86/tests/conv_NCHW_3X3_user_layouts.mlir#L12 

When I specify the layouts, pmlc-opt is generating no output. The command line I am using is here: https://github.com/plaidml/plaidml/blob/private/sanket/data_layout_opts/pmlc/target/x86/tests/conv_NCHW_3X3_user_layouts.mlir#L1 . 

` #K_map = affine_map<(K,C,R,S) -> (R, S, C, K)>
#NCHW_to_NHWC = affine_map<(N,C,H,W) -> (N,H,W,C)>
!I_memref = type memref<1x64x56x56xf32, #NCHW_to_NHWC>
 !K_memref = type memref<64x64x3x3xf32, #K_map>`

When I specify no layout https://github.com/plaidml/plaidml/blob/private/sanket/data_layout_opts/pmlc/target/x86/tests/conv_NCHW_3X3_user_layouts.mlir#L8, the code works just fine. 

`!I_memref = type memref<1x64x56x56xf32> 
!K_memref = type memref<64x64x3x3xf32>`

Thanks in advance.

flaub:
What is the overall goal of this PR?

brightlaboratory:
> What is the overall goal of this PR?

@flaub :
The goal is to be able to insert blocked data layouts such as NCHWc16 for tensors -- for Convolutions to start with. There are two ways that it can be done (and the various commits in this PR are attempting to do that):

1) If the user can specify the layouts associated with memrefs then the reorder-layouts pass (pmlc/dialect/pxa/transforms/reorder_layouts.cc) with make them "explicit": insert data copy operations from original layouts to target layouts, and have the subsequent reads to the tensor be done in the target layout. E.g.,   #NCHW_to_NHWC = affine_map<(N,C,H,W) -> (N,H,W,C)>
 !I_memref = type memref<1x64x56x56xf32, #NCHW_to_NHWC> . I have implemented the functionality to orchestrate this. However, a challenge that arose was, where and how to insert the user/oracle provided layouts.

That led the following approach that I am currently working on:
2) We recognize Convolutions and distinguish between the input and filter tensors, and for those tensors we select the blocked data layouts. It is implemented. Currently I am investigating incorrectness in output results for Resnet-50.

3) A more general solution would be do determine the best data layouts on-the-fly. That is, we don't recognize that something is a convolution but instead perform the best data selection based on for example data access patterns without being aware of what operator we are looking at. In fact, Konrad's data layout reordering pass does this. He chooses the heuristic of: the data layout should be conforming to the execution schedule. That is, the data should reside in memory in the order in which it is accessed. This heuristic makes sense as a general principle but doesn't lead us to blocked data layouts because there is an assumption that the execution order is already optimal which is not always the case.

jbruestle (commented):
In addition the inline comments, I think the added logging is a bit overboard, especially the logs that just print which function is being run.  Generally, lots of the code changes are not really 'production quality', so I think some cleanup is in order.

Regarding the main layout pass and it's associated tests:  There are no comments there yet because it's very complex and I haven't been able to grok it yet at all.  I think we should schedule a long meeting to go over the theory and what's being done under the hood.  Based on earlier discussions with Adam Straw, I think the code can probably be decomposed into a set of smaller transformations and be make more general and more readable.  As is, if we land it, I think we will end up in a situation where no one else in our team understands what it is doing.

pmlc/target/x86/tests/conv_sequence.mlir
```diff
@@ -0,0 +1,188 @@
...
+
+// PLAIDML_VERBOSE=5 build-x86_64/Release/bin/pmlc-opt -convert-linalg-to-loops -pxa-reorder-layouts="allow-reorder=true make-user-layouts-explicit=true" -canonicalize -x86-stencil-tpp-gemm pmlc/target/x86/tests/conv_sequence.mlir > output
+
+// WithOUT data reordering: build-x86_64/Release/bin/pmlc-opt -convert-linalg-to-loops -canonicalize -x86-stencil-tpp-gemm -x86-convert-pxa-to-affine --normalize-memrefs --simplify-affine-structures  -lower-affine  -canonicalize -convert-scf-to-std -x86-convert-std-to-llvm pmlc/target/x86/tests/conv_sequence.mlir | build-x86_64/Release/bin/pmlc-jit -e conv
+
+
+// WITH data reordering: build-x86_64/Release/bin/pmlc-opt -convert-linalg-to-loops -pxa-reorder-layouts="allow-reorder=true make-user-layouts-explicit=true" -canonicalize -x86-stencil-tpp-gemm -x86-convert-pxa-to-affine --normalize-memrefs --simplify-affine-structures  -lower-affine  -canonicalize -convert-scf-to-std -x86-convert-std-to-llvm pmlc/target/x86/tests/conv_sequence.mlir | build-x86_64/Release/bin/pmlc-jit -e conv
+
+// cat ref | grep -v sizes > ref.cleaned 
+// cat out | grep -v sizes > out.cleaned
```
flaub, line 12: These comments probably don't belong in this file.

pmlc/target/x86/tests/conv_sequence.mlir
```diff
@@ -0,0 +1,188 @@
...
+      %O_ud = memref.cast %638 : memref<1xf32> to memref<*xf32>
+      call @print_memref_f32(%O_ud) : (memref<*xf32>) -> ()
+    }
+
+//  %O_ud = memref.cast %45 : memref<1x56x56x256xf32> to memref<*xf32>
+//  call @print_memref_f32(%O_ud) : (memref<*xf32>) -> ()
+
+   return
+}
+// CHECK: pxa.generic
```
flaub, line 186: This doesn't seem to test anything?

pmlc/target/x86/tests/conv_sequence.mlir
```diff
@@ -0,0 +1,188 @@
...
+
+
+     // bias
+    %arg3 = affine.parallel (%arg111) = (0) to (64) reduce ("assign") -> (memref<64xf32>) {
+      %637 = pxa.reduce assign %cst_1, %origArg3[%arg111] : memref<64xf32>
+      affine.yield %637 : memref<64xf32>
+    }
+
+//  %O_ud = memref.cast %arg3 : memref<64xf32> to memref<*xf32>
+//  call @print_memref_f32(%O_ud) : (memref<*xf32>) -> ()
```
flaub, line 34: These comments should be removed.

flaub (changes requested):
Agreed that there is too much logging, or at least we need to be more tactical about it, perhaps using higher log levels. It's definitely not OK to land this since it has an `exit()` in the middle of the code.

pmlc/dialect/pxa/transforms/reorder_layouts.cc
```diff
@@ -1549,14 +1594,33 @@ mlir::Optional<ReorderDesc> optimizeLayoutForReads(
...
+   * be NCHWc16
+   *
+   * 2. If a tensor appears in a generic op not recognized by our pattern
+   * matchers such as Convolution pattern matchers, then we would like to derive
+   * optimal layouts from first principles.
+   */
+
+  /* If the flag - makeUserLayoutsExplicit is set, then we adopt strategy 1).
+   * If the said flag is not set, then we go for strategy 2)
+   */
```
flaub, line 1792: Use C++ style comments (avoid C style comments, i.e. /* */

plaidml/bridge/keras/__init__.py
```diff
@@ -1782,7 +1785,15 @@ def variable(value, dtype=None, name=None, constraint=None):
             #     'Casting to requested dtype in variable, received {} and requested {}'.format(
             #         value.dtype, dtype))
             value = value.astype(dtype)
-        return _KerasNode('variable', name=name, var=value)
+        if (str(name) == 'kernel'):
```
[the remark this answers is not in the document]
brightlaboratory: I have removed this code now.

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -649,6 +649,66 @@ TEST_F(CppEdsl, Convolution) {
...
+
+  auto I = Placeholder(DType::FLOAT32, {1, 64, 56, 56});
+  auto K1 = Constant(makeBuffer(DType::FLOAT32, {64, 64, 3, 3}, data1), "K1");
+  auto K2 = Constant(makeBuffer(DType::FLOAT32, {64, 64, 3, 3}, data2), "K2");
+  auto conv1 = Convolution3(I, K1);
+  auto conv2 = Convolution3(conv1, K2, "NCHW", "KCHW");
+  auto program = makeProgram("convolution", {I}, {conv2});
+
+  // clang-format off
+  // CHECK-LABEL: CppEdsl.Convolution_NCHW
```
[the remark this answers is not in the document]
brightlaboratory: I have undone the code changes to edsl_test.cc. The mlir test files test the functionality of blocking of data layouts now.

pmlc/dialect/pxa/analysis/strides.cc
```diff
@@ -142,6 +142,19 @@ StrideInfo &StrideInfo::operator*=(int64_t factor) {
   return *this;
 }
 
+// Divide the offset and all strides by a constant.
+StrideInfo &StrideInfo::operator/=(int64_t factor) {
+  offset /= factor;
+  if (factor == 0) {
```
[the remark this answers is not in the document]
brightlaboratory: I have undone changes to strides.cc and strides.h.

pmlc/dialect/pxa/transforms/stencil.cc
```diff
@@ -143,13 +145,22 @@ void StencilBase::recursiveBindIndex(SmallVector<BlockArgument, 8> &boundIdxs,
       assert(requirements[currIdx].predicates.size() == values.size() &&
              "Each predicate entry must have one function per I/O op");
       for (unsigned i = 0; i < values.size(); i++) {
+        IVLOG(4, "Calling getStrideInfo");
         auto strideInfo = getStrideInfo(values[i]);
+        if (!strideInfo) {
+          IVLOG(4, "StrideInfo is NULL ");
+          exit(1);
```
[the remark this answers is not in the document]
brightlaboratory: I have undone changes to stencil.cc and have removed all exit()s :)

test_cnn_model.sh
```diff
@@ -0,0 +1,4 @@
+MODEL=$1
```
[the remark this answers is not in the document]
brightlaboratory: I have removed the .sh files.

pmlc/target/x86/tests/conv_sequence.mlir
brightlaboratory: I have added additional tests - running the same file and checking for the results. I felt the comments stating how to run the code and a good way to check for correctness, would be may still be useful if someone is investigating it later.

pmlc/dialect/pxa/transforms/reorder_layouts.cc
brightlaboratory: Yes. For the subsequent comments I added, I used "//".

pmlc/target/x86/tests/conv_sequence.mlir
brightlaboratory: Done.

pmlc/target/x86/tests/conv_sequence.mlir
brightlaboratory: Yeah, I have added more "CHECK"s now.

brightlaboratory:
> In addition the inline comments, I think the added logging is a bit overboard, especially the logs that just print which function is being run. Generally, lots of the code changes are not really 'production quality', so I think some cleanup is in order.
> 
> Regarding the main layout pass and it's associated tests: There are no comments there yet because it's very complex and I haven't been able to grok it yet at all. I think we should schedule a long meeting to go over the theory and what's being done under the hood. Based on earlier discussions with Adam Straw, I think the code can probably be decomposed into a set of smaller transformations and be make more general and more readable. As is, if we land it, I think we will end up in a situation where no one else in our team understands what it is doing.

Currently I have disabled reordering pass in the x86 pipeline (because without hoisting of the weight conversion code happening, the performance will be worse). The three tests I have added exercise the blocked data layout reordering functionality.  With regards to logging - I felt the names of the functions sometimes are necessary to know (and hence to be printed) to make sense of the other statements appearing in the logs. Therefore, I have left them in there for now at the log level of 4.

flaub (approved):

jbruestle (approved):
Looks good!

PR #1528, opened 2020-11-03
Vectorize memory ops

abialas1:
Things that are missing in llvm upstream, before we can have pull this in:

- New SPIRV Op OpVectorExtractDynamic and lowering of vector.extract_element to SPIRV (https://reviews.llvm.org/D90679)
- Updates to Vector size verification (https://reviews.llvm.org/D90683)

Currently only vector_load supported, WIP support for vector write as well

### DIFF
--- a/pmlc/target/intel_gen/subgroup_broadcast.cc
+++ b/pmlc/target/intel_gen/subgroup_broadcast.cc
@@ -11,9 +11,11 @@
 #include "mlir/IR/StandardTypes.h"
 #include "mlir/Interfaces/VectorInterfaces.h"
 #include "mlir/Pass/Pass.h"
+#include "mlir/Support/DebugStringHelper.h"
 
 #include "pmlc/dialect/stdx/ir/ops.h"
 #include "pmlc/target/intel_gen/pass_detail.h"
+#include "pmlc/util/logging.h"
 #include "pmlc/util/tags.h"
 
 using namespace mlir; // NOLINT[build/namespaces]
@@ -34,7 +36,7 @@ public:
       : loop(loop), vectorSize(vectorSize), useBlockOps(useBlockOps) {}
 
   bool isVectorTypeValid(VectorType type) {
-    return type.getRank() == 1 && type.getDimSize(0) == vectorSize;
+    return type.getRank() == 1 && type.getDimSize(0) % vectorSize == 0;
   }
 
   template <typename T>
@@ -81,6 +83,7 @@ public:
   }
 
   LogicalResult devectorizeTransferRead(vector::TransferReadOp op) {
+    IVLOG(3, "Orginal Op: " << debugString(*op));
     if (!isVectorTypeValid(op.getVectorType()) || op.indices().size() < 1)
       return failure();
     OpBuilder builder(op);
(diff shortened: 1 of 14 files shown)


abialas1:
Hi Jeremy, Please take a look at this PR. It does not break build, as the required new ops are inside llvm upstream. This optimization is disabled in pipeline right now and can be enabled after llvm upstream update. If you do not have any concerns I believe it could land on plaidml-v1 now.

jbruestle (approved):
Looks good overall.  Could use more lit tests on the changes to subgroup broadcast lowering, but not a blocker.

pmlc/target/intel_gen/subgroup_broadcast.cc
```diff
@@ -97,16 +100,56 @@ class DevectorizeImpl {
...
+                op.getLoc(), op.memref(), idxs);
+        devectorizeVectorOp(newBlockReadOp.getOperation());
+        op.replaceAllUsesWith(newBlockReadOp.getResult());
+        IVLOG(3, "Load Op: " << debugString(*newBlockReadOp));
+      }
+      // Case3: Used buffer was allocated inside the kernel, no block reads
+      // allowed. Check if element type of the allocated memory is vector, if so
+      // then it is needed to use vector TransferReadOp for writing
+    } else if (op.memref().getDefiningOp() &&
+               dyn_cast<AllocOp>(op.memref().getDefiningOp())
```
flaub, line 134: Don't use `dyn_cast` if the result is not being used, instead this should be `isa`

flaub: Better yet, use a matcher :)

pmlc/target/intel_gen/subgroup_broadcast.cc
```diff
@@ -151,13 +207,48 @@ class DevectorizeImpl {
...
+    return success();
+  }
+
+  LogicalResult devectorizeExtractMap(vector::ExtractMapOp op) {
+    OpBuilder builder(op);
+    auto idVal = op.id();
+    IVLOG(3, "oldExtractOp: " << debugString(*op));
+    // It is needed to use i32 type for extract element index. Use cast in case
+    // it comes from index
+    if (op.id().getType().dyn_cast<IndexType>()) {
```
flaub, line 240: Same here, use `isa` instead of `dyn_cast`

pmlc/target/intel_gen/subgroup_broadcast.cc
abialas1: Thanks Frank, actually in this case I need the dyn_cast to get Alloc's element type. But there are some other places that could be moved to matcher. I am working on adding vectorization to write ops, and will include these changes in next PR.

abialas1:
Had to rebase the code due to conflicts. Please re-approve

jbruestle:
The isa thing is a minor point, but I think it's worth fixing.

abialas1:
Ok, makes sense. I changed dyn_casts to isa. I uploaded my latest changes with refactors and addition to vector reduce ops. As expected it gives minor boost (~2%) on my machine. Most benefits comes from block reads on gen9.

jbruestle (approved):

PR #1530, opened 2020-11-03
fix non-deterministic behavior with affine map, arg ordering

adstraw:


jbruestle (commented):
So, this change has a rather annoying impact on clients due to the need to always pass 'relative', which seems non-intuitive.  Perhaps we could just order blocks via their depth under the FuncOp?

pmlc/dialect/pxa/analysis/strides.cc
```diff
@@ -213,22 +213,41 @@ StrideRange StrideInfo::range() const {
...
+    auto block = kvp.first.getOwner();
+    assert(block);
+    auto parent = block->getParentOp();
+    while (parent) {
+      nestLevel++;
+      parent = parent->getParentOp();
+    }
+
+    ordered.emplace(
+        nestLevel * 10000 + kvp.first.getArgNumber(),
```
adstraw, line 229: 10000 is a magic number here, basically it represents a max number of arguments per level.  there may be a better way to do this.

pmlc/dialect/pxa/analysis/strides.cc
```diff
@@ -213,22 +214,40 @@ StrideRange StrideInfo::range() const {
...
+    int64_t loopDepth = 0;
+    auto block = kvp.first.getOwner();
+    auto parent = block->getParentOp();
+    while (!dyn_cast<FuncOp>(parent)) {
+      loopDepth++;
+      parent = parent->getParentOp();
+    }
+
+    ordered.emplace(
+        loopDepth * strides.size() + kvp.first.getArgNumber(),
```
jbruestle, line 229: 1) This code is incorrect, it is not guaranteed to give each index a unique number since the number of indexes in an affine.parallel is a) not the same from loop to loop in a nest, b) Not related to strides.size(), which could in fact be either bigger or smaller than the # of indexes.
2) Regardless, this mapping of two integers to 1 is a weird hacky thing to do anyway, just make a map of: 
std::map<std::pair<unsigned, unsigned>, std::pair<mlir::BlockArgument, int64_t>>
Pair is lexicographically ordered, so that should work just fine.  Also, the restriction to use int64_t everywhere is really only for things which might conceivable just to do arithmetic with indexes (like strides, sizes, etc), for things like loop depth or arg number, unsigned is fine.

jbruestle (approved):

pmlc/dialect/pxa/analysis/strides.cc
jbruestle: Whoops, didn't mean to approve!  But other than the comment inline, looks good.

jbruestle (approved):
Looks good.

PR #1532, opened 2020-11-05
Disable tile.MulOp fold pattern with 0 when result() and rhs() have different datatypes

YangleiZouIntel:
Sometimes we want the result of a Tensor multiple 0 to be still a Tensor such as in [a workaround for embedding bag offsets sum ](https://github.com/Flex-plaidml-team/openvino/pull/5/files#diff-4e93753e3ae0a26abb5bfdebfd81e3ccbddf047c1d89c6cefac41c840b1a5f8eR76).

The following test currently does not get compiled. 
I removed the corresponding MulOp fold pattern in this patch so we do not need the workaround in the example above. Please advice if there is a better way to fix this.

TEST_F(CppEdsl, MulZero) {
  auto I = Placeholder(DType::FLOAT32, {3, 3});
  auto O = I * 0;
  auto program = makeProgram("mul_zero", {I}, {O});
}

### DIFF
--- a/pmlc/dialect/tile/ir/eltwise.cc
+++ b/pmlc/dialect/tile/ir/eltwise.cc
@@ -64,6 +64,13 @@ OpFoldResult DivOp::fold(ArrayRef<Attribute> operands) {
 }
 
 OpFoldResult MulOp::fold(ArrayRef<Attribute> operands) {
+  // mul(x, 0) -> 0
+  if (lhs().getType() == rhs().getType()) {
+    if (matchPattern(rhs(), m_Zero())) {
+      return rhs();
+    }
+  }
+
   // mul(x, 1) -> x
   if (matchPattern(rhs(), m_One())) {
     return lhs();
--- a/pmlc/dialect/tile/ir/eltwise.cc
+++ b/pmlc/dialect/tile/ir/eltwise.cc
@@ -70,7 +70,6 @@ OpFoldResult MulOp::fold(ArrayRef<Attribute> operands) {
       return rhs();
     }
   }
-
   // mul(x, 1) -> x
   if (matchPattern(rhs(), m_One())) {
     return lhs();
--- a/pmlc/dialect/tile/ir/eltwise.cc
+++ b/pmlc/dialect/tile/ir/eltwise.cc
@@ -64,10 +64,6 @@ OpFoldResult DivOp::fold(ArrayRef<Attribute> operands) {
 }
 
 OpFoldResult MulOp::fold(ArrayRef<Attribute> operands) {
-  // mul(x, 0) -> 0
-  if (matchPattern(rhs(), m_Zero())) {
-    return rhs();
-  }
   // mul(x, 1) -> x
   if (matchPattern(rhs(), m_One())) {
     return lhs();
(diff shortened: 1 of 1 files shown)


flaub:
I don't understand the problem here. Are you saying there's a compilation error in your example? Can you show us the error you get?

flaub (changes requested):
This is an important optimization. I'd like to understand a little better what the actual problem. I don't see how this folding would be incorrect.

tzerrell:
Is the linked line of code the workaround? If so, what do you think the non-workaround version should be?

Or does the linked line require a workaround? If so, what workaround did you add to make it work?

tzerrell:
Oh, is the concern with the broadcasting behavior? I.e., in the following line you are `select`ing with both outcomes being 0... are you doing this in order to broadcast to the shape of the `slices[0]` tensor?

YangleiZouIntel:
> I don't understand the problem here. Are you saying there's a compilation error in your example? Can you show us the error you get?

Sorry I did not make it clear. I mean the tile program in this example does not get compiled to llvm kernels.

The issue is about a Tensor muliply 0. For the following edsl example,
```
TEST_F(CppEdsl, MulZero) {
auto I = Placeholder(DType::FLOAT32, {3, 3});
auto O = I * 0;
auto program = makeProgram("mul_zero", {I}, {O});
}
```
It gets the following error:
```
error: type of return operand 0 ('tensor<!tile.six>') doesn't match function result type ('tensor<3x3xf32>') in function @main
```
The IR goes wrong at the first CanonicalizerPass in
https://github.com/plaidml/plaidml/blob/plaidml-v1/pmlc/ast/builder.cc#L510

IR before the first CanonicalizerPass
```
module @mul_zero {
  func @main(%arg0: tensor<3x3xf32>) -> tensor<3x3xf32> {
    %c0 = tile.constant(0 : i64) : tensor<!tile.six>
    %0 = tile.mul %arg0, %c0 : (tensor<3x3xf32>, tensor<!tile.six>) -> tensor<3x3xf32>
    return %0 : tensor<3x3xf32>
  }
}
```
IR After the first CanonicalizerPass, the return type is tile.six, which does not match the function return type.
```
module @mul_zero {
  func @main(%arg0: tensor<3x3xf32>) -> tensor<3x3xf32> {
    %c0 = tile.constant(0 : i64) : tensor<!tile.six>
    return %c0 : tensor<!tile.six>
  }
}
```

haoyouab:
> Oh, is the concern with the broadcasting behavior? I.e., in the following line you are `select`ing with both outcomes being 0... are you doing this in order to broadcast to the shape of the `slices[0]` tensor?

Yes. The most straight-forward non-workaround version I could come up with is

`if (default_index == -1) {
  Os.push_back(slices[0] * 0);
}`

because when `withDefaultIndex = false` the 'empty bag' should be filled with zeros. Please point me out if there's any better way to do this as I'm not yet very familiar with the code base to know where to go. That's probably why I used this weird `select` approach to fill a tensor of the same shape as `slices[0]` with zeros.

Here's what I got when I used the non-workaround version:

`error: 'tile.contract' op tensor #0 must be eltwise-any, but got 'tensor<!tile.six>'`

I'm not sure if this is the expected behavior when doing `some tensor * 0`. Multiplying any other constants seems fine, though.

YangleiZouIntel:
> This is an important optimization. I'd like to understand a little better what the actual problem. I don't see how this folding would be incorrect.

I have updated this patch to disable this fold pattern only when lhs() and rhs() have different datatypes, i.e Tensor and int, where the semantics of the MulOp depends on both lhs and rhs datatypes, in this situation, directly replacing MulOp by rhs() is likely to cause datatype dismatch in the following/parent ops.

flaub:
This is fine as a workaround but I think there's a deeper issue here that I'll investigate.

flaub (approved):

PR #1541, opened 2020-11-06
TF Enablement

dgkutnic:
Summary of changes:
* Enable 3 full-sized networks (I3D, ResNext-50, ResNet-152) via the TF XLA backend
  * Note: Currently ResNext and ResNet are timing out in CI, this is because we run XLA to get the frozen HLO representation of the networks, and this takes a long time. We can talk about what to do here, whether it's storing the HLO representation in the `plaidml/depot` repo or just enabling these tests as a `py_binary` for now.
* Add documentation
* Add support for 3 new XLA ops (`Clamp`, `Log1p`,  `RoundNearestAfz`)
* Infer tensor layouts for convolutions instead of just assuming NXC (todo: do this for the rest of the ops where layout matters)


flaub:
What's the status of this?

dgkutnic:
Hi @flaub, sorry for keeping this open for so long.

The current opens are:
* Builds fail on macos
    1. On macos, TF's bazel rules expect `--cpu=darwin` while our build rules expect `--cpu=darwin_x86_64`. I think the easiest way of working around this is to overwrite the `macos` config in TF to expect `--cpu=darwin_x86_64`.
    1. A struct is being reinitialized in the TF code and clang doesn't like this. I can overwrite this code so that the struct only gets initialized once, but I'm not sure whether this is the cleanest way of doing things. 
* Reference Implementations
    1. For the purposes of the TR, we used OpenVino as a reference implementation and measured those numbers locally, so that doesn't belong in the PR. However, we could include a build rule that generates the NN collateral for someone to run `model-optimizer` on. Does this seem useful?
    1. The current reference numbers in the PR are TF 1.x numbers, which are not really a useful comparison point since they're much slower than OV and even TF 2.x. Should we just remove these?

dgkutnic:
EISW-1044 TF Multicore CPU enablement work #implemented

flaub:
@dgkutnic Should we drop this?

PR #1542, opened 2020-11-07
Remove src memref from comp.alloc

earhart:
Combining the copy and the alloc prevents us from hoisting the alloc, and doesn't get us a synchronization event for the data copy.  If this optimization turns out to be important, I think it'd be better handled as part of the lowering to LLVM.


jbruestle (approved):
Looks good.

PR #1543, opened 2020-11-07
Return OpenVINO plugin building with CMake

tzerrell:
Move the OpenVINO PlaidML plugin to this repository using CMake to build. This is currently a draft as it's not yet functional (tests fail with `undefined symbol: CreatePluginEngine`), but it does build. The build files will need to be cleaned up as well.


flaub:
@tzerrell I got things building and functional tests are running. I had to comment out some of the Convolution BackProp tests because they're causing a segfault. I think there's actually a bug here because when I run in debug mode I see xsmm complaining about invalid parameters (ldb larger than k). Also, I needed to add OpenVINO as a submodule, I couldn't figure out a way around it.

tzerrell:
Thanks @flaub! Too bad a submodule is necessary, but this LGTM.

And thanks for the heads up about conv backprop

flaub:
I'll try and get this to work without a submodule, I have one more thing to try.

flaub (approved):

PR #1544, opened 2020-11-07
Introduce eDSL/Tile layer op

flaub:


plaidml/edsl/edsl.h
```diff
@@ -361,6 +362,8 @@ class Tensor {
   TensorLens lens_;
 };
 
+using Tensors = std::vector<Tensor>;
```
jbruestle, line 365: I'm not sure about this name:  I think it makes the code a bit hard to read since Tensor and Tensors are very similar looking.  Maybe TensorList or TensorVec?

plaidml/edsl/edsl.h
```diff
@@ -1273,5 +1275,64 @@ inline Tensor pragma(const Tensor& tensor, const std::string& op, const PragmaAt
 ///
 inline Tensor trace(const Tensor& x, const std::string& msg) { return pragma(x, "trace", {{"msg", Value(msg)}}); }
 
+using LayerBodySingleFn = std::function<Tensor()>;
+using LayerBodyMultiFn = std::function<Tensors()>;
+
+inline Tensors layer(const std::string& op, const Tensors& operands, const Dictionary& attrs,
```
jbruestle, line 1281: Aren't the operands derivable via the use/def pairs that cross the block boundary (which we have to compute regardless)?  If so, why make the user explicitly list them?

jbruestle (commented):
Two inline comments.

plaidml/edsl/edsl.h
```diff
@@ -30,6 +30,7 @@ namespace details {
 struct Deleter {
   void operator()(plaidml_dim_expr* ptr) { ffi::call_void(plaidml_dim_expr_free, ptr); }
   void operator()(plaidml_expr* ptr) { ffi::call_void(plaidml_expr_free, ptr); }
+  void operator()(plaidml_exprs* ptr) { ffi::call_void(plaidml_exprs_free, ptr); }
```
tzerrell, line 33: I will echo the Tensor/Tensors concern here for plaidml_expr/plaidml_exprs; I had to read this line a few times to see how it differed from the above. That said, I think we've been doing fine with plaidml_expr vs plaidml_exprs for a while so perhaps that's an argument for this being ok

jbruestle (approved):
Approving as is to unblock downstream work.  Remaining issue to be fixed in followup commit.

PR #1546, opened 2020-11-09
Add OpenMP workaround pass

earhart:


LLITCHEV (approved):
Looks fine to me.

pmlc/target/x86/pipeline.cc
```diff
@@ -104,25 +84,124 @@ struct ConvertStandardToLLVMPass
...
+    funcOp.walk([&](omp::ParallelOp parOp) {
+      // The values used to construct the smuggling struct.
+      SmallVector<LLVM::LLVMType, 8> fieldTypes;
+      SmallVector<Value, 8> fieldSrcs;
+
+      // A mapping from original values to the OpOperands within the
+      // parallel region that use those values, for value replacement.
+      std::unordered_map<void *, std::list<OpOperand *>> valueMap;
+
+      // Find values that cross the loop
```
jbruestle, line 117: Couldn't we use getUsedValuesDefinedAbove?  That would simplify this code quite a bit.  We will almost certainly have to do that to upstream this.

earhart: We need the `OpOperand`s, but maybe we can use `visitUsedValuesDefinedAbove`?  I'll try it out.

pmlc/target/x86/pipeline.cc
```diff
@@ -104,25 +84,124 @@ struct ConvertStandardToLLVMPass
...
+              break;
+            }
+            region = region->getParentRegion();
+          }
+          if (definedInsideParOp) {
+            continue;
+          }
+
+          // Otherwise, we need to smuggle the value through an alloca'd struct.
+          auto [it, inserted] = valueMap.try_emplace(value.getAsOpaquePointer(),
```
flaub, line 156: Please don't use C++17 features, I have to remove them when I go to integrate with downstream users.

pmlc/target/x86/pipeline.cc
```diff
@@ -104,25 +84,124 @@ struct ConvertStandardToLLVMPass
...
+    LLVM::LLVMFuncOp funcOp = getOperation();
+    OpBuilder builder{&getContext()};
+    funcOp.walk([&](omp::ParallelOp parOp) {
+      // The values used to construct the smuggling struct.
+      SmallVector<LLVM::LLVMType, 8> fieldTypes;
+      SmallVector<Value, 8> fieldSrcs;
+
+      // A mapping from original values to the OpOperands within the
+      // parallel region that use those values, for value replacement.
+      std::unordered_map<void *, std::list<OpOperand *>> valueMap;
```
flaub, line 115: We should never use `void*` for the key, what are we trying to do here?

pmlc/target/x86/pipeline.cc
flaub: Here's a snippet that I've used in multiple places, so it probably belongs in a utility, but seems similar to what you're trying to accomplish:

```
      llvm::SetVector<Value> sinkCandidates;
      getUsedValuesDefinedAbove(band.getLoopBody(), sinkCandidates);

      llvm::SetVector<Operation*> sunkOperations;
      for (Value candidate : sinkCandidates) {
        collectScalarOps(candidate, &sunkOperations);
      }

      BlockAndValueMapping map;
      for (Operation* op : sunkOperations) {
        OpBuilder builder(band.getLoopBody());
        Operation* clonedOp = builder.clone(*op, map);
        // Only replace uses within the current band.
        for (auto result : llvm::enumerate(op->getResults())) {
          auto replacement = clonedOp->getResult(result.index());
          for (auto& use : llvm::make_early_inc_range(result.value().getUses())) {
            if (use.getOwner()->getParentOfType<AffineParallelOp>() == band) {
              use.set(replacement);
            }
          }
        }
      }
    }
```

pmlc/target/x86/pipeline.cc
earhart: Which downstream users?  If this is going to be an ongoing thing, should we disable c++17 in our build?  It might be easier than trying to remember which features landed in which version...?

pmlc/target/x86/pipeline.cc
earhart: Duplicate detection.  Honestly, I just copied the pattern from MLIR's `BlockAndValueMapping`.

PR #1547, opened 2020-11-10
Better use of MLIR builtins

earhart:


flaub:
Great, thanks!

LLITCHEV (approved):
Much simpler when using built-in functions.

Issue #1491, opened 2020-10-21
Add axis parameter to Gather


tzerrell:
Add an axis parameter to allow gather to be performed over a user-selected axis.

Issue #1490, opened 2020-10-21
Add interpolated gather


tzerrell:
Allow gather to take non-integal indices by adding an interpolation mode. Should likely include at least linear & cubic interpolation, probably also nearest interpolation.

PR #1548, opened 2020-11-10
Add axis parameter and interpolation mode to GatherOp

YangleiZouIntel:
This PR
1.adds axis parameter to GatherOp  https://github.com/plaidml/plaidml/issues/1491

2.adds interpolation mode to GatherOp, including linear, cubic and five nearest modes. https://github.com/plaidml/plaidml/issues/1490 
Cubic interpolation algorithm follows [the same algorithm used in OpenVINO nGraph](https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=1163711)

3.gather related Keras backend tests remain unchanged (support up to 2 parameters).
    edsl gather op supports up to six parameters (tensor, indices, axis, interpolation mode, nearest mode, cube coeff).

4.added support for indices on & out of boundary (idx<0 or idx>axis_length-1), treat the data as if edge-padding was applied to the data. (the same algorithm used in OpenVINO nGraph)

4.added OpenVINO interpolation Ops in this PR since it is closely related to intepolated gather.

5.added a few edsl gather tests, passed on CPU and intel_gen vulkan, disabled on intel_gen_ocl since OclFloorF and OclCeilF  Ops have not been added to mlir spirv dialect.


pmlc/dialect/tile/ir/ops.td
```diff
@@ -90,10 +90,14 @@ def GatherOp : MaterializeOp<"gather"> {
   let summary = "special gather operation";
   let arguments = (ins
     RankedTensorOf<[AnyScalar]>:$tensor,
-    RankedTensorOf<[AnyInteger]>:$dims);
+    RankedTensorOf<[AnyScalar]>:$indices,
```
tzerrell, line 93: I _think_ we have been trying to standardize our spelling to `indexes` instead of `indices`. But perhaps that's different in the MLIR codebase? Does anyone know if upstream MLIR has an opinion on the indexes/indices spelling?

tzerrell (commented):
I'll wait for tests to pass to do a detailed review, but at a glance this looks good.

pmlc/dialect/tile/ir/ops.td
flaub: I've been attempting to use `indices` everywhere we spell it out, and using `idxs` for shorthand variable names.

pmlc/ast/builder.cc
```diff
@@ -607,6 +608,53 @@ struct ProgramBuilder {
...
+  Value makeGatherOp(ExprNodeIntrinsic *node, ArrayRef<Value> operands) {
+    auto operands_size = operands.size();
+    TensorShape shape = evaluator.getShape(node);
+    RankedTensorType resultType = builder.getRankedTensorType(shape);
+    auto op = builder.create<tile::GatherOp>(
+        loc, resultType, ValueRange{operands[0], operands[1]},
+        ArrayRef<NamedAttribute>{});
+
+    switch (operands_size) {
+    case 5: {
```
flaub, line 620: What are these magic numbers?

flaub: Basically, I don't think we should try to do different logic based on different numbers of operands at this stage. Can we force the C API to always supply the full number of operands and default those to sensible values above?

pmlc/ast/ast_ops.cc
```diff
@@ -59,24 +59,29 @@ struct IndexOp : Intrinsic {
 struct GatherOp : Intrinsic {
   TensorShapes getShapes(Evaluator *evaluator, ArrayRef<ExprNodePtr> operands,
                          ArrayRef<TensorShape> shapes) const final {
-    if (operands.size() != 2) {
-      throw std::runtime_error("'gather' requires 2 arguments.");
+    auto operands_size = operands.size();
+    if (operands_size < 2 || operands_size > 5) {
+      throw std::runtime_error("'gather' requires 2-5 arguments.");
```
flaub, line 64: I'd prefer that the `GatherOp` at the ast level always expects a fixed number of operands. This means that the `edsl.h` should add default values if we don't want to force the user to specify all 5 operands in every call.

plaidml/edsl/edsl.h
```diff
@@ -806,7 +806,13 @@ inline Tensor floor(const Tensor& x) { return intrinsic("floor", x); }
...
 /// \return Tensor
 ///
-inline Tensor gather(const Tensor& x, const Tensor& y) { return intrinsic("gather", x, y); }
+
+inline Tensor gather(const Tensor& x, const Tensor& y, const std::vector<Tensor>& attributes = {}) {
+  std::vector<Tensor> args = {x};
+  args.emplace_back(y);
+  args.insert(args.end(), attributes.begin(), attributes.end());
+  return intrinsicCall("gather", args);
+}
```
flaub, line 815: We should use a fluent/builder-style API if we want the user to specify optional attributes/values to the op. See how we did this for something like a convolution.

pmlc/dialect/tile/ir/ops.td
```diff
@@ -90,10 +90,14 @@ def GatherOp : MaterializeOp<"gather"> {
...
+    DefaultValuedAttr<IndexAttr,"0">:$axis,
```
flaub, line 94: Please add a space after the `,`

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -890,46 +891,225 @@ struct GatherOpConversion : public OpConversionPattern<tile::GatherOp> {
...
+        loc, f32Type, rewriter.getFloatAttr(f32Type, 0.5));
+    index = rewriter.create<mlir::AddFOp>(loc, index, cstHalfF32);
+    index = rewriter.create<mlir::FPToUIOp>(loc, index, i32Type).getResult();
+    index =
+        rewriter.create<mlir::IndexCastOp>(loc, index, indexType).getResult();
+    srcOps.at(dim) = index;
+    return rewriter.create<mlir::LoadOp>(loc, tensor, srcOps);
+  }
+
+  Value buildLinearInterpolationOps(Location loc,
```
flaub, line 973: Hmm, I wonder if we could write `.mlir` files instead and just link in these functions instead of generating them here. It's a bit hard to read in C++ format, but would be much cleaner in a `.mlir` file.

I don't think we need to modify this PR, but in the future, I'd like to explore how we can write complex code like this in a more maintainable fashion.

flaub:
This is a great start!

pmlc/ast/builder.cc
YangleiZouIntel: Changed to fixed number (5) of operands for gather.

plaidml/edsl/edsl.h
YangleiZouIntel: Use a builder-style API for gather now with default attributes values.  Currently the same default values are used in 3 places, 1, edsl c++ API. 2, edsl python API. 3. tile GatherOp (in tile/ir/ops.td file).

pmlc/ast/ast_ops.cc
YangleiZouIntel: Changed to fixed number (5) of operands for gather.  Added default values to edsl c++ API, but also need to add default values to edsl python API (to support keras backend tests). I wonder if we should use a new function`plaidml_expr_gather` (like`plaidml_expr_pragma`) instead of `intrinsicCall` (which calls `plaidml_expr_intrinsic`).

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
YangleiZouIntel: Yes, that would be much more mantainable.

plaidml/edsl/edsl.h
```diff
@@ -809,7 +813,37 @@ inline Tensor floor(const Tensor& x) { return intrinsic("floor", x); }
...
+
+  Tensor build() const {
+    std::vector<Tensor> args = {x_, y_, axis_, mode_, cubic_coeff_};
+    return intrinsicCall("gather", args);
+  }
+
+ private:
+  Tensor x_;
+  Tensor y_;
+  Tensor axis_ = Tensor(0);
```
flaub, line 843: Don't use a Tensor for these values, they are serialized in a `Value` when it crosses the C API, which natively stores an integer.

plaidml/edsl/edsl.h
```diff
@@ -809,7 +813,37 @@ inline Tensor floor(const Tensor& x) { return intrinsic("floor", x); }
...
+  Tensor mode_ = Tensor(static_cast<uint64_t>(InterpolationMode::linear));
+  Tensor cubic_coeff_ = Tensor(-0.5);
```
flaub, line 845: Same with these, use `int64_t` and `double` respectively.

flaub, line 845: My previous comment here was a mistake; these are eDSL primitive instead of oplib calls. I apologize. However I'm now wondering if we should use `Value` for eDSL primitive operands instead of `Tensor`.

plaidml/bridge/openvino/ops/embedding_bag_offsets_sum.cpp
```diff
@@ -56,7 +56,7 @@ static OpRegistration reg("EmbeddingBagOffsetsSum", [](const Context& ctx) {
   auto batch = offsets.size();
   offsets.push_back(num_indices);
 
-  auto I_gathered = gather(I, indices);
+  auto I_gathered = gather(I, indices).build();
```
flaub, line 59: You could also use ```
Tensor I_gathered = gather(I, indices);
```

**Comment by tzerrell** [comment]
This looks generally good; I see it's still marked draft, so I presume it's not quite ready for a careful review?

**Comment by YangleiZouIntel** [comment]
> This looks generally good; I see it's still marked draft, so I presume it's not quite ready for a careful review?

Right now all the gather edsl tests and keras backend tests passed on CPU and intel_gen_vulkan, (right now some spirv ocl ops(OclFloor, OclCeil et al) haven't been added to mlir, so interpolated gather is not currently supported by intel_gen_ocl and I added those tests to intel_gen_opencl edsl skip list ). I am adding a few more tests and comparing the interpolation results with ngraph, This should be finished soon and I will make it ready for review.

File: plaidml/edsl/edsl.h [hunk 10]
```diff
@@ -25,6 +27,8 @@ class TensorDim;
 class TensorIndex;
 class Value;
 
+using pmlc::util::InterpolationMode;
```
flaub, line 30: We can't do this. `pmlc` is not visible to users of the PlaidML API. We need to make a duplicate copy of this enumeration.

plaidml/edsl/edsl.h
```diff
@@ -14,6 +14,8 @@
 #include "plaidml/core/core.h"
 #include "plaidml/edsl/ffi.h"
 
+#include "pmlc/util/enums.h"
```
flaub, line 17: Sorry I didn't notice this before, but we can't do this here because users of the PlaidML API only need to depend on `libplaidml.so` and the `plaidml` header files. They won't have access to the `pmlc`, which is internal to the API.

pmlc/ast/builder.cc
```diff
@@ -738,6 +739,38 @@ struct ProgramBuilder {
...
 
+  Value makeGatherOp(ExprNodeIntrinsic *node, ArrayRef<Value> operands) {
+    TensorShape shape = evaluator.getShape(node);
+    RankedTensorType resultType = builder.getRankedTensorType(shape);
+    auto op = builder.create<tile::GatherOp>(
+        loc, resultType, ValueRange{operands[0], operands[1]},
+        ArrayRef<NamedAttribute>{});
+
+    FloatAttr cubicCoeff;
+    if (!m_Constant(&cubicCoeff).match(operands[4].getDefiningOp())) {
```
flaub, line 750: use `matchPattern()` instead of `m_Constant().match()`. The `matchPattern()` takes an `Operation*` or `Value`. So it'd be like:
```
if (!matchPattern(operands[4], m_Constant(&cubicCoeff)) {
...
}
```

pmlc/ast/builder.cc
```diff
@@ -738,6 +739,38 @@ struct ProgramBuilder {
         .result();
   }
 
+  Value makeGatherOp(ExprNodeIntrinsic *node, ArrayRef<Value> operands) {
+    TensorShape shape = evaluator.getShape(node);
+    RankedTensorType resultType = builder.getRankedTensorType(shape);
+    auto op = builder.create<tile::GatherOp>(
+        loc, resultType, ValueRange{operands[0], operands[1]},
```
flaub, line 746: Use `operands.take_front(2)` instead of `ValueRange{operands[0], operands[1]}`

pmlc/ast/builder.cc
```diff
@@ -738,6 +739,38 @@ struct ProgramBuilder {
...
+          "'mode' primitive expects argument 4 to be a constant integer");
+    }
+    op.setAttr("mode", mode);
+
+    IntegerAttr axis;
+    if (!m_Constant(&axis).match(operands[2].getDefiningOp())) {
+      throw std::runtime_error(
+          "'axis' primitive expects argument 3 to be a constant integer");
+    }
+    op.setAttr("axis", builder.getIndexAttr(axis.getInt()));
```
flaub, line 769: I think it'd be better if we had a single builder function that took all the parameters during construction of the op instead of setting the attributes after constructing the op.

pmlc/ast/builder.cc
```diff
@@ -738,6 +739,38 @@ struct ProgramBuilder {
...
+    if (!m_Constant(&mode).match(operands[3].getDefiningOp())) {
+      throw std::runtime_error(
+          "'mode' primitive expects argument 4 to be a constant integer");
+    }
+    op.setAttr("mode", mode);
+
+    IntegerAttr axis;
+    if (!m_Constant(&axis).match(operands[2].getDefiningOp())) {
+      throw std::runtime_error(
+          "'axis' primitive expects argument 3 to be a constant integer");
```
flaub, line 767: All of these validation checks should happen earlier, so that if an exception is thrown, the call stack will eventually trace back to the line where the eDSL op was constructed. By throwing it here, the call stack will look like the error happened in `buildProgram`, which will be harder for users to determine where the actual error in their own code was.

To fix this, do as much validation as possible in the `ast_ops.cc` code.

flaub: Also, the error message here should be more like:

"'gather' primitive expects the 'axis' argument to be a constant integer"

YangleiZouIntel: Fixed.

pmlc/ast/builder.cc
YangleiZouIntel: Replaced with a single builder function.

pmlc/ast/builder.cc
YangleiZouIntel: Fixed

plaidml/edsl/edsl.h
YangleiZouIntel: Removed this dependency.

pmlc/ast/builder.cc
YangleiZouIntel: Fixed

plaidml/bridge/openvino/ops/embedding_bag_offsets_sum.cpp
YangleiZouIntel: Fixed

plaidml/edsl/edsl.h
YangleiZouIntel: Removed this dependency and added a copy of the enum in edsl.h.

plaidml/bridge/openvino/ops/embedding_bag_packed_sum.cpp
```diff
@@ -32,7 +32,7 @@ static OpRegistration reg("EmbeddingBagPackedSum", [](const Context& ctx) {
     with_weights = true;
   }
 
-  auto I_gathered = gather(I, indices);
+  Tensor I_gathered = edsl::gather(I, indices);
```
flaub, line 35: To be consistent, drop the `using namespace plaidml::edsl` above and use `edsl::Tensor` here.

plaidml/edsl/__init__.py
```diff
@@ -842,19 +842,31 @@ def floor(x):
...
+            2: FLOOR,
+            3: CEIL,
+            4: SIMPLE
+        cube_coeff (float): The coefficient that controls the cubic interpolation.
 
     Returns:
         Tensor: The result of the ``gather`` operation.
     """
-    return intrinsic('gather', x, y)
+    return intrinsic('gather', x, y, axis, interpolation_mode, nearest_mode, cube_coeff)
```
flaub, line 869: We should add enums to the python bindings as well.

plaidml/edsl/edsl.h
```diff
@@ -802,14 +802,86 @@ inline Tensor exp(const Tensor& x) { return intrinsic("exp", x); }
...
+  _LAST,
+};
+
+enum class NearestMode : uint64_t {
+  ROUND_PREFER_FLOOR,
+  ROUND_PREFER_CEIL,
+  FLOOR,
+  CEIL,
+  SIMPLE,
+  _LAST,
```
flaub, line 818: Drop the `_LAST`

plaidml/edsl/edsl.h
```diff
@@ -802,14 +802,86 @@ inline Tensor exp(const Tensor& x) { return intrinsic("exp", x); }
 ///
 inline Tensor floor(const Tensor& x) { return intrinsic("floor", x); }
 
+enum class InterpolationMode : uint64_t {
+  NEAREST,
+  LINEAR,
+  CUBIC,
+  _LAST,
```
flaub, line 809: Drop this

pmlc/ast/ast_ops.cc
```diff
@@ -59,22 +67,48 @@ struct IndexOp : Intrinsic {
...
+    if (!rank) {
       throw std::runtime_error(
           "'gather' requires first operand to have at least one dimension.");
     }
-    if (idxs.elementType != DataType::si32) {
-      // TODO: Handle other integer types?  Floor floats?
-      throw std::runtime_error("'gather' requires the data type for the second "
-                               "argument to be INT32.");
+    auto axis = getIntegerValue(evaluator, operands[2]);
+    if ((!axis) || (axis.getValue() >= rank) || (axis.getValue() < 0)) {
```
flaub, line 82: Drop the extra parens

pmlc/ast/builder.cc
```diff
@@ -740,6 +741,40 @@ struct ProgramBuilder {
...
+      throw std::runtime_error(
+          "'gather' primitive expects the 'nearestMode' argument "
+          "to be a constant integer");
+    }
+    if (!matchPattern(operands[5], m_Constant(&cubeCoeff))) {
+      throw std::runtime_error(
+          "'gather' primitive expects the 'cubeCoeff' argument "
+          "to be a constant float");
+    }
+    auto cubeCoeffFloat = static_cast<float>(cubeCoeff.getValueAsDouble());
```
flaub, line 770: Why is this cast necessary? We should store constants in the highest precision possible.

pmlc/dialect/tile/ir/ops.td
```diff
@@ -92,11 +92,26 @@ def GatherOp : MaterializeOp<"gather"> {
   let summary = "special gather operation";
   let arguments = (ins
     RankedTensorOf<[AnyScalar]>:$tensor,
-    RankedTensorOf<[AnyInteger]>:$dims);
+    RankedTensorOf<[AnyScalar]>:$indices,
+    DefaultValuedAttr<IndexAttr, "0">:$axis,
+    DefaultValuedAttr<InterpolationMode, "InterpolationMode::linear">:$interpolationMode,
+    DefaultValuedAttr<NearestMode, "NearestMode::round_prefer_floor">:$nearestMode,
+    DefaultValuedAttr<F32Attr, "-0.75">:$cubeCoeff
```
flaub, line 99: Use `F64Attr`

pmlc/ast/builder.cc
YangleiZouIntel: Fixed, set cubeCoeff as f64 attributes and added support for f16, f32 ,f64. f16 works on CPU. MLIR spirv not fully support f16 right now.

plaidml/edsl/__init__.py
YangleiZouIntel: Added interpolation enums in this file.

plaidml/bridge/openvino/ops/embedding_bag_packed_sum.cpp
YangleiZouIntel: Fixed

pmlc/ast/ast_ops.cc
```diff
@@ -73,13 +78,19 @@ struct GatherOp : Intrinsic {
     }
     auto tensor = shapes[0];
     auto idxs = shapes[1];
+    if (isDataTypeFloat(idxs.elementType)) {
+      if (!isDataTypeFloat(tensor.elementType)) {
```
flaub, line 82: Could you tighten this to a single if statement?

plaidml/bridge/openvino/ops/interpolate.cpp
```diff
@@ -0,0 +1,141 @@
...
+      THROW_IE_EXCEPTION << "Unsupported Interpolate InterpolateMode";
+      break;
+  }
+}
+
+edsl::NearestMode get_plaidml_nearest_mode(ngraph::op::v4::Interpolate::NearestMode nearest_mode, bool is_downsample) {
+  switch (nearest_mode) {
+    case ngraph::op::v4::Interpolate::NearestMode::round_prefer_floor:
+      return edsl::NearestMode::ROUND_PREFER_FLOOR;
+      break;
```
flaub, line 53: Drop the `break;`, it's redundant.

plaidml/bridge/openvino/ops/interpolate.cpp
```diff
@@ -0,0 +1,141 @@
...
+    case ngraph::op::v4::Interpolate::NearestMode::ceil:
+      return edsl::NearestMode::CEIL;
+      break;
+    case ngraph::op::v4::Interpolate::NearestMode::simple:
+      if (is_downsample) {
+        return edsl::NearestMode::CEIL;
+      } else {
+        return edsl::NearestMode::SIMPLE;
+      }
```
flaub, line 68: This could be a single statement:

```
return is_downsample ? edsl::NearestMode::CEIL : edsl::NearestMode::SIMPLE;
```

pmlc/ast/ast_ops.cc
YangleiZouIntel: Fixed

plaidml/bridge/openvino/ops/interpolate.cpp
YangleiZouIntel: Removed

plaidml/bridge/openvino/ops/interpolate.cpp
YangleiZouIntel: Fixed

tzerrell (commented):
This looks good to me, but @mars1026 or @flaub or @jbruestle can one of you review this as well when you have a chance?

plaidml/edsl/edsl.h
```diff
@@ -802,20 +802,85 @@ inline Tensor exp(const Tensor& x) { return intrinsic("exp", x); }
...
+  gather& axis(int axis) {
+    axis_ = Tensor(axis);
+    return *this;
+  }
+
+  ///
+  /// Set the interpolation mode for gather.
+  ///
+  gather& interpolationMode(InterpolationMode mode) {
+    interpolation_mode_ = Tensor(static_cast<uint64_t>(mode));
```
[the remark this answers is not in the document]
flaub: At the EDSL primitive level, all arguments to a primitive are Tensors (i.e. `plaidml_expr*`). So I think this is correct.

tzerrell (approved):

tzerrell (approved):
LGTM

PR #1550, opened 2020-11-11
Fix subproject usage of OpenVINO

flaub:


tzerrell (commented):
LGTM (but I see the build's failing)

jbruestle (approved):
w00t.

PR #1551, opened 2020-11-11
Sync latest OV op changes

tzerrell:
Get the changes made to the OV repo sync'd into this repo. This should be up to date with commit ~`3b8b1a36c`~`02c1b88` on that repo.

~This is currently a draft because this increases testing time by a factor of 3-4x. I'm investigating options to address this. It is otherwise ready for review.~ This adds a new target `PlaidMLFuncTestsSmoke` which is run in `check-smoke`. The longer tests remain in `PlaidMLFuncTests` which is run in `check-test`.

In a followup, I want to examine why the headers of `depth_to_space.cpp` need to be included in a lint-failing order.

### DIFF
--- a/plaidml/bridge/openvino/CMakeLists.txt
+++ b/plaidml/bridge/openvino/CMakeLists.txt
@@ -21,7 +21,7 @@ if(NOT InferenceEngineDeveloperPackage_FOUND)
   FetchContent_Declare(
     openvino
     GIT_REPOSITORY https://github.com/plaidml/openvino.git
-    GIT_TAG        507de2630fb341b4732ace986af22cc6f71ca459
+    GIT_TAG        02c1b88ca69553ab90ad15b18328e6e3169ae205
   )
   if(LOCAL_OPENVINO_DIR)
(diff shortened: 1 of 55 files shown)


tzerrell:
Smoke test functionality added. Remains a draft because a few more recent ops have not yet been synced, and because ExtractImagePatches is failing its smoke tests. The failure appears to have been introduced by the interaction of ExtractImagePatches and #1556; I'll examine it further tomorrow.

flaub:
LGTM

tzerrell:
Should I change the per-PR CMake CI to `check-smoke` instead of `check-test`? The CI time hasn't come in yet but when I ran it locally just the OV part of `check-test` came in at just under 18 minutes, and that's only going to get longer (`check-smoke` ran in under a minute). But I don't know if there's other `check-test` stuff that we need.

flaub:
Let's monitor it, but right now it took 13 minutes on buildkite. Luckily it's running in parallel with the other tests, so it actually doesn't effect overall PR build-time.

flaub (approved):

PR #1553, opened 2020-11-11
Add ABI dialect

earhart:


pmlc/dialect/abi/ir/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+pml_tblgen_library(
+  NAME ops-gen
+  TD_FILE ops.td
+  OPTS -dialect=comp
+  OUTS
+    -gen-op-decls ops.h.inc
+    -gen-op-defs ops.cc.inc
+    -gen-dialect-decls dialect.h.inc
+)
```
flaub, line 22: Thanks for adding this too!

pmlc/dialect/abi/ir/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+    dialect.cc
+  DEPS
+    ::ops-gen
+  PUBLIC
+)
+
+pml_tblgen_library(
+  NAME ops-gen
+  TD_FILE ops.td
+  OPTS -dialect=comp
```
flaub, line 17: This is probably wrong

pmlc/dialect/abi/ir/ops.td
```diff
@@ -0,0 +1,85 @@
...
+      return mlir::cast<DoneOp>(finiRegion().back().getTerminator());
+    }
+  }];
+}
+
+def ABI_DoneOp : ABI_Op<"done", [Terminator, HasParent<"pmlc::dialect::abi::LoopOp">]> {
+  let summary = "Terminates an abi.loop operation body or fini region";
+}
+
+def ABI_CreateNetworkOp : ABI_Op<"create_network", [Terminator, HasParent<"pmlc::dialect::abi::LoopOp">]> {
```
flaub, line 80: I think we can drop the full namespace, recently MLIR fixed namespacing so that ops get defined within the proper namespace.

pmlc/dialect/abi/ir/CMakeLists.txt
earhart: Oops -- thanks for the catch.  (What's the right way to test these?)

pmlc/dialect/abi/ir/ops.td
```diff
@@ -0,0 +1,85 @@
...
+  IsolatedFromAbove, HasParent<"ModuleOp">, RecursiveSideEffects]> {
+  let summary = "Repeat runs of a network";
+  let description = [{
+    The "abi.loop" operation represents repeated runs of a program,
+    taking new inputs and generating new outputs each time.
+    This allows loop-invariant code motion to move setup and teardown
+    operations outside of the loop, so that their cost is ammortized over
+    many runs of the program.
+  }];
+  let regions = (region AnyRegion:$initRegion, AnyRegion:$bodyRegion, AnyRegion:$finiRegion);
```
flaub, line 29: Hmm, why multiple regions? Seems like we want one region and have the code be hoisted to either pre- or post- this op.

pmlc/dialect/abi/ir/ops.td
```diff
@@ -0,0 +1,85 @@
...
+      std::vector<mlir::Type> result;
+      auto arrayAttr = getAttrOfType<mlir::ArrayAttr>(getNetworkFieldTypesAttrName());
+      if (arrayAttr) {
+        for (auto attr : arrayAttr) {
+          if (auto tyAttr = attr.dyn_cast<mlir::TypeAttr>()) {
+            result.emplace_back(tyAttr.getValue());
+          }
+        }
+      }
+      return result;
```
flaub, line 43: Please move C++ definitions to a `.cc` file, we'd like to minimize the amount of C++ code in `.td` files.

earhart: Will do

pmlc/dialect/abi/ir/ops.td
```diff
@@ -0,0 +1,85 @@
...
+      return mlir::cast<CreateNetworkOp>(initRegion().back().getTerminator());
+    }
+
+    DoneOp getFiniTerminator() {
+      return mlir::cast<DoneOp>(finiRegion().back().getTerminator());
+    }
+  }];
+}
+
+def ABI_DoneOp : ABI_Op<"done", [Terminator, HasParent<"pmlc::dialect::abi::LoopOp">]> {
```
jbruestle, line 76: Typically, the name is always 'YieldOp' for things like this (or abi.yield in opcode form).

pmlc/dialect/abi/ir/ops.td
jbruestle, line 80: Why not use yield for all three regions?

jbruestle (commented):
Overall design points: 1) Why not just have 'init' be everything before the op, and fini be everything after and the body be the body?  2) Presuming there is a good reason for not doing that, why have different terminators?

pmlc/dialect/abi/ir/ops.td
earhart: Several reasons:

1. This approach makes the abi-to-functions pass simpler, since there's a direct mapping from regions to functions, the region arguments correspond to the function arguments, &c -- no need to go sniffing around to find the loop.

1. This approach is also automatically correct-by-construction: an abi.loop is an isolated-from-above child of a ModuleOp; there's no correct way to give it any other parent, to have multiple abi.loop operations within a single function, &c.

1. This approach makes the semantics of what's going on more clear: we're recasting a function into a separate initialization, execution, and finalization section, at the top level, and then transforming it from there.  The alternative, having a single function wrapping a loop, still "looks like" a single function that you call and get some result from; it's not obvious that additional lowering is needed in order to match the ABI expected by JitExecutable.

pmlc/dialect/abi/ir/ops.td
earhart: The init region is logically creating and returning a network object; doing so is what terminates the region.  The others absolutely must not return anything.  We could combine the operations and use some explicit validation logic, but ops are cheap.

pmlc/dialect/abi/ir/ops.td
earhart: They're not yielding a value... and `abi.terminator` would imply that it could be used in the other regions, which isn't really the case.

earhart:
@jbruestle See the commend thread on the multiple region definitions in ops.td

pmlc/dialect/abi/ir/ops.td
flaub: Wait, so this is intended to replace a `func` op?

earhart: Yup.  The conversion takes the body of a `func`, puts it into an `abi.loop`'s body region, sets up arguments, and calls it a day.

I'm thinking the next PR in this series will be that conversion pass, with a lit test to show what's going on -- think I should combine it with the current PR?

pmlc/dialect/abi/ir/ops.td
jbruestle: I see now that createNetwork is variadic any type input and the Done is not.  Still, from a convention/idiomatic perspective a single Yield op that is variadic any and must have 0 inputs for the other two regions would be more in line with all the other upstream dialects, and it's a trivial validation either way.

flaub:
I think I need to see more examples before I can decide if this a good design or not. Can you write up a few `.mlir` files and the before and after? It's unclear what form we expect things to be in before we get to this dialect and then what form we expect below.

pmlc/dialect/abi/ir/ops.td
jbruestle: Or alternately, make two: YieldOp (abi.yield) and Terminate (abi.terminate), but at least use standard naming.

flaub: Where would `createNetwork` be used?

pmlc/dialect/abi/ir/ops.td
jbruestle: Yeah, I also didn't catch that, although it's all in the PR (HasParent<"ModuleOp"> for example).  Question: Does this design actually work with LICM?  I realize there is some concerns on 'destructor' motion, so we might need our own transforms anyway, but I'd like LICM to work if possible.

flaub: I agree that we should be able to use the standard LICM pass. If we need to handle the split of alloc/dealloc, I think that's a separate thing that we can address in some other pass.

pmlc/dialect/abi/ir/ops.td
jbruestle: It's the 'yield' for init...  But I agree that the name is questionable.

flaub:
If this is meant to replace `std.func`, why not call it `abi.func`?

pmlc/dialect/abi/ir/ops.td
jbruestle: scf.for, scf.if, etc always uses scf.yield for both value returning termination and for non-value returning termination.  Yield does not imply values are being sent IMO.

flaub:
Should we just use `api.func` from the start and not even bother with `std.func` until we get down to lowering to LLVM?

pmlc/dialect/abi/ir/ops.td
jbruestle: Basically for scf.if, both branches must yield identical (possibly 0) results, which will be the result of the if, for scf.for, yield is used to pass loop carried variables, etc.  i.e. yield is a generic terminator and the semantics depend on the parent op.

flaub:
I'm also not sure this is really related to the ABI. It's more like a way to represent a repeated invocation of some program. So maybe we could improve the name of this dialect somehow. I don't have any other great suggestions tho.

pmlc/dialect/abi/ir/ops.td
jbruestle: I'm open to the idea of making abi.loop be the primary entry point and having it live directly in the module, but I do worry a bit about losing all the functionality that exists on FuncOp.

flaub: Agreed, how to we represent attributes on arguments? This is something FuncOp supports. Is there a FunctionLike interface or trait that we can use to acquire the features we want from standard FuncOp?

earhart:
I'm thinking of "ABI" as "The interface that things that use the network have to understand and deal with -- both the API and the details of how objects are handed across the boundary to the LLVM-produced executable code".  Part of that interface is the whole concept of an init/exec/fini split.

I'll add some .mlir files.

pmlc/dialect/abi/ir/ops.td
earhart: SGTM -- yield for the init region, terminate for the others.

pmlc/dialect/abi/ir/ops.td
earhart: It does not work with the standard LICM pass as-is -- I was saving adding LoopLikeInterface for a subsequent PR (although I do have it implemented.

Also, as @flaub notes, LICM doesn't handle alloc/dealloc; it also doesn't handle anything with memory side-effects.  I wound up adding my own pass, along with an interface to add support for "move operations to epilog".

TL;DR: Later.

earhart: What's the use case for attributes on arguments at this stage of the lowering?  (NB This is pretty close to the point where we're going to be converting to `llvm.func` and generating code.)

earhart: ... also, note that `FunctionLike` implies a single region, a set of arguments and a set of results, &c; we might want to add argument attribute support, but this op isn't `FunctionLike`.

pmlc/dialect/abi/ir/ops.td
earhart: Done

earhart:
I haven't been calling it `abi.func` because it's not `FunctionLike`.  OTOH, I don't have a strong opinion here.

pmlc/dialect/abi/ir/ops.td
jbruestle: Ok on further thought here is my feelings on LICM: I think we could make most (all?) things work using LICM via cleverness (i.e. making allocs no-side-effect and adding deallocs to fini only after we are done, after all, we already begin with allocs only and add deallocs later).  However, this is sometimes awkward since really we want 'idempotent' ops to be moved to do-once, not just 'no-side-effect' exactly, and things like 'alloc' actually *are* marked as side-effect, so even if we wanted to use LICM, we would need a special version of alloc.  Moreover, LICM is < 130 lines of code (including a lot of comments and boilerplate), so it's not like partially duplicating bits of it is a major engineering cost.

Conclusion, reuse is probably not worth it here.  At that point, the issue of 1 region vs 3 regions becomes less relevant from that perspective, and there is definitely something nice about having abi.loop be top level in the module.

flaub:
How do we deal with constant buffers? These are program arguments marked with `tile.const` attributes at the top level.

pmlc/dialect/abi/ir/example.mlir
```diff
@@ -0,0 +1,349 @@
...
+    xsmm.gemm.invoke.f32 %0, %arg2[%c0, %c0] = %arg0[%c0, %c0], %arg1[%c0, %c0] : (memref<8x16xf32>, memref<16x32xf32>) -> memref<8x32xf32>
+    return
+  }
+  func @plaidml_rt_thread_num() -> index
+}
+
+// After lowering to ABI:
+
+module @dot {
+  "abi.loop"() ( {
```
flaub, line 53: Can you make a custom printer/parser for this op? It's pretty hard to read in generic form.

flaub:
Will this work for targets that don't use LLVM?

pmlc/dialect/abi/ir/example.mlir
jbruestle: Agree, checkout the Declarative Assembly Form bit in: 'https://mlir.llvm.org/docs/OpDefinitions/'. It's actually very easy, even for things with regions now.

pmlc/dialect/abi/ir/ops.td
earhart: FWIW: LICM works on arbitrary LoopLike ops, and uses the op for manipulations, so the multiple regions and top-level aspect aren't incompatibilities at all; I used LICM early on, and only implemented a separate hoisting pass to handle ops with side-effects and where I needed move-to-epilog semantics.

But I still (clearly :-) like having abi.loop as a top-level op corresponding to "How we're going to run this network".

earhart:
> How do we deal with constant buffers? These are program arguments marked with `tile.const` attributes at the top level.

I've added an example for this -- they're passed to the `init` region, and then forwarded to the `body` and `fini` regions, making operations over constant buffers hoistable to the `init` region.

earhart:
> Will this work for targets that don't use LLVM?

Currently, the code I have lowers `abi.loop` directly into LLVMIR.  Going to standard functions and then to LLVMIR requires a few additional operations and types, but I think it's probably worth doing at some point, so that we have more flexibility in the lowering -- i.e. for targets where we're not using LLVMIR, or for heterogeneous targets.

pmlc/dialect/abi/ir/ops.td
```diff
@@ -0,0 +1,70 @@
...
+    This allows loop-invariant code motion to move setup and teardown
+    operations outside of the loop, so that their cost is ammortized over
+    many runs of the program.
+  }];
+  let regions = (region AnyRegion:$initRegion, AnyRegion:$bodyRegion, AnyRegion:$finiRegion);
+  let arguments = (ins TypeArrayAttr:$networkFieldTypes);
+  let builders = [OpBuilder<"">];
+  let skipDefaultBuilders = 1;
+  let assemblyFormat = [{
+     `init` $initRegion `yielding` $networkFieldTypes `to` `body` $bodyRegion `and` `to` `fini` $finiRegion attr-dict
```
flaub, line 34: I'd prefer that we use less verbiage and keep this more like assembly language where we have ops with symbols instead of using verbs in between parts.

I also think we should include block arguments in the custom syntax so that the default printing of block arguments is avoided.

earhart: Verbiage: decreased.

I'm not sure how to include block arguments in the custom syntax -- got an example of where this is done?

jbruestle (approved):

pmlc/dialect/abi/ir/ops.td
flaub: Checkout the AffineParallelOp for an example.

PR #1556, opened 2020-11-12
Integrate layer op into OpenVINO

flaub:


plaidml/bridge/openvino/CPPLINT.cfg
```diff
@@ -0,0 +1 @@
+filter=-build/include_order,-build/include_subdir,-build/namespaces,-runtime/references
```
tzerrell, line 1: Should I unwind the various existing `//NOLINT` comments for these? (as a low priority task) -- in the OV ops directory, primarily, as it looks like you've done this already for the core files.

jbruestle (approved):

tzerrell (commented):
LGTM (but I see test failures)

flaub:
Found a few more issues when running all the func tests. Will push fixes soon.

plaidml/bridge/openvino/CPPLINT.cfg
flaub: Up to you, I'm not sure if it matters much. WDYT about ignoring these lints in the first place?

tzerrell: I see arguments both ways -- these keep the code cleaner but require us to either divert from OV style or add a bunch of NOLINTs.

Maybe remove just `include_order`? I've seen some oddities in the order of `#includes` and I think it might be good to have those explicitly called out with at a minimum NOLINT comments

tzerrell (approved):

PR #1558, opened 2020-11-13
Add lower-to-abi pass

earhart:


pmlc/dialect/abi/transforms/passes.td
```diff
@@ -0,0 +1,20 @@
+#ifndef __PMLC_DIALECT_ABI_PASSES__
+#define __PMLC_DIALECT_ABI_PASSES__
+
+include "mlir/Pass/PassBase.td"
+
+def LowerToABIPass : Pass<"lower-to-abi", "mlir::ModuleOp"> {
```
flaub, line 6: The convention is for passes defined in a dialect to have the command line arg be prefixed with the name of the dialect. So it would be: `--abi-lower-to-abi` in this case.

pmlc/dialect/abi/tests/dot.mlir
```diff
@@ -0,0 +1,101 @@
...
+    return
+  }
+  func @plaidml_rt_thread_num() -> index
+}
+
+// CHECK-LABEL: module @dot {
+//  CHECK-NEXT:   abi.loop init  {
+//  CHECK-NEXT:   ^bb0(%arg0: !llvm.ptr<i8>):  // no predecessors
+//  CHECK-NEXT:     abi.yield
+//  CHECK-NEXT:   } yield [] body  {
```
flaub, line 54: This is confusing, can we just make the syntax be:

```
abi.loop init {
} exec {
} fini {
}
```

pmlc/dialect/abi/transforms/lower_to_abi.cc
```diff
@@ -0,0 +1,132 @@
...
+  auto loc = mainFunc.getLoc();
+  auto builder = mlir::OpBuilder{mainFunc};
+  auto loopOp = builder.create<abi::LoopOp>(loc);
+
+  // Initialize the loop's body by taking the main function's body.
+  loopOp.bodyRegion().takeBody(mainFunc.getBody());
+
+  // A function's blocks can end with std.return; since std.return expects its
+  // parent to be std.func, we need to replace any cloned top-level std.return
+  // operations with abi.done.
```
flaub, line 49: Typo, i think `abi.done` got renamed?

pmlc/util/ids.h
```diff
@@ -0,0 +1,10 @@
+// Copyright 2020 Intel Corporation
+#pragma once
+
+// A collection of well-known identifiers used in codegen.
```
flaub, line 4: Do we really need a whole separate file just for this?

I think this probably belongs in `compiler/program.h`

pmlc/dialect/abi/transforms/lower_to_abi.cc
```diff
@@ -0,0 +1,132 @@
...
+  // intrinsically parameters of the body entry block.  Constant arguments do
+  // not vary; we pass them to the initializtion block, which can then pass them
+  // on to the body as needed.
+  //
+  // At this point, all of the arguments should be memrefs, except for a
+  // possible initial device argument -- which, if present, is always passed
+  // to the initialization block.
+  auto *bodyEntryBlock = loopOp.getBodyEntryBlock();
+  unsigned argIdx = 0;
+  mlir::Identifier constAttrId = builder.getIdentifier("tile.const");
```
flaub, line 72: This seems out of place

flaub: And also, why do we need to create an identifier? can we just use the string version directly?

pmlc/dialect/abi/transforms/lower_to_abi.cc
```diff
@@ -0,0 +1,132 @@
...
+      returnOp.erase();
+    }
+  }
+
+  // Add an entry block for the init region.
+  auto *initEntryBlock = builder.createBlock(&loopOp.initRegion());
+
+  // Non-constant arguments will vary from invocation to invocation; they are
+  // intrinsically parameters of the body entry block.  Constant arguments do
+  // not vary; we pass them to the initializtion block, which can then pass them
```
flaub, line 64: typo: initialization

pmlc/dialect/abi/transforms/lower_to_abi.cc
```diff
@@ -0,0 +1,132 @@
...
+      auto ty = arg.getType();
+      auto newArg = bodyEntryBlock->insertArgument(networkArgs.size(), ty);
+      arg.replaceAllUsesWith(newArg);
+      bodyEntryBlock->eraseArgument(argIdx + 1);
+      initEntryBlock->addArgument(ty);
+      networkArgs.emplace_back(
+          initEntryBlock->getArgument(initEntryBlock->getNumArguments() - 1));
+      networkTypes.emplace_back(ty);
+    }
+    ++argIdx;
```
flaub, line 109: Use `llvm::enumerate` instead of tracking the argIdx independently?

flaub: Actually, you can just use `bodyEntryBlock->getArgNumber()`

pmlc/dialect/abi/transforms/passes.td
earhart: Done

pmlc/dialect/abi/tests/dot.mlir
earhart: We need the field types somewhere; it kinda makes sense to put them after the init region.  (There are some hacky alternatives, but it's simplest to just keep the authoritative version in an attribute.)

pmlc/dialect/abi/transforms/lower_to_abi.cc
earhart: Fixed; thanks!

pmlc/util/ids.h
earhart: Files are cheap, and `JitExecutable` will need access to these identifiers; it'd be nice to not take a dependency on `compiler`.

pmlc/dialect/abi/transforms/lower_to_abi.cc
earhart: Is there another good way of distinguishing constant weight tensors?

Using an identifier just means a little less work for the compiler, since it's slightly faster to compare interned strings (although honestly, it's not a huge cost).

pmlc/dialect/abi/transforms/lower_to_abi.cc
earhart: Thanks!  Fixed.

pmlc/dialect/abi/transforms/lower_to_abi.cc
earhart: Done; thanks!

pmlc/dialect/abi/tests/dot.mlir
flaub: Ok, so:
```
abi.loop init (%arg0 : type, ...) {
} exec (%arg0 : type, ...) {
} fini (%arg0: type, ...) {
}
```

pmlc/util/ids.h
flaub: JitExecutable should depend on compiler shouldn't it? We need the `Program` to be passed along, that's where the constant buffers live.

pmlc/dialect/abi/transforms/lower_to_abi.cc
flaub: You should look at the definition of the string-based `getStringAttr`, it calls `getIndentifier` under the hood. There's never a time where the compiler does more work because it's comparing strings, identifiers are always used. 

My point is that if there's no need to have an identifier, it's just more boilerplate that should be avoided.

flaub: And to be clear, the 'out of place' comment is just that this line of code isn't used into a little bit later, so I'd move this line close to its usage. However, if we just use the string-based call, this whole line would go away.

pmlc/dialect/abi/tests/dot.mlir
earhart: It's a little tricky with exec, since there's no clear division between the block arguments coming from the initialized network object and the per-execution block arguments, other than by looking at the networkFieldTypes attribute, which then needs to be stored somewhere.  (We could look at the fini region; that's the hacky alternative I mentioned earlier, but that seems kind of hacky.  An earlier version of this code actually looked at the terminator of the init region, but again, that seemed hacky; I'd rather just be very explicit by having a single authoritative attribute.)

Is there an easy way to use the declarative asm syntax to get nice function-style arguments like this?  If so, I'd love to use it.

pmlc/dialect/abi/transforms/lower_to_abi.cc
earhart: I did in fact trace through the code here.  It looked to me like it grounded out in an llvm::lower_bound and then a strcmp, but perhaps I missed something.  I can move the identifier creation, though.

pmlc/util/ids.h
earhart: JitExecutable currently depends on compiler, but IMHO that's an antipattern: `Program::compile` isn't an intrinsic part of a `Program` object, it's a member of Program because it has to do with programs and humans like categorizing things; it could be easily be a free function.  (I do this all the time myself, which is why I'm on somewhat guard against it.  :-)

I also don't think these identifiers are necessarily part of Program -- unless we want to constrain Program to always providing these particular interfaces?  But I think of them as a handshake between a particular target and JitExecutable.

flaub: The constant buffers lifetime are tied to the `Program`. We need to pass the `program` to an `executable` so that the program can be executed. I don't see the antipattern here.

pmlc/dialect/abi/tests/dot.mlir
jbruestle: So I think readability of the IR is very important here.  With regard to the types attribute, the more that I think about it, since the 'networkFieldTypes' is 1-to-1 to the fini params (and a validation error on mismatch), I think we should remove the attribute and instead add an accessor to the op that grabs the fini arg types, which will end up being equally easy to use in the code, and doesn't duplicate data that is required to match anyway.  With regard to the textual format, the ODS assembly is somewhat poorly documented, and so I'd try it out and see what it can do.  I believe that it might just be:
let assemblyFormat = [{ `init` $initRegion  `body` $bodyRegion `fini` $finiRegion attr-dict }];
If that doesn't work (i.e. formats argument anonyingly, has ^bb goo, etc), I'd mess around a bit to try to fix it, and if you can't work around it, write a full custom parser/printer.  Basically, the default format is very annoying to read/write, and it's worth a few hours of pain writing parser/printer logic to fix it.

earhart: The current impl uses ODS assembly.  :-). I don't think I have the time right now to spend a few hours writing the parser/printer logic for something beyond what ODS can do.  But replacing networkFieldTypes with the "use the fini types" hack is pretty straightforward; I'll go ahead and implement it.

flaub: There are already traits that allow the ODS assembly format to be smarter by saying things like "these operand types must match those result types", so perhaps we can use some existing traits for this.

pmlc/util/ids.h
earhart: The antipattern is that `Program::compile()` exists.  It would be better for `Program` to be moved somewhere else, and to think of compilation as a transformation applied to a `Program` datastructure instead of something that's intrinsic to the definition of a `Program`.

(Or is there another reason why `Program` is defined under `pmlc/compiler`?)

pmlc/dialect/abi/tests/dot.mlir
jbruestle: Once the networkFieldTypes bit is in I will fix the formatting on this branch.

pmlc/util/ids.h
flaub: I still don't see the issue.

flaub (approved):

flaub:

PR #1560, opened 2020-11-17
Adding Gen pipeline options

adstraw:
Usage e.g.  PLAIDML_TARGET="intel_gen_ocl_spirv{use-block-ops=false spirv-version=120}"

### DIFF
--- a/pmlc/compiler/program.cc
+++ b/pmlc/compiler/program.cc
@@ -149,13 +149,13 @@ void Program::compile(StringRef targetNameAndOptions, bool collectPasses,
                         /*out=*/llvm::errs());
   }
 
-  auto targetNameSize = targetNameAndOptions.find('{');
-  auto targetName = targetNameAndOptions.substr(0, targetNameSize);
-  auto targetOptions = targetNameAndOptions.substr(targetNameSize);
+  auto begOpts = targetNameAndOptions.find('{');
+  auto targetName = targetNameAndOptions.substr(0, begOpts);
+  auto targetOptions = targetNameAndOptions.substr(begOpts);
 
-  // TODO: this is a bit hacky
   if (!targetOptions.empty()) {
-    targetOptions = targetOptions.substr(1, targetOptions.size() - 2);
+    auto endOpts = targetOptions.find('}');
+    targetOptions = targetOptions.substr(1, endOpts - 2);
   }
 
   target = resolveTarget(targetName);
--- a/pmlc/compiler/program.cc
+++ b/pmlc/compiler/program.cc
@@ -153,9 +153,11 @@ void Program::compile(StringRef targetNameAndOptions, bool collectPasses,
   auto targetName = targetNameAndOptions.substr(0, begOpts);
   auto targetOptions = targetNameAndOptions.substr(begOpts);
 
+  // if target options are specified
   if (!targetOptions.empty()) {
+    // trim off curly braces
     auto endOpts = targetOptions.find('}');
-    targetOptions = targetOptions.substr(1, endOpts - 2);
+    targetOptions = targetOptions.substr(1, endOpts - 1);
   }
 
   target = resolveTarget(targetName);
--- a/pmlc/compiler/program.cc
+++ b/pmlc/compiler/program.cc
@@ -108,9 +108,9 @@ static StringRef getDiagKindStr(DiagnosticSeverity kind) {
   llvm_unreachable("Unknown DiagnosticSeverity");
 }
 
-void Program::compile(StringRef targetName, bool collectPasses,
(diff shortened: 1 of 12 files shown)


pmlc/target/intel_gen_ocl_spirv/spirv_target.cc
```diff
@@ -15,12 +15,20 @@ namespace {
...
+  IntelGenOclAddSpirvTarget() = default;
+  explicit IntelGenOclAddSpirvTarget(unsigned spirvVersion) {
+    this->spirvVersion = spirvVersion;
+  }
   void runOnOperation() {
     auto target_env = getOperation().getAttrOfType<spirv::TargetEnvAttr>(
         spirv::getTargetEnvAttrName());
     if (!target_env) {
+      auto version = spirv::Version::V_1_5;
+      if (spirvVersion == 120) {
```
abialas1, line 27: Do we want to fix on 1.5 and 1.2 only or make it more generic?

pmlc/target/x86/register.cc
```diff
@@ -20,7 +20,10 @@ static PassPipelineRegistration<> passPipelineReg(kPassPipelineTargetName,
 
 class Target : public compiler::Target {
 public:
-  void buildPipeline(mlir::OpPassManager &pm) { pipelineBuilder(pm); }
+  void buildPipeline(mlir::OpPassManager &pm, llvm::StringRef targetOptions) {
+    assert(targetOptions.empty());
```
abialas1, line 24: Is this assert needed?

abialas1:
Looks good to me, added some minor comments

pmlc/conversion/gpu_to_spirv/gpu_to_spirv.cc
```diff
@@ -31,6 +31,22 @@ struct StdxSubgroupBroadcastOpConversion final
     : public SPIRVOpLowering<stdx::SubgroupBroadcastOp> {
   using SPIRVOpLowering<stdx::SubgroupBroadcastOp>::SPIRVOpLowering;
 
+  LogicalResult
```
jbruestle, line 34: Maybe template on the op to generate to reduce code duplication

jbruestle (commented):
Seems good to me minus my one comment, adding Frank for additional visibility.

pmlc/target/intel_gen_ocl_spirv/spirv_target.cc
adstraw: I feel like this is OK.  The `auto version = ` can signify the default (latest) version and we can add any other versions needed, currently just 1_2, in the conditional statements below.  Not sure how to make it more general, either.  I don't think there's a way to pass `spriv::Version` through the PassOptions directly.  Meaning, we have to translate from int (or string) to `spriv::Version` at some level.  Meaning, we need this case logic at some level.

pmlc/conversion/gpu_to_spirv/gpu_to_spirv.cc
adstraw: Done.

pmlc/target/x86/register.cc
adstraw: Nope, removed.

pmlc/compiler/program.cc
```diff
@@ -149,8 +149,19 @@ void Program::compile(StringRef targetName, bool collectPasses,
...
+  auto begOpts = targetNameAndOptions.find('{');
+  auto targetName = targetNameAndOptions.substr(0, begOpts);
+  auto targetOptions = targetNameAndOptions.substr(begOpts);
+
+  // if target options are specified
+  if (!targetOptions.empty()) {
+    // trim off curly braces
+    auto endOpts = targetOptions.find('}');
+    targetOptions = targetOptions.substr(1, endOpts - 1);
+  }
```
flaub, line 161: This is the only part that gives me pause. I wonder if we can use the existing parse pipeline functionality of MLIR for this.

flaub:
LGTM, but we might want to come up with a more 'MLIR' way to parse the pipeline options. When I have some cycles I can investigate this, but I don't think it should block landing this.

pmlc/compiler/program.cc
adstraw: This code separates target name from options.  E.g. PLAIDML_TARGET="mytarget{myoption1=true myoption2=7}" you get two strings targetName = "mytarget" and targetOptions = "myoption1=true myoption2=7".  I did not find an MLIR utility that does this.  I do parse the targetOptions with an MLIR utility called `createFromString`.

jbruestle (approved):
Looks good!

PR #1562, opened 2020-11-18
Fix precision when IE/OV disagree

tzerrell:
Fix the precision issues when InferenceEngine provides/requests an input/output precision different from the precision on the Parameter or Result node in nGraph. Marked as draft for now as it doesn't yet resolve the FP16 issues, but it does already fix the other single_layer_test precision issues we've seen.

@abialas1 I'm not sure what test case you were fixing in your branch https://github.com/plaidml/plaidml/tree/abialas/benchmark_app_gpu; does this address what you need for that? I'd also be happy to run your test(s) myself if you let me know what they are.

### DIFF
--- a/plaidml/bridge/openvino/plaidml_builder.cpp
+++ b/plaidml/bridge/openvino/plaidml_builder.cpp
@@ -34,6 +34,10 @@ class ProgramBuilder {
 
   const InferenceEngine::ICNNNetwork& network;
 
+  // cache network input/output info
+  InferenceEngine::InputsDataMap networkInputs;
+  InferenceEngine::OutputsDataMap networkOutputs;
+
   // Lets us look up the PlaidML tensor by the name of the node that produces it and the index of which output it is
   std::map<std::pair<std::string, size_t>, plaidml::edsl::Tensor> tensorMap;
 
@@ -41,7 +45,10 @@ class ProgramBuilder {
   std::map<std::string, plaidml::edsl::Tensor> tensorIONameMap;
 };
 
-ProgramBuilder::ProgramBuilder(const InferenceEngine::ICNNNetwork& network) : network(network) {}
+ProgramBuilder::ProgramBuilder(const InferenceEngine::ICNNNetwork& network) : network(network) {
+  network.getInputsInfo(networkInputs);
+  network.getOutputsInfo(networkOutputs);
+}
 
 plaidml::Program ProgramBuilder::build() {
   std::shared_ptr<const ngraph::Function> func = network.getFunction();
@@ -58,15 +65,10 @@ plaidml::Program ProgramBuilder::build() {
     }
   }
 
-  InferenceEngine::InputsDataMap networkInputs;
-  network.getInputsInfo(networkInputs);
   std::vector<plaidml::edsl::Tensor> inputs;
   for (const auto& kvp : networkInputs) {
     inputs.push_back(tensorIONameMap.at(kvp.first));
   }
-
-  InferenceEngine::OutputsDataMap networkOutputs;
-  network.getOutputsInfo(networkOutputs);
   std::vector<plaidml::edsl::Tensor> outputs;
   for (const auto& kvp : networkOutputs) {
     outputs.push_back(tensorIONameMap.at(kvp.first));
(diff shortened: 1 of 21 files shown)


plaidml/bridge/openvino/plaidml_builder.cpp
```diff
@@ -34,14 +34,21 @@ class ProgramBuilder {
 
   const InferenceEngine::ICNNNetwork& network;
 
+  // cache network input/output info
+  InferenceEngine::InputsDataMap networkInputs;
+  InferenceEngine::OutputsDataMap networkOutputs;
```
flaub, line 39: It's better not to put state onto the class if it's only ever used in a single function.

plaidml/bridge/openvino/plaidml_builder.cpp
```diff
@@ -91,8 +93,12 @@ void ProgramBuilder::handleConstant(const std::shared_ptr<ngraph::Node>& node) {
 
 void ProgramBuilder::handleParameter(const std::shared_ptr<ngraph::Node>& node) {
   IE_ASSERT(node->get_output_size() == 1);
-  std::vector<int64_t> dims{node->get_shape().begin(), node->get_shape().end()};
-  plaidml::DType type = to_plaidml(node->get_element_type());
+  // TODO: Decide if we want to compare to the nGraph dims & type and issue warnings
+  // std::vector<int64_t> ng_dims{node->get_shape().begin(), node->get_shape().end()};
+  // plaidml::DType ng_type = to_plaidml(node->get_element_type());
+  auto inputDesc = networkInputs[node->get_friendly_name()]->getTensorDesc();
```
flaub, line 99: Can't we get the tensor descriptor directly from the current node?

plaidml/bridge/openvino/plaidml_builder.cpp
tzerrell: These are used in every call to `handleParameter` (for inputs) or `handleOutput` (for outputs); my impression was `get*putsInfo` were vaguely expensive functions, hence the caching, but I can certainly call those on each use instead.

plaidml/bridge/openvino/plaidml_builder.cpp
tzerrell: The whole thing this PR is trying to fix is that IE can provide/request input/output tensors of different data types than the types of the Parameter and Result nodes in the nGraph network representing those tensors. I.e. a Result node might be fp16 but the IE TensorDesc for that node might be fp32; in this case we insert a cast to fp32 that is not present in the network graph.

tzerrell:
Many FP16 tests are working, and most errors appear unrelated to this (e.g., failure b/c we don't currently lower f16 acos). However, some errors persist; and I'm investigating further.

Notably, FP16 Ceiling produces an accuracy error where we produce -10 and the nGraph interpreter produces -9. I worry that this may be due to the inputs being provided as f32, so we cast to f16 to run the f16 network, which may be introducing a precision error where we are slightly below -10 and the nGraph interpreter (which I believe never switches to f16 even on f16 networks) has values slightly above -10, causing the Ceilings to differ. If my diagnosis is correct (unclear), I'm not sure what if anything we would want to do about it.

plaidml/bridge/openvino/plaidml_builder.cpp
flaub: Yes, but I'm wondering if we can get the Tensor Descriptor from the node, and not look at the shape/type. But I could see how maybe the network info differs from the node and that they are out of sync.

plaidml/bridge/openvino/plaidml_builder.cpp
flaub: I see, I didn't realize they were used also in `handleParameter` until after I commented here.

abialas1:
Hi Tim, Thanks, that covers my case. I do see now the same behavior as with changes from my branch. I am using benchmark_app with resnet50 ran on Windows, this is my only testing case right now.

tzerrell:
FP16 Convolution still failing a number of cases for unclear reasons; I need to continue investigating. Some activations also fail for FP16 due to lack of lowering. Added skips for these failing tests, so this could be landed if we want.

flaub:
This dropped from my radar, but is this still needed?

tzerrell:
Whoops, thought I had closed this. Thanks!

This is largely addressed by #1629, with some tests added in #1644. FP16 testing isn't complete but it seems unrelated to the #1629 change so no need to keep this open.

PR #1564, opened 2020-11-19
Extract Layer operation from tile dialect into its own dialect

mars1026:


flaub:
LGTM, I can approve if it's ready

jbruestle (approved):
Nice work.

PR #1566, opened 2020-11-19
Openvino Mish activation via eltwise implementation

mwyi:


tzerrell (approved):
We will eventually need an intrinsic (since log(1+exp) has major numeric issues) but this is good for now.

PR #1567, opened 2020-11-19
Openvino SoftPlus activation via eltwise

mwyi:


tzerrell (approved):
As with Mish (https://github.com/plaidml/plaidml/pull/1566), this looks good but will eventually need an intrinsic (they could potentially share the intrinsic, it's the log(1+exp) portion of Mish that's concerning).

PR #1569, opened 2020-11-20
workaround for multiple openmp loadings

drgone:
see: https://docs.openvinotoolkit.org/2018_R5/_docs_MO_DG_Known_Issues_Limitations.html


brianretford (approved):
Great work, thanks.

PR #1570, opened 2020-11-20
Cast result branches of select

tzerrell:
The 2 output branches of `select` need to be of the same type. In the long run, we may want to handle this before we reach the layers with select by using FLOATX type (or INTX or UINTX) for creating constant tensors, but for now, we can match to the target type.


plaidml/bridge/openvino/ops/logical_not.cpp
```diff
@@ -17,8 +17,8 @@ namespace PlaidMLPlugin {
 static OpRegistration reg("logicalnot", [](const Context& ctx) {
   IE_ASSERT(ctx.operands.size() == 1);
   auto I = ctx.operands.at(0);
-  auto T = edsl::Tensor(true);
-  auto F = edsl::Tensor(false);
+  auto T = edsl::cast(edsl::Tensor(true), DType::BOOLEAN);
+  auto F = edsl::cast(edsl::Tensor(false), DType::BOOLEAN);
```
flaub, line 21: I'm actually puzzled by this. I think this is actually a bug in the eDSL. The principle of least surprise would say that `edsl::Tensor(true)` ought to have a element type of `DType::BOOLEAN`, but I'm guessing you had to do this because right now it thinks `true` is actually a `i1` which then considers this to be an arbitrary precision integer.

flaub: @jbruestle WDYT?

jbruestle: I would agree that Tensor(true) should be boolean.

flaub:
No, we can't use `FLOATX`, `INTX`, or `UINTX`, these are the arbitrary precision types. `select` requires that the true and false cases are concrete types.

plaidml/bridge/openvino/ops/logical_not.cpp
flaub: This whole thing can actually be replaced by the `logical_not` eDSL primitive.

plaidml/bridge/openvino/ops/prelu.cpp
```diff
@@ -18,7 +19,8 @@ static OpRegistration reg("prelu", [](const Context& ctx) {
   IE_ASSERT(ctx.operands.size() == 2);
   auto I = ctx.operands.at(0);
   auto slope = ctx.operands.at(1);
-  auto O = select(I < 0.0, slope * I, I);
+  DType target_type = to_plaidml(ctx.layer->get_default_output().get_element_type());
+  auto O = select(I < 0.0, cast(slope * I, target_type), cast(I, target_type));
```
flaub, line 23: As long as `I` does not have an arbitrary precision type, this is unnecessary.

tzerrell:
I believe that all comments have been addressed and this is ready for review. @cnamrata15 or @flaub can you give this another look?

cnamrata15 (approved):

flaub:
LGTM

PR #1571, opened 2020-11-20
Add ShapeOf with skipped test

tzerrell:
Adds a ShapeOf operator. Tests fail because constant tensors have their nodes renamed by nGraph optimizations and the PlaidML plugin can't find them as outputs (this is https://github.com/plaidml/openvino/issues/88).

However, I suspect this will prove useful in a full network context where the ShapeOf result is unlikely to be an output. That said, this is untested, so it's hard to be certain.


cnamrata15 (approved):
LGTM

PR #1572, opened 2020-11-24
Adjustments for downstream compiler

flaub:
The biggest change here is the replacemnet of C++ static intializers with registration functions that must be called near the start of a program.

Other changes are made to make the plaidml_builder compatible with the ICompiler interface.


pmlc/ast/builder.cc
```diff
@@ -813,10 +815,13 @@ struct ProgramBuilder {
...
 std::shared_ptr<Program> buildProgram(llvm::StringRef name,
                                       const ProgramArguments &args) {
-  enableGlobalDialectRegistry(true);
-  registerDialect<dialect::tile::TileDialect>();
-  registerDialect<dialect::layer::LayerDialect>();
-  registerDialect<StandardOpsDialect>();
+  static std::once_flag once;
+  std::call_once(once, []() {
+    enableGlobalDialectRegistry(true);
+    registerDialect<dialect::tile::TileDialect>();
```
brianretford, line 821: This seems like sort of an odd entry point - when does this list need to get updated and how will ppl know about it?

brianretford (approved):
Would like to understand why buildProgram has init functions, otherwise looks good

pmlc/ast/builder.cc
flaub: This will likely change soon, once we update with the latest llvm. This has been an area that's gone under a bunch of changes this year. I think dialect registration is now obsolete in upstream MLIR. But for now, the idea is to register the 'initial' dialects that we expect users to specify programs in. If we were to support escape hatches, we'd need to add those dialects here.

PR #1577, opened 2020-11-27
fix shuffle_channels

XingHongChenIntel:
Accroding to Pr https://github.com/plaidml/openvino/pull/112, fix the op code style.


tzerrell:
Why is this PR closed?

YangleiZouIntel:
> Why is this PR closed?

This PR is based on our forked repo, not from plaidml/plaidml, so auto CI check is not enabled.... I  was thinking about adding these kind of code clean up/ code style thing for all ops later in a single patch would be better than many small patches.

XingHongChenIntel:
fix this code in PR https://github.com/plaidml/plaidml/pull/1641

PR #1579, opened 2020-11-27
Add ReverseSequence op

XingHongChenIntel:
Added ReverseSequence op, and requires changes in plaidml/openvino, please see https://github.com/plaidml/openvino/pull/116.


tzerrell (commented):
Hmm, it is unfortunate (from a perf & clarity of code perspective) that we need to repeatedly slice & concatenate to accomplish this. However, I don't currently see a way around doing so. I will continue thinking on this but it looks reasonable for now.

XingHongChenIntel:
> Hmm, it is unfortunate (from a perf & clarity of code perspective) that we need to repeatedly slice & concatenate to accomplish this. However, I don't currently see a way around doing so. I will continue thinking on this but it looks reasonable for now.

since we have gather op, I replaced slice op by gather op. After testing, it speed up all test process.

tzerrell (approved):
Great, I agree this is a better approach.

PR #1583, opened 2020-12-01
Adding tuple pack/unpack ops + lowerings

jbruestle:
Pack + unpack support in preparation for init/fini goo.  Note: pack mallocs, unpack doesn't free, the assumption here is that returned allocations will have their lifetime managed by outer API.


flaub (approved):
Nice!

PR #1587, opened 2020-12-02
Cmake on Windows

abialas1:


cmake/third_party/volk.cmake
```diff
@@ -0,0 +1,8 @@
+FetchContent_Declare(
+  volk
+  URL      https://github.com/zeux/volk/archive/2638ad1b2b40f1ad402a0a6ac55b60bc51a23058.zip
+  URL_HASH SHA256=4a5fb828e05d8c86f696f8754e90302d6446b950236256bcb4857408357d2b60
+)
+FetchContent_MakeAvailable(volk)
+
+target_include_directories(volk PRIVATE ${vulkan_headers_SOURCE_DIR}/include)
```
flaub, line 8: Add missing ending newline

plaidml/CMakeLists.txt
```diff
@@ -26,6 +26,24 @@ pml_cc_library(
...
+    OUTPUT_NAME plaidml
+    SOVERSION 1
+  DEPS
+    plaidml::core::obj
+    plaidml::edsl::obj
+    plaidml::exec::obj
+    plaidml::op::obj
+    pmlc::rt::all_runtimes
+    pmlc::target::all_obj
+)
```
flaub, line 45: Add indentation

plaidml/CMakeLists.txt
```diff
@@ -26,6 +26,24 @@ pml_cc_library(
     pmlc::rt::all_runtimes
     pmlc::target::all_obj
 )
+else()
+pml_cc_library(
+  NAME plaidml
+  TYPE STATIC
+  LINKOPTS
+  PROPS
+    LINK_DEPENDS ${WINDOWS_DEF}
```
flaub, line 35: Not sure we need a `.def` file if we're just going to statically compile. Also, the intended use case for this library is for python, and so it pretty much has to be a `MODULE` or `SHARED` type.

flaub (commented):
Thanks for adding this. I do think there's more work to get everything to work properly on windows, but I do think this is a good PR to land because it actually gets vulkan/opencl added. I would even rename this PR and focus on just adding that support and then later perhaps deal with Windows separately.

plaidml/CMakeLists.txt
abialas1: I got a problem with this one. As I am using this for plaidmlPlugin for OpenVino I believe it should be statically linked. I did not see other solutions for different plugins. I was thinking on building both static and shared libs, so the shared one can be used in python and static for the c++ content.

flaub: That makes sense to me (having dual static/shared).

plaidml/CMakeLists.txt
```diff
@@ -7,42 +7,59 @@ add_subdirectory(op)
...
-  LINKOPTS
-  PROPS
+# For Windows build both static and shared lib
+  pml_cc_library(
+    NAME plaidml
+    TYPE STATIC
+    LINKOPTS
+    PROPS
+      OUTPUT_NAME plaidml
+      SOVERSION 1
```
flaub, line 37: I don't think we need these for STATIC lib

plaidml/CMakeLists.txt
```diff
@@ -7,42 +7,59 @@ add_subdirectory(op)
...
-pml_cc_library(
-  NAME plaidml
-  TYPE STATIC
-  LINKOPTS
-  PROPS
+# For Windows build both static and shared lib
+  pml_cc_library(
+    NAME plaidml
+    TYPE STATIC
+    LINKOPTS
```
flaub, line 34: Remove this, there are no linkopts specified

plaidml/CMakeLists.txt
```diff
@@ -6,26 +6,61 @@ add_subdirectory(op)
...
+      plaidml::exec::obj
+      plaidml::op::obj
+      pmlc::rt::all_runtimes
+      pmlc::target::all_obj
+  )
+
+  pml_cc_library(
+    NAME plaidml_dll
+    TYPE SHARED
+    LINKOPTS
```
flaub, line 50: I think we need to specify the `.def` file here, but I don't remember the exact flag syntax. You could just drop the `paidml_dll` target for now since we have nothing that tests it here anyways.

pmlc/rt/opencl/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+    opencl_runtime.cc
+    opencl_wrappers.cc
+    opencl_utils.h
+    opencl_utils.cc
+  DEPS
+    pmlc::compiler
+    pmlc::rt
+    MLIRPass
+	MLIRSPIRV
+	mlir_runner_utils
```
flaub, line 18: missing indentation

flaub (approved):

PR #1582, opened 2020-12-01
Add buffer transfer optimization to gpu_to_comp pass 

YangleiZouIntel:
1. adds buffer transfer optimization at comp level  https://github.com/plaidml/plaidml/pull/1356#issuecomment-700854644
    OpenCL resnet50 overall inference latency improved from 74ms to 61ms (about 18% improvement) on my local machine with Gen9 . Resnet50 total number of memory transfers reduced from 473 to 230

2. optimized comp dialect representation to help support memory optimization and share comp dialect representation between ocl, vulkan and L0

Example: PLAIDML_DEVICE=vulkan.0 PLAIDML_TARGET=intel_gen bazelisk run plaidml/bridge/keras:backend_test -- TestBackendOps.testConv1d 

Below shows one of the conv1d test with input data shape  <2x3x1xf32>,<1x8x3xf32>, <1x8x1xf32>

Before 
```
func @main(%arg0: !comp.device, %arg1: memref<2x3x1xf32>, %arg2: memref<1x8x3xf32>, %arg3: memref<1x8x1xf32>) {
    %cst = constant 0.000000e+00 : f32
    %c2 = constant 2 : index
    %c4 = constant 4 : index
    %c5 = constant 5 : index
    %c6 = constant 6 : index
    %c7 = constant 7 : index
    %c0 = constant 0 : index
    %c10 = constant 10 : index
    %c3 = constant 3 : index
    %c8 = constant 8 : index
    %c1 = constant 1 : index
    %0 = alloc() : memref<1x10x3xf32>
    %1 = comp.create_execenv %arg0 : (!comp.device) -> !comp.execenv<ocl:0,(11)>
    %2 = comp.alloc %1 : (!comp.execenv<ocl:0,(11)>) -> memref<1x10x3xf32, 11>
    %3 = comp.schedule_write %0 to %2 on %1 : (memref<1x10x3xf32>, memref<1x10x3xf32, 11>, !comp.execenv<ocl:0,(11)>) -> !comp.event<ocl>
    comp.wait %3 : !comp.event<ocl>
    %4 = "comp.schedule_func"(%1) ( {
      "gpu.launch_func"(%c1, %c1, %c1, %c1, %c10, %c3, %2) {kernel = @main_kernel::@main_kernel} : (index, index, index, index, index, index, memref<1x10x3xf32, 11>) -> ()
      "comp.schedule_end"() : () -> ()
    }) : (!comp.execenv<ocl:0,(11)>) -> !comp.event<ocl>
[...]

PR #1590, opened 2020-12-02
Add buffer transfer optimization for OpenCL

YangleiZouIntel:
This is the part of PR https://github.com/plaidml/plaidml/pull/1582 only for OpenCL, Vulkan related code will be added in a separate PR.  

1. Adds GatherGpuLaunchFuncsPass before gpu_to_comp pass, gathering consecutive gpu.lauch_func ops separated by std.alloc Op

2. Adds MinimizeBufferTransfersPass after  gpu_to_comp pass, remove redundant memory copies.

3. Changes on gpu_to_comp pass that share execEnv with consecutive kernel calls.


jbruestle (approved):
Looks good, thanks for splitting the PR.

jbruestle (approved):

jbruestle:
Hmm, so some of the tests are failing due to problems with our CI machines, but at least one of the failures is due to that fact that we now have cmake support and the new passes aren't being added to the cmake config, so we need to add that as well.  We will probably transition to all cmake at some point, but for now we need to always update both the bazel and the cmake rules.

YangleiZouIntel:
> Hmm, so some of the tests are failing due to problems with our CI machines, but at least one of the failures is due to that fact that we now have cmake support and the new passes aren't being added to the cmake config, so we need to add that as well. We will probably transition to all cmake at some point, but for now we need to always update both the bazel and the cmake rules.

I see. I have updated the cmake files. It seems mobilenet fails on my local machine (llvm_cpu) on latest plaidml-v1 branch with the same error code as the CI server...

flaub:
I believe the build issue is related to a recent h5py change. This PR should fix the issue: https://github.com/plaidml/plaidml/pull/1591

pmlc/conversion/gpu_to_comp/gpu_to_comp.cc
```diff
@@ -105,121 +113,173 @@ RewriteLaunchFunc::matchAndRewrite(gpu::LaunchFuncOp op,
...
+  return mlir::success();
+}
 
+template <typename T>
+mlir::LogicalResult
+RewriteLaunchFunc::getConsecutiveOps(T op, std::vector<T> &ops) const {
+  // Collect consecutive ops of type T starting from a given op.
+  auto block = op.getOperation()->getBlock();
+  for (auto itOp = block->begin(); itOp != block->end(); itOp++) {
+    if (mlir::isa<T>(itOp) && op == mlir::cast<T>(*itOp)) {
```
adstraw, line 159: Again, it might be more clear to do this in two loops:  1) find `op` in the block and 2) find all consecutive `ops` with same type T following `op`.  There may be an MLIR standard way to do step 1 as well.

adstraw (changes requested):

pmlc/conversion/gpu/gather_gpu_launch_funcs.cc
```diff
@@ -0,0 +1,70 @@
...
+  void runOnOperation() override;
+
+private:
+  void gatherAllocOpsBetweenConsecutiveLaunchOps(
+      mlir::Block &block,
+      std::vector<std::vector<mlir::AllocOp>> &allocOpVectors,
+      std::vector<gpu::LaunchFuncOp> &firstLaunchOpVector);
+};
+
+void GatherGpuLaunchFuncsPass::gatherAllocOpsBetweenConsecutiveLaunchOps(
```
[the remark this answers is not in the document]
YangleiZouIntel: Fixed.

pmlc/conversion/gpu/gather_gpu_launch_funcs.cc
```diff
@@ -0,0 +1,70 @@
...
+    mlir::Block &block, std::vector<std::vector<mlir::AllocOp>> &allocOpVectors,
+    std::vector<gpu::LaunchFuncOp> &firstLaunchOpVector) {
+  for (auto itOp = block.begin(); itOp != block.end(); itOp++) {
+    if (mlir::isa<gpu::LaunchFuncOp>(itOp)) {
```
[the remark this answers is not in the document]
YangleiZouIntel: Thanks, changed following your suggestion, and I found using getNextNode() makes the code much more succinct.

jbruestle (approved):
Looks good!

pmlc/dialect/comp/transforms/minimize_buffer_transfers.cc
```diff
@@ -0,0 +1,154 @@
...
+    }
+  });
+
+  func.walk([&](comp::ScheduleRead op) {
+    auto copyMode = getBufferCopyMode(op.getOperation(), op.hostMem());
+    if ((copyMode & BufferCopyMode::DeviceToHost) == BufferCopyMode::NoCopy) {
+      auto users = op.getResult().getUsers();
+      for (auto user : users) {
+        if (mlir::isa<comp::Wait>(user)) {
+          removeWaitOpEvent(mlir::cast<comp::Wait>(*user), op.getResult());
```
flaub, line 134: These sorts of things should do casting once, like:

```
if (auto waitOp = mlir::dyn_cast<comp::Wait>(user)) {
  removeWaitOpEvent(waitOp, op.getResult());
}
```

```

Comment (flaub) [reply pmlc/dialect/comp/transforms/minimize_buffer_transfers.cc:134]: I notice this pattern can be applied in much of this PR.

Comment (YangleiZouIntel) [reply pmlc/dialect/comp/transforms/minimize_buffer_transfers.cc]: Fixed

File: pmlc/conversion/gpu/gather_gpu_launch_funcs.cc [hunk 5]
```diff
@@ -0,0 +1,82 @@
...
+    }
+    op = op->getNextNode();
+  }
+  return op;
+}
+
+void GatherGpuLaunchFuncsPass::gatherAllocOpsBetweenConsecutiveLaunchOps(
+    mlir::Block &block, std::vector<std::shared_ptr<Move>> &moves) {
+  mlir::Operation *op = &(*block.begin());
+  while (op != nullptr) {
```
adstraw, line 47: This still feels unnecessarily complex.  I may be oversimplifying but I read this code as doing:

```
op = advanceToNextLaunchOp()
while (op != nullptr)
  if LaunchOp:
    create a Move 
    move->insertionOp = op
    moves.push_back(move)
    op = op->getNextNode()
  else if AllocOp:
    moves.back()->ops.push_back(op)
    op = op->getNextNode()
  else
     op = advanceToNextLaunchOp()
```

pmlc/conversion/gpu/gather_gpu_launch_funcs.cc
```diff
@@ -0,0 +1,82 @@
...
+
+void GatherGpuLaunchFuncsPass::runOnOperation() {
+  auto module = getOperation();
+  module.walk([&](mlir::FuncOp funcOp) {
+    if (funcOp.isExternal()) {
+      return;
+    }
+    auto &blocks = funcOp.getBlocks();
+    for (auto &block : blocks) {
+      std::vector<std::shared_ptr<Move>> moves;
```
adstraw, line 68: Nitpick, but why not return the vector of Moves rather than pass by reference?

adstraw (approved):
I am approving so we can make forward progress here but I feel there could be some cleanup in a future commit.

pmlc/conversion/gpu/gather_gpu_launch_funcs.cc
adstraw: Revised:  I don't think you need the first `advanceToNextLaunchOp` call as it is handled in the else case.

pmlc/conversion/gpu_to_comp/gpu_to_comp.cc
```diff
@@ -105,121 +113,171 @@ RewriteLaunchFunc::matchAndRewrite(gpu::LaunchFuncOp op,
...
+  }
+  return mlir::success();
+}
 
+template <typename T>
+mlir::LogicalResult
+RewriteLaunchFunc::getConsecutiveOps(T op, std::vector<T> &ops) const {
+  // Collect consecutive ops of type T starting from a given op.
+  auto operation = op.getOperation();
+  while (operation != nullptr) {
```
adstraw, line 158: This too could be simplified, right?
```
while op isa T:
  ops.push_back(op)
  op = op->getNextNode()
return mlir::success()
```

pmlc/conversion/gpu/gather_gpu_launch_funcs.cc
YangleiZouIntel: Thanks. Here we are not collecting from every LaunchOp, we are collecting from the first LaunchOp that are not following otherLaunchOps or AllocOp.
An example would be, change from 

```
AllocOp
LaunchOp
AllocOp
LaunchOp
AllocOp
LaunchOp
SomeOtherOp
SomeOtherOp
AllocOp
LaunchOp
AllocOp
LaunchOp
```
To
```
AllocOp
AllocOp
AllocOp
LaunchOp
LaunchOp
LaunchOp
SomeOtherOp
SomeOtherOp
AllocOp
AllocOp
LaunchOp
LaunchOp
```
Here we will generate two moves, the first move for the first three allocOps before the first LaunchOp...we have to skip the second and third launchOp, that is why I add function `gatherAllocOpsStartingFromOp`

pmlc/conversion/gpu_to_comp/gpu_to_comp.cc
YangleiZouIntel: Yes, I will simplify this. Maybe it is better I make the update in the vulkan PR.

pmlc/conversion/gpu/gather_gpu_launch_funcs.cc
YangleiZouIntel: Changed in [PR1605](https://github.com/plaidml/plaidml/pull/1605/files#diff-df1e828d87ae2c07b4b3674b23ed4c4e560635639564f3acb37d3bcdf56ca021R27)

pmlc/conversion/gpu_to_comp/gpu_to_comp.cc
YangleiZouIntel: Changed in [PR1605](https://github.com/plaidml/plaidml/pull/1605/files#diff-6663ff64acb7df7101c5a996e292cb741f317686d2130b8a3a11a24946d8b724R158)

PR #1592, opened 2020-12-02
change tile.box op constructor to accept array of types, instead of values

mars1026:


pmlc/ast/builder.cc
```diff
@@ -681,16 +681,19 @@ struct ProgramBuilder {
       operands.push_back(builder.lookupNode(operand));
     }
     llvm::SetVector<Value> results;
+    llvm::SetVector<Type> resultTypes;
```
flaub, line 684: I'm not sure this is correct. This SetVector will collapse multiple types into a single entry. But we definitely want to be able to support multiple results with the same type.

mars1026: good point - what I really want is a plain vector, copying the types of the results set.

flaub: Yes, but be careful, because we *do* want multiple values to be collapsed. So you might need to have a 2nd loop to construct the plain vector.

pmlc/ast/builder.cc
```diff
@@ -684,14 +684,17 @@ struct ProgramBuilder {
     for (const ExprNodePtr &result : node->results) {
       results.insert(builder.lookupNode(result));
     }
+    llvm::SmallVector<Type, 4> resultTypes;
+    for (const Value &val : results) {
```
flaub, line 688: Don't use by-ref with Value.

flaub (approved):

PR #1594, opened 2020-12-03
Enable init/fini on CPU path

jbruestle:
Initial 'it does something' version of init-fini code.

### DIFF
--- a/pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
+++ b/pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
@@ -216,9 +216,14 @@ struct PackLowering : public ConvertOpToLLVMPattern<stdx::PackOp> {
                   ConversionPatternRewriter &rewriter) const override {
     auto op = cast<stdx::PackOp>(baseOp);
     Location loc = op.getLoc();
+    auto i8PtrType = LLVMType::getInt8Ty(rewriter.getContext()).getPointerTo();
+    if (op.getNumOperands() == 0) {
+      auto nullPtr = rewriter.create<LLVM::NullOp>(loc, i8PtrType);
+      rewriter.replaceOp(op, {nullPtr});
+      return success();
+    }
     // Get the relevant types
     auto structType = getStructType(typeConverter, op.getOperandTypes());
-    auto i8PtrType = LLVMType::getInt8Ty(rewriter.getContext()).getPointerTo();
     // Get the size of the struct type and malloc
     auto sizeofStruct = getSizeInBytes(loc, structType, rewriter);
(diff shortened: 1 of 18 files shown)


pmlc/conversion/stdx_to_llvm/stdx_to_llvm.cc
```diff
@@ -315,8 +315,8 @@ void populateStdXToLLVMConversionPatterns(LLVMTypeConverter &converter,
                   TanLowering,     //
                   UnpackLowering   //
                   >(converter);
-  converter.addConversion([](TupleType type) -> Optional<Type> {
-    // Tuples look like i8 pointers.  I'd like this to be void*, but MLIR
+  converter.addConversion([](stdx::ArgpackType type) -> Optional<Type> {
+    // Opaque types look like i8 pointers.  I'd like this to be void*, but MLIR
```
flaub, line 319: Adjust comment to say "Argpack types..."

pmlc/dialect/tile/transforms/BUILD
```diff
@@ -22,6 +22,7 @@ plaidml_cc_library(
         "pad_ranges.cc",
         "padding.cc",
         "pass_detail.h",
+	"split_main.cc",
```
flaub, line 25: whitespace

jbruestle:
Now passes all local tests, running on CI.

jbruestle:
This change adds support for init/fini on the CPU path only, with no additional optimizations or hoisting (so it's basically a no-op in terms of performance), but it sets the stage for adding hoisting of code from main -> init (or for alloc/dealloc ops from main -> init/fini).  The next commits will be:

- General hoisting support (which should immediately give us tensor constant folding when applied at the tile level)
- Adding SplitMain to the GPU pipelines, fix any bugs
- Breaking kernel launching in comp into compilation + launch so the compilation can be hoisted

Technically, this change could be separated into two PR's:  One to change stdx.pack/unpack semantics, and another to actually hook up init/fini, but I figure it's small enough to go in as one PR if no one disagrees.

flaub (dismissed):

pmlc/dialect/tile/transforms/CMakeLists.txt
```diff
@@ -9,10 +9,12 @@ pml_cc_library(
     pad_constraints.cc
     pad_ranges.cc
     padding.cc
+    split_main.cc
   DEPS
     LLVMSupport
     MLIRIR
     pmlc::dialect::tile::ir
+    pmlc::dialect::pxa::ir
```
flaub, line 17: Why do we depend on pxa? Seems like this should be stdx?

flaub (approved):

PR #1597, opened 2020-12-03
Align runtime registration for opencl and vulkan

abialas1:


pmlc/rt/vulkan/vulkan_runtime.cc
```diff
@@ -54,6 +59,11 @@ class VulkanRuntime final : public pmlc::rt::Runtime {
   std::vector<std::shared_ptr<VulkanDevice>> devices;
 };
 
-pmlc::rt::RuntimeRegistration<VulkanRuntime> reg{"vulkan"};
+extern void registerSymbols();
+
+void registerRuntime() {
+  registerSymbols();
+  registerRuntime("vulkan", std::make_shared<VulkanRuntime>());
```
flaub, line 66: We could just put a try/catch around this line I think.

flaub: Actually, perhaps we should do it like:
```
try {
  registerRuntime("vulkan", std::make_shared<VulkanRuntime>());
  registerSymbols();
} catch (std::exception& ex) {
  IVLOG(2, "Failed to register 'vulkan' runtime: " << ex.what());
}
```

This way we only call `registerSymbols` if the runtime can be constructed.

abialas1: Yes could be, however I wanted to be consistent with the OpenCL runtime as is. If you think it is important I can move it to this place.

pmlc/rt/opencl/opencl_runtime.cc
```diff
@@ -29,6 +30,13 @@ class OpenCLRuntime final : public pmlc::rt::Runtime {
...
 };
 
-pmlc::rt::RuntimeRegistration<OpenCLRuntime> reg{"opencl"};
+extern void registerSymbols();
+extern void registerMemrefCasts();
+
+void registerRuntime() {
+  registerSymbols();
+  registerMemrefCasts();
+  registerRuntime("opencl", std::make_shared<OpenCLRuntime>());
```
flaub, line 39: Might as well try/catch here too.

abialas1: Ok let me change both vulkan and ocl to what you proposed

flaub (approved):

PR #1599, opened 2020-12-03
Run smoke test on nightly build

mwyi:


flaub (approved):
Seems fine to me, although there's probably lots of other improvements we could make to the nightly pipeline.

flaub (approved):

tzerrell:
As context, we're hoping to soon run benchmark_app and eventually accuracy_checker in nightly to give us an automated means of seeing progress / noticing regression in OV full network support.

PR #1600, opened 2020-12-03
layer.box op may contain side effects; do not tag NoSideEffect

mars1026:
Original definition of layer dialect BoxOp was erroneously marked [NoSideEffect], but this is not a reasonable guarantee.


LLITCHEV (approved):
Looks fine to me.

flaub:
Well, this might depend on the context. If we are using value-only semantics, like in tile dialect, I think `[NoSideEffect]` might hold.

flaub:
There should probably be a lit test to go along with changes like these.

PR #1602, opened 2020-12-04
Fix Squeeze in OpenVINO plugin

tzerrell:
Fix OV op `Squeeze`, which was dropped from the op list. Also enabled testing of squeeze and unsqueeze to prevent regression.  Note that floating point tests fail with an nGraph error (which is probably why we didn't previously have them), but integer tests work.

A bit of detail on the error: it is `Check 'm_element_type.is_integral_number()' failed` coming from `ngraph::op::Constant::get_axis_set_val()`. According to the OV docs, axis values for squeeze/unsqueeze must be integers, and we use this `get_axis_set_val` function on that assumption, but when testing float values, the test also provides axes as floats, leading to this error.


flaub (approved):
Thanks for this!

PR #1606, opened 2020-12-08
Add OpenVINO Gather

YangleiZouIntel:
Add negative axis support to edsl::gather (nGraph Inference Engine gather does not support negative axis right now).


tzerrell (approved):
Thanks!

PR #1609, opened 2020-12-10
Add CumSum op

XingHongChenIntel:
Add CumSum op. And I wonder if you prefer to add exclusive and reverse modes to oplib cumsum?


plaidml/bridge/openvino/ops/cum_sum.cpp
```diff
@@ -0,0 +1,74 @@
...
+  auto pad_I = op::explicit_padding(I, lo_pad, hi_pad).padval(edsl::Constant(0));
+  return edsl::gather(pad_I, edsl::index({dims[axis]}, 0)).axis(axis);
+}
+
+}  // namespace
+
+namespace PlaidMLPlugin {
+
+void registerCumSum() {
+  registerOp("cumsum", [](const Context& ctx) {
```
flaub, line 55: nit: Use `"CumSum"`

plaidml/bridge/openvino/ops/cum_sum.cpp
```diff
@@ -0,0 +1,74 @@
...
+edsl::Tensor reverse_tensor(edsl::Tensor reverse_crop, int64_t seq_axis) {
+  std::vector<edsl::TensorDim> dims(reverse_crop.rank());
+  reverse_crop.bind_dims(dims);
+  std::vector<edsl::TensorIndex> I_idxs(reverse_crop.rank());
+  std::vector<edsl::TensorIndex> O_idxs(I_idxs);
+  O_idxs[seq_axis] = dims[seq_axis] - 1 - I_idxs[seq_axis];
+  return edsl::Contraction().outShape(dims).outAccess(O_idxs).assign(reverse_crop(I_idxs));
+}
+
+edsl::Tensor exclued_first(edsl::Tensor I, int64_t axis) {
```
flaub, line 38: typo: `exclude_first`

plaidml/bridge/openvino/ops/cum_sum.cpp
XingHongChenIntel: thanks, fixed this.

plaidml/bridge/openvino/ops/cum_sum.cpp
XingHongChenIntel: thanks, fixed this.

flaub (commented):
LGTM, I'll let Tim weigh in.

plaidml/bridge/openvino/ops/cum_sum.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+namespace {
+template <typename T>
+std::vector<T> cast_constant_operand(size_t operand_idx, ngraph::Node* layer) {
+  auto* ngraph_const = ngraph::as_type<ngraph::op::Constant>(layer->get_input_node_ptr(operand_idx));
+  if (ngraph_const) {
+    return ngraph_const->cast_vector<T>();
+  } else {
+    THROW_IE_EXCEPTION
+        << "Dynamic slicing not currently supported by PlaidML plugin; all of indices, offsets and default index"
```
tzerrell, line 24: This op just uses axis, so update this message to something like "Dynamic axis not currently  supported by PlaidML plugin"

tzerrell (commented):
Looks generally good!

I think following your suggestion of moving exclusive to the oplib would allow us to avoid the gather here (instead just shifting the contraction by 1). I think doing so would be worthwhile. Reverse is less important IMO, it's fine in either location.

This could also be landed with just a change to the exception message, but if we do so we should open an issue for moving exclusive to oplib cumsum.

XingHongChenIntel:
> Looks generally good!
> 
> I think following your suggestion of moving exclusive to the oplib would allow us to avoid the gather here (instead just shifting the contraction by 1). I think doing so would be worthwhile. Reverse is less important IMO, it's fine in either location.
> 
> This could also be landed with just a change to the exception message, but if we do so we should open an issue for moving exclusive to oplib cumsum.

very thanks for your comments. now, I move `exclusive` mode to oplib CumSum, add op test and apply this new oplib cumsum to our op.

Issue #1493, opened 2020-10-21
Upgrade Scatter


tzerrell:
Scatter needs to be extended to cover more cases. We probably need to review exactly what needs to be added, but at a minimum we should be able to specify an axis and scatter blocks of elements (i.e. instead of just single elements).

haoyouab:
Closed with #1610. Possible followup tracked in #1612.

PR #1610, opened 2020-12-10
add axis parameter and update feature to Scatterop

haoyouab:
1. Added axis parameter and modes to support normal scatter and various scatter update operations (e.g. N-dimension and element-wise update) to address #1493.

2. Added a few edsl scatter tests and passed on intel_gen and cpu.

3. Openvino ops `ScatterUpdate`, `ScatterElementsUpdate` and `ScatterNDUpdate` and their tests were added.

Note:

For now scatter (current `edsl::scatter` or `ScatterMode::NORMAL`) and scatter_update ( `ScatterMode::UPDATE_SLICE, UPDATE_ELT, UPDATE_ND`) are both implemented in `ScatterOp` to provide a unified API:
```
scatter(data, indices, updates).axis(...).mode(...)
```
According to [tf.scatter_nd](https://www.tensorflow.org/api_docs/python/tf/scatter_nd), only the shape of `data` is needed in `ScatterMode::normal` while `ScatterMode::update_*` (see [tf.compat.v1.scatter_update](https://www.tensorflow.org/api_docs/python/tf/compat/v1/scatter_update) and [tf.scatter_nd_update](https://www.tensorflow.org/api_docs/python/tf/tensor_scatter_nd_update)) needs `data` itself, so passing the `data` could be an unnecessary overhead in normal mode. Maybe putting the normal scatter and the scatter update into two separate ops like `ScatterOp` (which accepts parameter shape, indices, updates) and `ScatterUpdateOp` (which accepts parameter data, indices, updates) is better?


flaub (commented):
Nice work

tzerrell (approved):
Great!

I think we can land this now, and I've opened an issue (#1612) to consider splitting scatter.

PR #1611, opened 2020-12-10
Access qualifier for read only buffers pass

abialas1:
Added this pass on spirv level so we would not need to propagate the read only attribute all the way down. Also we need to set it per kernel as some kernel use same buffers for read, others for write.
I tried to add the NoWrite enum to spirv dialect instead of adding magic number, but seems like it is kind of pandora's box when I tried to add Function Parameter Attribute Adds to SPIRVBase.td using provided mlir scripts. There are tons of new Extensions to be added (which seems need to be done manually), not sure if it is related directly with this change or lack of new updates lately. I also got problems with lack of definitions for the NVidia's extensions, so decided to abandon this and fix for magic number for now.

### DIFF
--- a/pmlc/target/intel_gen_ocl_spirv/access_qualifiers.cc
+++ b/pmlc/target/intel_gen_ocl_spirv/access_qualifiers.cc
@@ -1,0 +1,121 @@
+// Copyright 2020, Intel Corporation
+
+#include "pmlc/target/intel_gen_ocl_spirv/pass_detail.h"
+#include "pmlc/target/intel_gen_ocl_spirv/passes.h"
+
+#include "mlir/Dialect/SPIRV/SPIRVOps.h"
+#include "mlir/Dialect/SPIRV/TargetAndABI.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/Support/LLVM.h"
+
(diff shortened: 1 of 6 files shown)


jbruestle (approved):
This is great.  FYI, I am currently reworking the Affine -> GPU lowering on the host side to make hoisting possible.  Specifically, I'm going to do conversion to comp during kernel outlining, and as part of that, I'm going to annotate buffer use as RO, WO, RW at the comp level.  This is needed to allow hoisting of the weight transfers into init (otherwise, we can't tell if kernels trash the weights, thus requiring reinitialization every inference).

pmlc/target/intel_gen_ocl_spirv/access_qualifiers.cc
```diff
@@ -0,0 +1,121 @@
...
+    return true;
+  }
+
+  void runOnOperation() {
+    spirv::ModuleOp module = getOperation();
+    spirv::FuncOp func = getEntryPoint(module);
+    if (!func)
+      return;
+
+    for (auto &arg : func.getArguments()) {
```
flaub, line 55: This is a `BlockArgument` and thus should use value-like semantics.

PR #1614, opened 2020-12-11
Error surfacing

jd-bartlett96:
Surfaced errors from `builder.cc` to be thrown in construction phase. Below is a brief explanation for each error of `builder.cc` for what I did or why I nothing.

- builder.cc:96 - Invalid data type
	Only error-causing cases are "invalid" and "bf16"
	Surfaced to Placeholder constructor

- builder.cc:123 - Missing placeholder during program build
	Checks that exprMap was built correctly; building-phase specific, no surfacing necessary

- builder.cc:126 - ExprNode not found
	Checks that exprMap was built correctly; building-phase specific, no surfacing necessary

- builder.cc:135 - Element node not found
	Checks that exprTuples was built correctly; building-phase specific, no surfacing necessary

- builder.cc:139 - Out of range ordinal for element node
	Checks that exprTuples was built correctly; building-phase specific, no surfacing necessary

- builder.cc:165 - Unsupported VarNode
	Surfaced to checkAttribute function in ffi.cc

- builder.cc:201 - Implicit operand detected within layer body
	thrown for missing operand to layer...don't immediately see a way to surface this to the construction phase but it may be possible

- builder.cc:250 - Invalid expression in AstTraversal:push
	Bulding-phase specific; don't think surfacing is necessary

- builder.cc:447 - NYI: AffineOp::Max
	Surfaced to evaluator; couldn't find an example of using this to test

- builder.cc:449 - NYI: AffineOp::Min
	Surfaced to evaluator; couldn't find an example of using this to test

- builder.cc:568 - Output not found while building program
	Bulding-phase specific; don't think there's a reasonable way to surface this to the construction phase	

- builder.cc:603 - Program build failure
	Bulding-phase specific; no surfacing necessary

- builder.cc:760 - 'gather' primitive expects the 'axis' argument to be a constant integer
	Could not find api input that would raise this error; if axis input is castable to int, it gets cast, otherwise a compiletime error is thrown

- builder.cc:764 - 'gather' primitive expects the 'interpolationMode' argument to be a constant integer
	Could not find api input that would raise this error; no way for interpolationMode to be set to anything other than an a constant int

- builder.cc:770 - 'gather' primitive expects the 'nearestMode' argument to be a constant integer
	Could not find api input that would raise this error; no way for nearestMode to be set to anything other than an a constant int

- builder.cc:774 - 'gather' primitive expects the 'cubeCoeff' argument to be a constant float
	Could not find api input that would raise this error; if cubeCoeff is castable to float, it gets cast, otherwise a compiletime error is thrown

- builder.cc:803 - 'index' primitive expects at least one operand
	Could not find api input that would raise this error; axis operand is always included; if user does not specify an axis a compiletime error is thrown. However, an empty dim list causes a segfault during makeProgram so a construction-phase check for this was added.

- builder.cc:809 - 'index' primitive expects argument 1 to be a constant integer
	Could not find api input that would raise this error; if axis input is castable to int, it gets cast, otherwise a compiletime error is thrown
	
- builder.cc:820 - 'prng' primitive expects at least one operand
	Could not find api input that would raise this error; not providing operands will throw a compiletime error

- builder.cc:840 - Unknown EDSL primitive
	It might be possible to check that AbstractOperation::lookup() returns something earlier (perhaps at intrinsicCall in edsl.h), but I think it would require creating a dummy MLIRContext, which a) I'm not sure is possible there and b) seems likely unnecessary as a user running into this error seems unlikely (Not sure what they'd be doing to raise this error)


tzerrell:
By the way, the scatter PR (https://github.com/plaidml/plaidml/pull/1610) introduced some more builder errors, but I suspect no work is needed for them for the same reasons no work is needed for the gather errors.

plaidml/edsl/edsl.h
```diff
@@ -1161,6 +1169,7 @@ inline Tensor Tensor::operator!() const { return intrinsicCall("logical_not", {*
   inline Tensor operator _op_(uint64_t lhs, const Tensor& rhs) { return intrinsic(_fn_, lhs, rhs); }      \
   inline Tensor operator _op_(double lhs, const Tensor& rhs) { return intrinsic(_fn_, lhs, rhs); }
 
+// clang-format off
```
flaub, line 1172: What's this for?

plaidml/edsl/ffi.cc
```diff
@@ -334,6 +349,7 @@ plaidml_expr* plaidml_expr_pragma(  //
     llvm::StringMap<ast::VarNodePtr> attrs;
     for (size_t i = 0; i < nattrs; i++) {
       plaidml_attr* attr = raw_attrs[i];
+      checkAttribute(attr->value->node);
```
flaub, line 352: I think we should allow any VarNode here, why is this needed?

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -2024,6 +2024,25 @@ TEST_F(CppEdsl, LayerUnusedOperand) {
   runProgram(program);
 }
 
+TEST_F(CppEdsl, BadDataType) {
+  // auto A = Placeholder(DType::BFLOAT16, {10, 20});
+  // Tensor O = A + A;
+  // EXPECT_ANY_THROW({ makeProgram("BadDataType", {A}, {O}); });
+  EXPECT_ANY_THROW({ auto A = Placeholder(DType::BFLOAT16, {10, 20}); });
```
flaub, line 2031: If BFLOAT16 isn't yet supported, let's just drop it from the enum.

pmlc/.clang-format
```diff
@@ -1,2 +1,2 @@
 BasedOnStyle: LLVM
-AlwaysBreakTemplateDeclarations: Yes
+AlwaysBreakTemplateDeclarations: true
```
flaub, line 2: I copied this file directly from MLIR, why do we need to change it?

pmlc/ast/eval.cc
```diff
@@ -178,9 +178,13 @@ int64_t Evaluator::evaluate(const DimNode *node) {
         case AffineOp::Div:
           return lhs / rhs;
         case AffineOp::Max:
-          return std::max(lhs, rhs);
+          // Surfacing of error due to not being implemented at builder level
+          throw std::runtime_error("NYI: AffineOp::Max");
```
flaub, line 182: Not sure I understand this one, if you comment out the implementation, then yes it's NYI.

plaidml/edsl/edsl.h
```diff
@@ -700,6 +700,10 @@ inline Tensor Placeholder(             //
     DType dtype,                       //
     const std::vector<int64_t>& dims,  //
     const std::string& name = "") {
+  // Surfacing of builder.cc:96 - Invalid data type
```
flaub, line 703: Remove this comment, it will get stale really fast.

plaidml/edsl/edsl.h
jd-bartlett96: For some reason clang-format wanted to add a space here (between `OPS` and `(`:
`PLAIDML_EDSL_DEFINE_TENSOR_BINARY_OPS (^, "bit_xor");`
which was then caught by cpplint, preventing me from being able to push my commit. If there's a preferred way to handle this lmk

pmlc/.clang-format
jd-bartlett96: This was giving me an error during the pre-commit check, "invalid boolean type" or something like that.

plaidml/edsl/edsl.h
flaub: Make sure you use the `.cenv` conda environment, it has a specific version of clang-format that we should all be using.

pmlc/ast/eval.cc
jd-bartlett96: Investigating this now because it appears that I misinterpretted what was going on between here and the builder level; there's a similar "NYI" message at the builder level for min and max AffineOps, and I was aiming to surface this to where those ops are added at the construction level. But it looks like things that are AffineOp::Max here don't necessarily correspond to that at the build level, as seen by resnet50_test failing in the merge checks because it contains construction-level AffineOp::Max but no builder-level AffineOp::Max. Taking a second look at this (and min) now.

pmlc/.clang-format
jbruestle: You probably have the wrong version of clang format.  Are you running the precommit hooks from the conda env?

flaub: Are you running `git commit` from the `.cenv` conda environment?

jd-bartlett96: Yes :/

pmlc/ast/eval.cc
flaub: OK, I remember this now. There are 2 ways things that the evaluator is used for. 
1. DimNode
2. PolyNode

For the `DimNode` case, min/max is sensible and is implemented here.
For the `PolyNode` case, it's not clear what the right implementation would be. This is a question for @tzerrell, it might just be invalid in general.

plaidml/edsl/ffi.cc
flaub: I see that in the builder, we say unsupported, but I think we should just implement it at some point. I think the public interface should actually support it.

plaidml/edsl/edsl.h
```diff
@@ -920,6 +924,10 @@ inline Tensor ident(const Tensor& x) { return intrinsic("ident", x); }
 /// \return Tensor
 ///
 inline Tensor index(const std::vector<TensorDim>& dims, size_t axis) {
+  if (dims.size() < 1) {
+    // Prevents empty dim list from causing segfault during program build
+    throw std::runtime_error("index() must be provided non-empty dims list");
+  }
```
flaub, line 930: Note that we'll have to do this kind of validation for all language bindings (so `python` also today). Is there a way we can do this check behind the ffi? I'm thinking in `ast_ops.cc`

plaidml/edsl/ffi.cc
jd-bartlett96: I didn't see any VarNode that would trigger it, but there is a default error message thrown in the builder if an unknown varnode is specified; I surfaced that because it could be, but if it is not worth checking every use for that/there is indeed no input that would trigger it, I can remove it.

plaidml/edsl/tests/edsl_test.cc
jd-bartlett96: At that point, do you think it's worth keeping the check for the "INVALID" dtype, or is it safe to assume a user isn't going to try and use that and expect it to work?

plaidml/edsl/ffi.cc
flaub: I think it's the string type that would make it fail. I just don't think we should do a `checkAttribute` at this level. It shouldn't be possible in the edsl to create an invalid attribute.

plaidml/edsl/tests/edsl_test.cc
flaub: Perhaps we could check for the invalid dtype deeper, in the ffi. Generally, I try to do validation inside the ffi so that every language benefits from it.

flaub: Actually, not in the ffi, more like in the construction of an AST node.

pmlc/ast/eval.cc
tzerrell: Probably min/max of a PolyNode is straight-up invalid; the min (or max) of two (affine) polynomials is not (affine) polynomial itself, but is instead _piecewise_ (affine) polynomial. IIRC this is not something we want anywhere, but also is something that is supported with the full semi-affine generality of MLIR Affine dialect... I could definitely be mis-remembering either part of that though. @jbruestle might know more about what we intend & are able to support here.

pmlc/.clang-format
jd-bartlett96: (In case anyone runs into this in the future, I believe my issue was that I'd configured & activated .cenv from the `master` branch then was trying to use that to commit to this branch off `plaidml-v1`. Resolved by removing .cenv then recreating it with `./configure` on `plaidml-v1`)

pmlc/ast/eval.cc
jd-bartlett96: Regardless, it sounds like (for now at least) keeping this error as-is at the builder level is appropriate as it properly distinguishes the invalid/nyi min/max operations. My attempt at surfacing these errors has been removed.

flaub (approved):

PR #1619, opened 2020-12-14
Vectorize mem fixes

abialas1:
Added some fixes to vectorize mem I found:
1. Make the optimization checks to be stricter (if any vector load or vector reduce in the loop cannot be vectorized then do not do optimization)
2. Check if the YieldOp type matches the loop result type, if not then update it
3. Change after code review from access qualifiers


pmlc/dialect/pxa/transforms/vectorize_mem.cc
```diff
@@ -320,6 +325,11 @@ struct VectorizeMemImpl {
     auto results = loopOp.results();
     for (auto res : results) {
       res.setType(nemMemrefType);
+      // Update also AffineYieldOp if type does not match
+      for (auto yieldOp : loopOp.getOps<AffineYieldOp>()) {
+        if (yieldOp.getOperand(0).getType() != nemMemrefType)
```
jbruestle, line 330: Seems like the original code had a typo (I presume nemMemrefType is supposed to be newMemrefType). Probably should fix the name as part of this commit.

jbruestle (commented):
Looks good and makes sense, tiny typo fix request.

pmlc/dialect/pxa/transforms/vectorize_mem.cc
```diff
@@ -295,9 +300,9 @@ struct VectorizeMemImpl {
         VectorType::get({memOpsPlan.subgroupSize * tileSize},
                         vectorReduce.getVectorType().getElementType());
 
-    auto nemMemrefType =
+    auto memMemrefType =
```
flaub, line 303: This should probably be renamed to `newMemrefType`

abialas1: you are right! let me change it

flaub (approved):

PR #1621, opened 2020-12-15
Simplify arithmetic pass

abialas1:
This simple pass gets rid of redundant ops from the reduce assign 0 <memref> -> reduce addf <memref> that is side effect of the fusion pass. Had to move MemrefAccess class to analysis folder to reuse it from the dataflow opt.


jbruestle (approved):
Looks good.

PR #1624, opened 2020-12-16
Comp refactor

jbruestle:
OpenCL sort of works, seeing what breaks in CI


pmlc/conversion/gpu/kernel_outlining.cc
```diff
@@ -120,25 +133,185 @@ static gpu::GPUFuncOp outlineKernelFuncImpl(gpu::LaunchOp launchOp,
...
+    OpBuilder builder(func.getBody());
+
+    // Add in comp device parameter to function
+    auto oldFuncTy = func.getType();
+    auto deviceTy = builder.getType<comp::DeviceType>();
+    SmallVector<Type, 4> inputs{deviceTy};
+    inputs.insert(inputs.end(), oldFuncTy.getInputs().begin(),
+              oldFuncTy.getInputs().end());
+    auto newFuncTy = builder.getFunctionType(inputs, oldFuncTy.getResults());
+    (void) newFuncTy;
```
flaub, line 156: ??

pmlc/dialect/comp/ir/ops.td
```diff
@@ -86,50 +99,59 @@ def COMP_DestroyExecEnv : COMP_Op<"destroy_execenv",
...
+    $kernel `on` $execEnv attr-dict `:` functional-type(operands, results)
+  }];
+}
 
-def COMP_ScheduleFunc : COMP_Op<"schedule_func",
-                                [COMP_ScheduleSingleBlock,
-                                 DeclareOpInterfaceMethods<COMP_ExecEnvOpInterface>,
-                                 DeclareOpInterfaceMethods<COMP_ScheduleOpInterface>]>
-                  , Arguments<(ins COMP_AnyExecEnv:$execEnv,
+def COMP_ScheduleCompute : COMP_Op<"schedule_compute",
```
flaub, line 130: Maybe `schedule_kernel` just so its consistent with `alloc_kernel`, `dealloc_kernel`.

pmlc/rt/opencl/opencl_wrappers.cc
```diff
@@ -11,115 +11,162 @@ namespace pmlc::rt::opencl {
 
 extern "C" {
 
-void *oclCreate(void *device) {
-  return new OpenCLInvocation(static_cast<OpenCLDevice *>(device));
+void *ocl_create_execenv(void *device) {
+  IVLOG(1, "ocl_create_execenv device=" << device);
+  void *result = new OpenCLInvocation(static_cast<OpenCLDevice *>(device));
+  IVLOG(1, "  ->" << result);
+  return result;
```
flaub, line 18: I wonder if we should do a try/catch here and then come up with a way to report errors gracefully.

pmlc/dialect/comp/ir/ops.td
earhart: IMHO -- this operation doesn't change the state of the kernel at all; it's not something one does to the kernel, and you can do it multiple times with the same kernel with various inputs, producing different completion events.  It's more like you're enqueuing some compute task to be run, and the kernel is an input to that compute task.

flaub: We could call it `launch_kernel`?

pmlc/conversion/comp_to_llvm/BUILD
```diff
@@ -35,6 +34,7 @@ plaidml_cc_library(
     ],
     deps = [
         ":gen-pass",
+	"//pmlc/conversion/stdx_to_llvm",
```
flaub, line 37: Tabs

pmlc/rt/opencl/opencl_wrappers.cc
jbruestle: Probably, but saving for the next PR at least.

pmlc/dialect/comp/ir/ops.td
jbruestle: I agree with Rob on this one.  You might schedule computations with different grid sizes, buffers, etc, so it's not really one-to-one with a kernel, although each compute does have a unique kernel, so I think schedule_compute is a better name.

jbruestle: Also other things that make events all use schedule_....

jbruestle:
Fairly nontrivial refactor of the lower level of the system.  Basically, the core goal here was really to split kernel creation from kernel launching, and to add addition 'effect' information so that it would be possible to hoist comp operations into init, but I additionally refactored a few other things that had some technical debt.  Specifically:

1) We do conversion to comp during outlining rather than in two stages (outlining -> launch to comp).  This simplified a number of things.  We also compute which buffers are read/written to, and only do transfers as needed.
2) I drastically simplified comp_to_llvm to create a more or less 1-1 mapping from comp ops to lllvm function calls
3) I fixed the OpenCL runtime wrappers to use these new calls (including the core change of making kernel creation explicit).

The changes (including the core kernel create/launch seperation) have *not* been applied to Vulkan, but in the interest of getting results on the OpenCL side I have turned off Vulkan CI after talking with folks about it.

Finally, it is possible that the changes may have broken the effectiveness of some of the COMP optimizations (although not in any way that CI caught), but I will be looking into that when I consider hoisting.

jbruestle:
Well, it looks like I must have pessimized something, since CI failed for perf, I'll take a look at it tomorrow when I have time.

flaub (approved):

PR #1626, opened 2020-12-17
Printing locations for EDSL errors

jd-bartlett96:
Only a few functions have been implemented so far, but I'm opening this draft PR for feedback on the current design. 

Remaining unsolved limitations of this approach include:
- A more robust method of checking for the existence of `std::source_location` or `std::experimental::source_location`. `__has_include` would be ideal but [I believe it is only available for >=C++17](https://en.cppreference.com/w/cpp/preprocessor/include)
- A way to obtain the `source_location` for variadic functions like `bind_dims(Ts...)`. I found [this fix](https://cor3ntin.github.io/posts/variadic/#source_location) for that issue, but could not get it to work, I believe due to [this bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79501)


plaidml/core/ffi.h
```diff
@@ -82,6 +98,8 @@ void plaidml_string_free(  //
 typedef struct {
   size_t code;
   plaidml_string* msg;
+  const char* origin_file;
+  uint32_t origin_line;
```
flaub, line 102: Seems like maybe we don't need this since the `call` wrappers already have the `loc` available directly.

flaub: This would be necessary if we want to somehow include file/line info in the originating throws that occur in the c++ AST code. But since you're just wrapping the exception from the caller side, then we probably don't need to pass this info down.

plaidml/core/core.h
```diff
@@ -21,6 +21,20 @@ inline std::string str(plaidml_string* ptr) {
...
 
+template <typename T, typename F, typename... Args>
+T call(edsl_source_location loc, F fn, Args... args) {
+  plaidml_error err;
+  err.origin_file = loc.file_name();
+  err.origin_line = loc.line();
+  auto ret = fn(&err, args...);
+  if (err.code) {
+    std::string fullmsg = "Exception at " + std::string(err.origin_file) + ":" + std::to_string(err.origin_line) +
+                          " with message: " + str(err.msg);
```
flaub, line 32: Could use a `std::stringstream` to build up this error message.

jd-bartlett96:
@flaub @tzerrell please take a look at the change I just pushed to the classes in `op.h` ... this allows the op classes to interact correctly with the modified operator overloads. I'm a bit worried that this could cause issues for anyone trying to add a new op; if they forget/don't know to include the `operator edsl::LocatedTensor` it will cause compiler issues that may be pretty inscrutable to them. 

Also, I saw `pmlc/dialect/oplib/tests/BasicTensorOp.td` which seems to be automatically generating an op class? If so, that process would need to be modified to insert the `operator edsl::LocatedTensor`, but I'm not seeing where that `BasicTensorOp` is actually being generated.

I'm thinking it could make sense to implement an abstract class for ops that ensures a `operator edsl::LocatedTensor` function is included along with the `operator edsl::Tensor` function?

plaidml/edsl/tests/edsl_error_test.cc
```diff
@@ -0,0 +1,61 @@
+// Copyright 2020 Intel Corporation
```
flaub, line 1: Could this get merged into the `edsl_test.cc`, or were you thinking that one is already too long? I'm fine to start splitting up the source file I guess, but I want to have a single binary instead of having lots of individual test executables.

plaidml/op/op.h
```diff
@@ -725,5 +725,49 @@ inline edsl::Tensor variance(const edsl::Tensor& I, const edsl::Value& axes = ed
...
+  OpTensor(explicit_padding op) : tensor(edsl::Tensor(op)) {}  // NOLINT
+  OpTensor(lrn op) : tensor(edsl::Tensor(op)) {}               // NOLINT
+  OpTensor(mvn op) : tensor(edsl::Tensor(op)) {}               // NOLINT
+  OpTensor(l2norm op) : tensor(edsl::Tensor(op)) {}            // NOLINT
+  OpTensor(relu op) : tensor(edsl::Tensor(op)) {}              // NOLINT
+  OpTensor(repeat op) : tensor(edsl::Tensor(op)) {}            // NOLINT
+  OpTensor(slice op) : tensor(edsl::Tensor(op)) {}             // NOLINT
+};
+
+#define PLAIDML_OPLIB_OP_CONVERSION(_op_)                                                               \
```
flaub, line 740: Why do we need to duplicate all of these?

plaidml/edsl/tests/edsl_error_test.cc
jd-bartlett96: My plan was to add tests here for more errors, as I add the `source_location` functionality to more of the API, and yes I was thinking that might start to be a bit long to just include with `edsl_test.cc`. I'll see how long it actually ends up being, but yes I can merge the binaries.

plaidml/op/op.h
jd-bartlett96: So, when I modified the operator overloads replacing the `Tensor` inputs with `LocatedTensor`, there were broader ramifications than I originally realized. All those operators work fine for `Tensor` operands, which now get implicitly converted to `LocatedTensors`, but there were a number of non-Tensor classes that were relying on those overloads to be cast into `Tensors`, and with the change to `LocatedTensors` no longer will be; ie, the compiler does not automatically convert them to `Tensors` then to `LocatedTensors`. One example of this is the `Contraction` class, which I was able to handle pretty easily by just adding a `Contraction` constructor to `LocatedTensor`. The real issue is for all of the classes defined in `op.h`; since `op.h` depends on `edsl.h`, I couldn't just include `op.h` and add the necessary constructors to `LocatedTensor` without creating a circular dependency. 

I'm not a huge fan of my fix here, but it was the best I could come up with. Other options I thought of would be to individually define all of the operators for each `op.h` class, which made the code messier without really offering any benefit to this approach, or to pull the `LocatedTensor` definition and associated operators into its own header file, which might be a bit neater, but I thought would lead to some awkward inclusion requirements (ie, anywhere `edsl.h` or `op.h` is included, after those inclusions). The ideal fix would be to make the relevant `op.h` classes available to `edsl.h` somehow but I couldn't figure out a way to do this.

plaidml/core/core.h
```diff
@@ -47,7 +47,14 @@ T call(edsl_source_location loc, F fn, Args... args) {
   auto ret = fn(&err, args...);
   if (err.code) {
     std::stringstream ss;
-    ss << "Exception at " << loc.file_name() << ":" << std::to_string(loc.line()) << " with message: " << str(err.msg);
+    if (!strstr(loc.file_name(), "edsl/edsl.h")) {
```
flaub, line 50: Avoid using `strstr` for c++ code, you can use `std::string::find` as an alternative.

plaidml/core/core.h
```diff
@@ -69,7 +76,14 @@ void call_void(edsl_source_location loc, F fn, Args... args) {
...
+      ss << "Exception at " << loc.file_name() << ":" << std::to_string(loc.line());
+    } else {
+      ss << "Exception at "
+         << "??"
+         << ":" << std::to_string(0);
+    }
+    ss << " with message: " << str(err.msg);
```
flaub: I see this common pattern of code duplicated. Having a small helper function might be useful.

flaub: I wonder if we could accomplish this with a custom exception class derived from `std::exception`

plaidml/core/core.h
```diff
@@ -5,12 +5,32 @@
...
 #include <stdexcept>
 #include <string>
 #include <vector>
 
 #include "plaidml/core/ffi.h"
 
+#if __has_include(<source_location>)
+#include <source_location>
+using edsl_source_location = std::source_location;
+#define ERRORTRACING true
```
flaub, line 18: Not sure this should go into the public header. It seems only be used by our own internal tests.

plaidml/core/core.h
```diff
@@ -5,12 +5,32 @@
...
 #include <functional>
 #include <memory>
+#include <sstream>
 #include <stdexcept>
 #include <string>
 #include <vector>
 
 #include "plaidml/core/ffi.h"
 
+#if __has_include(<source_location>)
```
flaub, line 15: You mentioned that this only works on C++17 and beyond, I think that's going to be a blocker. I think we need to support C++11 and beyond.

jd-bartlett96: I am fairly confident that if `__has_include()` isn't defined, it will default to the `#else` block automatically. I tested that with a `#if nonsensefunction(morenonsense)` which worked as desired. If you think I got this functionality wrong lmk and I can look at alternatives.

plaidml/core/core.h
jd-bartlett96: Ok; I can remove this and just make replace its use in `edsl_test.cc` with a similar `__has_include` check.

plaidml/core/core.h
jd-bartlett96: Sounds good; I'll try out making a custom exception and if I discover some compelling reason not to do that I'll make a helper function.

plaidml/core/core.h
dgkutnic: Instead of being "fairly confident," would it make sense to just test this on c++11 and see what happens?
In my mind this could be something as simple as building the target (and perhaps a small unit test) with `bazel build --cxxopt='-std=c++11'` but perhaps there are other dependencies I'm missing; perhaps Frank has a better idea.

jd-bartlett96: I was unable to build with c++11 due to llvm (and coincidentally, my code, for the same reason): both use `enable_if_t` which is not available in c++11. I believe the fix to that for my code would be a simple as changing it to `enable_if<..., true>`, but can't easily verify that works since nothing depending on llvm will build for me if I specify `--cxxopt='-std=c++11'`

jd-bartlett96: However `--cxxopt='-std=c++14'` builds & runs fine which further verifies my usage of `__has_include()` since it shouldn't be defined for c++14

flaub: Right, so in this situation, we need to figure out how to build just the edsl public portion with C++11 since LLVM does require C++14. I don't have a good solution to offer right now, it might require some tricky CMake code. It might be good enough to test with C++14 for now. Also @dgkutnic can we get some requirements on whether C++14 would be acceptable?

dgkutnic: @flaub from previous discussions I've had w/ others, my impression was that C++11 and above was the requirement. However, I'll take the AR to cycle back on this and get clarity for what we need for a minimum viable product. Ideally, we could get this landed with C++14 and above for the initial pass, and work later on the C++11 backwards compatibility when necessary. Happy to take this conversation offline.

plaidml/core/core.h
jd-bartlett96: I was getting a bizarre response when I tried to implement the Exception class; everything compiles & runs but the printed error is nonsense (eg "\xC0\x17\x16\x6"). If I put a print the error message in `what()` method right before I return it, it looks fine, but the message doesn't make it through the `return`. I tried several implementations, copying tutorials verbatim, etc; nothing resolved this. Couldn't find any mention of this issue online. So, I used a helper function instead, but if you've seen anything like this before & know how it might be fixed please lmk.

flaub: The trick with `what()` is that it must return a buffer that lives beyond the lifetime of the method. So you'd need to have a `std::string` member to store the memory. This string would be built up in the constructor of the exception. Then `what()` returns `msg.c_str()`

plaidml/core/core.h
```diff
@@ -160,11 +209,19 @@ class TensorShape {
...
     if (sizes.size() != strides.size()) {
-      throw std::runtime_error("Sizes and strides must have the same rank.");
+      std::stringstream ss;
+      if (!strstr(loc.file_name(), "edsl/edsl.h")) {
+        ss << "Exception at " << loc.file_name() << ":" << std::to_string(loc.line());
+      } else {
+        ss << "Exception at ??:0";
+      }
+      ss << " with message: Sizes and strides must have the same rank.";
+      throw std::runtime_error(ss.str());
```
flaub, line 221: Could this use the new helper?

plaidml/core/core.h
jd-bartlett96: I tested C++11 with `source_location` out locally outside of `edsl.h` with this:
```
#include <iostream>

#if __has_include(<source_location>)
#include <source_location>
using edsl_source_location = std::source_location;
#elif __cplusplus>=201402 and __has_include(<experimental/source_location>)
#include <experimental/source_location>
using edsl_source_location = std::experimental::source_location;
#else
struct edsl_source_location {
  static edsl_source_location current() { return edsl_source_location{}; }
  constexpr uint_least32_t line() const noexcept { return 0; }
  constexpr uint_least32_t column() const noexcept { return 0; }
  constexpr const char* file_name() const noexcept { return "??"; }
  constexpr const char* function_name() const noexcept { return "??"; }
};
#endif

int main() {
    auto loc = edsl_source_location::current();
    std::cout<<loc.file_name()<<":"<<loc.line()<<std::endl;
}
```

There was an issue with `__has_include`, but not the one expected; `__has_include` itself actually works fine back through C++11 (ie, regardless of C++ version `edsl_source_location` always gets defined). However, apparently with C++11 some experimental version of the `source_location` header is found, but not one which compiles (this appears to be a [known issue](https://features.isocpp.open-std.narkive.com/RkxjCiAU/sg10-checking-has-include-on-its-own-is-not-sufficient) with `__has_include`). The additional check to ensure C++>=14 handles this. So, I have vetted the definition of `source_location` for C++11-17 locally. 

As for the `enable_if` functionality that allows `source_location` to be passed through operator overloads, I switched from `enable_if_t` (which was only added in C++14) to `enable_if` (which should be fine for C++11). I've checked this for C++14-17, however I can't do a meaningful check for C++11 since llvm does not compile with C++11.

TLDR, the code I'm introducing _should_ work for C++11-17; it's been thoroughly tested for C++14-17; it's been tested to the extent possible for C++11 (barring implementing the tricky CMake code Frank mentioned).

plaidml/edsl/edsl.h
```diff
@@ -1150,16 +1252,60 @@ inline Tensor Tensor::operator-() const { return intrinsicCall("neg", {*this});
...
+
+  template <typename T, typename = enable_if_t<has_tensor<T>::value>>
+  LocatedTensor(T tensor, edsl_source_location loc = edsl_source_location::current())  // NOLINT
+      : tensor(tensor), loc(loc) {}
+
+  LocatedTensor(Tensor tensor, edsl_source_location loc = edsl_source_location::current())  // NOLINT
+      : tensor(tensor), loc(loc) {}
+};
+
+#ifdef _WIN32
```
flaub, line 1284: I think perhaps this should be `#ifdef MSVC`? or whatever the appropriate define to use for detecting MSVC instead of WIN32 (imagine if someone was trying to compile the EDSL using gcc-mingw32 on windows).

plaidml/core/core.h
flaub: SGTM

plaidml/core/core.h
```diff
@@ -3,14 +3,51 @@
...
+#endif
+
+class ffi_exception : public std::exception {
+  std::string msg_;
+  edsl_source_location loc;
+
+ public:
+  explicit ffi_exception(const std::string& message, edsl_source_location loc) : loc(loc) {
+    std::stringstream ss;
+    if (!strstr(loc.file_name(), "edsl/edsl.h")) {
```
flaub, line 39: Avoid using `strstr` in C++ code, you can use `std::string::find` as an alternative.

plaidml/core/core.h
```diff
@@ -3,14 +3,51 @@
...
+  constexpr uint_least32_t line() const noexcept { return 0; }
+  constexpr uint_least32_t column() const noexcept { return 0; }
+  constexpr const char* file_name() const noexcept { return "??"; }
+  constexpr const char* function_name() const noexcept { return "??"; }
+};
+#endif
+
+class ffi_exception : public std::exception {
+  std::string msg_;
+  edsl_source_location loc;
```
flaub, line 34: Add `private:` just so that it's explicit (even though the default for `class` is to have private members).

plaidml/core/core.h
```diff
@@ -3,14 +3,51 @@
...
+    std::stringstream ss;
+    if (!strstr(loc.file_name(), "edsl/edsl.h")) {
+      ss << "Exception at " << loc.file_name() << ":" << std::to_string(loc.line());
+    } else {
+      ss << "Exception at ??:0";
+    }
+    ss << " with message: " << message;
+    msg_ = ss.str();
+  }
+  virtual ~ffi_exception() noexcept {}
```
flaub, line 47: Is this needed?

plaidml/core/core.h
jd-bartlett96: Ah my bad; I'd removed that but must've copied it back in during my `Exception` experimentation

plaidml/edsl/edsl.h
```diff
@@ -389,7 +413,7 @@ class IndexedTensor {
   ///
   /// Performs a multiplication combination within a contraction.
   ///
-  IndexedTensor operator*(const IndexedTensor& rhs) const { return IndexedTensor(PLAIDML_COMBO_OP_MUL, {*this, rhs}); }
+  IndexedTensor operator*(const IndexedTensor rhs) const { return IndexedTensor(PLAIDML_COMBO_OP_MUL, {*this, rhs}); }
```
flaub, line 416: Was this intentional?

jd-bartlett96: Nope, artifact of an old change I undid; fixed

plaidml/core/core.h
jd-bartlett96: No; removed it.

flaub (approved):

PR #1627, opened 2020-12-17
Hoisting added to gpu ocl pipeline

abialas1:
Excluded constant ops from hoisting, due to constant folding problems in the lower dialect canonicalizations.
Changed ‘maybeCrossFunc’ to ‘inInit’ per proposition from Jeremy.

### DIFF
--- a/pmlc/transforms/hoisting.cc
+++ b/pmlc/transforms/hoisting.cc
@@ -6,6 +6,7 @@
 #include "mlir/Support/DebugStringHelper.h"
 
 #include "pmlc/dialect/stdx/ir/ops.h"
+#include "pmlc/dialect/tile/ir/ops.h"
 #include "pmlc/transforms/pass_detail.h"
 #include "pmlc/util/logging.h"
 
@@ -49,7 +50,7 @@ private:
   UnpackOp mainUnpack;
   UnpackOp finiUnpack;
   // List of values to consider as boundry crossing in disconnect
-  DenseSet<Value> maybeCrossFunc;
+  DenseSet<Value> inInit;
 };
 
 bool HoistingState::matchesProtocol() {
@@ -89,7 +90,7 @@ void HoistingState::connectFunctions() {
   for (unsigned i = 0; i < initPack.getNumOperands(); i++) {
     mainUnpack.getResult(i).replaceAllUsesWith(initPack.getOperand(i));
     finiUnpack.getResult(i).replaceAllUsesWith(initPack.getOperand(i));
-    maybeCrossFunc.insert(initPack.getOperand(i));
+    inInit.insert(initPack.getOperand(i));
   }
 }
 
@@ -106,11 +107,14 @@ void HoistingState::doHoisting() {
       continue;
     if (!innerEffect.hasNoEffect())
       continue;
+    // Skip constant ops
+    if (isa<pmlc::dialect::tile::ConstantOp>(innerOp))
+      continue;
     // Check if all operands (if any) are in init
     IVLOG(3, "Checking operands " << debugString(innerOp));
     bool allOperandsInInit = true;
     for (auto operand : innerOp.getOperands()) {
-      if (!maybeCrossFunc.count(operand)) {
+      if (!inInit.count(operand)) {
         allOperandsInInit = false;
         break;
       }
@@ -122,7 +126,7 @@ void HoistingState::doHoisting() {
     // Yes?  Hoist to init, add results as possible cross function results
     innerOp.moveBefore(initPack);
     for (auto result : innerOp.getResults()) {
-      maybeCrossFunc.insert(result);
+      inInit.insert(result);
     }
   }
 }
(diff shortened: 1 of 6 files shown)


pmlc/transforms/hoisting.cc
```diff
@@ -106,11 +107,14 @@ void HoistingState::doHoisting() {
...
       continue;
+    // Skip constant ops
+    if (isa<pmlc::dialect::tile::ConstantOp>(innerOp))
+      continue;
     // Check if all operands (if any) are in init
     IVLOG(3, "Checking operands " << debugString(innerOp));
     bool allOperandsInInit = true;
     for (auto operand : innerOp.getOperands()) {
-      if (!maybeCrossFunc.count(operand)) {
+      if (!inInit.count(operand)) {
```
jbruestle, line 117: In addition to skipping constants, we need to allow constant operands even if they are not inInit here.

pmlc/transforms/hoisting.cc
```diff
@@ -122,7 +126,7 @@ void HoistingState::doHoisting() {
     // Yes?  Hoist to init, add results as possible cross function results
     innerOp.moveBefore(initPack);
```
jbruestle, line 127: Additionally, when moving an op, if it uses constants, we need to also copy them into init.

jbruestle (changes requested):
Currently this change as is also prevents hoisting of any ops that use constants, I added comments where additional code is needed.  Clearly I also need to write some hoisting tests, since CI was happy with this.

abialas1:
thanks, let me address these tomorrow

jbruestle:
Absent a proper lit test, looking at the IR for //plaidml/edsl/tests:cc_test for --gtest_filter=CppEdsl.MixedAdd would be good to verify hoisting works, since the initial add is hoisted at the tile level for that test.

abialas1:
I looked at the generated code for CppEdsl.MixedAdd with and without my changes, and it looks the same. The problem here is that these constants are part of the function's arguments, so there is no actual constantOp being created. I can try to add some test tomorrow for this, just wondering if user can actually create constantOp on the edsl level, that would not be part of the function arguments? If it is not the case then the orginal change should be sufficient?

jbruestle:
Sorry, wrong test.  The example I was thinking of was BackendTests.testResizeImagesBilinear, which actually hoists a contraction which relies on the init constant for the contraction, and post hoisting looks like this:

module @get_value {
  func @init() -> !stdx.argpack {
    %cst = tile.constant(0.33333333333333331 : f64) : tensor<f32>
    %cst_0 = tile.constant(2.500000e-01 : f64) : tensor<f32>
    %cst_1 = tile.constant(0.000000e+00 : f64) : tensor<f32>
    %0 = tile.contract assign, none, %cst_1, %cst {lowerBounds = #map0, sink = #map1, srcs = [#map2], upperBounds = #map3} : tensor<f32>, tensor<f32> -> tensor<3xf32>
    %1 = tile.contract add, none, %cst_1, %0 {cons = #set0, lowerBounds = #map4, sink = #map5, srcs = [#map6], upperBounds = #map7} : tensor<f32>, tensor<3xf32> -> tensor<5xf32>
    %2 = tile.contract assign, none, %cst_1, %cst_0 {lowerBounds = #map0, sink = #map1, srcs = [#map2], upperBounds = #map8} : tensor<f32>, tensor<f32> -> tensor<4xf32>
    %3 = tile.contract add, none, %cst_1, %2 {cons = #set1, lowerBounds = #map4, sink = #map5, srcs = [#map9], upperBounds = #map10} : tensor<f32>, tensor<4xf32> -> tensor<7xf32>
    %4 = tile.contract assign, mul, %cst_1, %1, %3 {lowerBounds = #map4, sink = #map11, srcs = [#map5, #map12], upperBounds = #map13} : tensor<f32>, tensor<5xf32>, tensor<7xf32> -> tensor<5x7xf32>
    %5 = stdx.pack %cst_1, %4 : tensor<f32>, tensor<5x7xf32>
    return %5 : !stdx.argpack
  }
  func @main(%arg0: !stdx.argpack, %arg1: tensor<1x1x2x3xf32>) -> tensor<1x1x3x8xf32> {
    %0:2 = stdx.unpack %arg0 : tensor<f32>, tensor<5x7xf32>
    %1 = tile.contract add, mul, %0#0, %arg1, %0#1 {cons = #set2, lowerBounds = #map14, sink = #map15, srcs = [#map16, #map17], upperBounds = #map18} : tensor<f32>, tensor<1x1x2x3xf32>, tensor<5x7xf32> -> tensor<1x1x6x12xf32>
    %2 = tile.contract assign, none, %0#0, %1 {lowerBounds = #map19, sink = #map20, srcs = [#map20], upperBounds = #map21} : tensor<f32>, tensor<1x1x6x12xf32> -> tensor<1x1x3x8xf32>
    return %2 : tensor<1x1x3x8xf32>
  }
  func @fini(%arg0: !stdx.argpack) {
    %0:2 = stdx.unpack %arg0 : tensor<f32>, tensor<5x7xf32>
    return
  }
}

However, honestly, I think it's unreasonable to block a commit based on an example that appears in no lit tests, and none of this actually effects any real network, so I'm going to OK this commit.  I need to make modifications to the hoisting pass soon anyway to hoist Alloc/Dealloc pairs and idempotent things to get all of the comp stuff to hoist anyway, so I can fix the constant thing then if it's actually blocking anything.

jbruestle (approved):

fini:
LOL I was mentioned and notified on this because my username is fini :) Had no clue what was going on. Cheers!

PR #1631, opened 2020-12-18
Weights reorders moved to init function

abialas1:
In this checkin I assume that when stdx.unpack op is present then init/main/fini functions are valid and contain single pack\unpack op.
I am going to add some lit tests later.

### DIFF
--- a/pmlc/dialect/pxa/transforms/reorder_layouts.cc
+++ b/pmlc/dialect/pxa/transforms/reorder_layouts.cc
@@ -292,10 +292,10 @@ naiveScheduleModel(mlir::ArrayRef<mlir::AffineParallelOp> loopNest) {
 // ============================================================================
 // Helper function to get pack\unpack ops.
 // ============================================================================
-template <typename funcType>
-void getPackOp(funcType &packOp, mlir::FuncOp &funcOp) {
+template <typename OpType>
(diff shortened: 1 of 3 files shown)


jbruestle (approved):
Looks good!  I think overall both this and the hoisting pass indicate that the init/main/fini representation is a little awkward, which I think would be improved by a block-based implementation, and probably we need some utility functions either way, but it's good to get examples of use cases first anyway.

pmlc/dialect/pxa/transforms/reorder_layouts.cc
```diff
@@ -286,24 +289,101 @@ naiveScheduleModel(mlir::ArrayRef<mlir::AffineParallelOp> loopNest) {
...
+  if (!packOps.empty())
+    packOp = *packOps.begin();
+}
+
+// ============================================================================
+// Helper function to replace unpackOps with updated types in sync with packOp.
+// ============================================================================
+pmlc::dialect::stdx::UnpackOp
+updateUnpackOp(pmlc::dialect::stdx::UnpackOp &unpackOp,
+               pmlc::dialect::stdx::PackOp &packOp) {
```
flaub, line 308: Ops should be passed by-value. Like `pmlc::dialect::stdx::UnpackOp unpackOp`

Issue #1612, opened 2020-12-11
Consider splitting updating vs writing Scatter


tzerrell:
The current Scatter (as of https://github.com/plaidml/plaidml/pull/1610) has a unified API with a parameter controlling whether to override or sum into existing results. For scatters where we know summing isn't needed, we currently need to pass a data tensor for API consistency with summing scatter (rather than for algorithmic reasons). We could instead split scatter into two versions and thereby avoid passing the data tensor to a non-updating scatter. Consider whether we want to make this change.

More details available in the linked PR.

haoyouab:
Another point that is worth noting is the behavior of updating scatter with duplicate indices.

According  to [tf.tensor_scatter_nd_update](https://www.tensorflow.org/api_docs/python/tf/tensor_scatter_nd_update):

> If indices contains duplicates, then their updates are accumulated (summed).

while [tf.compat.v1.scatter_update](https://www.tensorflow.org/api_docs/python/tf/compat/v1/scatter_update) shows some conflicts:

> If values in ref is to be updated more than once, because there are duplicate entries in indices, the order at which the updates happen for each value is undefined.

The current implementation  does not take summation into account (openvino ops either) in updating modes but I think we should be aware of this in case any changes on these.

flaub:
Is it possible to have a single primitive but then have 2 APIs that use this primitive?

tzerrell:
> Is it possible to have a single primitive but then have 2 APIs that use this primitive?

I think the answer is no, but perhaps MLIR can be flexible in ways I don't know about.

The fundamental distinction between the two cases is that the update case requires us to read an initial `data` tensor of values that will be updated (or used as the default values if never updated). Whereas the write case will create a new tensor  that is everywhere zero (except where scatter writes), so no initial `data` needs to be read.

I don't know how to unify that into a single primitive without forcing a `data` tensor to passed in the write case even though it will never be used. But perhaps there's a mechanism in MLIR for optional inputs that I'm unaware of? In which case we should use that in creating a unified primitive.

tzerrell:
I suppose it might also be possible to pass an empty `data` tensor in the write case (and be sure to get the sizing data from the output shape, _not_ the `data` tensor)? Not sure if that creates too confusing of an interface (i.e. because the shape of the `data` represents the output shape in the update case, but is unrelated to the output shape in the write case).

jbruestle:
[...]

PR #1636, opened 2020-12-21
add initialization tensor to scatter summing case

haoyouab:
As discussed in #1612, a unified API for `scatter` with an initialization tensor for all cases is preferred.


tzerrell (approved):
LGTM

PR #1638, opened 2020-12-21
Add ROIPooling op

XingHongChenIntel:
Main "ROI" background information, please ref https://arxiv.org/pdf/1803.07066.pdf and https://arxiv.org/pdf/1504.08083.pdf

According to openvino spec(https://docs.openvinotoolkit.org/latest/openvino_docs_ops_detection_ROIPooling_1.html), the ROIPooling have two pooling method.

- max
this is the basic ROIPooling operation method. please reference the below paper and openvino ngraph implement.

  > https://arxiv.org/pdf/1807.11590v1.pdf Fig.6 : Illustration of RoI Pooling, RoI Align and PrROI Pooling.
  > https://github.com/openvinotoolkit/openvino/blob/master/ngraph/core/reference/include/ngraph/runtime/reference/roi_pooling.hpp

- bilinear
this means that interpolating only a single value at each bin center (without pooling). please ref below paper and ngraph implement.

  > https://openaccess.thecvf.com/content_ICCV_2017/papers/He_Mask_R-CNN_ICCV_2017_paper.pdf comment 2
  > https://github.com/openvinotoolkit/openvino/blob/master/ngraph/core/reference/include/ngraph/runtime/reference/roi_pooling.hpp

Work with plaidml/openvino#123 to solve plaidml/openvino#78, change openvino commit to enable ROIPooling.


plaidml/bridge/openvino/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ if(NOT InferenceEngineDeveloperPackage_FOUND)
   FetchContent_Declare(
     openvino
     GIT_REPOSITORY https://github.com/plaidml/openvino.git
-    GIT_TAG        04750e225e633fd07ae28b2bf28c3e8202ad7c2b
+    GIT_TAG        ea65cfebb27b5fd9be8bb3c602746a79d03d5b9d
```
tzerrell, line 29: This will need to be updated with the hash of the now-landed commit

tzerrell (commented):
Looks good for now; we will want to revisit once we can handle the coordinate box sizes dynamically.

plaidml/bridge/openvino/CMakeLists.txt
XingHongChenIntel: ok, fix this

tzerrell (approved):

plaidml/bridge/openvino/ops/roi_pooling.cpp
```diff
@@ -0,0 +1,155 @@
...
+  auto* ngraph_const = ngraph::as_type<ngraph::op::Constant>(layer->get_input_node_ptr(operand_idx));
+  if (ngraph_const) {
+    return ngraph_const->cast_vector<T>();
+  } else {
+    THROW_IE_EXCEPTION
+        << "Dynamic coordinate is not currently supported by PlaidML plugin; all coordinate must be Constants. ";
+  }
+}
+
+edsl::Tensor crop_max_pooling(edsl::Tensor I, std::vector<float>& coord, int64_t pooled_h, int64_t pooled_w) {
```
flaub, line 28: Looks like `coord` should be `const std::vector<float>& coord`? Is this intended to be an in/out parameter?

XingHongChenIntel: `coord` don't intended to be an in/out parameter, so you are right, it should be  `const std::vector<float>& coord`. I fix this in PR https://github.com/plaidml/plaidml/pull/1641

PR #1639, opened 2020-12-21
add EmbeddingSegmentsSum op

haoyouab:
This is based on PR #1636.

Note: as discussed before, there's no 'zero-like' operation currently so `broadcast` op with `compute_shape` to get the shape of tensors is used.


tzerrell (approved):
LGTM

PR #1640, opened 2020-12-22
Add OpenVino plugin Onehot Op

LiyangLingIntel:
Implementation refers to plaidml#1173.


tzerrell (approved):
LGTM

plaidml/bridge/openvino/ops/one_hot.cpp
```diff
@@ -0,0 +1,56 @@
+// Copyright (C) 2020 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <iostream>
```
flaub, line 5: This should be dropped (is it still in use?)

plaidml/bridge/openvino/ops/one_hot.cpp
```diff
@@ -0,0 +1,56 @@
...
+using namespace InferenceEngine;  // NOLINT[build/namespaces]
+
+namespace PlaidMLPlugin {
+
+void registerOneHot() {
+  registerOp("onehot", [](const Context& ctx) {
+    IE_ASSERT(ctx.operands.size() == 4);
+
+    auto indices = ctx.operands.at(0);
+    auto depth = get_shape_from_constant_operand(1, ctx.layer);  // ctx.operands.at(1);
```
flaub, line 24: The comment at the end of the line should be dropped (or there should be an additional comment about it)

Issue #126, opened 2018-05-23
What's the backend engine on Mac?


PhoenixShield:
What's the backend engine the framework use while running on macOS with AMD GPU?
Is it OpenGL, Metal, or RoCM?

choongng:
We recommend Metal for Mac due to better performance but OpenCL will work as well.

PhoenixShield:
Thanks all! I'll try both.

lesves:
This should be in an FAQ or something like that.

PR #1643, opened 2020-12-23
Add OpenVINO plugin ROIAlign Op

LiyangLingIntel:
As ROIAlign still not been added in the latest OpenVINO upstream, current implementation of ROIAlign Op refers to 
* [OpenVINO ROIAlign Spec](https://docs.openvinotoolkit.org/latest/openvino_docs_ops_detection_ROIAlign_3.html)
* [Mask R-CNN](https://openaccess.thecvf.com/content_ICCV_2017/papers/He_Mask_R-CNN_ICCV_2017_paper.pdf)
* Partially refers to ROIPooling op #1638

One thing should be mensioned is attribute `sampling_ratio` is not in used in current implementation.
In Spec, its description is 
> *sampling_ratio* is the number of bins over height and width to use to calculate each output feature map element. If the value
is equal to 0 then use adaptive number of elements over height and width: `ceil(roi_height / pooled_h)` and `ceil(roi_width / pooled_w)` 

While in Spec there also says 
> Apply bilinear interpolation with 4 points in each bin and apply maximum or average pooling based on *mode* attribute to produce output feature map element.

and in [paper reference](https://openaccess.thecvf.com/content_ICCV_2017/papers/He_Mask_R-CNN_ICCV_2017_paper.pdf) footnote 2
> We sample four regular locations, so that we can evaluate either max or average pooling. 

If we use 4 points (2 on height and 2 on width) to generate each pooled feature map bin, `sampling_ratio` has to be fixed.
So in current implementation, I directly use 2*2 pooling size and assert `sampling_ratio` equals to 0.

Need openvino [Add ROIAlign Op #126](https://github.com/plaidml/openvino/pull/126).


tzerrell (commented):
LGTM but will need the OV reference updated to be able to utilize https://github.com/plaidml/openvino/pull/126

LiyangLingIntel:
> LGTM but will need the OV reference updated to be able to utilize [plaidml/openvino#126](https://github.com/plaidml/openvino/pull/126)

Updated.

tzerrell (approved):

PR #1646, opened 2020-12-25
Add OpenVINO plugin RegionYolo Op

YangleiZouIntel:
This PR needs https://github.com/plaidml/openvino/pull/127

Algorithm follows [OpenVino RegionYolo reference inplementation](https://github.com/openvinotoolkit/openvino/blob/master/ngraph/core/reference/include/ngraph/runtime/reference/region_yolo.hpp)

More information about OpenVINO yolo region layer:  [OpenVINO Region layer](https://docs.openvinotoolkit.org/2020.1/_docs_MO_DG_prepare_model_convert_model_tf_specific_Convert_YOLO_From_Tensorflow.html)

The region layer is first introduced in [darknet region layer](https://github.com/pjreddie/darknet/blob/master/src/region_layer.c)


plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/region_yolo.cpp
```diff
@@ -0,0 +1,51 @@
...
+const std::vector<size_t> num_regions = {5, 9};
+const size_t coords = 4;
+const int start_axis = 1;
+const int end_axis = 3;
+
+const auto testCase_yolov3 = ::testing::Combine(
+    ::testing::ValuesIn(inShapes_v3), ::testing::Values(classes[0]), ::testing::Values(coords),
+    ::testing::Values(num_regions[1]), ::testing::Values(do_softmax[1]), ::testing::Values(masks[2]),
+    ::testing::Values(start_axis), ::testing::Values(end_axis), ::testing::Values(InferenceEngine::Precision::FP32),
+    ::testing::Values(CommonTestUtils::DEVICE_PLAIDML));
```
tzerrell, line 32: As I learned [here](https://github.com/plaidml/plaidml/pull/1644#discussion_r548315375), the linter will put each element of a list on its own line if you end the list with a trailing comma. So this might look better if you put a comma between the `))` at the end of this line (and similarly for the other `::testing:Combine` lines).

tzerrell (commented):
Looks good! I have a linting comment and this will need https://github.com/plaidml/openvino/pull/127 updated & landed as well.

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/region_yolo.cpp
YangleiZouIntel: Thanks,  I have updated the vector initialization lists with a trailing comma and it works very well. 

It seems `::testing:Combine` does not support trailing comma. I added `//` for each line so the code is more readable.

tzerrell (commented):
LGTM, will approve once the OV reference is updated to the landed version of the commit with RegionYOLO

tzerrell (approved):

PR #1651, opened 2020-12-30
Fix split-main pass

shiyao00:


pmlc/dialect/tile/tests/split_main.mlir
```diff
@@ -0,0 +1,18 @@
+// RUN: pmlc-opt -tile-split-main="main-function=split" %s | FileCheck %s
+
+func @split(%arg0: tensor<16x16xf32>) -> tensor<16x16xf32> {
+  return %arg0 : tensor<16x16xf32>
+}
+
+// CHECK-LABEL: @init
```
flaub, line 7: CHECK-LABEL is typically used for dividing a file into multiple logical tests.

https://llvm.org/docs/CommandGuide/FileCheck.html#the-check-label-directive

shiyao00: removed

jbruestle (approved):

PR #1664, opened 2021-01-14
fix devector bug non-broadcast block read / write

adstraw:


abialas1 (approved):
Right, looks that this condition was wrong

PR #1667, opened 2021-01-14
Porting docs to cmake

jd-bartlett96:
Added docs to cmake project. Build them with:
```cmake --build . --target docs```

I also modified a `pml_cc_library` in `pmlc\dialect\pxa\ir\CMakeLists.txt`; this is unrelated to the docs but is fixing what I believe is a small error I noticed while building the CMake project (`interfaces.cc.inc` was being generated but not found by `interfaces.cc` causing build errors for some targets).

### DIFF
--- a/environment-windows.yml
+++ b/environment-windows.yml
@@ -3,6 +3,7 @@ channels:
 dependencies:
   - clang-tools
   - conda=4.6
+  - doxygen
   - numpy=1.16
   - pip=19
   - posix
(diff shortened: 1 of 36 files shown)

### CODE REVIEW

File: environment-windows.yml [hunk 1]
```diff
@@ -3,14 +3,21 @@ channels:
 dependencies:
   - clang-tools
   - conda=4.6
+  - doxygen
```
flaub, line 6: If you could pin these to specific versions, it will:
* make conda run much faster since it doesn't have to consider many options when trying to solve deps
* ensure reproducible builds in the future and for new clones

flaub (approved):

PR #1670, opened 2021-01-16
Add whole-block option for cache pass

shiyao00:


jbruestle (approved):
Looks good, nice clear tests as well.  Approving as is, however one question I have regarding a possible optimization: Do we need to enforce wholeInner on *all* dimensions, or just the low stride dimension?  At least on some hardware, I would imagine a 2d transfer will consist of N 1-d transfers, and each 1-d transfer must be contiguous, but the higher stride dimension wouldn't have that requirement.  So perhaps three modes: 1) No contig requirement 2) contig only for stride 1 dim, 3) contig for all dims?  Regardless, this is an improvement as is.

PR #1672, opened 2021-01-18
llvm update + vectorize mem and layouts reorder uncommented

abialas1:


jbruestle:
Hey, so due to requirements from multiple fronts, Frank and I have been working this week to pull in a much more recent version of upstream (https://github.com/plaidml/plaidml/pull/1676).  It should be in today I think.  You probably need to rebase this on top of that, although hopefully that will actually make this commit even smaller.

abialas1:
Thanks, makes sense. I rebased the code, had to revert some changes in extract/insertelement as got functional fails after uncommenting vectorize mem pass

flaub (commented):
LGTM

jbruestle (approved):

PR #1673, opened 2021-01-19
benchmark_app package

mwyi:


tzerrell (commented):
LGTM but I'm not a CMake expert and the build is failing.

flaub (approved):

flaub:
@mwyi Is this actually used anywhere? I think I need to revert it.

PR #1674, opened 2021-01-20
Pin pyyaml to 5.3.1

mwyi:
Pinned pyyaml to 5.3.1 as pypi was updated to 5.4 today (01-19-2021) and plaidml-v1 build fails.

### DIFF
--- a/environment-windows.yml
+++ b/environment-windows.yml
@@ -9,7 +9,7 @@ dependencies:
   - posix
   - pre_commit=1.17.0
   - python=3.7.0
-  - pyyaml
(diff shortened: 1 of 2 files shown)


drgone (approved):

environment-windows.yml
```diff
@@ -9,7 +9,7 @@ dependencies:
   - posix
   - pre_commit=1.17.0
   - python=3.7.0
-  - pyyaml
+  - pyyaml==5.3.1
```
flaub, line 12: I think for conda dependencies (these aren't pip deps), we need a single `=`

PR #1678, opened 2021-01-22
Add a verifier to ReshapeOp

tzerrell:
Add a verifier to the Tile dialect `ReshapeOp` to ensure it doesn't cast or change the overall size of the tensor. E.g., the following is not a legal reshape because the product of 1x112x112x64 is 802816, which does not equal 1x12544:
```
tile.reshape %{{.*}} : (tensor<1x112x112x64xf32>) -> tensor<1x12544xf32>
```

### DIFF
--- a/pmlc/.clang-format
+++ b/pmlc/.clang-format
@@ -1,2 +1,2 @@
 BasedOnStyle: LLVM
-AlwaysBreakTemplateDeclarations: Yes
+AlwaysBreakTemplateDeclarations: true
--- a/pmlc/.clang-format
+++ b/pmlc/.clang-format
@@ -1,2 +1,2 @@
 BasedOnStyle: LLVM
-AlwaysBreakTemplateDeclarations: true
+AlwaysBreakTemplateDeclarations: Yes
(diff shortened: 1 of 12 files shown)


pmlc/dialect/tile/ir/ops.cc
```diff
@@ -322,6 +322,35 @@ LogicalResult verifyContractionOp(ContractionOp op) {
...
+  if (inEltType != outEltType) {
+    return op.emitOpError(
+               "reshape op must not cast, but found input element type ")
+           << inEltType << " and output element type " << outEltType;
+  }
+  auto inShape = inType.getShape();
+  int64_t inTotalSize = 1;
+  for (auto dim : inShape) {
+    inTotalSize *= dim;
+  }
```
flaub, line 340: Use this instead:
```
int64_t inNumElements = inType.getNumElements();
```

tzerrell: Ah, great, thanks!

pmlc/.clang-format
```diff
@@ -1,2 +1,2 @@
 BasedOnStyle: LLVM
-AlwaysBreakTemplateDeclarations: Yes
+AlwaysBreakTemplateDeclarations: true
```
flaub, line 2: Which version of `clang-format` are you using? I believe we require version 9.

tzerrell: ah, apparently 6.something, I'll update, thanks

flaub (approved):

PR #1679, opened 2021-01-25
Update OpenVINO PlaidML plugin

YangleiZouIntel:
This patch works with https://github.com/plaidml/openvino/tree/update_openvino, which currently is based on [this OpenVINO commit](https://github.com/openvinotoolkit/openvino/tree/5408f611e9130555bc617266b5e77aace9e5f618).

we have finished updating/fixing the following ops, due to reference implementation, test setup changes in upstream.
 In addition, all op test files have been updated. 
- [x]  Activation (including two ops in opset5: Hsigmoid, Round) - Liyang
- [x] 	BatchNormInference (updated to opset5 version) - Xin
- [x] 	ConvolutionBackpropData - Xin
- [x] 	Eltwise - Xin
- [x] 	ExtractImagePatches - Xinghong
- [x] 	Interpolate - Yanglei
- [x] 	Lrn - Zhibin
- [x] 	MatMul - Zhibin
- [x] 	Pooling - Zhibin
- [x] 	ReduceOp - Liyang
- [x] 	ReorgYolo - Xinghong
- [x] 	ReverseSequence -Yanglei
- [x] 	ScatterElementsUpdate - Zhibin
- [x] 	ScatterNDUpdate - Zhibin
- [x] 	SqueezeUnsqueeze - Xinghong
- [x]  StridedSlice - Xinghong

We are currently working on 
- [ ] 	ROIAlign
- [ ] 	ROIPooling

This is a big patch, please let me know if you prefer to split it into small ones.


plaidml/bridge/openvino/ops/round.cpp
```diff
@@ -0,0 +1,49 @@
...
+  registerOp("Round", [](const Context& ctx) {
+    IE_ASSERT(ctx.operands.size() == 1);
+    auto I = ctx.operands.at(0);
+    auto* layer = ngraph::as_type<ngraph::opset5::Round>(ctx.layer);
+    auto roundMode = layer->get_mode();
+
+    switch (roundMode) {
+      case ngraph::op::v5::Round::RoundMode::HALF_TO_EVEN: {
+        auto unsignedI = op::abs(I);
+        auto intPart = edsl::cast(unsignedI, DType::INT32);
```
tzerrell, line 27: What happens if this overflows I32? Maybe this is a moot point because FP32 can't represent every integer beyond (IIRC) 2^24, so rounding some float in the vicinity of 2^50 is going to be a no-op and the rounding mode doesn't matter. But what about an FP64 case?

This might just mean we want a round mode on edsl::round in the long run. In the short term though, perhaps we need to determine the type of the cast by the type of the input so that FP64 gets INT64 for the intPart

plaidml/bridge/openvino/ops/squeeze.cpp
```diff
@@ -19,10 +19,8 @@ void registerSqueeze() {
   registerOp("squeeze", [](const Context& ctx) {
     IE_ASSERT(ctx.operands.size() == 2);
     auto I = ctx.operands.at(0);
-    auto axes = get_axis_set_from_constant_operand(1, ctx.layer);
```
tzerrell, line 22: [marking another `get_axis_set_from_constant_operand` -- see above comment]

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/activation.cpp
```diff
@@ -63,23 +75,49 @@ std::map<std::vector<size_t>, std::vector<std::vector<size_t>>> preluBasic = {
...
-// INSTANTIATE_TEST_CASE_P(smoke_Prelu_Param, ActivationParamLayerTest, basicPreluCases,
-//                         ActivationLayerTest::getTestCaseName);
+INSTANTIATE_TEST_CASE_P(                  //
+    smoke_Activation_Basic_Prelu,         //
+    ActivationLayerTest,                  //
+    basicPreluCases,                      //
+    ActivationLayerTest::getTestCaseName  //
+);
+
+// ActivationParamLayerTest fails in nGraph
```
tzerrell, line 114: Is this also commented out in CPU plugin version?

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/convolution_smoke.cpp
```diff
@@ -1,246 +0,0 @@
-// Copyright (C) 2019 Intel Corporation
```
tzerrell, line 1: Thanks for cleaning up the old *_smoke.cpp tests too!

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/roi_align.cpp
```diff
@@ -2,62 +2,45 @@
...
+    ::testing::Values(CommonTestUtils::DEVICE_PLAIDML)    //
 );
 
-INSTANTIATE_TEST_CASE_P(smoke, ROIAlignLayerTest,
-                        ::testing::Combine(smokeRoiAlignParams,                                  //
-                                           ::testing::ValuesIn(netPrecisions),                   //
-                                           ::testing::Values(CommonTestUtils::DEVICE_PLAIDML)),  //
-                        ROIAlignLayerTest::getTestCaseName);
-}  // namespace
+// INSTANTIATE_TEST_CASE_P(smoke_TestsROIAlign_max, ROIAlignLayerTest, ROIAlignCases_max,
```
tzerrell, line 45: Is this the part you're still working on? (and the commented out test cases in RoI Pooling as well)

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/split.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2020 Intel Corporation
+// Copyright (C) 2019 Intel Corporation
```
tzerrell, line 1: We should be updating copyright on files we're modifying to 2021, not reverting to 2019. Even in cases where we're largely basing the files on older non-PlaidML OpenVINO tests, the combination of those tests with the PlaidML cases is novel work we're doing right now.

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/strided_slice.cpp
```diff
@@ -7,181 +7,290 @@
...
+        {1, 1, 1, 1, 1},
+        {1, 0, 1, 1, 1},
+        {1, 0, 1, 1, 1},
+        {},
+        {0, 1, 0, 0, 0},
+        {},
+    },
+
+    /// those two case will run into segment fault, it is caused by openvino inference,
+    /// which is locate in 'MakeNextStageTask' of 'ie_infer_async_request_thread_safe_default.hpp'
```
tzerrell, line 259: Can we have a longer discussion on the problem here? A key question: does this also cause problems for the CPU plugin?

tzerrell (commented):
This is a big PR so I have a number of comments, but overall this looks very good! Some of the more minor issues I raised could likely be postponed for followup PRs if they become the only blockers.

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/roi_align.cpp
YangleiZouIntel: Yes, we are still working on updating Roi Pooling and RoI Align.

plaidml/bridge/openvino/ops/batch_norm_inference.cpp
```diff
@@ -21,11 +21,11 @@ void registerBatchNormInference() {
     // This is because the tests show that the input is expected to be the third input tensor. Presumably either the OV
```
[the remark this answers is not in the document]
YangleiZouIntel: Done.

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/activation.cpp
YangleiZouIntel: No, this is not commented in CPU plugin. However it fails when generating expected output from reference implementation. We will furthur check this issue.

plaidml/bridge/openvino/ops/reduce.cpp
```diff
@@ -19,71 +19,73 @@ void registerReduceOps() {
   registerOp("ReduceLogicalAnd", [](const Context& ctx) {
     IE_ASSERT(ctx.operands.size() == 2);
     auto I = ctx.operands.at(0);
-    std::vector<size_t> axes = get_axis_vector_from_constant_operand(1, ctx.layer);
+    std::vector<int64_t> axes = cast_constant_operand<int64_t>(1, ctx.layer);
     auto* layer = ngraph::as_type<ngraph::opset1::ReduceLogicalAnd>(ctx.layer);
+    I = edsl::cast(I, DType::FLOAT32);
```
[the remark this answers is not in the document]
LiyangLingIntel: Because ReduceLogical Ops require bool precision in which NGraph will send `i1` tensor to plaidml, but it seems plaidml tensor `sum`(`LogicalOr`) and `product`(`LogicalAnd`) with `i1` datatype will get incorrect result. Here I simply cast input to `fp32` type.
I think you are right, maybe we can cast input tensor to `UINT8` as `all` and `any` ops originally want and cast back when returning output. Like this
```cpp
auto I_i8 = edsl::cast(I, DType::UINT8);
auto O = op::all(I_i8, edsl::make_tuple(axes), layer->get_keep_dims());
return edsl::make_tuple(edsl::cast(O, I.dtype()));
```

plaidml/bridge/openvino/ops/reduce.cpp
```diff
@@ -19,71 +19,73 @@ void registerReduceOps() {
   registerOp("ReduceLogicalAnd", [](const Context& ctx) {
     IE_ASSERT(ctx.operands.size() == 2);
     auto I = ctx.operands.at(0);
-    std::vector<size_t> axes = get_axis_vector_from_constant_operand(1, ctx.layer);
```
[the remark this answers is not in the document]
YangleiZouIntel: This change is to support negative axes in the tests. Since size_t  is unsigned integer, we changed to int64_t.

plaidml/bridge/openvino/ops/reduce.cpp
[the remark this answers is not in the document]
tzerrell: Ok, makes sense.

I'll flag for the long term that there might be an additional problem in the oplib for op::all (and op::any) where the return type doesn't necessarily match the original type. OTOH, I think other frameworks might explicitly want the result of such an operation to be a different type the the inputs; I think it probably makes sense to leave oplib as-is, and cast back to the original type in the frontend (as you do here) where necessary.

So in summary, I like the approach you suggested in this comment.

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/activation.cpp
tzerrell: Ok, sounds good. Odd that the reference implementation fails here but not with the CPU plugin.

tzerrell (commented):
I think this PR is now sufficiently polished that we could land it as-is and finish the outstanding comments & the RoI ops in followup PRs.

tzerrell:
In the OV fork, can we merge the https://github.com/plaidml/openvino/tree/update_openvino branch that this PR uses into the `plaidml` branch? I've updated `plaidml-2021.1` to the current HEAD of `plaidml`, so we don't need to worry about losing our 2021.1 version.

(and then have this PR reference the resulting commit in the `plaidml` branch)

YangleiZouIntel:
> In the OV fork, can we ~merge~ move the https://github.com/plaidml/openvino/tree/update_openvino branch that this PR uses to be the `plaidml` branch? I've updated `plaidml-2021.1` to the current HEAD of `plaidml`, so we don't need to worry about losing our 2021.1 version.
> 
> (and then have this PR reference the resulting commit in the `plaidml` branch)

Thank you Tim. The OpenVINO plaidml branch is a protected branch which I do not have permission to force push changes in https://github.com/plaidml/openvino/tree/update_openvino to it. Could you or someone else please help?

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/split.cpp
YangleiZouIntel: Done.

plaidml/bridge/openvino/plaidml_util.hpp
```diff
@@ -23,6 +24,18 @@
...
 
+template <typename T>
+std::vector<T> cast_constant_operand(size_t operand_idx, ngraph::Node* layer) {
+  auto* ngraph_const = ngraph::as_type<ngraph::op::Constant>(layer->get_input_node_ptr(operand_idx));
+  if (ngraph_const) {
+    return ngraph_const->cast_vector<T>();
+  } else {
+    THROW_IE_EXCEPTION
+        << "Dynamic slicing not currently supported by PlaidML plugin; all of indices, offsets and default index"
+           "must be Constants.";
```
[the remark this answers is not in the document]
YangleiZouIntel: Done.

plaidml/bridge/openvino/ops/mat_mul.cpp
```diff
@@ -57,7 +57,32 @@ void registerMatMul() {
       B = op::unsqueeze(B, {0});
     }
 
-    return edsl::make_tuple(op::dot(A, B));
+    std::vector<edsl::TensorDim> A_dims(ndimsA);
```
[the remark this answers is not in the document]
YangleiZouIntel: The test case was added in https://github.com/plaidml/plaidml/pull/1684

plaidml/bridge/openvino/ops/reorg_yolo.cpp
```diff
@@ -21,7 +21,7 @@ void registerReorgYolo() {
     auto I = ctx.operands.at(0);
     // as openvino op request, the strides have to be divisible by 'H' and 'W'.
     auto strides = layer->get_strides()[0];
-    return edsl::make_tuple(op::reorg_yolo(I, strides, false));
+    return edsl::make_tuple(op::reorg_yolo(I, strides, true));
```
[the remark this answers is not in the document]
YangleiZouIntel: There was no reference implementation for reorg_yolo in nGraph. They added reorg_yolo reference implementation recently (included in this update). 

Unfortunately, this reference implementation does not have the `decrease` flag as plaidml op::reorg_yolo does and it does not match `op::reorg_yolo(I, strides, false)` or  `op::reorg_yolo(I, strides, true);` We though there was 3 different ways to fix this:
1. change the nGraph reference implementation to match `op::reorg_yolo(I, strides, true)`
2. change the nGraph reference implementation to match `op::reorg_yolo(I, strides, false)`
3. change plaidml `op::reorg_yolo `to match  nGraph reference implementation

I can not say which implementation is right or wrong because I think there were arguments about some details how this op should be implemented. But I believe plaidml `op::reorg_yolo` is more mature and advanced than the current nGraph reorg_yolo .

So we chose the 1st way to fix this issue. Some of the discussion between @XingHongChenIntel  and I can be found here: https://github.com/Flex-plaidml-team/openvino/pull/28/files (We can go the 2nd way as well.)

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/strided_slice.cpp
XingHongChenIntel: no, this don't cause problems for the CPU plugin, we are still further check this issue, and will fix this in a following PR.

YangleiZouIntel:
> I think this PR is now sufficiently polished that we could land it as-is and finish the outstanding comments & the RoI ops in followup PRs. [Edit: we should probably make the change from my next comment first, unless there's a reason that won't work. But I think that can just be done by moving the `plaidml` branch to the current commit of `update_openvino` so probably these could be done independently if no further changes are needed there.]

Yes, I agree we can land this PR as-is and I have marked it as ready for review. 

To sum up, the items we will have in followup PRs (Please add if I missed any items.):
1. Update RoI align
2. Update RoI pooling
3. Fix ActivationParamLayerTest
4. Two test cases in StridedSlice

tzerrell (approved):

plaidml/bridge/openvino/ops/reorg_yolo.cpp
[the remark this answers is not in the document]
tzerrell: Interesting, thanks for the detailed update and the link to the discussion that led to this choice. I think this is a good plan for now. I will look into what it would take to update the nGraph reference implementation of reorg_yolo and who such a change would impact.

tzerrell:
> To sum up, the items we will have in followup PRs (Please add if I missed anything.):
> 
> 1. Update RoI align
> 2. Update RoI pooling
> 3. Fix ActivationParamLayerTest
> 4. Fix two test cases in StridedSlice

Yes, I think this list is correct, with one addition:

On VariadicSplit testing, I noticed in going over the OV-repo changes that this update removed the test including a negative axis. The docs indicate that negative axes are legal for VariadicSplit, so one more followup will be to restore some sort of testing for VariadicSplit negative axis.

YangleiZouIntel:
> > To sum up, the items we will have in followup PRs (Please add if I missed anything.):
> > 
> > 1. Update RoI align
> > 2. Update RoI pooling
> > 3. Fix ActivationParamLayerTest
> > 4. Fix two test cases in StridedSlice
> 
> Yes, I think this list is correct, with one addition:
> 
> On VariadicSplit testing, I noticed in going over the OV-repo changes that this update removed the test including a negative axis. The docs indicate that negative axes are legal for VariadicSplit, so one more followup will be to restore some sort of testing for VariadicSplit negative axis.

Do you mean the [negative axis test in CPU plugin](https://github.com/plaidml/openvino/commit/bfa3251a7f0a3a10c4a3b086b4895849d5466f32#diff-26306da6180148c3eff8e6ddffdba08b91b00347b4f7b5338d50b2cbbd1e16d3L25) ? Fixed in https://github.com/plaidml/openvino/pull/130/files. Please check, thanks.

plaidml/bridge/openvino/plugin/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 set(TARGET_NAME "plaidmlPlugin")
 
+ie_cpack_set_library_dir()
```
flaub, line 3: What was this added for?

YangleiZouIntel: After one commit in OV, plugin has to explicitly set LIBRARY PATH, otherwise when cmake will give this error:
```
CMake Error at /home/yangleizou/plaidml/openvino/cmake/developer_package/plugins/plugins.cmake:107 
(install):
  install TARGETS given no LIBRARY DESTINATION for module target
  "plaidmlPlugin".
Call Stack (most recent call first):
  plaidml/bridge/openvino/plugin/CMakeLists.txt:3 (ie_add_plugin)
```
the function definition is here:
https://github.com/openvinotoolkit/openvino/blob/master/cmake/developer_package/packaging.cmake#L14

PR #1680, opened 2021-01-26
Remove some comments of BatchNormInference.

XinWangIntel:


XinWangIntel:
The order of parameters in BatchNormInference is changed in opset5 (but opset4 still keeps the old order). Remove the comments that are out of date for opset5 now.

YangleiZouIntel (approved):

PR #1682, opened 2021-01-26
PrioBox only supports fixed_ratio with one element now.

XinWangIntel:
Ngraph's implementation of number_of_priors use a weird way to process density (https://github.com/openvinotoolkit/openvino/blob/master/ngraph/core/src/op/prior_box.cpp#L116), this is different from the reference it provided (https://github.com/openvinotoolkit/openvino/blob/master/ngraph/core/reference/include/ngraph/runtime/reference/prior_box.hpp#L159) . This conflict makes the node output shape is different from the real output so it only work well when the fixed_ratio size is one.  Change the test case of PriorBox to follow this implicit rule and wait for their fix now.


tzerrell (commented):
Can we also throw an exception if we are given multiple elements for fixed_ratio? E.g., "PriorBox does not currently support multiple elements in fixed_ratio attribute"

XinWangIntel:
Add verifier to check fixed_ratio size when we are using fixed_size path.

tzerrell (approved):

PR #1684, opened 2021-01-27
fix matmul and add new test cases

haoyouab:
Fix the problem memtioned by Tim at https://github.com/plaidml/plaidml/pull/1679#discussion_r563909981.

I simply use `op::repeat` to fix the problem because when we need broadcast semantics over the leading "batch" dimensions, they are always equal to 1. I think in this case `op::repeat` and `op::broadcast` perfom the same functionality and `op::repeat` is less complex. What do you think @tzerrell ?

### DIFF
--- a/plaidml/bridge/openvino/ops/mat_mul.cpp
+++ b/plaidml/bridge/openvino/ops/mat_mul.cpp
@@ -53,9 +53,14 @@ void registerMatMul() {
     while (A.rank() < B.rank()) {
       A = op::unsqueeze(A, {0});
     }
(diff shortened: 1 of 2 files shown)


tzerrell (approved):
Broadcast is intended for when we have one tensor and a target broadcast shape given as integers; I agree that repeat is better here where we have two tensors.

That said, I think this could be optimized further by folding the repeat ops into setting the sizes and indices of the contraction. But, that is somewhat complicated and this has correct behavior. So we can do that optimization work in a followup PR.

PR #1686, opened 2021-02-01
Fix openvino plaidml util get_axis method

LiyangLingIntel:
The original implementation of `get_axis_set_from_constant_operand` and `get_axis_set_from_constant_operand` use [`op::Constant::get_axis_*_val()`](https://github.com/plaidml/openvino/blob/plaidml/ngraph/core/src/op/constant.cpp#L520) which directly change negative axis value to 0, and fail some tests. 
```cpp
AxisSet op::Constant::get_axis_set_val() const
{
    NGRAPH_CHECK(m_element_type.is_integral_number());
    std::vector<int64_t> out_axis_set = cast_vector<int64_t>();
    AxisSet output_axis_set;
    for (auto& axis : out_axis_set)
    {
        output_axis_set.insert(axis > 0 ? axis : 0);
    }
    return output_axis_set;
}
```
So, using [`ngraph::normalize_axes`](https://github.com/plaidml/openvino/blob/plaidml/ngraph/core/src/validation_util.cpp#L910) to make all axis to non-negative before casting to `AxisSet` and `AxisVector` can solve this issue.
```cpp
int64_t ngraph::normalize_axis(const std::string& node_description, std::int64_t axis, ...)
{
    // ...
    if (axis < 0)
    {
        axis = axis + tensor_rank;
    }

    return int64_t(axis);
}
```

~This adjustment works on my local machine(Ubuntu 18.04), but fails on CI([subcommand failed](https://buildkite.com/plaidml/plaidml-plaidml/builds/5253#19064a27-2a98-45d2-9402-e05cf794b9b9/1227-1372)). 
Need more time to fix.~

### DIFF
--- a/plaidml/bridge/openvino/ops/reduce.cpp
+++ b/plaidml/bridge/openvino/ops/reduce.cpp
@@ -19,7 +19,7 @@ void registerReduceOps() {
   registerOp("ReduceLogicalAnd", [](const Context& ctx) {
     IE_ASSERT(ctx.operands.size() == 2);
     auto I = ctx.operands.at(0);
-    std::vector<int64_t> axes = cast_constant_operand<int64_t>(1, ctx.layer);
+    std::vector<size_t> axes = get_axis_vector_from_constant_operand(1, ctx.layer);
     auto* layer = ngraph::as_type<ngraph::opset1::ReduceLogicalAnd>(ctx.layer);
     auto I_i8 = edsl::cast(I, DType::UINT8);
     auto O = op::all(I_i8, edsl::make_tuple(axes), layer->get_keep_dims());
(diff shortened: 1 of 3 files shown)


tzerrell (commented):
This looks like a good change to me. I presume you removed the reviewers because you need to fix some details before this is ready to land, is that correct?

LiyangLingIntel:
> This looks like a good change to me. I presume you removed the reviewers because you need to fix some details before this is ready to land, is that correct?

Yes, this change passed all tests on my machine, but failed CI build procedure for some reasons. 
It is OK now after changing the place of the header file `ngraph/validation_util.hpp` from `plaidml_util.hpp` to `plaidml_util.cpp`.

This PR is ready for review.

tzerrell (approved):

PR #1687, opened 2021-02-02
Update OpenVINO, including upstream fix on Roi Pooling and Roi Align

YangleiZouIntel:
This PR is based on  https://github.com/plaidml/openvino/tree/update-openvino (@tzerrell could you please help push this branch to plaidml branch?), which is rebased onto  https://github.com/openvinotoolkit/openvino/commit/3a86b3a17e3311a69687b8147b6f2317d60b8b82

This update includes upstream fix/update on bucketize, scatter_update, Roi Pooling and Roi Align.

### DIFF
--- a/plaidml/bridge/openvino/CMakeLists.txt
+++ b/plaidml/bridge/openvino/CMakeLists.txt
@@ -28,7 +28,7 @@ if(NOT InferenceEngineDeveloperPackage_FOUND)
   FetchContent_Declare(
     openvino
     GIT_REPOSITORY https://github.com/plaidml/openvino.git
-    GIT_TAG        78c7e637b2c00c8e071227e41b73681cd271a353
+    GIT_TAG        2a82d330a05eb06ccde9e7e38a1535c268ad58db
   )
   if(LOCAL_OPENVINO_DIR)
     set(openvino_SOURCE_DIR ${LOCAL_OPENVINO_DIR})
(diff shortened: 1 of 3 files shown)


tzerrell (commented):
This change looks good. Regarding the OV branch update, I think it might be better to merge rather than rebase upstream master updates going forward (since we no longer have to worry about adding version-specific commits like when did when transitioning from 2021.1 to master). The benefit of this is it will no longer be necessary to force-push the `plaidml` branch, and so it won't need admin approval from me to make the merge (just a normal review).

I'll open a PR with that work in the OV repo. If you like this approach and that PR looks good to you, you can review & land it and use its commit hash for this PR.

YangleiZouIntel:
> This change looks good. Regarding the OV branch update, I think it might be better to merge rather than rebase upstream master updates going forward (since we no longer have to worry about adding version-specific commits like when did when transitioning from 2021.1 to master). The benefit of this is it will no longer be necessary to force-push the `plaidml` branch, and so it won't need admin approval from me to make the merge (just a normal review).
> 
> I'll open a PR with that work in the OV repo. If you like this approach and that PR looks good to you, you can review & land it and use its commit hash for this PR.

Thanks, I have reviewed and landed the OV PR and updated the commit hash in this PR.

tzerrell (approved):

PR #1691, opened 2021-02-03
Initial CPackaging

jd-bartlett96:
My initial goal is to get a basic package that contains the `plaidml_edsl_tests_cc_test` binary and the required files to run it.

To create the package (from `build-opt` folder, in `.cenv`, with the cmake generated with `ninja`):

`ninja && sudo cpack -C "Release" -G "DEB"`

This generates the package `PlaidML-1.0.0-Linux-DEVKIT.deb` in `build-opt`. When installed (`sudo dpkg -i PlaidML-1.0.0-Linux-DEVKIT.deb`), it creates the package filesystem in `/usr/local/lib/plaidml`, with the cc_test binary at `/usr/local/lib/plaidml`

Issues:
- Currently, I have to comment out all `benchmark_app` related packing commands for this package to build; for some reason this makes the packaging take hours then fail with a message about exceeded file size
- This process produces a package that works when installed on the same machine the package was generated on, but when I try and install the resulting package on a new machine, when the binary is run it fails with
```
*** stack smashing detected ***: <unknown> terminated
Aborted (core dumped)
```


CMakeLists.txt
```diff
@@ -131,4 +141,65 @@ configure_lit_site_cfg(
...
+      /lib/x86_64-linux-gnu/libpthread.so.0     
+      /lib/x86_64-linux-gnu/libpthread-2.27.so  
+      /lib/x86_64-linux-gnu/librt.so.1  
+      /lib/x86_64-linux-gnu/librt-2.27.so
+      /lib/x86_64-linux-gnu/libdl.so.2
+      /lib/x86_64-linux-gnu/libdl-2.27.so
+      /lib/x86_64-linux-gnu/libz.so.1
+      /lib/x86_64-linux-gnu/libz.so.1.2.11
+      /lib/x86_64-linux-gnu/libtinfo.so.5
+      /lib/x86_64-linux-gnu/libtinfo.so.5.9
```
flaub, line 188: We don't want to package up the binaries that are part of the machine doing the building.

jd-bartlett96:
I've incorporated automatic detection of non-system libraries, and inclusion of system libraries as a dependency list that is checked at install time (and tells the user which missing libraries to install). 

The package can be generated from a fresh repo with:
```
source activate .cenv/
mkdir build-opt
cd build-opt
cmake .. -G Ninja -D CMAKE_BUILD_TYPE=Release
ninja && cmake ..
ninja && cmake ..
sudo cpack -C "Release" -G "DEB"
```

I specify this to highlight that after the cmake project is first built, 2 iterations of `ninja && cmake ..` are required to before cpack can be run. In general this is because ninja being run alters inputs cmake builds with, and vice versa, these 2 iterations are required for the results to stabilize such that cpack can find all the right files where it expects them. I am happy to provide more detail on this if necessary; I tried extensively to remove this requirement without success.

(In general though, if you've already built the cmake project frequently on your machine, the `sudo cpack -C "Release" -G "DEB"` command will run successfully on its own)

pmlc/target/x86/CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ pml_cc_library(
 
 add_custom_command(
   OUTPUT heatmap.cc
-  COMMAND python
+  COMMAND ${PYTHON_EXECUTABLE}
```
flaub, line 56: Why do we need this and where does `PYTHON_EXECUTABLE` come from?

jd-bartlett96: This was causing a build error for me because the command needed the python executable from `.cenv/` to run correctly due to the pystache requirement; on my build `python` was finding my machine's python instead which did not have pystache installed. 

I believe it gets set with `find_package(Python)` in `cmake/pml_format.cmake`

jd-bartlett96:
A functional header-only library can now be generated with `cpack`.

Some notes:
- In order to get this devkit cpackaging to work alongside the BENCHMARK_APP cpackaging, the devkit packaging must be done through the Openvino `ie_cpack` command along with the BENCHMARK_APP. Trying to separate these processes (eg, not sending the devkit component to `ie_cpack`) universally fails, either by the devkit component not being packaged or by the `cpack` command hanging for hours before failing.
  The only real negative to this is that there are some variables being changed within `ie_cpack`, namely the package names, which cannot be altered (as far as I can tell, I haven't been able to find the actual place `ie_cpack` is defined in whatever 3rd party package it lives in). Because of this, the resulting package name for the devkit is `inference-engine--Linux-devkit.tar.gz`; if a different name is desired it must be renamed manually.

- I've included the necessary files in `cmake\packaging_test_project` for a minimal example project that imports the devkit package and runs an external copy of the edsl tests

- With the exception of Boost, I bundled the binaries & headers for 3rd party dependencies into the package. For Boost, the cmake file for the package installation is included in the devkit package and automatically run when the devkit package is imported. I can provide more detail as to why, but in general I have not (yet) figured out how to include Boost in the same way as the other 3rd party packages, or to include references to the other packages the same way I have done for Boost. If this mixed approach is fine, great. If not, let me know which of the 2 approaches used is preferred so I can know which way to focus my attention on figuring out how to get it to work.

cmake/packaging_test_project/edsl_tests.cc
```diff
@@ -0,0 +1,2114 @@
+// Copyright 2020 Intel Corporation
```
flaub, line 1: This file seems to be a duplicate of the existing `edsl_tests.cc`, are there any differences here? I think having a duplicate file makes it much harder to maintain over time.

cmake/packaging_test_project/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+add_executable(edsl_test edsl_tests.cc)
+target_link_libraries(edsl_test PlaidML::plaidml_edsl_edsl)
+target_link_libraries(edsl_test PlaidML::plaidml_exec_exec)
+target_link_libraries(edsl_test PlaidML::plaidml_op_op)
+target_link_libraries(edsl_test PlaidML::plaidml_testenv)
+target_link_libraries(edsl_test PlaidML::pmlc_util_util)
+target_link_libraries(edsl_test PlaidML::gmock)
+target_link_libraries(edsl_test PlaidML::gtest)
+target_link_libraries(edsl_test PlaidML::half)
+target_link_libraries(edsl_test PlaidML::LLVMSupport)
```
flaub, line 21: Users should link only to a single `PlaidML` dependency, and this dependency should be the `libplaidml.so` + header files.

cmake/packaging_test_project/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+project("Devkit-Test")
+
+cmake_minimum_required(VERSION 3.18)
+
+set(CMAKE_C_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
+
+# MUST SET THIS W/ CORRECT EXTRACTION PATH
+set(PlaidML_DIR /path/to/PlaidML-Package/share/PlaidML/cmake)
```
flaub, line 9: Use `option` instead of `set` here

cmake/packaging_test_project/CMakeLists.txt
flaub: For gmock/gtest, we should include our own FetchContent directly, we should not include gmock/gtest as part of our package. For the other dependencies, they are all internal to `libplaidml.so`, with the exception of `plaidml_testenv`. This is the one part that is currently internal but we could refactor it to make it be external.

jd-bartlett96: Yes sorry the packaging_test_project is outdated now; it is still set up to work with the old package with all the static libraries/3rd party stuff included. Currently working on updating that to reflect the new package.

cmake/packaging_test_project/edsl_tests.cc
jd-bartlett96: The only difference is the main() function at the end so it can be compiled into an executable directly (to keep the CMakeLists of the test project minimal). If there is a good way to reference the actual `edsl_tests.cc` file from the test project and compile it into an executable (namely, without the use of the `pml_cc_test` function).

cmake/packaging_test_project/CMakeLists.txt
flaub: Currently, testenv uses LLVM, which makes it hard to export to devkit users. We could refactor testenv to remove this dependency. The half dependency is likely not easy to remove, so we could also include FetchContent in the devkit cmake so that users can fetch it directly (again we should not distribute our own version of half in our package).

The only contents for our package should be:
* CMakeLists.txt
* plaidml/*.h  (for core, edsl, exec, and op) (and only the public headers)
* plaidml/edsl/edsl_test.cc
* plaidml/edsl/edsl_test.py
* libplaidml.so
* libomp.so
* libmlir_runner_utils.so.12git

In the README we should say, the following system packages are required (for ubuntu, package names may vary for other distros):
* libtinfo5
* zlib1g

The CMakeLists.txt needs to do FetchContent for:
* gflags
* googletest (include gmock and gtest)
* half

cmake/pml_cc_library.cmake
```diff
@@ -116,10 +116,12 @@ function(pml_cc_library)
     target_include_directories(${_NAME} SYSTEM
       PUBLIC
         "$<BUILD_INTERFACE:${PML_COMMON_INCLUDE_DIRS}>"
+        "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>"
```
flaub, line 119: I don't think we want to generically do this, lots of headers aren't considered 'public' or 'external' to the project.

plaidml/core/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ pml_cc_library(
     pmlc::ast
     pmlc::compiler
     pmlc::util
+    pmlc::target::all
```
flaub, line 21: We can't do this. We need to allow for external projects to decide which targets they want to include.

cmake/packaging_test_project/CMakeLists.txt
flaub: Oops and the package needs to include the `testenv` source and header.

cmake/pml_packaging.cmake
```diff
@@ -0,0 +1,49 @@
...
+set(CPACK_GENERATOR "TGZ")
+set(CPACK_ARCHIVE_COMPONENT_INSTALL 1)
+set(CPACK_PACKAGE_NAME "PlaidML")
+set(CPACK_PACKAGE_VENDOR "Intel Corp")
+set(CPACK_PACKAGE_CONTACT "Intel")
+set(CPACK_PACKAGE_VERSION ${PLAIDML_VERSION})
+set(CPACK_COMPONENTS_IGNORE_GROUPS 1)
+set(CPACK_COMPONENTS_ALL devkit)
+
+set_target_properties(mlir_runner_utils PROPERTIES INTERFACE_LINK_LIBRARIES "")
```
flaub, line 13: I guess I'm worried that changing this property is a global thing and that it might break other situations, but it's unclear to me.

cmake/pml_packaging.cmake
```diff
@@ -0,0 +1,49 @@
...
+
+install(TARGETS plaidml_plaidml
+                omp
+                mlir_runner_utils
+        EXPORT ${PROJECT_NAME}_Targets
+        DESTINATION lib
+        COMPONENT devkit)
```
flaub, line 20: We might try installing files rather than targets. I haven't had much luck with getting targets to work.

CMakeLists.txt
```diff
@@ -131,4 +132,8 @@ configure_lit_site_cfg(
   ${CMAKE_CURRENT_SOURCE_DIR}/lit.cfg.py
 )
 
-ie_cpack(BENCHMARK_APP)
+if(BUILD_PLAIDML_DEVKIT)
+  include(pml_packaging)
+else()
+  ie_cpack(BENCHMARK_APP)
+endif()
```
flaub, line 139: This is a clever workaround, I like it. The downside though is that we might need to have multiple builds in CI. I can take a stab at some point to integrate to OpenVINO differently, which might remove the need for this.

cmake/pml_cc_library.cmake
```diff
@@ -201,4 +204,24 @@ function(pml_cc_library)
     # For example, foo/bar/ library 'bar' would end up as 'foo::bar'.
     add_library(${_PACKAGE_NS} ALIAS ${_NAME})
   endif()
+
+  get_property(INSTALL_TARGETS GLOBAL PROPERTY install_targets_property)
+  if(TARGET ${_NAME} AND NOT ${_NAME} MATCHES ".*openvino.*" AND NOT ${_NAME} IN_LIST INSTALL_TARGETS)
```
flaub, line 209: Hmm, I'm not feeling good about having library-specific goo in this generic `pml_cc_library` rule.

What if we added a whole new flag to `pml_cc_library` which means 'this is a public/external library'. Only in this case would we install headers for that library.

Alternatively, we could just list the headers directly from the packaging so that things are a little less magical and it's explicitly listed someplace exactly what's going into the public package.

cmake/pml_packaging.cmake
```diff
@@ -22,28 +22,42 @@ install(TARGETS plaidml_plaidml
...
         DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/cmake
         COMPONENT devkit)
 
-# TODO: Install any desired standalone executables or source code for reference
+install(FILES "${PROJECT_SOURCE_DIR}/cmake/third_party/CPM.cmake"
+              "${PROJECT_SOURCE_DIR}/cmake/third_party/easylogging.cmake"
+              "${PROJECT_SOURCE_DIR}/cmake/third_party/gflags.cmake"
+              "${PROJECT_SOURCE_DIR}/cmake/third_party/googletest.cmake"
+              "${PROJECT_SOURCE_DIR}/cmake/third_party/half.cmake"
+              "${PROJECT_SOURCE_DIR}/cmake/third_party/llvm-project.cmake"
```
flaub, line 49: We don't want devkit users to depend on LLVM, we should drop this.

cmake/pml_packaging.cmake
```diff
@@ -22,28 +22,42 @@ install(TARGETS plaidml_plaidml
...
+        INSTALL_DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/cmake)
+
 install(FILES "${PROJECT_BINARY_DIR}/${PROJECT_NAME}Config.cmake"
               "${PROJECT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake"
         DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/cmake
         COMPONENT devkit)
 
-# TODO: Install any desired standalone executables or source code for reference
+install(FILES "${PROJECT_SOURCE_DIR}/cmake/third_party/CPM.cmake"
+              "${PROJECT_SOURCE_DIR}/cmake/third_party/easylogging.cmake"
```
flaub: I don't think we need these deps. Actually, I think we need to tweak `gtest_main.cc` so that it doesn't use `pmlc/util`, which is what eventually pulls in the logging.

cmake/packaging/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+set(PLAIDML_INSTALL_DIR /path/to/PlaidML)
+
+# Include PlaidML package
+include("${PLAIDML_INSTALL_DIR}/share/cmake/PlaidMLConfig.cmake")
+
+# Include 3rd party packages necessary for edsl_tests
+include(FetchContent)
+include("${PLAIDML_INSTALL_DIR}/share/cmake/third_party/gflags.cmake")
+include("${PLAIDML_INSTALL_DIR}/share/cmake/third_party/googletest.cmake")
+include("${PLAIDML_INSTALL_DIR}/share/cmake/third_party/half.cmake")
```
flaub, line 17: I kinda think we should just copy/paste these or perhaps do a compile-time include here so that devkit users don't have to deal with these extra `.cmake` files

cmake/packaging/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# TODO: Remove LLVM dep for edsl_test
+include("${PLAIDML_INSTALL_DIR}/share/cmake/third_party/llvm-project.cmake")
+
+add_executable(edsl_test ${PLAIDML_INSTALL_DIR}/example_project/gtest_main.cc)
```
flaub: I think we should have a package dir structure like:
```
cmake
  -- cmake files go here (or we just compile-time include them potentially)
include
  -- headers go here
lib
  -- libs go here
src
  -- gtest_main.cc
  -- edsl_test.cc
  -- edsl_test.py
CMakeList.txt
```

cmake/packaging/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+target_link_libraries(edsl_test PlaidML::plaidml_plaidml
```
flaub, line 22: Can we make the target name simply be `PlaidML`?

cmake/pml_packaging.cmake
jd-bartlett96: Yep planning on it; just trying to get the external edsl_tests working correctly as they currently stand, then will look at removing the LLVM dependency.

cmake/packaging/CMakeLists.txt
jd-bartlett96: It can at very least be plaidml_plaidml; I think there is an option to rename it altogether. I'll look into that.

flaub:
@jd-bartlett96 I pushed some adjustments, WDYT?

flaub (approved):

PR #1694, opened 2021-02-04
Layouts for OV

abialas1:
Initial support for adding layouts based on the OV plugin.
The first pass marks all the layer::BoxOp ops with appropriate enum based layout via tags. It also adds contraction ops placeholders for physical data reordering, so that hoisting pass on tile dialect can move weights reorderings to init function.
Later on in tile to pxa, these tags are being switched to layout maps injected in the memrefs.

What is not done yet:
1. Following the tile_to_pxa we should call normalize memref pass, to clean the layout maps. Right now it does not work with pxa ops
2. Only some transformations are done here, need to expand with more ops and more layouts

### DIFF
--- a/pmlc/dialect/tile/ir/util.cc
+++ b/pmlc/dialect/tile/ir/util.cc
@@ -19,6 +19,7 @@
 
 #include "pmlc/dialect/tile/ir/ops.h"
 #include "pmlc/dialect/tile/ir/types.h"
+#include "pmlc/util/layout.h"
 #include "pmlc/util/logging.h"
 
 using namespace mlir; // NOLINT
@@ -271,6 +272,37 @@ Value createIdentity(OpBuilder &builder, Location loc, Type elementType,
   }
   return builder.create<tile::ConstantOp>(loc, elementType,
                                           getUnsignedIntegerIdentity(agg));
+}
+
+MemRefType updateMemRefWithLayoutMap(MLIRContext *context,
+                                     RankedTensorType memrefType,
+                                     Type elementType, TensorLayout layout) {
+  auto rankedTensorType = pmlc::dialect::tile::getRankedTensorType(memrefType);
+  auto originalShape = rankedTensorType.getShape();
+  auto shape = llvm::to_vector<8>(originalShape);
+
+  auto outRank = rankedTensorType.getRank();
+  auto spatialNum = outRank == 4 ? 2 : 0;
+
+  SmallVector<AffineExpr, 4> dimExprs;
+  dimExprs.reserve(outRank);
+  if (layout == TensorLayout::NCX) {
+    dimExprs.push_back(mlir::getAffineDimExpr(0, context));
+    for (auto spat = 0; spat < spatialNum; spat++) {
+      dimExprs.push_back(
+          mlir::getAffineDimExpr(outRank - spatialNum + spat, context));
+    }
+    dimExprs.push_back(mlir::getAffineDimExpr(1, context));
+  } else if (layout == TensorLayout::KCX) {
+    for (auto spat = 0; spat < spatialNum; spat++) {
+      dimExprs.push_back(
+          mlir::getAffineDimExpr(outRank - spatialNum + spat, context));
+    }
+    dimExprs.push_back(mlir::getAffineDimExpr(1, context));
+    dimExprs.push_back(mlir::getAffineDimExpr(0, context));
+  }
+  auto idMap = AffineMap::get(outRank, 0, dimExprs, context);
+  return MemRefType::get(shape, elementType, {idMap});
 }
 
 } // namespace pmlc::dialect::tile
--- a/pmlc/dialect/tile/ir/util.cc
+++ b/pmlc/dialect/tile/ir/util.cc
@@ -19,7 +19,6 @@
 
 #include "pmlc/dialect/tile/ir/ops.h"
 #include "pmlc/dialect/tile/ir/types.h"
-#include "pmlc/util/layout.h"
 #include "pmlc/util/logging.h"
 
 using namespace mlir; // NOLINT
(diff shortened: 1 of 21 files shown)


pmlc/util/layout.h
```diff
@@ -0,0 +1,38 @@
...
+namespace pmlc {
+
+enum class MLFramework {
+  Default, // No changes to layouts formats will be done
+  NGraph
+};
+
+MLFramework getMLFramework(llvm::StringRef opName);
+
+enum class TensorLayout { NXC, NCX, KCX };
```
jbruestle, line 19: We should use enum attributes: https://mlir.llvm.org/docs/OpDefinitions/#enum-attributes, and probably string enum attributes so that the IR is more readable (definitely worth the minor cost).

pmlc/dialect/tile/transforms/reorder_inputs.cc
```diff
@@ -0,0 +1,128 @@
...
+
+  for (unsigned i = 0; i < outRank; i++) {
+    dimExprs.push_back(mlir::getAffineDimExpr(i, op->getContext()));
+  }
+  auto idMap = AffineMap::get(outRank, 0, dimExprs, op->getContext());
+
+  std::stringstream ss;
+  ss << "reordered_" << arg.getArgNumber();
+
+  auto newOp = builder.create<ContractionOp>(
```
jbruestle, line 62: I don't quite understand this code.  Is it adding identity contractions between each arg of the box op and all of it's uses?  I imagine these would turn into transposes basically during lowering to pxa, which is good for the weights, but for internal activations I'd imagine we would want to add layout data directly so we don't induce transposes inline, or do those somehow optimize away.  Maybe some more lit tests would make a clear, I'm having a hard time reasoning about the whole pipeline herre.

pmlc/dialect/tile/ir/util.cc
```diff
@@ -273,4 +274,35 @@ Value createIdentity(OpBuilder &builder, Location loc, Type elementType,
                                           getUnsignedIntegerIdentity(agg));
 }
 
+MemRefType updateMemRefWithLayoutMap(MLIRContext *context,
+                                     RankedTensorType memrefType,
+                                     Type elementType, TensorLayout layout) {
+  auto rankedTensorType = pmlc::dialect::tile::getRankedTensorType(memrefType);
+  auto originalShape = rankedTensorType.getShape();
+  auto shape = llvm::to_vector<8>(originalShape);
```
jbruestle, line 282: Why do we need this?  Can't we just use rankedTensorType.getShape() or even memrefType.getShape() in the single use of 'shape' later?  Doesn't seem like rankedTensorType is really needed either, since memrefType.getRank() would also be fine.  Also is elementType just memrefType.getElementType(), in which case do we need to pass it in?  Finally, if all we are doing is changing the map on a memref type consider:

MemRefType newMemRef = MemRefType::Builder(oldMemRef).setAffineMaps({map});

Basicially MemRefType::Builder is a methods for 'modifying' the memref type.

pmlc/dialect/tile/transforms/reorder_inputs.cc
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+  return argPos;
+}
+
+void performDataReordering(OpBuilder &builder, Operation *op, BlockArgument arg,
+                           MLFramework framework, bool isConst) {
+  // Check if op is a BoxOp, if so scan which argument is used,
+  // and get its tensor type.
+  // Later on create the new contraciton op with the same sizes and additional
```
flaub, line 34: typo: contraction

pmlc/dialect/tile/transforms/reorder_inputs.cc
```diff
@@ -0,0 +1,128 @@
...
+  // layout tag. These ops would be moved to init as weights reorders in the
+  // later pass.
+  auto layerOp = dyn_cast<layer::BoxOp>(op);
+  if (!layerOp)
+    return;
+
+  auto argPos = getArgPos(*op, arg);
```
flaub, line 41: Use `arg.getArgNumber()`

pmlc/util/layout.cc
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+
+TensorLayout getLayoutType(MLFramework framework, StringRef opName,
+                           bool isConst) {
+  // TODO: Create a common list of primitives that would be used here and in the
+  // plugins so we know they are consistent. For now use namings from OV plugin
+  // Const are used to identify the primitives weights or parameters, rest is
+  // considered as main flow data type
+  if (framework == MLFramework::NGraph) {
```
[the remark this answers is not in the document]
abialas1: Good idea. Generally this is done in a very basic way now, there are few things to consider like the common base of the primitives (right now the layer names are defined in the plaidml bridge, some start with capital letters, some do not, and do not have consistent namings). It would be good to auto generate these somehow and keep consistency with the plugins. I can come up with some proposition for this, but if you have something specific in mind then let me know

pmlc/util/layout.h
abialas1: Will move theses attributes to enum.td. The IR will be stringified.

pmlc/dialect/tile/transforms/reorder_inputs.cc
abialas1: Yes, the idea was to create the contractions that would be moved to the init function on hoisting pass in tile dialect, and set their layout types. As you said those would be only transposes. This way we have these separated from the main function.  If it was not there then we will not read weights efficiently in the main function kernels, but have them transposed in main flow (not efficient). The reorderLayouts that is called after subgroup pass would modify the layout maps of those contractions instead of adding new ones. Yes the reordering of the activation is redundant, however I kept it to know that the pipeline would work. I was especially worried if any of the pass does not use strides and will have the analysis broken. With this explicit reorder on activation we kind of use old path with nxc for the analysis.

pmlc/dialect/tile/ir/util.cc
abialas1: Will remove redundancy. Actually we only need the affine maps here and set them in the tile_to_pxa. Will change that

pmlc/dialect/tile/transforms/reorder_inputs.cc
abialas1: actually I need a position of the argument in the certain op, not in the function. I did not see a function for that

PR #1695, opened 2021-02-05
add gather nd support

haoyouab:
Note:

- Gather op now has two modes: `normal` and `nd` which is similar to Scatter op.
- As [tf.gather_nd](https://www.tensorflow.org/api_docs/python/tf/gather_nd) doesn't reshape the leading `batch_dims` into one dimension and reshaping may incur more complexity, in `tile_to_pxa` conversion I simply follow what tensorflow does. Openvino specs said:

> In case non default value for batch_dims the output shape is calculated as (multiplication of indices.shape[:b]) + indices.shape[b:-1] + data.shape[(indices.shape[-1] + b):]

and the "multiplication" (i.e. reshape into one dimension) of the leading batch_dims is done at tile edsl level in `plaidml/bridge/openvino/ops/gather_nd.cpp`


tzerrell (approved):
Great, looks good!

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -1141,55 +1142,87 @@ struct GatherOpConversion : public OpConversionPattern<tile::GatherOp> {
...
+      }
+      srcOps.insert(srcOps.end(), idxs.begin(), idxs.end());
+      for (size_t i = idxDims - 1; i < dstDims; ++i) {
+        srcOps.push_back(loop.getIVs()[i]);
       }
-      srcOps.at(axis) = idx;
       interpVal = rewriter.create<mlir::LoadOp>(loc, tensor, srcOps);
+    } break;
+    default:
+      llvm_unreachable("unrecognized scatter mode");
```
flaub, line 1225: Shouldn't this say "gather mode"?

pmlc/ast/ast_ops.cc
```diff
@@ -43,6 +43,14 @@ static bool isDataTypeFloat(DataType type) {
          type == DataType::f64;
 }
 
+static bool isDataTypeInteger(DataType type) {
+  return type == DataType::i1 || type == DataType::si8 ||
+         type == DataType::ui8 || type == DataType::si16 ||
+	 type == DataType::ui16 || type == DataType::si32 ||
+	 type == DataType::ui32 || type == DataType::si64 ||
+	 type == DataType::ui64;
+}
```
flaub, line 52: Did you use `clang-format`?

tzerrell (commented):
Oops, forgot I wanted a comment typo corrected. And I see Frank has a couple of good comments as well. Let's fix in a followup PR.

PR #1702, opened 2021-02-13
Enable FP16 tests where already passing

tzerrell:
We have a number of ops that work with fp16 data but which we have not been testing with fp16 data. This adds fp16 tests for these ops to better cover & understand which ops are working in fp16. This also makes some non-fp16 test improvements, notably adding test coverage for logical & comparison ops, which were previously altogether untested.

This supersedes & replaces the work in https://github.com/plaidml/openvino/pull/98, which also added tests like these but was for code that is now outdated.

### DIFF
--- a/plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/activation.cpp
+++ b/plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/activation.cpp
@@ -10,17 +10,10 @@
 using namespace LayerTestsDefinitions;
 using namespace ngraph::helpers;
 namespace {
-// Common params
-const std::vector<InferenceEngine::Precision> inputPrecisions = {
-    InferenceEngine::Precision::FP32
-    // TODO: Fix Issue-27390
-    // InferenceEngine::Precision::I16,
-    // InferenceEngine::Precision::U8
-};
 
 const std::vector<InferenceEngine::Precision> netPrecisions = {
     InferenceEngine::Precision::FP32,
-    // InferenceEngine::Precision::FP16,
+    InferenceEngine::Precision::FP16,
 };
 
 const std::map<ActivationTypes, std::vector<std::vector<float>>> activationTypes = {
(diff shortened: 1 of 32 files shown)


plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/activation.cpp
```diff
@@ -56,6 +38,24 @@ const std::map<ActivationTypes, std::vector<std::vector<float>>> activationTypes
...
+    {Acosh, {}},
+    {Asin, {}},
+    {Asinh, {}},
+    {Atan, {}},
+    {Atanh, {}},
+    {Cosh, {}},
+    {Floor, {}},
+    {Sinh, {}},
+    {Tan, {}},
+    {Erf, {}},
```
flaub, line 58: Is there a plan to add fp16 and fp64 support for these ops?

tzerrell: What needs to be done is IMO reasonably clear: Add a pass for lowering/legalizing code of the form
```
%out = "stdx.acosh"(%in) : (f16) -> f16
```
for these eltwise ops.

This would presumably be different for different target devices. E.g., for the CPU case being tested here, I suspect the correct approach is to convert this to fp32 (since there won't be fp16 intrinsics). But other devices may want to retain this as fp16 code. Since this test is running on CPU, this list is the ops that don't lower on CPU specifically. I don't know whether this list has any relationship whatsoever with which ops work on fp16 on any other device.

That said, I don't believe anyone is actually assigned to do this work yet.

YangleiZouIntel: I think we might need to have different passes for different target devices to deal with fp16. We can start looking at the fp16 issues on CPU and then other target devices.

YangleiZouIntel (approved):

PR #1707, opened 2021-02-22
add stable option to argsort op

haoyouab:
This PR adds stable sort option and let users choose stable/unstable sort while using `edsl::argsort`. Notes are listed below.

1. No new algorithm is added but just modification of the original selection sort. Reference can be found in [wikipedia](https://en.wikipedia.org/wiki/Selection_sort). In short:
> Selection sort can be implemented as a stable sort if, rather than swapping in step 2, the minimum value is inserted into the first position and the intervening values shifted up. However, this modification either requires a data structure that supports efficient insertions or deletions, such as a linked list, or it leads to performing Θ(n2) writes.

Since we don't use linked list in this case, we should be aware that this modification can incur performance impact.

2. Stable sort is needed in Ops like `TopK` when comparing with OV reference implementation. [tf.argsort](https://www.tensorflow.org/api_docs/python/tf/argsort) has boolean `stable` parameter, too.

3. For now the argsort algorithm is working correctly but its time complexity is O(n^2). In the long term, we should implement an O(n*log(n)) argsort algorithm.

### DIFF
--- a/plaidml/edsl/tests/edsl_test.cc
+++ b/plaidml/edsl/tests/edsl_test.cc
@@ -2243,6 +2243,69 @@ TEST_F(CppEdsl, ArgSort1d) {
   // clang-format on
 }
 
+TEST_F(CppEdsl, ArgSort1dInt) {
+  auto I = Placeholder(DType::INT32, {20});
+  auto O = argsort(I, /*axis=*/0);
+  auto program = makeProgram("argsort", {I}, {O});
+  std::vector<int32_t> input = {
+      81, 95, 27, 43, 55, 56, 57, 5,  39, 7,   //
+      14, 67, 20, 66, 64, 71, 68, 54, 87, 80,  //
+  };
+  // indexed:
+  //    0: 81,  1: 95,  2: 27,  3: 43,  4: 55
+  //    5: 56,  6: 57,  7:  5,  8: 39,  9:  7
+  //   10: 14, 11: 67, 12: 20, 13: 66, 14: 64
+  //   15: 71, 16: 68, 17: 54, 18: 87, 19: 80
+  // sorted:
+  //    7:  5,  9:  7, 10: 14, 12: 20,  2: 27
+  //    8: 39,  3: 43, 17: 54,  4: 55,  5: 56
+  //    6: 57, 14: 64, 13: 66, 11: 67, 16: 68
+  //   15: 71, 19: 80,  0: 81, 18: 87,  1: 95
+  std::vector<int32_t> output = {
+      7, 9, 10, 12, 2, 8, 3, 17, 4, 5, 6, 14, 13, 11, 16, 15, 19, 0, 18, 1,  //
+  };
+  checkExact(program, {input}, {output});
+  // clang-format off
+  // CHECK-LABEL: CppEdsl.ArgSort1dInt
+  // CHECK: module @argsort
+  // CHECK: func @main(%[[ARG0:.*]]: tensor<20xsi32>) -> tensor<20xsi32>
+  // CHECK: %[[X0:.*]] = tile.argsort asc %[[ARG0]][0] : (tensor<20xsi32>) -> tensor<20xsi32>
+  // CHECK: return %[[X0]] : tensor<20xsi32>
+  // clang-format on
+}
+
+TEST_F(CppEdsl, ArgSort1dDup) {
+  auto I = Placeholder(DType::FLOAT32, {20});
+  auto O = argsort(I, /*axis=*/0);
+  auto program = makeProgram("argsort", {I}, {O});
+  // Duplicate element 81.69 at position 0 and 1.
+  std::vector<float> input = {
+      81.69, 81.69, 27.74, 43.69, 55.79, 56.79, 57.52, 7.11,  39.48, 5.9,    //
+      14.81, 66.23, 20.25, 66.05, 64.5,  71.07, 67.6,  54.42, 87.59, 80.02,  //
+  };
+  // indexed:
+  //    0: 81.69,  1: 81.69,  2: 27.74,  3: 43.69,  4: 55.79
+  //    5: 56.79,  6: 57.52,  7: 7.11,   8: 39.48,  9:  5.9
+  //   10: 14.81, 11: 66.23, 12: 20.25, 13: 66.05, 14: 64.5
+  //   15: 71.07, 16: 67.6,  17: 54.42, 18: 87.59, 19: 80.02
+  // sorted:
+  //    9:  5.9,   7:  7.11, 10: 14.81, 12: 20.25,  2: 27.74
+  //    8: 39.48,  3: 43.69, 17: 54.42,  4: 55.79,  5: 56.79
+  //    6: 57.52, 14: 64.5,  13: 66.05, 11: 66.23, 16: 67.6
+  //   15: 71.07, 19: 80.02,  0: 81.69,  1: 81.69, 18: 87.59
+  std::vector<int32_t> output = {
+      9, 7, 10, 12, 2, 8, 3, 17, 4, 5, 6, 14, 13, 11, 16, 15, 19, 0, 1, 18  //
+  };
+  checkExact(program, {input}, {output});
+  // clang-format off
+  // CHECK-LABEL: CppEdsl.ArgSort1dDup
+  // CHECK: module @argsort
+  // CHECK: func @main(%[[ARG0:.*]]: tensor<20xf32>) -> tensor<20xsi32>
+  // CHECK: %[[X0:.*]] = tile.argsort asc %[[ARG0]][0] : (tensor<20xf32>) -> tensor<20xsi32>
+  // CHECK: return %[[X0]] : tensor<20xsi32>
+  // clang-format on
+}
+
 TEST_F(CppEdsl, ArgSort2dAxis0) {
   auto I = Placeholder(DType::FLOAT32, {5, 4});
   auto O = argsort(I, /*axis=*/0);
@@ -2280,38 +2343,6 @@ TEST_F(CppEdsl, ArgSort2dAxis0) {
   // CHECK: func @main(%[[ARG0:.*]]: tensor<5x4xf32>) -> tensor<5x4xsi32>
   // CHECK:  %[[X0:.*]] = tile.argsort asc %[[ARG0]][0] : (tensor<5x4xf32>) -> tensor<5x4xsi32>
   // CHECK:  return %[[X0]] : tensor<5x4xsi32>
-  // clang-format on
-}
-
-TEST_F(CppEdsl, ArgSort1dDup) {
-  auto I = Placeholder(DType::FLOAT32, {20});
-  auto O = argsort(I, /*axis=*/0);
-  auto program = makeProgram("argsort", {I}, {O});
-  // Duplicate element 81.69 at position 0 and 1.
-  std::vector<float> input = {
-      81.69, 81.69, 27.74, 43.69, 55.79, 56.79, 57.52, 7.11,  39.48, 5.9,    //
-      14.81, 66.23, 20.25, 66.05, 64.5,  71.07, 67.6,  54.42, 87.59, 80.02,  //
-  };
-  // indexed:
-  //    0: 81.69,  1: 81.69,  2: 27.74,  3: 43.69,  4: 55.79
-  //    5: 56.79,  6: 57.52,  7: 7.11,   8: 39.48,  9:  5.9
-  //   10: 14.81, 11: 66.23, 12: 20.25, 13: 66.05, 14: 64.5
-  //   15: 71.07, 16: 67.6,  17: 54.42, 18: 87.59, 19: 80.02
-  // sorted:
-  //    9:  5.9,   7:  7.11, 10: 14.81, 12: 20.25,  2: 27.74
-  //    8: 39.48,  3: 43.69, 17: 54.42,  4: 55.79,  5: 56.79
-  //    6: 57.52, 14: 64.5,  13: 66.05, 11: 66.23, 16: 67.6
-  //   15: 71.07, 19: 80.02,  0: 81.69,  1: 81.69, 18: 87.59
-  std::vector<int32_t> output = {
-      9, 7, 10, 12, 2, 8, 3, 17, 4, 5, 6, 14, 13, 11, 16, 15, 19, 0, 1, 18  //
-  };
-  checkExact(program, {input}, {output});
-  // clang-format off
-  // CHECK-LABEL: CppEdsl.ArgSort1dDup
-  // CHECK: module @argsort
-  // CHECK: func @main(%[[ARG0:.*]]: tensor<20xf32>) -> tensor<20xsi32>
-  // CHECK: %[[X0:.*]] = tile.argsort asc %[[ARG0]][0] : (tensor<20xf32>) -> tensor<20xsi32>
-  // CHECK: return %[[X0]] : tensor<20xsi32>
   // clang-format on
 }
 
--- a/plaidml/edsl/tests/edsl_test.cc
+++ b/plaidml/edsl/tests/edsl_test.cc
@@ -2283,6 +2283,38 @@ TEST_F(CppEdsl, ArgSort2dAxis0) {
   // clang-format on
 }
 
+TEST_F(CppEdsl, ArgSort1dDup) {
+  auto I = Placeholder(DType::FLOAT32, {20});
+  auto O = argsort(I, /*axis=*/0);
+  auto program = makeProgram("argsort", {I}, {O});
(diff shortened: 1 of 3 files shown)


plaidml/edsl/__init__.py
```diff
@@ -774,19 +774,30 @@ class SortDirection(enum.IntEnum):
...
+
+    UNSTABLE = 1
+    """Sort a tensor in an unstable way."""
+
+
+def argsort(x, axis=-1, direction=SortDirection.ASC, stability=SortStability.UNSTABLE):
     """Returns the indices of a tensor ``x`` that give its sorted order along
     an ``axis``, where -1 represents the last axis, in either ascending or
-    descending ``direction``.
+    descending ``direction`` and with either stable or unstalbe ``stability``.
```
tzerrell, line 790: typo: `unstalbe` -> `unstable`

plaidml/edsl/__init__.py
```diff
@@ -774,19 +774,30 @@ class SortDirection(enum.IntEnum):
...
 
     Args:
         x (Tensor): The tensor to be sorted.
         axis (TensorDim): The axis to perform sorting on.
         direction (SortDirection): The direction of value comparison.
+        stability (SortStability): The stability of value comparison.
```
tzerrell, line 796: Maybe say "Whether the sort is stable" instead? Stability is more about whether certain reorderings are legal than about how comparisons are made.

plaidml/edsl/__init__.py
```diff
@@ -774,19 +774,30 @@ class SortDirection(enum.IntEnum):
     """Sort a tensor in descending order."""
 
 
-def argsort(x, axis=-1, direction=SortDirection.ASC):
+class SortStability(enum.IntEnum):
+    """Stability for a sort operation."""
+
+    STABLE = 0
+    """Sort a tensor in a stable way."""
```
tzerrell, line 781: For clarity, perhaps instead give a definition of a stable sort, e.g. "Preserve the initial order of equivalent values".

plaidml/edsl/__init__.py
```diff
@@ -774,19 +774,30 @@ class SortDirection(enum.IntEnum):
...
+
+    UNSTABLE = 1
+    """Sort a tensor in an unstable way."""
```
tzerrell, line 784: Similarly, could define, e.g. "Permit reordering of equivalent values".

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -2243,6 +2243,70 @@ TEST_F(CppEdsl, ArgSort1d) {
...
+  //    5: 56.79,  6: 57.52,  7: 7.11,   8: 39.48,  9:  5.9
+  //   10: 14.81, 11: 66.23, 12: 20.25, 13: 66.05, 14: 64.5
+  //   15: 71.07, 16: 67.6,  17: 54.42, 18: 87.59, 19: 80.02
+  // sorted:
+  //    9:  5.9,   7:  7.11, 10: 14.81, 12: 20.25,  2: 27.74
+  //    8: 39.48,  3: 43.69, 17: 54.42,  4: 55.79,  5: 56.79
+  //    6: 57.52, 14: 64.5,  13: 66.05, 11: 66.23, 16: 67.6
+  //   15: 71.07, 19: 80.02,  1: 81.69,  0: 81.69, 18: 87.59
+  std::vector<int32_t> output = {
+      9, 7, 10, 12, 2, 8, 3, 17, 4, 5, 6, 14, 13, 11, 16, 15, 19, 1, 0, 18  //
```
tzerrell, line 2298: Hmm, do we actually want to be testing that an unstable sort will _necessarily_ reorder equivalent values?

That said I'm not certain of the best way to test an unstable sort... perhaps use Slice to remove the last few (unstable) values?

plaidml/edsl/tests/edsl_test.py
```diff
@@ -666,6 +666,106 @@ def test_argsort_1d(self):
...
+            5,
+            6,
+            14,
+            13,
+            11,
+            16,
+            15,
+            19,
+            1,
+            0,
```
tzerrell, line 764: See my earlier comment on testing that enforces reordering in an unstable sort.

tzerrell (commented):
Looks generally good to me although I don't have a good sense of what sorting algorithms are reasonable for our use cases. @mars1026 does the stable algorithm look reasonable to you?

Beyond that, I have a comment on testing, and some minor documentation suggestions.

plaidml/edsl/tests/edsl_test.cc
haoyouab: Or I can just remove the 'unstable' tests.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -1015,42 +1015,41 @@ struct ArgSortOpConversion : public OpConversionPattern<tile::ArgSortOp> {
       minVal = rewriter.create<LoadOp>(loc, minValVar, zeroIndex).getResult();
       // Is compVal smaller than minVal? If so, update the allocs
       Value orderPred;
-      if (elementType.isSignedInteger()) {
```
mars1026, line 1018: Why are these cases being reordered?

pmlc/dialect/tile/ir/ops.td
```diff
@@ -97,7 +97,8 @@ def ArgSortOp : MaterializeOp<"argsort"> {
   let arguments = (ins
     RankedTensorOf<[AnyScalar]>:$tensor,
     IndexAttr:$axis,
-    SortDirection:$direction);
+    SortDirection:$direction,
+    SortStability:$stability);
   let results = (outs RankedTensorOf<[SI32]>:$result);
   let assemblyFormat = [{
     $direction $tensor `[` $axis `]` attr-dict `:` functional-type(operands, $result)
```
flaub, line 104: We should adjust this custom format so that it includes the stability.

flaub:
My basic feeling is that we shouldn't change the argsort interface unless absolutely necessary. In this sense, I feel like we should only support stable sorting. It seems like we can always write a better algorithm later, or even make a parallel version, but that should all be implementation details behind the argsort interface.

mars1026:
On Feb 23, 2021, at 2:03 PM, Frank Laub <notifications@github.com<mailto:notifications@github.com>> wrote:



My basic feeling is that we shouldn't change the argsort interface unless absolutely necessary. In this sense, I feel like we should only support stable sorting. It seems like we can always write a better algorithm later, or even make a parallel version, but that should all be implementation details behind the argsort interface.

The hitch is that stability has a cost; it limits our choice of algorithms and not toward the better-performing options. Requiring the client to request stability when they need it therefore preserves preserves implementation flexibility.  This is the approach TensorFlow has taken with their argsort implementation: the 'stable' parameter defaults to False, and you have to specify stable=True if you want that guarantee - even though they have yet to provide a non-stable sort implementation.

plaidml/edsl/edsl.h
```diff
@@ -805,8 +811,10 @@ enum class SortDirection {
 /// \param direction SortDirection
 /// \return Tensor
 ///
-inline Tensor argsort(const Tensor& x, int axis = -1, SortDirection direction = SortDirection::ASC) {
-  return intrinsic("argsort", x, static_cast<int64_t>(axis), static_cast<int64_t>(direction));
+inline Tensor argsort(const Tensor& x, int axis = -1, SortDirection direction = SortDirection::ASC,
+                      SortStability stability = SortStability::STABLE) {
```
tzerrell, line 815: Based on both TensorFlow's approach and my own intuition, I think we should instead default to the (probably more performant) unstable sort and require the user to explicitly request a stable sort.

tzerrell:
If we do decide to leave the interface as-is and force a single choice between stable & unstable, it's unclear to me whether stable is the correct choice. We're adding it because the OpenVINO TopK tests require a stable sort to pass; however, the _documentation_ for TopK does not mention stability, so it's unclear if this is a requirement or just a (misplaced) assumption of the tests. See OV docs: https://docs.openvinotoolkit.org/latest/openvino_docs_ops_sort_TopK_3.html

jbruestle:
Personally I'd rather only have a stable sort if the perf gap isn't significant.  Some things definitely *require* stable sort, so if we only picked one, that would be the way to go.  I'd also actually tend to make it default if we allow both, since if you get unstable when you were expecting stable, you results will be wrong, whereas if you get stable when you were OK with unstable, all you are doing is losing some perf.  So default stable seems safer to me.  As for the per difference, the perf results here are interesting:

https://stackoverflow.com/questions/810951/how-big-is-the-performance-gap-between-stdsort-and-stdstable-sort-in-practic

Basically, there is a measurable win on the CPU from unstable sort, but it's not huge.  Not sure about other platforms.

tzerrell:
Ok, I'm convinced. Going to stable-sort-only sounds reasonable to me.

jbruestle:
Also, we can add an 'unstable' option later as a perf thing if we want.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
haoyouab: Using the test case below:
```
TEST_F(CppEdsl, ArgSort1dInt) {
  auto I = Placeholder(DType::INT32, {10});
  auto O = argsort(I, /*axis=*/0);
  auto program = makeProgram("argsort", {I}, {O});
  std::vector<int32_t> input = {
    0, 1, 8, 5, 5, 2, 0, 7, 7, 10,
  }
  std::vector<int32_t> output = {
    0, 6, 1, 5, 3, 4, 7, 8, 2, 9,
  }
  checkExact(program, {input}, {output});
```
Will raise an error:
```
error: 'std.cmpf' op operand #0 must be floating-point-like, but got 'i32'
  note: see current operation: %13 = "std.cmpf"(%11, %12) {predicate = 4 : i64} : (i32, i32) -> i1
```
Integer input tensor is not tested and it seems this will run into the "Assume float" case which causes type mismatch. I guess `elementType.isSignlessInteger()` is missing so I simply follow what `convertAgg` function at line 552 does. In short:
```
if (type.isa<FloatType>()) {
  ...
} else if (type.isSignedInteger()) {
  ...
} else {
  ...
}
```
And the error's gone. Please let me know if there's anything else missing.

haoyouab:
> If we do decide to leave the interface as-is and force a single choice between stable & unstable, it's unclear to me whether stable is the correct choice. We're adding it because the OpenVINO TopK tests require a stable sort to pass; however, the _documentation_ for TopK does not mention stability, so it's unclear if this is a requirement or just a (misplaced) assumption of the tests. See OV docs: https://docs.openvinotoolkit.org/latest/openvino_docs_ops_sort_TopK_3.html

Yes there's no mention about whether `TopK` uses stable sort or not. My assumption is that it's dependent on implementation. In openvino, I run test cases of `TopK` and the original selection sort by Mars failed while the stable one worked correctly so my best guess is OV is implemented in a stable way (selection sort failed anyway).

haoyouab:
I agree that we should keep **ONLY** stable sort in `argsort` for now. I will remove the `stability` flag and make stable the default behavior.

Besides, I found something interesting when I looked into the OV ref implementation of `TopK`. Please check [TopK OV reference implementation](https://github.com/plaidml/openvino/blob/plaidml/ngraph/core/reference/include/ngraph/runtime/reference/topk.hpp#L136) for details. Basically `std::sort` is used, which does not guarantee to be stable. It's OK for the values tensor but when re-arrange indices based on this (see [this](https://github.com/plaidml/openvino/blob/plaidml/ngraph/core/reference/include/ngraph/runtime/reference/topk.hpp#L144)), the indices of two equivalent values could be re-ordered(or not, there's no guarantee). As I mentioned above, in the test cases they are not re-ordered but I think this is dependent on implementations. Perhaps we should change it to `std::stable_sort` to make sure that. Any ideas?

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
tzerrell: I think it makes sense for these cases and the cases in line 552 to match. So I think this is a good format for this PR.

If we want to make adjustments, I propose we make a separate PR and change both of these sections (and any other similar sections) in that PR.

tzerrell (commented):
IMO this PR is essentially ready for approval, but I'm going to hold off on clicking "Approve" so we can land the sort->stable_sort reference PR and include that repo reference update with in this.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -1015,42 +1020,41 @@ struct ArgSortOpConversion : public OpConversionPattern<tile::ArgSortOp> {
...
-        CmpIPredicate dir{};
+      if (elementType.isa<FloatType>()) {
+        CmpFPredicate dir{};
```
flaub, line 1024: Why is there a trailing `{}`?

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -1066,16 +1070,27 @@ struct ArgSortOpConversion : public OpConversionPattern<tile::ArgSortOp> {
       // unsorted region, by setting the insertion point after its block.
       rewriter.setInsertionPointAfter(minLoop);
 
-      // Swap the sort position with the minimum position.
       auto finalMinPos =
           rewriter.create<LoadOp>(loc, minIdxVar, zeroIndex).getResult();
       ops[axis] = finalMinPos;
       auto finalMinIdx = rewriter.create<LoadOp>(loc, result, ops).getResult();
+
+      // Move every elements between [sortIV,finalMinPos) a step forward
```
flaub, line 1078: Could rephrase this as: "Move all elements ..." or "Move every element ..."

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
flaub: This follows other code that also falls back to unsigned if it's not float or signed. Just FYI, we don't use signless integer types at the edsl/tile level. 

Also, nit: I'd try to break up this function, it's kind of long. Maybe make a helper function that just returns the comparison op (like something that returns the value for `orderPred`).

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
haoyouab: Actually I didn't add this but just reordered the cases. Anyway it looks unnecessary so I will remove it.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
haoyouab: The test case above falls back to the last case (presumably unsigned) but before this change, there was a case that explicitly used `else if (type.isUnsignedInteger())` and it didn't match the test. This is why I guess it's a `SignlessInteger` case. If it's not, I'm not sure why it falls back to the last case then.

As for making a helper function, I updated the code and please check.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
haoyouab: Changed to "Move every element ...".

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -1105,6 +1085,48 @@ struct ArgSortOpConversion : public OpConversionPattern<tile::ArgSortOp> {
...
+      case tile::SortDirection::asc:
+        pred = CmpFPredicate::OLT;
+        break;
+      case tile::SortDirection::desc:
+        pred = CmpFPredicate::OGT;
+        break;
+      }
+      return rewriter.create<mlir::CmpFOp>(loc, pred, compVal, minVal)
+          .getResult();
+    } else if (type.isSignedInteger()) {
```
flaub, line 1104: no need for `else if` since the first `if` is returning. just `if` would be sufficient

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
flaub: I'm curious why we're seeing a signless integer in this case, but it's not a blocker for this PR

haoyouab: I will try to figure it out.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
haoyouab: OK. Then some other similar code block (like line 552) may need this change, too.

Besides, I'm planning to add the `ArgSort1DInt` test case to this PR.

tzerrell (approved):
This looks good!

PR #1708, opened 2021-02-23
Fix layer op

YangleiZouIntel:
When layer op has multiple return values, sometimes the order of the values in layer.return is not correct. 
This would cause error in TopK (not submitted yet). 
This patch fixes this issue.

### DIFF
--- a/plaidml/edsl/tests/edsl_test.cc
+++ b/plaidml/edsl/tests/edsl_test.cc
@@ -2048,6 +2048,19 @@ TEST_F(CppEdsl, LayerMissingOperand) {
   EXPECT_ANY_THROW({ makeProgram("LayerMissingOperand", {A, B}, {O}); });
 }
 
+TEST_F(CppEdsl, LayerMultipleReturnValues) {
+  auto A = Placeholder(DType::FLOAT32, {10, 5});
+  TensorVec tuple = layer("two_output", {A}, {}, [&]() {
+    Tensor idxs = argsort(A, 0);
+    Tensor vals = gather(A, idxs);
+    TensorVec outputs = {vals, idxs};
+    return outputs;
+  });
+
+  auto program = makeProgram("LayerMultipleReturnValues", {A}, tuple);
+  runProgram(program);
+}
+
 TEST_F(CppEdsl, LayerEmbeddedConst) {
   auto A = Placeholder(DType::FLOAT32, {10, 20});
   Tensor O = layer("sum", {A}, [&]() {  //
--- a/plaidml/edsl/tests/edsl_test.cc
+++ b/plaidml/edsl/tests/edsl_test.cc
@@ -2058,6 +2058,16 @@ TEST_F(CppEdsl, LayerMultipleReturnValues) {
   });
 
   auto program = makeProgram("LayerMultipleReturnValues", {A}, tuple);
+  // clang-format off
+  // CHECK-LABEL: CppEdsl.LayerMultipleReturnValues
+  // CHECK: module @LayerMultipleReturnValues
+  // CHECK: func @main(%[[ARG0:.*]]: tensor<10x5xf32>) -> (tensor<10x5x5xf32>, tensor<10x5xsi32>) {
+  // CHECK:   %[[X0:.*]]:2 = layer.box "two_output" (%[[ARG1:.*]]) = (%[[ARG0]]) : (tensor<10x5xf32>) -> (tensor<10x5x5xf32>, tensor<10x5xsi32>) {
+  // CHECK:      %[[X1:.*]] = tile.argsort asc %[[ARG1]][{{[0-9]*}}] : (tensor<10x5xf32>) -> tensor<10x5xsi32>
+  // CHECK:      %[[X2:.*]] = tile.gather %[[ARG1]] %[[X1]] {{{.*}}} : (tensor<10x5xf32>, tensor<10x5xsi32>) -> tensor<10x5x5xf32>
+  // CHECK:      layer.return %[[X2]], %[[X1]] : tensor<10x5x5xf32>, tensor<10x5xsi32>
+  // CHECK:   return %[[X0]]#0, %[[X0]]#1 : tensor<10x5x5xf32>, tensor<10x5xsi32>
+  // clang-format on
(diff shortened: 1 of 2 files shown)


flaub:
We need some tests for this change.

pmlc/ast/builder.cc
```diff
@@ -720,10 +720,14 @@ struct ProgramBuilder {
...
       assert(op && "Unexpected block argument");
       Operation *clonedOp = bodyBuilder.clone(*op, mapper);
-      if (results.contains(value)) {
-        for (Value result : clonedOp->getResults()) {
-          innerResults.push_back(result);
+
+      auto iter = std::find(results.begin(), results.end(), value);
+      if (iter != results.end()) {
+        if (clonedOp->getNumResults() != 1) {
+          throw std::runtime_error("Program build failure.");
```
flaub, line 727: This error message will be confusing to users.

YangleiZouIntel: Thanks. I updated this section of code with support for ops that has multiple results and avoided throwing an error here.

YangleiZouIntel:
> We need some tests for this change.

Please see this [multiple-return-value test](https://github.com/plaidml/plaidml/pull/1708/files#diff-55aa61b72febc9ab6e112cb4e26668c807521b41def522f17aada5a7d047b9ddR2051).

Before this patch is applied, this edsl test generates error:

```
module @LayerOperandOrder  {
  func @main(%arg0: tensor<10x5xf32>) -> (tensor<10x5x5xf32>, tensor<10x5xsi32>) {
    %0:2 = layer.box "test" (%arg1) = (%arg0) : (tensor<10x5xf32>) -> (tensor<10x5x5xf32>, tensor<10x5xsi32>) {
      %1 = tile.argsort asc %arg1[0] : (tensor<10x5xf32>) -> tensor<10x5xsi32>
      %2 = tile.gather %arg1 %1 {axis = 0 : index, batchDims = 0 : index, cubeCoeff = -7.500000e-01 : f64,
           interpolationMode = 1 : i64, mode = 0 : i64, nearestMode = 0 : i64} 
           : (tensor<10x5xf32>, tensor<10x5xsi32>) -> tensor<10x5x5xf32>
      layer.return %1, %2 : tensor<10x5xsi32>, tensor<10x5x5xf32>
    }
    return %0#0, %0#1 : tensor<10x5x5xf32>, tensor<10x5xsi32>
  }
}

error: type of return operand 0 ('tensor<10x5xsi32>') doesn't match function result type ('tensor<10x5x5xf32>') 
in function @main

note: see current operation: "std.return"(%0, %1) : (tensor<10x5xsi32>, tensor<10x5x5xf32>) -> ()
unknown file: Failure

C++ exception with description "Exception at plaidml/testenv.cc:143 with message: Compilation failure" 
thrown in the test body.

```

After this patch is applied, this edsl test passes:
```
module @LayerMultipleReturnValues  {
  func @main(%arg0: tensor<10x5xf32>) -> (tensor<10x5x5xf32>, tensor<10x5xsi32>) {
    %0:2 = layer.box "test" (%arg1) = (%arg0) : (tensor<10x5xf32>) -> (tensor<10x5x5xf32>, tensor<10x5xsi32>) {
      %1 = tile.argsort asc %arg1[0] : (tensor<10x5xf32>) -> tensor<10x5xsi32>
      %2 = tile.gather %arg1 %1 {axis = 0 : index, batchDims = 0 : index, cubeCoeff = -7.500000e-01 : f64, 
           interpolationMode = 1 : i64, mode = 0 : i64, nearestMode = 0 : i64} 
           : (tensor<10x5xf32>, tensor<10x5xsi32>) -> tensor<10x5x5xf32>
      layer.return %2, %1 : tensor<10x5x5xf32>, tensor<10x5xsi32>
    }
    return %0#0, %0#1 : tensor<10x5x5xf32>, tensor<10x5xsi32>
  }
}

```

pmlc/ast/builder.cc
```diff
@@ -721,13 +721,13 @@ struct ProgramBuilder {
...
-          throw std::runtime_error("Program build failure.");
-        }
-        size_t index = std::distance(results.begin(), iter);
-        innerResults[index] = clonedOp->getResult(0);
+      auto itLayerResults = std::find(results.begin(), results.end(), value);
+      if (itLayerResults != results.end()) {
+        auto opResults = op->getResults();
+        auto itOpResults = std::find(opResults.begin(), opResults.end(), value);
+        size_t idxOpResults = std::distance(opResults.begin(), itOpResults);
+        size_t idxLayerResults = std::distance(results.begin(), itLayerResults);
```
flaub, line 729: Can we use `OpResult::getResultNumber` instead of `std::distance`?

plaidml/edsl/tests/edsl_test.cc
```diff
@@ -2048,6 +2048,19 @@ TEST_F(CppEdsl, LayerMissingOperand) {
...
+  auto A = Placeholder(DType::FLOAT32, {10, 5});
+  TensorVec tuple = layer("two_output", {A}, {}, [&]() {
+    Tensor idxs = argsort(A, 0);
+    Tensor vals = gather(A, idxs);
+    TensorVec outputs = {vals, idxs};
+    return outputs;
+  });
+
+  auto program = makeProgram("LayerMultipleReturnValues", {A}, tuple);
+  runProgram(program);
```
flaub, line 2061: Can you add a lit test? I'd like to see the before/after of this case.

pmlc/ast/builder.cc
flaub: We should also probably use a `SmallVector` rather than a `SetVector` for `results` since we don't just want to check set membership, we actually want to know which result we're dealing with.

plaidml/edsl/tests/edsl_test.cc
YangleiZouIntel: Yes, added a lit test.

pmlc/ast/builder.cc
YangleiZouIntel: Yes, both fixed.

flaub (approved):
Thanks, LGTM

PR #1711, opened 2021-02-25
Add deformableconvolution

DingZhangIntel:
This PR adds multi-D deformable convolution and depends on https://github.com/plaidml/openvino/pull/142 . 
Currently there is no reference implementation in OV, we compared 2D results with deformable convolution in MXNet (Currently there is only 2D algorithm implemented in MXNet).


plaidml/bridge/openvino/ops/deformable_convolution.cpp
```diff
@@ -0,0 +1,285 @@
...
+#include "plaidml/op/op.h"
+
+using namespace plaidml;          // NOLINT[build/namespaces]
+using namespace InferenceEngine;  // NOLINT[build/namespaces]
+using namespace edsl;
+
+namespace {
+
+// Compute pad_before and the size of output.
+// This function is based on (the functon compute_padding_and_output_size() in plaidml/op/lib/ops.cc).
```
tzerrell, line 20: Hmm, perhaps it would be possible to template the compute_padding_and_output_size function allowing it to take and produce `size_t` (or even something other integer like `int64_t`) instead of `TensorDim`. If that could be made to work that would be better as it would reduce code duplication.

plaidml/bridge/openvino/ops/deformable_convolution.cpp
```diff
@@ -0,0 +1,285 @@
...
+    }
+    pad_before = (max + lower_term) / 2;
+    size_t output_size = ((I_eff + stride - 1) / stride);
+    std::pair<size_t, size_t> result(pad_before, output_size);
+    return result;
+  }
+  THROW_IE_EXCEPTION << "Unexpected autopadding mode.";
+}
+
+std::vector<int> cast_vector(std::vector<int64_t> vec) {
```
tzerrell, line 52: Is `cast_vector` needed? It looks like these vectors are just being used as parameters in `compute_deformable_convolution ` -- can we not just change the type in that function to vectors of int64_t instead of vectors of ints?

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/deformable_convolution.cpp
```diff
@@ -0,0 +1,316 @@
...
+INSTANTIATE_TEST_CASE_P(Convolution3D_SameUpper, DeformableConvolutionLayerTest,
+                        ::testing::Combine(                                            //
+                            conv3DParams_SameUpper,                                    //
+                            ::testing::ValuesIn(netPrecisions),                        //
+                            ::testing::Values(std::vector<size_t>({1, 1, 4, 4, 4})),   //
+                            ::testing::Values(std::vector<size_t>({1, 24, 4, 4, 4})),  //
+                            ::testing::Values(CommonTestUtils::DEVICE_PLAIDML)),       //
+                        DeformableConvolutionLayerTest::getTestCaseName);
+
+INSTANTIATE_TEST_CASE_P(Convolution3D_SameLower, DeformableConvolutionLayerTest,
```
tzerrell, line 307: Can you prefix at least one of the test names with `smoke_` so at least one DeformableConvolution gets run during `check-smoke`? I appreciate that you're not cluttering `check-smoke` with lots of tests, but every op that we can test at all should have at least one test in `check-smoke`.

plaidml/bridge/openvino/ops/deformable_convolution.cpp
```diff
@@ -0,0 +1,285 @@
...
+  edsl::Tensor deformed_input = deformed_input_vec[0];
+  // Compute DeformableConvolution.
+  edsl::Tensor result = op::convolution(deformed_input, F)
+                            .strides(std::vector<int>{F_shape.begin() + 2, F_shape.end()})
+                            .autopad_mode(plaidml::op::AutoPadMode::VALID)
+                            .input_layout(plaidml::op::TensorLayout::NCX)
+                            .filter_layout(plaidml::op::TensorLayout::KCX)
+                            .autogroup_mode(plaidml::op::AutoGroupMode::EXPLICIT)
+                            .group_layout(plaidml::op::GroupLayout::IN_C)
+                            .group_layout(plaidml::op::GroupLayout::IN_K)
```
tzerrell, line 215: One of these is being overwritten by the other (I think this ultimately sets the group layout to IN_K but I'm not 100% certain of that). Please remove the redundant GroupLayout and only use the relevant one.

Actually, if I'm reading the tests correctly, OpenVINO doesn't currently support grouped deformable conv, so we can't be sure what the intended layout is (I would guess `plaidml::op::GroupLayout::SEPARATE` as that's what we need for `GroupConvolution`; but if so, we need a special case for `G == 1` because we probably don't want to introduce an extra Group dimension in a non-grouped convolution). If OV doesn't itself support grouped deformable conv, we should **instead throw an error if the group size != 1** and not set `group_layout` at all. That way, once OV support is added, any users will hit our error rather than assuming that the group_layout is correct (even though we just made a guess about the intended layout).

tzerrell (commented):
Great! The core operation looks good to me. I have some style / interface suggestions.

plaidml/bridge/openvino/ops/deformable_convolution.cpp
DingZhangIntel: Ok, it make sense, and i will fix it.

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/deformable_convolution.cpp
DingZhangIntel: I will add the test names with `smoke_`.

plaidml/bridge/openvino/ops/deformable_convolution.cpp
DingZhangIntel: ok, I will submit another pr which templates the compute_padding_and_output_size function.

plaidml/bridge/openvino/ops/deformable_convolution.cpp
DingZhangIntel: I use cast_vector because that the function op::broadcast just supports vector<int> as parameter, if i don't use cast_vector, there will be too many static_cast<int> function when i use the function op::broadcast.

tzerrell: Makes sense, seems ok. If you think it would simplify this code in other ways, I don't see any problems with changing `broadcast` to use `int64_t` instead of `int`. But your current approach is also fine.

plaidml/bridge/openvino/ops/deformable_convolution.cpp
YangleiZouIntel: We found out that we are able to reuse the original function in `plaidml/op/lib/ops.cc` in this op. We added the declaration in `plaidml/op/lib/ops.h` [(please see here)](https://github.com/plaidml/plaidml/pull/1711/files#diff-622f562b4f246e0fa4c4dc8fceb9be1e8e9ecd7109aa962e53c807591a7a43e6R40). But we also have to add  `plaidml::op::lib` as a dep of openvino plugin [(please see this line)](https://github.com/plaidml/plaidml/pull/1711/files#diff-56805e100050a20c6d5efc0b97e1ec1182f43cd55cccb1152e1e5545b961ab6bR22).

plaidml/bridge/openvino/ops/deformable_convolution.cpp
YangleiZouIntel: Done, we have removed the `cast_vector` function and changed the `op::broadcast` to use `int64_t` instead of `int` and fixed all related tests.

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/deformable_convolution.cpp
YangleiZouIntel: Done

plaidml/bridge/openvino/ops/deformable_convolution.cpp
YangleiZouIntel: Done

plaidml/bridge/openvino/ops/deformable_convolution.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include "ngraph/opsets/opset.hpp"
+#include "ngraph/opsets/opset4.hpp"
+
+#include "plaidml/op/lib/ops.h"
+#include "plaidml/op/op.h"
+
+using namespace plaidml;           // NOLINT[build/namespaces]
+using namespace InferenceEngine;   // NOLINT[build/namespaces]
+using namespace plaidml::edsl;     // NOLINT[build/namespaces]
+using namespace plaidml::op::lib;  // NOLINT[build/namespaces]
```
flaub, line 17: Which functions are we using from `op::lib`? I'm worried that we're going to get ODR violations if we do this. I'd suggest that whatever functions we want to use from the lib we just expose as a normal op.

plaidml/bridge/openvino/ops/deformable_convolution.cpp
tzerrell: @flaub Here's the function being used out of `plaidml::op::lib` that you asked about below (specifically `compute_padding_and_output_size`). It's used as a utility for `conv` and `pool` in the op library, and we need the same utility here for deformable convolution.

We could potentially expose this in `plaidml/op/op.h`, although it is a utility for writing operations that works on TensorDims, rather than an operation proper that works on Tensors like the rest of `op.h`. (Nevertheless, TensorDims can be wrapped as Values so I think exposing it like this should at least be possible.) We could also potentially move it to some shared utility location, although I'm not quite sure where that would be (a new `plaidml/op/util.h` perhaps?) or whether there would be additional problems from moving it to such a location.

plaidml/bridge/openvino/ops/deformable_convolution.cpp
tzerrell: see the comment I tagged you in above

plaidml/bridge/openvino/ops/deformable_convolution.cpp
tzerrell: Given this extra complexity, I think we can return to the original design with  a DeformableConv-specific compute_padding function. We can keep in mind for the future that a possible improvement will be to reuse the code by following one of the paths from my previous comment.

plaidml/op/lib/ops.cc
```diff
@@ -351,7 +351,7 @@ std::pair<TensorDim, TensorDim> compute_padding_and_output_size(  //
     bool use_ceil_for_output_shape) {
   // Effective input and filter sizes are the sizes after dilations are
   // accounted for. So a 4x3 filter dilated by (3, 2) has an effective filter
-  // size of 11 and 5 for its 2 spatial dims
+  // size of 10 and 5 for its 2 spatial dims
```
tzerrell, line 354: Thanks for this comment fix, by the way!

tzerrell (approved):
Looks good, thanks!

PR #1713, opened 2021-02-26
Add OV Range op

tzerrell:
Adds a Range op (that only accepts Constants as inputs). Not tested as the "constant propagation to output causes crash" bug prevents the unit tests from working, but should hopefully be useful in practical cases. Marked as a draft until we've tried it in a practical case.


tzerrell:
Still no CI due to the bug mentioned above, but I have tested a network with a Range op in it, and with the changes from my newest commit it did not crash, so IMO this is "tested enough" to land.

YangleiZouIntel:
LGTM

flaub:
@tzerrell Did we decide that we didn't need the range op after all since we do a constant folding pass? Should we close this one?

tzerrell:
Yes, good point, a constant-only Range op like this is pointless with constant folding, which we now have. And we don't yet support a dynamically-shaped Range op.

PR #1715, opened 2021-02-26
Add OpenVINO network Constant Folding

tzerrell:
Add the OpenVINO network-level ConstantFolding pass to the PlaidML OV plugin.

Note that we also considered adding a ConvertReduceToPooling pooling pass. However, this requires us to include ~`transformations/common_optimizations/common_optimizations.hpp`~ [**edit:** actually the file is `transformations/op_conversions/convert_reduce_to_pooling.hpp`], and our build is having trouble finding that file. If we decide we still want this pass, I propose we follow up in another PR that adds the pass and makes the necessary build changes.


flaub (approved):
Thanks!

PR #1716, opened 2021-02-27
offset-based batch reduce gemm implementation

taknevski:
0. Main change to PxaGemmOp is encoding numbatches as an array instead of a scalar, to allow for additional reduction dimensions.
1. Added implementation of offset based BRGEMM op in xsmm_lowering.cc
2. Added lowering of PxaGemmOp to offset-based BRGEMM when there are additional reduction indices, by default it lowers to just GEMM, or batched GEMM when the number of batches is greater than 1.
3. Added reduction indices' stride information to PxaGemmOp's tile map.
4. Added test cases that test Xsmm and Pxa lowering for offset-based batch reduce GEMM op.


jbruestle (commented):
Overall, I think the approach is good, and the transforms make sense to me.  Some inline comments.  Also, looks like this commit broke some tests, apparently in a meaningful way (see https://buildkite.com/plaidml/plaidml-plaidml/builds/5395, should be available by clicking on the buildkite/plaidml-plaidml/pr bit on the main page of the PR).  Unfortunately, it appears to be a mac only failure, and I'm guessing that you don't currently have access to an easy way to repro it, so I'm trying to reproduce it myself on our CI machines.

pmlc/dialect/pxa/ir/ops.td
```diff
@@ -90,7 +90,7 @@ def PxaGemmOp : PXA_OpWithPP<"gemm"> {
     AffineMapAttr:$bAccessMap,
     AffineMapAttr:$bTileMap,
     I64ArrayAttr:$tile,
-    I64Attr:$numBatches,
```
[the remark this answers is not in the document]
taknevski: Yes you are right, the first dimension of numBatches is the batch size over k. Sure I can modify the implementation for separate batches over the other dimensions. I guess we need to decide on what the right interface is for the numBatches field. I'll add the required comments.

pmlc/target/x86/tests/xsmm_lowering_3x3.mlir
```diff
@@ -0,0 +1,20 @@
+// RUN: pmlc-opt -tile-compute-bounds -convert-tile-to-pxa  --pass-pipeline='func(x86-affine-stencil-xsmm{threads=1 batched=true})'  -x86-convert-pxa-to-affine -lower-affine \
```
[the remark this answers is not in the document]
taknevski: Sure I can add this.

jbruestle (dismissed):
Here is the stack trace for the mac failure:  It's in test: Suite/ConvolutionTest.Basic/3

* thread #3, stop reason = EXC_BAD_ACCESS (code=2, address=0x700008ba2f50)
  * frame #0: 0x0000000105e456ec
    frame #1: 0x000000010230c683 op_test`__kmp_invoke_microtask + 147
    frame #2: 0x00000001022e30e1 op_test`__kmp_invoke_task_func + 129
    frame #3: 0x00000001022e269b op_test`__kmp_launch_thread + 235
    frame #4: 0x000000010230ce51 op_test`__kmp_launch_worker(void*) + 161
    frame #5: 0x00007fff67ec2109 libsystem_pthread.dylib`_pthread_start + 148
    frame #6: 0x00007fff67ebdb8b libsystem_pthread.dylib`thread_start + 15

The test that fails looks strongly related to the actual change, but it looks like a tricky one to track down.

jbruestle:
Basically it looks like the jitted code is reading/writing out of bounds at some point, but it probably works by accident on other platforms.

taknevski:
> Basically it looks like the jitted code is reading/writing out of bounds at some point, but it probably works by accident on other platforms.

I'll debug this issue and get back to you, thanks for alerting me

jbruestle (approved):
Looks good!  Not sure why the IRV10 stuff couldn't run, will try to trigger a rebuild.  Thanks for working on this.

pmlc/target/x86/tests/xsmm_call_3x3.mlir
```diff
@@ -0,0 +1,228 @@
...
+  %c3 = constant 3 : index
+  %N = dim %A, %c0 : memref<?x?x?x?xf32>
+  %P = dim %A, %c1 : memref<?x?x?x?xf32>
+  %Q = dim %A, %c2 : memref<?x?x?x?xf32>
+  %K = dim %A, %c3 : memref<?x?x?x?xf32>
+
+
+
+
+
```
flaub, line 86: Drop extra whitespace here and in other spots in this file

pmlc/dialect/xsmm/ir/ops.cc
```diff
@@ -85,6 +85,43 @@ ParseResult parseGemmInvokeF32Op(OpAsmParser &parser, OperationState &result) {
       parser.resolveOperands(b.indices, indexType, result.operands));
 }
 
+ParseResult parseBRGemmOffsInvokeF32Op(OpAsmParser &parser,
```
flaub, line 88: Can we use the ODS assembly format instead of hand writing a printer/parser?

See: https://mlir.llvm.org/docs/OpDefinitions/#declarative-assembly-format

pmlc/dialect/pxa/transforms/stencil_gemm.cc
```diff
@@ -222,13 +242,26 @@ class StencilGEMM : public StencilBase {
     int64_t numBatches = 1;
     int64_t kRange = getIdxRange(perm.indexes[2]);
     IVLOG(3, "kRange: " << kRange);
+    SmallVector<BlockArgument, 4> Aindices, Bindices;
```
flaub, line 245: These should be named `aIndices` and `bIndinces` to match our code style.

PR #1720, opened 2021-03-02
Fix ExtractImagePatches dtype crash

tzerrell:
Fixes a crash with an obscure exception ("Value cannot be coerced into Tensor") to something clearer when ExtractImagePatches is performed on a tensor with an unsupported dtype; also expands the list of supported dtypes for this op.

This also makes some conceptually-unrelated logging upgrades that I used while looking for the source of the original bug.


plaidml/bridge/openvino/ops/extract_image_patches.cpp
```diff
@@ -74,12 +74,28 @@ void registerExtractImagePatches() {
 
     edsl::Tensor kernel_tensor;
     switch (input_tensor.dtype()) {
+      case DType::FLOAT16:
+        // TODO: Investigate whether there's a way to create fp16 directly
+        kernel_tensor = edsl::cast(create_kernel_tensor<float>(filter_row, filter_col, input_tensor), DType::FLOAT16);
```
flaub, line 79: Should be ok to use the `half` data type here

tzerrell: Including half here led to build errors until I set `HALF_ENABLE_CPP11_TYPE_TRAITS`. I'm not sure if `add_compile_definitions` is the right style for such definitions, so let me know if I should be doing this in some other way.

tzerrell:
@jd-bartlett96 this is an upcoming fix like I just mentioned to you directly.

flaub:
@tzerrell Adjusted slightly to use the `DEFINES` parameter of `pml_cc_library`

flaub (approved):

PR #1718, opened 2021-03-02
Use standalone build of OpenVINO

flaub:
* Add `vendor/openvino` as a submodule
* Build openvino as a standalone project
* Use the `./configure` script to build openvino
* Trigger the IRv10 models pipeline
* Add a `testkit` package

PR #1721, opened 2021-03-03
add TopK op

haoyouab:
TopK can basically work except for case `ngraph::opset4::TopK::SortType::NONE` whose behavior is **UNDEFINED** (check [OV docs](https://docs.openvinotoolkit.org/latest/openvino_docs_ops_sort_TopK_3.html)) but actually in [OV reference implementation](https://github.com/plaidml/openvino/blob/plaidml/ngraph/core/reference/include/ngraph/runtime/reference/topk.hpp#L116), it uses `std::nth_element`. I think we should discuss about if it requires too much effort or even if it's worth doing this in `topk.cpp` considering it's undefined.


tzerrell (approved):
~LGTM, but this will need to be updated with the changes from #1718 (and will fail CI until that happens). I'll either do that update myself or put together instructions for it later today.~ This is wrong, see below

tzerrell:
Oh, whoops, I misread the cause of the CI failure. I'll dig into what's wrong here.

tzerrell:
Oh, and I realize I didn't write down my thoughts about `SortType::NONE`: I think the current approach is good. Because the docs say it has undefined behavior, I do not think we should attempt to emulate existing behavior.

haoyouab:
> Oh, and I realize I didn't write down my thoughts about `SortType::NONE`: I think the current approach is good. Because the docs say it has undefined behavior, I do not think we should attempt to emulate existing behavior from the reference implementation.

OK. Then let's keep the `SortType::NONE' part as is if no one has a disagreement.

plaidml/bridge/openvino/ops/topk.cpp
```diff
@@ -0,0 +1,84 @@
...
+    auto* layer = ngraph::as_type<ngraph::opset4::TopK>(ctx.layer);
+    IE_ASSERT(ctx.operands.size() == 2);
+    auto I = ctx.operands.at(0);
+    auto K = layer->get_k();
+    auto axis = layer->get_axis();
+    auto mode = layer->get_mode();
+    auto sort = layer->get_sort_type();
+    long ndims = I.rank();
+
+    edsl::Tensor values, indices;
```
flaub, line 31: Define these when they have values down below

haoyouab: Done.

plaidml/bridge/openvino/ops/topk.cpp
```diff
@@ -0,0 +1,84 @@
...
+      case Mode::MIN:
+        direction = Direction::ASC;
+        break;
+      default:
+        THROW_IE_EXCEPTION << "invalid topk sort mode";
+    }
+
+    edsl::Tensor values, indices;
+    values = op::sort(I, axis, direction);
+    indices = edsl::argsort(I, axis, direction);
```
flaub, line 45: nit: better to define a variable with it being initialized at construction.

```
edsl::Tensor values = op::sort(I, axis, direction);
edsl::Tensor indices = edsl::argsort(I, axis, direction);
```

haoyouab: OK. It's done.

YangleiZouIntel:
> Oh, whoops, I misread the cause of the CI failure. I'll dig into what's wrong here.
Thanks Tim, somehow CI seems to treat @haoyouab 's account as a public contributor, please help check. I updated this PR and it passed CI now.

flaub:
@YangleiZouIntel  @haoyouab When you submit a PR, I believe the git config for username and email comes into play. We might need to ensure that the account that is registered with buildkite is the same as the author/email that the git commit is set to.

tzerrell:
@haoyouab we've updated your email in buildkite to match the one you used to make these commits. Please let us know if you hit further problems like this.

tzerrell (approved):

haoyouab:
Sorry about this. It's my fault because I previously use my personal email on github and I recently changed it to intel email. Buildkite  shoud be working now. Thanks!

tzerrell:
No worries, glad we were able to get you updated in buildkite!

PR #1722, opened 2021-03-03
Throw on opset version mismatch

tzerrell:
There are some operations that have different versions from different opsets, but for which we only support one of these versions. For these operations, this change throws when we receive a different version of the operation (instead of segfaulting which was the previous behavior).


flaub:
I'm not sure what the right answer is, someone on the OpenVINO team would be better suited. However, I notice that in other plugins the following passes are used:

```
        passManager.register_pass<ngraph::pass::ConvertOpSet3ToOpSet2>();
        passManager.register_pass<ngraph::pass::ConvertOpSet2ToOpSet1>();
        passManager.register_pass<ngraph::pass::ConvertOpSet1ToLegacy>();
        passManager.register_pass<ngraph::pass::ConstantFolding>();
```

What is the actual difference in opsets?

tzerrell:
Oh interesting, thanks for that suggestion. The differences tend to be details: is something an input or an attribute, or maybe a new input/attribute, or maybe a boolean attribute was turned into an enum with additional options. I'll investigate what happens on example networks with those passes.

tzerrell:
Where conversion passes exist, this now adds conversions for the opset version of ops where we don't support all versions. I've retained the error messages in case a pass doesn't apply.

I also added a fix for Interpolate -- there's one size/scale per interpolated axis, but there are no sizes/scales for axes that aren't interpolated, so we need to use different variables to iterate over the axes of interpolation vs. the overall axes of the tensor.

flaub (approved):

PR #1727, opened 2021-03-15
Update to latest LLVM (with patches)

flaub:


pmlc/dialect/pxa/ir/ops.cc
```diff
@@ -38,19 +38,119 @@ ParseResult parseKeywordIntoEnumAttr(OpAsmParser &parser,
   return success();
 }
 
-/// Implements `map` and `operands` composition and simplification to support
-/// `makeComposedAffineApply`. This can be called to achieve the same effects
-/// on `map` and `operands` without creating an AffineApplyOp that needs to be
-/// immediately deleted.
+/// Replace all occurrences of AffineExpr at position `pos` in `map` by the
```
jbruestle, line 41: Curious what is going on here?

jbruestle (approved):
Great work, one question, but didn't seem worth blocking such a big and needed commit.

pmlc/dialect/pxa/ir/ops.cc
flaub: The change here is that originally we had copied some helper code from upstream because currently that code is not exposed as public functions. Other public code was refactored and so this private code needed to be copied again.

In the future we should figure out how to patch upstream or add our code to upstream to avoid this.

PR #1732, opened 2021-03-24
PXA store prototype

shiyao00:
- Add pxa.store and pxa.vector_store
- Modify PxaReduceOpInterface to PxaWriteOpInterface which is shared by pxa.store and pxa.reduce

### DIFF
--- a/pmlc/dialect/pxa/ir/ops.cc
+++ b/pmlc/dialect/pxa/ir/ops.cc
@@ -165,11 +165,14 @@ struct SimplifyAffineOp : public OpRewritePattern<AffineOpTy> {
 
   LogicalResult matchAndRewrite(AffineOpTy affineOp,
                                 PatternRewriter &rewriter) const override {
-    static_assert(std::is_same<AffineOpTy, PxaReduceOp>::value ||
-                      std::is_same<AffineOpTy, PxaVectorReduceOp>::value ||
-                      std::is_same<AffineOpTy, PxaLoadOp>::value ||
-                      std::is_same<AffineOpTy, PxaVectorLoadOp>::value,
-                  "affine reduce/vector_reduce or load op expected");
+    static_assert(
+        std::is_same<AffineOpTy, PxaReduceOp>::value ||
+            std::is_same<AffineOpTy, PxaVectorReduceOp>::value ||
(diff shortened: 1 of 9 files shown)


pmlc/dialect/pxa/ir/ops.cc
```diff
@@ -474,6 +494,54 @@ OpFoldResult PxaReduceOp::fold(ArrayRef<Attribute> cstOperands) {
...
+  return failure(
+      parser.parseOperand(val) || parser.parseComma() ||
+      parser.parseOperand(out) ||
+      parser.parseAffineMapOfSSAIds(idxs, mapAttr, "map", result.attributes) ||
+      parser.parseOptionalAttrDict(result.attributes) ||
+      parser.parseColonType(type) || parser.addTypeToList(type, result.types) ||
+      parser.resolveOperand(val, type.getElementType(), result.operands) ||
+      parser.resolveOperand(out, type, result.operands) ||
+      parser.resolveOperands(idxs, indexTy, result.operands));
+}
```
flaub, line 529: We should use the new assembly format support instead of hand writing this: https://mlir.llvm.org/docs/OpDefinitions/#declarative-assembly-format

pmlc/dialect/pxa/ir/ops.td
```diff
@@ -56,26 +64,26 @@ def PxaReduceOp : PxaReduceOpBase<"reduce"> {
...
-      build($_builder, $_state, mem.getType(), agg, val, mem, map, idxs);
-    }]>
-  ];
+  let builders = [OpBuilder<(ins "mlir::AtomicRMWKind"
+                             : $agg, "mlir::Value"
+                             : $val, "mlir::Value"
+                             : $mem, "mlir::AffineMap"
+                             : $map, "mlir::ValueRange"
+                             : $idxs),
+                            [{
```
flaub, line 78: The original style matches closer to upstream. I think we should try to match upstream style as much as possible to make submission to the pxa dialect to MLIR core easier.

flaub:
We'll need to have lit tests

flaub:
Generally I think we should avoid making big style changes to the .td, perhaps we can confine those changes to an NFC

PR #1733, opened 2021-03-24
Add non-affine store for PXA dialect

shiyao00:


pmlc/dialect/pxa/ir/ops.td
```diff
@@ -273,6 +273,49 @@ def PxaVectorLoadOp : PxaLoadOpBase<"vector_load"> {
   let hasFolder = 1;
 }
 
+class PxaStoreOpBase<string mnemonic, list<OpTrait> traits = []> :
+    PXA_OpWithPP<mnemonic, !listconcat(traits, [MemRefsNormalizable])> {
+
+  code extraClassDeclarationBase = [{
+    mlir::Value getValueToStore() { return value(); }
+    unsigned getValueOperandIndex() { return 0; }
```
flaub, line 281: These are already generated by the ODS/tablegen, we shouldn't need them.

pmlc/dialect/pxa/ir/ops.td
```diff
@@ -273,6 +273,49 @@ def PxaVectorLoadOp : PxaLoadOpBase<"vector_load"> {
   let hasFolder = 1;
 }
 
+class PxaStoreOpBase<string mnemonic, list<OpTrait> traits = []> :
+    PXA_OpWithPP<mnemonic, !listconcat(traits, [MemRefsNormalizable])> {
+
+  code extraClassDeclarationBase = [{
+    mlir::Value getValueToStore() { return value(); }
```
flaub, line 280: I don't think we need to make a wrapper, we can just call `value()`

pmlc/dialect/pxa/ir/ops.td
```diff
@@ -273,6 +273,49 @@ def PxaVectorLoadOp : PxaLoadOpBase<"vector_load"> {
...
+    unsigned getValueOperandIndex() { return 0; }
+    void setValueToStore(mlir::Value value) { setOperand(getValueOperandIndex(), value); }
+
+    mlir::Value getMemRef() { return memref(); }
+    unsigned getMemRefOperandIndex() { return 1; }
+    void setMemRef(mlir::Value value) { setOperand(getMemRefOperandIndex(), value); }
```
flaub, line 286: ODS already generates a setter, just use `op.memref(newValue)`

jbruestle (changes requested):
Can we add a reduction type (i.e. assign, addi, etc)?  Basically more like atomicrmw or pxa.reduce.  In that case, we can get plain 'store' via 'assign', but we also add atomic add, which is actually what scatter often uses (and it removes a load + add + store for the example you sent originally).  Also it makes thing parallel-safe if we lower to atomicrmw, which in fact scatter should require (right now I think it just produces incorrect results in some cases, but we don't have enough tests).

shiyao00:
> Can we add a reduction type (i.e. assign, addi, etc)? Basically more like atomicrmw or pxa.reduce. In that case, we can get plain 'store' via 'assign', but we also add atomic add, which is actually what scatter often uses (and it removes a load + add + store for the example you sent originally). Also it makes thing parallel-safe if we lower to atomicrmw, which in fact scatter should require (right now I think it just produces incorrect results in some cases, but we don't have enough tests).

Done

pmlc/dialect/pxa/ir/ops.td
shiyao00: Removed

jbruestle (approved):

PR #1734, opened 2021-03-25
Hack fix to fusion

jbruestle:
Checking effect on perf/correctness.


brightlaboratory (approved):
Looks great :) Thanks for the quick fix.

flaub:
I don't think we want to land this right?

jbruestle:
Correct, while it does fix the issue, it creates a major compile time slowdown, and the effect on runtime with / without fusion is negligible (basically in the noise of our current tests).

PR #1739, opened 2021-03-31
PXA store transformation

shiyao00:


pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -230,6 +230,22 @@ struct PxaVectorReduceOpConversion
...
 
+struct PxaStoreOpConversion : public OpConversionPattern<pxa::PxaStoreOp> {
+  using OpConversionPattern<pxa::PxaStoreOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(pxa::PxaStoreOp op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const final {
+    auto memref = op.memref();
+    auto resultType = memref.getType().cast<MemRefType>().getElementType();
+    rewriter.create<AtomicRMWOp>(op.getLoc(), resultType, op.agg(), op.value(),
```
jbruestle, line 241: We should probably add a special case for agg 'assign' to make a plain store, or verify that some later pass (perhaps in LLVM) does a reduction to store.  To my understanding, atomicrmw for assign basically does an atomic swap and returns the original value (prior to the assign), which still requires a barrier I think.  Since we never read the 'old' value, maybe MLIR/LLVM already convert to store.  If not, since PXA semantics for parallel 'assign' only require that the output is *any* of the written values, reduction to std.store is legal in our case and is likely quite a bit more performant.

jbruestle (commented):
Looks great, I'm very glad to see this fixed.  One minor perf comment inline.

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
shiyao00: Done.

jbruestle (dismissed):

jbruestle (approved):

PR #1740, opened 2021-04-01
WIP: Math vectorization

LiyangLingIntel:
Update vectorize pass, add functions to deal with Ops from Math Dialect.

Currently add a new flag `vectorzeMathOp` for vectorize pass to decide vectorize Math Ops directly or create loop for these ops with scalar operand type. 

Not sure if this is a good way to do with it, please add comments for this implementation.


shiyao00 (approved):

alheinecke:
@tzerrell @shiyao00 : does it make sense to merge this PR as we are now marching towards using TPP for everything. Would there be a short-term benefit from merging?

PR #1741, opened 2021-04-02
add DetectionOutput op

haoyouab:
`DetectionOutput` basically works for most of the attributes except the following three. Please check the note below.

`keep_top_k`: Now it's not working with multiple batches. The reason is that `keep_top_k` gets the top K elements after NMS process (note that `top_k` is passed into NMS as `max_output_boxes_per_class`) but the output numbers of every batches could be different. For example, there are 2 batches and the number of output boxes is 10 (3 in batch 1, 5 in batch 2 and 2 -1s) but we could not get this information from current NMS op. In this way we can't sort those scores per batch.

TODO: implement `keep_top_k` inside `NMS`.

`share_location`: is a flag that denotes if bounding boxes are shared among different classes. If it's set to false, we can't pass `boxes` and `scores` into NMS directly because according to the [OV specs of NMS](https://docs.openvinotoolkit.org/latest/openvino_docs_ops_sort_NonMaxSuppression_5.html) that means the boxes are shard among classes.

TODO: implement `share_location` in `NMS` or split the `scores` by class in `DetectionOutput`.

`decrease_label_id`: There are two styles of NMS: Caffe and MxNet. Neither of them is the same as NMS described in [OV specs](https://docs.openvinotoolkit.org/latest/openvino_docs_ops_sort_NonMaxSuppression_5.html). Please check [Caffe reference](https://github.com/weiliu89/caffe/blob/ssd/src/caffe/util/bbox_util.cpp#L1901) and [MxNet reference](https://github.com/apache/incubator-mxnet/blob/ccb08fb5caa5ae95d785fbed34c7fa66aaf53390/src/operator/contrib/multibox_detection.cc#L152). In summary, they use different kinds of fast version NMS. The Caffe NMS is much more similar to the OV NMS so the tests passed in most cases (but still differentiates in some special cases) so I'm using Caffe style for now. The MxNet NMS is very different so it's not supported now.

TODO: implement `decrease_label_id` inside `NMS` (perhaps by providing different modes).

In summary, some other versions/modifications of `NMS` should be provided. I think it's probably better to implement these attributes in `NMS` in `oplib` and provide users with extra parameters/modes. If modifying the `NMS` is acceptable, I will submit PRs later to support these attributes.


plaidml/bridge/openvino/ops/detection_output.cpp
```diff
@@ -0,0 +1,223 @@
...
+#include "plaidml/op/op.h"
+
+using namespace plaidml;          // NOLINT[build/namespaces]
+using namespace InferenceEngine;  // NOLINT[build/namespaces]
+using namespace plaidml::edsl;
+
+namespace PlaidMLPlugin {
+
+edsl::Tensor DecodeBoxes(edsl::Tensor priors, edsl::Tensor prior_variances, edsl::Tensor location,
+                         std::string code_type, int input_height, int input_width, int batch, int num_priors,
```
flaub, line 19: Use `const std::string&`, or use `llvm::StringRef`

plaidml/bridge/openvino/ops/detection_output.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+#include "plaidml/op/op.h"
+
+using namespace plaidml;          // NOLINT[build/namespaces]
+using namespace InferenceEngine;  // NOLINT[build/namespaces]
+using namespace plaidml::edsl;
+
+namespace PlaidMLPlugin {
+
+edsl::Tensor DecodeBoxes(edsl::Tensor priors, edsl::Tensor prior_variances, edsl::Tensor location,
```
flaub, line 18: Functions should be like `decodeBoxes`

plaidml/bridge/openvino/ops/detection_output.cpp
```diff
@@ -0,0 +1,223 @@
...
+    edsl::Tensor decoded_bboxes;
+    if (with_add_pred) {
+      // Update confidence if there are 5 inputs.
+      Tensor IX = edsl::index({edsl::TensorDim(num_priors)}, 0);
+      IX = IX * 2 + 1;
+      edsl::Tensor arm_conf = edsl::gather(ArmConfidence, IX).axis(1);
+      arm_conf = op::repeat(op::unsqueeze(arm_conf, {-1})).count(num_classes).axis(2);
+      confidence = edsl::select(arm_conf < objectness_score, cast(edsl::Tensor{0.0f}, DType::FLOAT32), confidence);
+
+      // Decode bouding boxes
```
tzerrell, line 149: Typo: `bouding` -> `bounding`

tzerrell (commented):
> I think it's probably better to implement these attributes in NMS in oplib and provide users with extra parameters/modes. If modifying the NMS is acceptable, I will submit PRs later to support these attributes.

Yes, it is definitely acceptable and indeed preferable to modify OpLib NMS. I think the ideal outcome would be to have a single OpLib NMS op that we use for all the OpenVINO nms-type ops (this, NonMaxSuppression, and Proposal). This will necessarily involve modifying oplib NMS to accommodate differences in features.

My opinion is that it is generally preferable to have most of the functionality in the OpLib NMS op, and to have the OpenVINO/PlaidML bridge be a relatively thin wrapper. That said, we shouldn't do so if it comes at the cost of making the code more complex or harder to understand & maintain. In that spirit, I have made a suggestion about moving `decodeBoxes` into the oplib. It may also be possible to other code into the oplib as well, such as the code for filling prior_boxes & prior_variances -- I have not considered those possibilities in detail though.

It may make sense to make those changes in followup PRs instead. What does need to get fixed in this PR is the correctness error in unit tests. If the error is occurring due to a case we don't yet support, we should throw an exception when that case occurs rather than allowing a correctness error. Or if it's just a bug, fixing the error would be even better.

> The MxNet NMS is very different so it's not supported now.

This makes sense, we can come back and handle the other NMS style in a different PR.

haoyouab:
> What does need to get fixed in this PR is the correctness error in unit tests. If the error is occurring due to a case we don't yet support, we should throw an exception when that case occurs rather than allowing a correctness error. Or if it's just a bug, fixing the error would be even better.

I think it's related to the stability of sort algorithm again. Please check [OV reference implementation](https://github.com/plaidml/openvino/blob/plaidml/ngraph/core/reference/include/ngraph/runtime/reference/detection_output.hpp#L652) of `DetectionOutput`. It's using `std::sort` which is not stable (`std::stable_sort` is used somewhere else in the same source file though). Changing it to `std::stable_sort` can pass the tests so I can submit a PR in plaidml/openvino if this `std::sort` is not for some specific purpose.

plaidml/bridge/openvino/ops/detection_output.cpp
haoyouab: Done. Use `const std::string&` now.

plaidml/bridge/openvino/ops/detection_output.cpp
haoyouab: Done.

plaidml/bridge/openvino/ops/detection_output.cpp
haoyouab: Fixed.

plaidml/bridge/openvino/ops/detection_output.cpp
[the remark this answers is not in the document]
haoyouab: > Part of this function appears fairly similar to the center_point_box parameter in the OpLib NMS.

Yes there are two types of box encoding in both functions, namely `corner` and `center`. In both functions they are doing pretty much the same thing but there are some differences, too. In terms of feasibility, I think merging these two functions is not complicated. There are a few differences we need to consider though.

- The different outputs of `NonMaxSuppression` and `DetectionOputput`. That is, the output of `NonMaxSuppression` is in the form of

```
[batch_index, class_index, box_score]
```
while the other one is
```
[batch_index, class_index, box_score, x_min, y_min, x_max, y_max]
```
If we merge `decodeBoxes` into OpLib then we should have two different types of output based on something like decode_mode. I'm not sure if this is acceptable/preferred.

- When there are 5 inputs of `DetectionOutput`, at [line 151](https://github.com/plaidml/plaidml/pull/1741/files#diff-1a9d0f648e4a5a282f62d552d60faaa7fb700552308e9d569c71d27eea535d75R151) we have to decode twice to get `decoded_boxes` (first with `ArmLocation` and then `Location`). Of course we can pass `ArmLocation` into OpLib NMS and do decoding twice inside of NMS but this seems to be a detection-output-specific parameter. It seems to me `decodeBoxes` should be a relatively standalone fuction.

What do you think?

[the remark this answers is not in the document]
tzerrell: Thanks for that explanation. In my opinion, it would be make sense to have 3 separate parameters determining:
* Corner vs. Center
* Output format
* Whether additional ARM prediction inputs are used

We wouldn't be using all 8 possible combinations of these 3 parameters -- at least not yet -- so it would be reasonable to just throw if an unsupported/untested choice of parameters is used. But in my opinion these 3 parameters are not intrinsically correlated, and so I think a design separating them is preferable unless there's some reason that separating them would be hard to implement.

Additionally, I think having a `decodeBoxes` helper function is the oplib is reasonable. I think at the moment this would only be used by NMS, and at the moment I can't think of other ops that would use it. But, it's not inherently linked to NMS, so it's reasonable to make it available in case someone else does have a use for it.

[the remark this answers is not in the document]
haoyouab: Basically done. Please check.

tzerrell (commented):
Thanks for moving these parts into the oplib!

In order to run CI you will need to merge the latest from `plaidml-v1`, we've made some adjustments to move more fully to CMake that are needed by CI. Please reach out to me if you encounter any problems running from an updated commit.

haoyouab:
Updated to the latest commit and the build is smooth. 

Just wonder why there's always something building even if I didn't change anything. The command is as below
```
ninja -C build-x86_64/Release/
```
If there's nothing changed at all, shouldn't it show something like `ninja: no work to do.`?

tzerrell:
> If there's nothing changed at all, shouldn't it show something like ninja: no work to do.?

I'm not certain, but I know the recent CMake changes included some functionality for triggering python tests; perhaps something in here isn't transparent to `ninja` and so it has to rebuild every time. Is this rebuild taking a long time for you?

tzerrell (approved):

haoyouab:
> I'm not certain, but I know the recent CMake changes included some functionality for triggering python tests; perhaps something in here isn't transparent to `ninja` and so it has to rebuild every time. Is this rebuild taking a long time for you?

I see. And no it just takes a few seconds.

flaub:
The rebuilds appear to be due to tblgen rules within LLVM. Since they run quickly, I haven't focused on resolving it. If it becomes too slow, for sure I'll try and get this fixed, but for now it seems pretty benign.

PR #1743, opened 2021-04-02
Add python support to cmake

flaub:


flaub:
This is a first step in replacing all bazel functionality with cmake.

earhart (dismissed):

PR #1748, opened 2021-04-08
Drop unmaintained conversions, dialects, passes, and targets

flaub:


earhart (approved):
All those moments will be lost in time, like tears in rain. Time to die.

PR #1749, opened 2021-04-09
Fix Scatter type check

tzerrell:
Fixes scatter failure with integers.


flaub (approved):

flaub (changes requested):
Since it takes 27 mins to run `check-smoke` on debug builds, I propose we do a few things:
1. leave the `check: core` on debug builds
2. add a lit test for integer cases of scatter so that this case gets picked up by `check-core`

flaub (approved):

PR #1763, opened 2021-04-26
Add PSROIPooling and interpolated gatherND

DingZhangIntel:


tzerrell (commented):
Apologies for my delay in review. I've added a few minor inline comments. But at a high level, I'm thinking about whether there is a reasonable way to access the 2nd input (`coords`) in all the ways we need it without forcing it to be constant. I'm also considering whether the very high dimension tensors can be avoided... if they helpfully represent information that would otherwise be hard to make tractable we should certainly use them, but I worry that that they could be a sign of unnecessary complexity. So I'm considering whether that complexity is necessary. I'll update further once I've given these more consideration.

plaidml/edsl/__init__.py
```diff
@@ -907,12 +913,13 @@ def gather(x,
         cube_coeff (float): The coefficient that controls the cubic interpolation.
         gather_mode (Enum): The type of gather op.
         batch_dims (int): The number of leading batch dimensions.
+        padding_mode (Enum): The type of padding only interpolation_mode is linear.
```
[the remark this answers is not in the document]
YangleiZouIntel: Thanks, fixed. Currently this padding mode is added to linear interpolation mode, I wonder if we need to add it to all mode (non-interpolation, nearest, linear, cubic mode) which I think is valuable for a few other ops.

plaidml/edsl/edsl.h
```diff
@@ -987,11 +988,17 @@ class gather {
...
+    return *this;
+  }
+
   ///
   /// Construct gather.
   ///
   Tensor build(edsl_source_location loc = edsl_source_location::current()) const {
-    std::vector<Tensor> args = {x_, y_, axis_, interpolation_mode_, nearest_mode_, cube_coeff_, mode_, batch_dims_};
+    std::vector<Tensor> args = {x_,          y_,    axis_,       interpolation_mode_, nearest_mode_,
+                                cube_coeff_, mode_, batch_dims_, padding_mode_};
```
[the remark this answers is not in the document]
YangleiZouIntel: Yes. I have put each tensor on a separate line.

plaidml/edsl/edsl.h
```diff
@@ -1702,7 +1710,8 @@ inline TensorVec layer(const std::string& op, const TensorVec& operands, const D
 
 inline Tensor layer(const std::string& op, const TensorVec& operands, const Dictionary& attrs,
                     const LayerBodySingleFn& fn, edsl_source_location loc = edsl_source_location::current()) {
-  return layer(op, operands, attrs, [&]() { return TensorVec{fn()}; }, loc)[0];
+  return layer(
+      op, operands, attrs, [&]() { return TensorVec{fn()}; }, loc)[0];
```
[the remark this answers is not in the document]
YangleiZouIntel: Yes, the format is a little wired. There is no semantic change here. 
However, I have tested clang-format-9.0.0, 9.0.1-12 and 10.0.0, all versions changed this line. Not sure if I missed anything thing.

YangleiZouIntel:
> Apologies for my delay in review. I've added a few minor inline comments. But at a high level, I'm thinking about whether there is a reasonable way to access the 2nd input (`coords`) in all the ways we need it without forcing it to be constant. I'm also considering whether the very high dimension tensors can be avoided... if they helpfully represent information that would otherwise be hard to make tractable we should certainly use them, but I worry that that they could be a sign of unnecessary complexity. So I'm considering whether that complexity is necessary. I'll update further once I've given these more consideration.

Thanks for the comments. Reimplemented this op without forcing the 2nd input to be constant. Using interpolated gatherND to avoid very high dimensional tensors and unnecessary gathers.

pmlc/conversion/tile_to_pxa/specials.cc
```diff
@@ -675,6 +702,39 @@ struct GatherOpConversion : public OpConversionPattern<tile::GatherOp> {
     auto valuePlusHalf = rewriter.create<mlir::AddFOp>(loc, value, half);
     return floorFPToSI(loc, rewriter, valuePlusHalf, integerType);
   }
+
+  Value processPaddingMode(Location loc, ConversionPatternRewriter &rewriter,
+                           Value result, Value idx, IndexBounds bounds,
+                           Type elementType, PaddingMode paddingMode) const {
+    switch (paddingMode) {
+    case PaddingMode::edge:
+      break;
```
tzerrell, line 711: How does this accomplish edge padding? Does interpolation work for gather with an out-of-bounds read by getting the closest value (i.e. at the edge)? That's not what I had thought but perhaps I'm mistaken.

DingZhangIntel: In previous version, the closest value will be read when the index is out-of-bounds, So it doesn't need any change when PaddingMode is edge. But in some operations, for example deformabel_convolution (https://github.com/plaidml/openvino/blob/plaidml/ngraph/core/reference/include/ngraph/runtime/reference/deformable_convolution.hpp, lines 157 to 160) zero will be read when the index is out-of-bounds. So, i added the zero_padding_mode to process this case.

YangleiZouIntel: Gather with an out-of-bounds read by getting the closest value (i.e. at the edge) is true for the linear interpolation (not true for cubic or nearest mode). The `paddingMode::edge` was the default behavior for interpolated gather (the same as ngraph ref implementation of interpolation). That is when the index is out-of-bounds, the data will be treated as if it was edge padded. 

For example, for linear interpolated gather with index at -0.3 for a 1D data with 2 element {data[0]=3; data[1]=4;}, the data will be treated as if the data[-1]=data[0]=3, so any linear interpolation  between -1 and 0 will be 3 (It looks like it gathers the nearest edge, but it is actually calculated by data[-0.3]=data[-1]*0.7+data[0]*0.3=3).

For cubic interpolated gather with index at -0.3 for a 1D data with 4 element {data[0]=3; data[1]=4;data[2]=3; data[3]=2;}
the result of data[-0.3] will be calculated by using four points, data[-1],data[0],data[1],data[2], since data[-1] is out-of-bounds, it will be set to data[-1]=data[0]=3, then use these four values to calculate data[-0.3] and it does not equal the edge value data[0].

As Ding mentioned, some ngraph ref implementation like deformableconvolution uses different out-of-bounds values than the interpolation op. They return 0 when the index is out of bounds. Ding added this function to return zero for linear interpolation when the index is out of bounds. I have applied it to all interpolation mode.

I think it is a little confusing to use the word `PaddingMode`, since in `PaddingMode::edge` we treat the data as if it was edge padded, in `PaddingMode::zero` we simply return zero when index is out of bounds (This is a different then treating the data as if it was zero padded). I think a better word for this mode would be `OutOfBoundsMode::data_edge_padding`  `OutOfBoundsMode::return_zero`

pmlc/conversion/tile_to_pxa/specials.cc
```diff
@@ -675,6 +702,39 @@ struct GatherOpConversion : public OpConversionPattern<tile::GatherOp> {
     auto valuePlusHalf = rewriter.create<mlir::AddFOp>(loc, value, half);
     return floorFPToSI(loc, rewriter, valuePlusHalf, integerType);
   }
+
+  Value processPaddingMode(Location loc, ConversionPatternRewriter &rewriter,
+                           Value result, Value idx, IndexBounds bounds,
+                           Type elementType, PaddingMode paddingMode) const {
```
YangleiZouIntel, line 708: The input `Value result` for this function is already the interpolated value treat the data as if it was edge padded. So if the `paddingMode` is edge, we can just return this value.

tzerrell (approved):

PR #1764, opened 2021-04-26
Fix DetectionOutput tests

tzerrell:
~The DetectionOutput tests are sporadically failing. We discussed this in the context of #1741 and I had thought this was fixed at the time, but clearly my understanding there was wrong. So, this PR disables those tests for now (to unblock CI), and meanwhile I'm looking into what happened in #1741 and related PRs to see whether we didn't actually land the fix, or if the fix was incomplete, or if there's some additional bug.~

This fixes a bug in DetectionOutput testing by updating the OpenVINO reference to a version that includes the bugfix.


jd-bartlett96 (dismissed):
LGTM

jd-bartlett96 (approved):

PR #1765, opened 2021-04-27
Fix memory allocation in padding of convolution inputs

jd-bartlett96:
Resolving padding memory allocation issues


tzerrell (commented):
This looks good to me. I am comfortable approving & landing this, but I know you discussed this work with Frank previously, so do you want to wait for his review as well?

pmlc/.clang-format
```diff
@@ -1,2 +1,2 @@
 BasedOnStyle: LLVM
-AlwaysBreakTemplateDeclarations: Yes
+AlwaysBreakTemplateDeclarations: true
```
flaub, line 2: Why is this needed? This seems to indicate you are using a different version of clang-format, we require version 9 at the moment.

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -851,40 +851,36 @@ struct CastOpConversion : public OpConversionPattern<tile::CastOp> {
...
     TileToPXATypeConverter typeConverter;
-    auto oldResultType = op.result().getType();
-    auto resultType =
-        typeConverter.convertType(oldResultType).cast<MemRefType>();
     auto operand = operands[0];
-    if (resultType == operand.getType()) {
+    if (op.result().getType() == operand.getType()) {
       rewriter.replaceOp(op, operand);
       return success();
     }
```
flaub, line 858: It seems like canonicalization should handle this case, I think we should probably just remove it.

jd-bartlett96: SGTM; yeah I was unable to produce anything that triggered that if statement so I agree.

pmlc/rt/boundscheck.cc
```diff
@@ -17,13 +17,13 @@ extern "C" void _mlir_ciface_plaidml_rt_bounds_check(intptr_t index,
...
-        "_mlir_ciface_plaidml_rt_bounds_check",
-        reinterpret_cast<void *>(_mlir_ciface_plaidml_rt_bounds_check));
-  }
-};
-static Registration reg;
-} // namespace
+namespace pmlc::rt {
+
+void registerBoundsCheck() {
+  using pmlc::rt::registerSymbol;
```
flaub, line 23: Why do a `using` here if we just use the name once?

pmlc/.clang-format
jd-bartlett96: I added this because I can't get the formatting to run/commit checks to pass without this change. My clang-format version is 6 so yes that's probably the issue. However, I was under the impression the cenv should set the correct version, is this not the case? (Or, is it supposed to be the case and failing for me for some reason?)

Regardless I will update my clang-format and undo this.

pmlc/rt/boundscheck.cc
jd-bartlett96: I was just following the same format used elsewhere (eg registerPrng) but happy to remove it

pmlc/.clang-format
flaub: `clang-format` is not specified in our `environment.yml`. I haven't been able to find a suitable package for conda to manage it. So right now, we have to manually install it.

pmlc/dialect/tile/transforms/pad_constraints.cc
```diff
@@ -120,7 +120,10 @@ void PadConstraintsPass::runOnFunction() {
     // Check if it's a block argument or unpack source, and if so add an IdentOp
     // to copy the value.
     if (def.isa<BlockArgument>() ||
-        dyn_cast_or_null<stdx::UnpackOp>(def.getDefiningOp())) {
+        isa_and_nonnull<stdx::UnpackOp>(def.getDefiningOp()) ||
+        isa_and_nonnull<tile::GatherOp>(def.getDefiningOp()) ||
+        isa_and_nonnull<tile::ScatterOp>(def.getDefiningOp()) ||
+        isa_and_nonnull<tile::PrngOp>(def.getDefiningOp())) {
```
flaub, line 126: You can say:
```
isa_and_nonnull<stdx::UnpackOp, tile::GatherOp, tile::ScatterOp, tile::PrngOp>(def.getDefiningOp())
```

pmlc/conversion/tile_to_pxa/tile_to_pxa.cc
```diff
@@ -851,40 +851,31 @@ struct CastOpConversion : public OpConversionPattern<tile::CastOp> {
                   ConversionPatternRewriter &rewriter) const override {
     auto loc = op.getLoc();
     TileToPXATypeConverter typeConverter;
```
flaub, line 853: Is this still used?

jd-bartlett96: Nope good catch; removed this.

flaub (approved):

PR #1766, opened 2021-04-28
cmake rules for examples

1sanz:


flaub:
I'm wondering why we have separate example tests from the edsl tests themselves?

plaidml/edsl/examples/example_tests.py
```diff
@@ -12,13 +12,13 @@
...
-from plaidml.edsl.examples.gemv import *
-from plaidml.edsl.examples.max_pool_1d import *
-from plaidml.edsl.examples.quantize import *
+from complex_conv_2d import *
+from conv_1d import *
+from conv_2d_dilated import *
+from gemm import *
+from gemv import *
+from max_pool_1d import *
+from quantize import *
```
flaub, line 21: We should try to avoid using `from xxx import *` in python. Also, we should use package names relative to the root, to avoid collisions.

tzerrell:
> I'm wondering why we have separate example tests from the edsl tests themselves?

My understanding was that tests in the documentation were going in example tests and other tests were going in edsl tests ... but it seems entirely reasonable to me to just merge everything into edsl tests instead if that's your preferred architecture.

plaidml/edsl/examples/CMakeLists.txt
```diff
@@ -19,7 +19,16 @@ pml_cc_test(
...
+  SRC 
+    complex_conv_2d.py
+    conv_1d.py
+    conv_2d_dilated.py
+    conv_2d_dilated.py
+    gemm.py
+    gemv.py
+    max_pool_1d.py
+    quantize.py
+    example_tests.py
```
flaub, line 31: I don't think `pml_py_test` supports multiple sources, I'm surprised this works.

plaidml/edsl/examples/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+    gemv.py
+    max_pool_1d.py
+    quantize.py
+)
+
+pml_py_test(
+  NAME examples_py_test
+  SRC 
+    example_tests.py
+  CHECKS
```
flaub, line 37: Need:

```
DEPS
  ::py
```

plaidml/edsl/examples/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+  DEPS
+    plaidml::testenv
+    pmlc::testing
+  CHECKS
+    core
+    smoke
+    test
+)
+
+pml_py_library(
```
flaub, line 20: This needs:

```
DEPS
  plaidml::py
```

plaidml/edsl/examples/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+pml_cc_test(
+  NAME examples_cc_test
```
flaub, line 2: Drop the `examples_` prefix

flaub (approved):

PR #1772, opened 2021-05-11
POC: XSMM/TPP integration

egeor:
Exemplary unary TPPs in xsmm dialect (f32 exp and relu)


cmake/third_party/xsmm.cmake
```diff
@@ -1,8 +1,7 @@
 message("Fetching xsmm")
 FetchContent_Declare(
   xsmm
-  URL      https://github.com/hfp/libxsmm/archive/592d72f2902fa49fda51a4221e83f7f978936a9d.zip
-  URL_HASH SHA256=2595c625e9e63099de8bea4876a7e5835f0c5f560049db6de31611da2539d448
+  URL      https://github.com/hfp/libxsmm/archive/refs/heads/master.zip
```
flaub, line 4: We should point to a specific commit instead of pointing to master (and we should also have a hash specified).

pmlc/dialect/xsmm/ir/ops.td
```diff
@@ -44,6 +44,50 @@ def XSMM_GemmInvokeF32Op : XSMM_Op<"gemm.invoke.f32"> {
...
+  let printer = ?;
+  let parser = ?;
+  let assemblyFormat = "$tile `,` $tileld attr-dict";
+}
+
+def XSMM_UnaryReluInvokeF32Op : XSMM_Op<"unary_relu.invoke.f32"> {
+  let summary = "invoke a previously queried LIBXSMM unary kernel";
+  let arguments = (ins I64:$ptr,
+    Arg<F32MemRef, "the input memref to load from", [MemRead]>:$i,
+    Arg<F32MemRef, "the output memref to store to", [MemWrite]>:$o,
```
flaub, line 82: Avoid using single-letter variable names (makes it hard to refactor, search, etc)

flaub:
@alheinecke @egeor I've updated this to use a single unary dispath/invoke runtime function. It uses tblgen enums and a mapping between MLIR types to `libxsmm_datatype`. I plan to do a follow-up PR which will add a relu primitive to the eDSL.

jbruestle (approved):

PR #1778, opened 2021-05-21
Update to latest MLIR

flaub:
Also improves the `check-${CHECK}` targets so that dependencies are properly defined. This ensures that, for instance, `libplaidml.so` is built before running python tests (which depend on `libplaidml.so`).

### DIFF
--- a/pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
+++ b/pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
@@ -25,6 +25,13 @@ namespace stdx = dialect::stdx;
 
 namespace {
 
+static void splitAffineMaps(AffineMap from, SmallVectorImpl<AffineMap> &into) {
+  for (AffineExpr expr : from.getResults()) {
+    into.push_back(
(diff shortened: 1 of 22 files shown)


flaub:
This draft PR will be ready for review when the corresponding PR in plaidml/llvm-project has landed and the pointer/hash has been bumped.

https://github.com/plaidml/llvm-project/pull/39

pmlc/conversion/pxa_to_affine/pxa_to_affine.cc
```diff
@@ -25,6 +25,13 @@ namespace stdx = dialect::stdx;
 
 namespace {
 
+static void splitAffineMaps(AffineMap from, SmallVectorImpl<AffineMap> &into) {
```
vinograd47, line 28: IMHO, this function can be moved to `pmlc/util/util.h`, since it is used in several places.

vinograd47 (commented):
LGTM, the remark is minor

jbruestle (approved):

PR #1780, opened 2021-05-21
Fixing SAME_UPPER/SAME_LOWER padding

jd-bartlett96:
Fixing rounding-related error in output size calculation for SAME_UPPER/SAME_LOWER autopad modes.

### DIFF
--- a/plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/pooling_smoke.cpp
+++ b/plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/pooling_smoke.cpp
@@ -20,11 +20,10 @@ const std::vector<InferenceEngine::Precision> netPrecisions = {
 const std::vector<std::vector<size_t>> kernels = {
     {3, 3},
     {3, 5},
-    {2, 2},
 };
 const std::vector<std::vector<size_t>> strides = {
     {1, 1},
-    {2, 2},
+    // {1, 2},
 };
 const std::vector<std::vector<size_t>> padBegins = {
     {0, 0},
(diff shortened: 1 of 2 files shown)


tzerrell (commented):
Did you also review what happened in the Valid or Explicit padding cases in situations where a nonzero ceil_term would change the output shape? I’m wondering whether the problem is specific to Same padding or if we are fundamentally misusing ceil_term in general and there’s a larger fix to be made.

jd-bartlett96:
> Did you also review what happened in the Valid or Explicit padding cases in situations where a nonzero ceil_term would change the output shape? I’m wondering whether the problem is specific to Same padding or if we are fundamentally misusing ceil_term in general and there’s a larger fix to be made.

As far as I can tell the Valid & Explicit paddings are fine; the ceil_term works as intended there. I believe the only error was using that term for the same upper/lower pads, which according to https://docs.openvinotoolkit.org/latest/openvino_docs_ops_pooling_MaxPool_1.html should not be affected by the rounding type (nothing is rounded). 

I should note that there are "failures" for certain test inputs for Valid & Explicit paddings, which do not appear to be real issues, or are problems with the OV side. Specifically, inputs with strides>1 cause OV to fail to run the test case with the message 
"The end corner is out of bounds at axis 3"
which appears to be erroneous & is already noted in the Avg_Pool test cases. 

Additionally, for some test inputs with Explicit Padding, the test fails with a message like 
"Relative comparison of values expected: -3.4028234663852886e+38 and actual: -inf at index 30 with threshold 0.0099999997764825821 failed"
which I assume is not actually problematic (I believe this is related to the op filling in pads used in the pooling with -inf, where OV appears to use some very-low number instead of true -inf).

tzerrell (approved):
Great, thanks for confirming. LGTM!

PR #1783, opened 2021-05-27
Change clamp in StridedSlice and open some tests 

XinWangIntel:


plaidml/bridge/openvino/ops/strided_slice.cpp
```diff
@@ -71,11 +71,14 @@ void registerStridedSlice() {
         continue;
       }
 
+      // The abs bounds in different direction are not equal sine SliceOp uses 0 vs -1 as begin
+      // The slice region is a half-open interval [...), out-of-bounds values are clamped in differnt way
+      // TODO: Check sign of "steps" when zero dimension is allowed in openvino
```
tzerrell, line 76: I think if we're in a size-0 dim case from the OV side, we will need to make changes elsewhere to patch that as a size-1 dim, and this code can work as normal; so I think changes will be needed elsewhere instead. I suppose it doesn't hurt to leave the TODO here just in case though.

tzerrell (approved):

PR #1784, opened 2021-05-27
Make stdx/llvm lowering type-agnostic

jd-bartlett96:


tzerrell (commented):
Is this sufficient to fix the failed lowerings you were hitting? Or does this get you one step further and then fail (e.g. perhaps an intrinsic does not exist for `resType`)?

jd-bartlett96:
> Is this sufficient to fix the failed lowerings you were hitting? Or does this get you one step further and then fail (e.g. perhaps an intrinsic does not exist for `resType`)?

This resolves the failures I was seeing, which were all trying to lower a F16 type, which does appear to have an existing intrinsic. Should a check also be included to check that the intrinsic exists and cast into/out of a working type if it does not?

tzerrell:
> > Is this sufficient to fix the failed lowerings you were hitting? Or does this get you one step further and then fail (e.g. perhaps an intrinsic does not exist for `resType`)?
> 
> This resolves the failures I was seeing, which were all trying to lower a F16 type, which does appear to have an existing intrinsic. Should a check also be included to check that the intrinsic exists and cast into/out of a working type if it does not?

Great! Adding such a check would be ideal, but considering the previous crash it would still be an improvement to land this as-is if there are difficulties with that.

flaub:
Adding a lit test might also be good

pmlc/conversion/stdx_to_llvm/tests/elementwise.mlir
```diff
@@ -0,0 +1,25 @@
+// RUN: pmlc-opt -convert-stdx-to-llvm -canonicalize %s | FileCheck %s
+
+module {
```
flaub, line 3: You can drop the enclosing `module`

tzerrell:
I added lit tests; this needs a review from someone other than me.

One particular concern: In the lit test for `pow` on `i32` data, it lowered to `powf` -- I don't know enough about LLVM to know if this is correct or buggy, but the lit test as-written is expecting `powf` to be produced.

flaub:
I'm also concerned about that. I think a further lowering would fail or worse crash if we don't have proper data types. 

FYI, upstream now has the pass I think we want: https://github.com/plaidml/llvm-project/blob/develop/mlir/include/mlir/Conversion/MathToLibm/MathToLibm.h

tzerrell (changes requested):
Thanks for the reference! Marking as "Request changes" as this is likely broken in the integer case.

PR #1787, opened 2021-05-31
Update LLVM to latest MLIR

vinograd47:
* Use `CHECK-DAG` for constants checks.
* Remove `EDSC` usage.

Related PR: https://github.com/plaidml/llvm-project/pull/40

### DIFF
--- a/cmake/third_party/llvm-project.cmake
+++ b/cmake/third_party/llvm-project.cmake
@@ -32,8 +32,8 @@ else()
   message("Fetching LLVM")
   FetchContent_Declare(
     llvm-project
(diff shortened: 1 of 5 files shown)


flaub (approved):

vinograd47:
@flaub Just interesting, how do you get `URL_HASH` value? Is there some GitHub API or is it visible somewhere?

flaub:
I know of 2 ways:
1. run `shasum` manually on the tarball/zip file (before it's expanded)
2. use a known 'wrong' hash and let cmake eventually complain about the hash not matching. At that time, it will print the expected vs current hash.

PR #1789, opened 2021-06-04
Lower relu -> TPP

flaub:
* Add special case to oplib relu impl to use new eDSL primitive.
* Add lowering: `tile.relu -> stdx.relu`
* Add `-x86-tpp-patterns` pass to convert `stdx.relu` to `pxa.generic` when possible.
* Add fallback lowering for `stdx.relu -> std.cmpf, std.select` when TPP patterns fail.

### DIFF
--- a/plaidml/op/lib/ops.cc
+++ b/plaidml/op/lib/ops.cc
@@ -2554,19 +2554,18 @@ Value relu(const Value& value) {
   if (args.size() != 4) {
     throw std::runtime_error("relu expects 4 arguments");
   }
-  auto I = args[0].as_tensor();
-  auto alpha = args[1];
-  auto max_value = args[2];
-  auto threshold = args[3].as_float();
-  Tensor A;
-  if (alpha.is_none()) {
-    A = Tensor(0.0);
-  } else {
-    A = alpha.as_tensor();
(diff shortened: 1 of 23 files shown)


plaidml/op/lib/ops.cc
```diff
@@ -2554,19 +2554,18 @@ Value relu(const Value& value) {
...
+  Value threshold = args[3];
+  double T = threshold.is_none() ? 0.0 : threshold.as_float();
+  if ((alpha.is_none() || (alpha.is_float() && alpha.as_float() == 0.0)) && max_value.is_none() && T == 0.0) {
+    return Value{edsl::relu(I)};
+  }
+  Tensor A = alpha.is_none() ? Tensor(0.0) : alpha.as_tensor();
+  Tensor O = select(I < T, edsl::cast(A * (I - T), I.dtype()), I);
   if (!max_value.is_none()) {
-    auto M = cast(max_value.as_tensor(), I.dtype());
+    Tensor M = cast(max_value.as_tensor(), I.dtype());
```
flaub, line 2568: @tzerrell FYI, adding special case to oplib. Should be transparent but I made a few tweaks to the oplib::relu interface.

jbruestle (approved):

PR #1806, opened 2021-06-22
Add OpenVino Deformable PSROI Pooling Op

LiyangLingIntel:
Implemented OpenVino Deformable PSROI Pooling Op.

One thing needs to mension is, 
in [DeformablePSROIPooling](https://docs.openvinotoolkit.org/latest/openvino_docs_ops_detection_DeformablePSROIPooling_1.html) OpenVino Spec, the third input is 4D input tensor of type T and shape `[NUM_ROIS, 2*NUM_CLASSES, group_size, group_size]` with transformation values , while the description of `part_size` is the size of H and W dimensions of the third input (offsets). 
After referring to [openvino reference implementation](https://github.com/plaidml/openvino/blob/plaidml/ngraph/core/reference/include/ngraph/runtime/reference/deformable_psroi_pooling.hpp#L120), the third tensor shape should be `[NUM_ROIS, 2*NUM_CLASSES, part_size, part_size]`.


plaidml/bridge/openvino/ops/deformable_psroi_pooling.cpp
```diff
@@ -0,0 +1,161 @@
...
+    IE_ASSERT(channels_in == output_dim * group_size * group_size);
+
+    auto rois_shape = rois.compute_shape().sizes();
+    auto num_rois = static_cast<int>(rois_shape[0]);
+    auto dim_one = edsl::TensorDim(1);
+
+    rois = edsl::round(rois);
+    auto roi_idx = edsl::index({edsl::TensorDim(2)}, 0) + 1;
+    // Left top ROI corner - (x1, y1)
+    edsl::Tensor roi_lt = edsl::cast(edsl::gather(rois, roi_idx).axis(1), DType::FLOAT32);
```
tzerrell, line 46: Much as in #1801, I think edsl::gather applied to edsl::index may be less efficient than op::slice. The use here seems fairly small so we could land this anyway, but this is a performance concern to consider in the future. (There is a larger gatherND + index case later which is more substantial, but I think there likely isn't any way to rewrite that using slice given how the index inputs are combined.)

tzerrell (commented):
Great! Generally looks good, a couple of in-line comments

plaidml/bridge/openvino/tests/functional/shared_tests_instances/single_layer_tests/deformable_psroi_pooling.cpp
```diff
@@ -0,0 +1,52 @@
...
+    ::testing::Values(std::vector<size_t>{3, 8, 16, 16}),  // data input shape
+    ::testing::Values(std::vector<size_t>{10, 5}),         // rois input shape
+    // Empty offsets shape means test without optional third input
+    ::testing::Values(std::vector<size_t>{}),                                                // offsets input shape
+    ::testing::Values(2),                                                                    // output_dim
+    ::testing::Values(2),                                                                    // group_size
+    ::testing::ValuesIn(std::vector<float>{1.0, 0.5, 0.0625}),                               // spatial scale
+    ::testing::ValuesIn(std::vector<std::vector<int64_t>>{{1, 1}, {2, 2}, {3, 3}, {2, 3}}),  // spatial_bins_x_y
+    ::testing::ValuesIn(std::vector<float>{0.0, 0.01, 0.5}),                                 // trans_std
+    ::testing::Values(2));                                                                   // part_size
```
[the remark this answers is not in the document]
LiyangLingIntel: Added

plaidml/bridge/openvino/ops/deformable_psroi_pooling.cpp
LiyangLingIntel: Yes, so now, I only replaced `edsl::gather` + continuous `edsl::index` result with `op::slice` and also updated this in `detection_output` and `region_yolo` in #1801. 

For those large `Gather`/`GatherND` cases, I think we can reduce dimensions of gather indices using `scanOp` when it's landed(currently using c++ for loop will significantly increase compilation cost when tensor size is large).

tzerrell (approved):

PR #1808, opened 2021-06-24
Replace PxaGemmOp with PxaGenericOp

flaub:


jbruestle (approved):

vinograd47:
@flaub It looks like this change introduce circular dependency between `pmlc::dialect::pxa::ir` and `pmlc::dialect::pxa::analysis` due to `StrideRange` class usage. In local builds I've faced with linker issues, it complains about unresolved `StrideRange` methods in `ops.cc` from `pxa::ir`. Could you please take a look at this?

PR #1810, opened 2021-06-25
Add openvino missing tests 

LiyangLingIntel:
Add openvino missing tests for `broadcast`, `fake_quantize`, `normalized_l2`, `pad`.

Fix which fails test.


tzerrell (approved):
LGTM, thanks!

PR #1815, opened 2021-07-02
Replace remarks with logging

flaub:


jbruestle (commented):
So in general, logging is more consistent with other PlaidML code, but one nice thing about remarks is that they localize the error to the specific IR (which would be particularly useful if we had lit tests for all the various fusion failure case).  If there a specific reason you moved things to logs?

earhart:
+1 -- if anything, I'd rather see us move to using MLIR's logging infrastructure everywhere, and eliminate easylogging.

flaub:
@jbruestle When compiling a full network (like resnet50), there's a lot of spam if we use remarks. Also, the remarks get dumped at a time that differs from the logs, and so it's difficult to line up other adjacent log entries. If we need context for a log entry, we can always use `debugString()`.

PR #1817, opened 2021-07-05
Add AlgebraicOpt pass

flaub:
First pattern to be implemented does the following.

From:
```
%zero = tile.constant(0.0 : f64)
%0 = tile.contract add, mul, %zero, %I, %K
%1 = tile.add %0, %B
%2 = tile.relu %1
```

Into:
```
%0 = tile.contract add, mul, %B, %I, %K
%2 = tile.relu %0
```

### DIFF
--- a/pmlc/dialect/tile/transforms/algebraic_opt.cc
+++ b/pmlc/dialect/tile/transforms/algebraic_opt.cc
@@ -1,0 +1,79 @@
+// Copyright 2020, Intel Corporation
+
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/Matchers.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Support/DebugStringHelper.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#include "pmlc/dialect/tile/ir/ops.h"
+#include "pmlc/dialect/tile/transforms/pass_detail.h"
+#include "pmlc/dialect/tile/transforms/passes.h"
+#include "pmlc/util/logging.h"
+
+using namespace mlir; // NOLINT
+
+namespace pmlc::dialect::tile {
+
+namespace {
+
+// From:
+// %zero = tile.constant(0.0 : f64)
+// %0 = tile.contract add, mul, %zero, %I, %K
+// %1 = tile.add %0, %B
+// %2 = tile.relu %1
+// Into:
+// %0 = tile.contract add, mul, %B, %I, %K
+// %2 = tile.relu %0
+struct AddInitPattern final : public OpRewritePattern<AddOp> {
+  using OpRewritePattern<AddOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(AddOp op, PatternRewriter &rewriter) const {
+    if (ContractionOp contractOp =
+            dyn_cast_or_null<ContractionOp>(op.lhs().getDefiningOp())) {
+      if (!contractOp->hasOneUse())
+        return failure();
+      if (isa_and_nonnull<ContractionOp>(op.rhs().getDefiningOp()))
+        return failure();
+      if (op.result().getType() != contractOp.result().getType())
+        return failure();
+
+      FloatAttr init;
+      if (!matchPattern(contractOp.init(), m_Constant(&init)))
+        return failure();
+      if (init.getValueAsDouble() != 0.0)
+        return failure();
+
+      contractOp.setOperand(0, op.rhs());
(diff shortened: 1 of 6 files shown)


pmlc/dialect/tile/transforms/algebraic_opt.cc
```diff
@@ -0,0 +1,79 @@
...
+// %2 = tile.relu %0
+struct AddInitPattern final : public OpRewritePattern<AddOp> {
+  using OpRewritePattern<AddOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(AddOp op, PatternRewriter &rewriter) const {
+    if (ContractionOp contractOp =
+            dyn_cast_or_null<ContractionOp>(op.lhs().getDefiningOp())) {
+      if (!contractOp->hasOneUse())
+        return failure();
+      if (isa_and_nonnull<ContractionOp>(op.rhs().getDefiningOp()))
```
jbruestle, line 37: I'm not sure I understand why RHS can't be a contraction.

pmlc/dialect/tile/transforms/algebraic_opt.cc
```diff
@@ -0,0 +1,79 @@
...
+// %1 = tile.add %0, %B
+// %2 = tile.relu %1
+// Into:
+// %0 = tile.contract add, mul, %B, %I, %K
+// %2 = tile.relu %0
+struct AddInitPattern final : public OpRewritePattern<AddOp> {
+  using OpRewritePattern<AddOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(AddOp op, PatternRewriter &rewriter) const {
+    if (ContractionOp contractOp =
```
jbruestle, line 33: I think we also need to verify that the contract op we are rewriting is has 'add' as it's AggOp.

jbruestle (commented):
See two inline comments.

pmlc/dialect/tile/transforms/algebraic_opt.cc
flaub: There's probably a better way to handle this case, but the case I was running into was `concat`. 

```
%0 = tile.contract %zero, ...
%1 = tile.contract %zero, ...
%2 = tile.add %0, %1
```

The result would be clearly broken:
```
%0 = tile.contract, %1, ...
%1 = tile.contract, %zero, ...
```

pmlc/dialect/tile/transforms/algebraic_opt.cc
flaub: OK, but shouldn't `assign` be fine too?

pmlc/dialect/tile/transforms/algebraic_opt.cc
```diff
@@ -0,0 +1,96 @@
...
+// %2 = tile.relu %1
+// Into:
+// %0 = tile.contract add, mul, %B, %I, %K
+// %2 = tile.relu %0
+struct AddInitPattern final : public OpRewritePattern<AddOp> {
+  using OpRewritePattern<AddOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(AddOp op, PatternRewriter &rewriter) const {
+    return failure(
+        failed(matchAndRewritePermutation(op, rewriter, op.lhs(), op.rhs())) ||
```
jbruestle, line 34: I think if lhs,rhs fails but rhs,lhs succeeds, we want the result to be success, this would make it a failure I believe (and actually not vheck rhs,lhs I think due to short circuit behavior).  I think you just need &&.

pmlc/dialect/tile/transforms/algebraic_opt.cc
```diff
@@ -0,0 +1,96 @@
...
+        failed(matchAndRewritePermutation(op, rewriter, op.rhs(), op.lhs())));
+  }
+
+  LogicalResult matchAndRewritePermutation(AddOp op, PatternRewriter &rewriter,
+                                           Value thisOperand,
+                                           Value otherOperand) const {
+    if (ContractionOp contractOp =
+            dyn_cast_or_null<ContractionOp>(thisOperand.getDefiningOp())) {
```
jbruestle, line 42: Nit:  Seems like we should cast to contractOp and early return failure if it's null rather than using if just to avoid extra indention and follow LLVM style.

jbruestle (commented):
One bug, one nit.

pmlc/dialect/tile/transforms/algebraic_opt.cc
flaub: I inverted this to check for success

jbruestle (approved):
Looks good!

PR #1841, opened 2021-07-22
Add PLAIDML_NEW_PIPELINE flag

flaub:


brightlaboratory:
@flaub : The difference between the "NEW" and the default pass pipelines seems to be in the placement of the fusion pass -- it is called just after tpp pass in the "NEW" order while it is called later in the original pass order.

A couple of suggestions:
1) Since the placement of `createStencilTppGemmPass` remains unchanged, `if (util::getEnvVar("PLAIDML_NEW_PIPELINE") == "1")` should be moved after calling the stenciling pass.

2) `PLAIDML_NEW_PIPELINE` could be changed to something like - `PLAIDML_EARLY_FUSION` (so the name of the environment variable itself conveys something about what's different about the new pipeline). Potentially, the same flag could be used to control when the `fusion` pass is called: the `createTileAccumulatePass` and `createCPUThreadPass` pass invocations could be left outside of the `if` conditional.

jbruestle (approved):

PR #1848, opened 2021-08-02
Edge Padding Op

mwyi:


plaidml/op/op_test.cc
```diff
@@ -382,6 +382,38 @@ TEST_F(OpTest, ExplicitPaddingNoOp) {
   checkExact(program, {I_input}, {O_output});
 }
 
+TEST_F(OpTest, ExplicitPaddingEdge) {
+  auto I = Placeholder(DType::FLOAT32, {1, 2, 3}, "A");
+  auto O = op::explicit_padding(I, {1, 2, 1}, {1, 2, 1}).mode(op::PadMode::EDGE);
```
tzerrell: Change one of these padding parameters to e.g. `{2, 1, 1}` to also test:

1.  That we aren't swapping lo and hi pads
2. That we aren't applying padding in reverse order (which a palindrome for both doesn't test)

plaidml/op/lib/ops.cc
```diff
@@ -1642,7 +1642,34 @@ Value explicit_padding(const Value& value) {
...
+      std::vector<TensorIndex> Lo_idxs(I.rank());
+      std::vector<TensorIndex> Hi_idxs(I.rank());
+      std::vector<TensorDim> Lo_dims(I.rank());
+      std::vector<TensorDim> Hi_dims(I.rank());
+
+      for (size_t i = 0; i < I.rank(); ++i) {
+        std::vector<TensorDim> O_dims(I.rank());
+        O.bind_dims(O_dims);
+
+        std::vector<TensorIndex> lo_idxs = Lo_idxs;
```
tzerrell: I find having both `lo_idxs` and `Lo_idxs` as variable names confusing. Perhaps rename as `src_lo_idxs` and `dst_lo_idxs` respectively?

tzerrell (commented):
Looks good! See inline comments.

tzerrell (approved):

PR #1850, opened 2021-08-03
Conversion from Linalg to PXA

shiyao00:
This is a smoke implementation for almost all Linalg ops. The only op that is not implemented is tiled_loop, which is usually not directly used by the original network. We need more test cases for corner cases.


pmlc/conversion/linalg_to_pxa/utils.cc
```diff
@@ -0,0 +1,77 @@
...
+
+  // Some original ops use outputs as operands, and some ops return outputs
+  if (outputs.empty()) {
+    // For the ops only return outputs, we have to initialize a tensor for its
+    // output.
+    for (auto outputType : outputTypes) {
+      auto shapedType = outputType.cast<ShapedType>();
+      auto outputShape = shapedType.getShape();
+      auto elemType = shapedType.getElementType();
+      auto initOp = builder.create<InitTensorOp>(builder.getUnknownLoc(),
```
flaub: We should try to avoid using the unknown location. If you pass in the original op, we should use its location.

pmlc/conversion/linalg_to_pxa/linalg_to_pxa.cc
```diff
@@ -0,0 +1,503 @@
...
+  Operation *relatedOp;
+};
+
+struct ConvOpConversion : public OpConversionPattern<ConvOp> {
+  using OpConversionPattern<ConvOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(ConvOp op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const final {
+    op.emitError("The program should not contain linalg.conv");
```
jbruestle: For all of the cases where we are adding the 'Generalization' patterns (conv, padding, etc), we should probably just remove these patterns.  Since those ops are not in the 'legal' set, if they are not generalized, the pass will fail with a nice error saying that conv is not a legal op.

pmlc/conversion/linalg_to_pxa/pooling.cc
```diff
@@ -0,0 +1,142 @@
...
+    // Do not replace the use of the output in the new generic op
+    op.output().replaceUsesWithIf(genericOp.getResult(0), [&](OpOperand &use) {
+      return genericOp.getOperation() != use.getOwner();
+    });
+    rewriter.eraseOp(op);
+    return success();
+  }
+};
+
+void populateLinalgPoolingOpGeneralizationPatterns(
```
jbruestle, line 127: This is a great approach (generalize all we can and then lower generic ops to PXA).  That said, I would imagine we could probably upstream these patterns to be part of the upstream linalg generalization passes.

jbruestle (commented):
Looks great!  Looks like there are some type bugs in your asserts according to debug build, and some inline comments.

pmlc/conversion/linalg_to_pxa/test/conv.mlir
```diff
@@ -0,0 +1,20 @@
...
+  func @test_conv(%arg0: memref<128x128x3xf32, offset: ?, strides: [?, ?, 1]>, %arg1: memref<3x3x3xf32, offset: ?, strides: [?, ?, 1]>, %arg2: memref<?x?x?xf32, offset: ?, strides: [?, ?, 1]>) {
+    linalg.conv(%arg0, %arg1, %arg2) {strides = [2]}: memref<128x128x3xf32, offset: ?, strides: [?, ?, 1]>, memref<3x3x3xf32, offset: ?, strides: [?, ?, 1]>, memref<?x?x?xf32, offset: ?, strides: [?, ?, 1]>
+    return
+  }
+}
+
+// CHECK: #map = affine_map<(d0, d1, d2)[s0, s1, s2] -> (d0 * s1 + s0 + d1 * s2 + d2)>
+// CHECK-LABEL: func @test_conv
+// CHECK-SAME: (%[[arg0:.*]]: memref<128x128x3xf32, #map>, %[[arg1:.*]]: memref<3x3x3xf32, #map>, %[[arg2:.*]]: memref<?x?x?xf32, #map>)
+// CHECK: affine.parallel (%[[arg3:.*]], %[[arg4:.*]], %[[arg5:.*]], %[[arg6:.*]], %[[arg7:.*]]) = (0, 0, 0, 0, 0) to (3, -1, 3, 128, 128) reduce ("addf")
```
flaub: Is it legal for the bounds to go from `0` to `-1`? That seems odd to me.

pmlc/conversion/linalg_to_pxa/test/conv.mlir
```diff
@@ -0,0 +1,20 @@
+// RUN: pmlc-opt -convert-linalg-to-pxa %s | FileCheck %s
+
+module  {
+  func @test_conv(%arg0: memref<128x128x3xf32, offset: ?, strides: [?, ?, 1]>, %arg1: memref<3x3x3xf32, offset: ?, strides: [?, ?, 1]>, %arg2: memref<?x?x?xf32, offset: ?, strides: [?, ?, 1]>) {
+    linalg.conv(%arg0, %arg1, %arg2) {strides = [2]}: memref<128x128x3xf32, offset: ?, strides: [?, ?, 1]>, memref<3x3x3xf32, offset: ?, strides: [?, ?, 1]>, memref<?x?x?xf32, offset: ?, strides: [?, ?, 1]>
```
flaub: Are we required to use `?` for offset and strides? Can we make this work when we have no layout?

shiyao00: I just copied the test case from Linalg's test cases. I've removed ? in the case.

pmlc/conversion/linalg_to_pxa/test/conv.mlir
shiyao00: I think that's because of ? which was removed.

pmlc/conversion/linalg_to_pxa/pooling.cc
shiyao00: Will do when I have time. There are a lot of changes.

pmlc/conversion/linalg_to_pxa/linalg_to_pxa.cc
shiyao00: Removed these rewriters, and added them in the illegal op list.

pmlc/conversion/linalg_to_pxa/utils.cc
shiyao00: Fixed.

jbruestle (approved):

pmlc/conversion/linalg_to_pxa/passes.td
```diff
@@ -0,0 +1,20 @@
...
+#define __PMLC_CONVERSION_LINALG_TO_PXA_PASSES__
+
+include "mlir/Pass/PassBase.td"
+
+def LowerLinalgToPXA : Pass<"convert-linalg-to-pxa", "mlir::ModuleOp"> {
+  let summary = "Convert Linalg dialect to PXA dialect";
+  let constructor = "pmlc::conversion::linalg_to_pxa::createLowerLinalgToPXAPass()";
+  let dependentDialects = [
+    "mlir::AffineDialect",
+    "mlir::linalg::LinalgDialect",
```
flaub, line 11: Since this is the input dialect, I don't think we need to specify it in the dependent dialects.

pmlc/conversion/linalg_to_pxa/linalg_to_pxa.cc
```diff
@@ -0,0 +1,450 @@
...
+};
+
+// Most of Linalg ops are converted to GenericOp first.
+struct GenericOpConversion : public OpConversionPattern<GenericOp> {
+  using OpConversionPattern<GenericOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(GenericOp op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const final {
+    GenericOpAdaptor adaptor(operands, op.getOperation()->getAttrDictionary());
```
flaub, line 182: `op.getOperation()->` can always be replaced with `op->`

PR #1856, opened 2021-08-17
Improve TLS support

flaub:
Fix issue found when trying to run keras predict in separate thread(s).


tzerrell (commented):
Looks reasonable to me, but I would like someone else's eyes on threading / logging in Python / the FFI (e.g. @jbruestle's)

PR #1859, opened 2021-08-26
Connect ReturnOp to output arguments instead of allocated memref

shiyao00:


pmlc/conversion/linalg_to_pxa/linalg_to_pxa.cc
```diff
@@ -400,7 +423,12 @@ struct YieldOpConversion : public OpConversionPattern<linalg::YieldOp> {
   LogicalResult
   matchAndRewrite(linalg::YieldOp op, ArrayRef<Value> operands,
                   ConversionPatternRewriter &rewriter) const final {
-    rewriter.replaceOpWithNewOp<AffineYieldOp>(op, operands);
+    // We should put AffineYieldOp after the replaced linalg::YieldOp.
+    // Otherwise, the use of Block::getTerminator() may return the to-be-removed
+    // linalg::YieldOp. So we do not use replaceOp or replaceOpWithNewOp here.
+    rewriter.setInsertionPointAfter(op);
+    auto affineYieldOp = rewriter.create<AffineYieldOp>(op.getLoc(), operands);
+    rewriter.eraseOp(op);
```
flaub: I'm not sure I understand what's happening here.

shiyao00: Like Tile->PXA, we add the new arguments for outputs in the function prototype. So we need to replace the allocated memref with the new output function argument finally. In ReturnOp conversion, we use pxa::getIndirectDef(operand) in which Block::getTerminator() is used to get the AffineYieldOp in a loop body.

If we directly replace linalg::YieldOp with AffineYieldOp here, AffineYieldOp is before linalg::YieldOp when linalg::YieldOp is not removed (to-be-removed). Later, before linalg::YieldOp is really removed, Block::getTerminator() invoked by  pxa::getIndirectDef(operand) returns the last op in block, i,e., linalg::YieldOp, instead of AffineYieldOp. Then the cast<AffineYieldOp> in pxa::getIndirectDef(operand)  crashes.

So we need to insert AffineYieldOp AFTER linalg::YieldOp and erase linalg::YieldOp then. This is a trick to avoid the above problem.

flaub: I'm concerned that this introduces a coupling between different conversion patterns. Could we replace the yield op as part of the GenericOpConversion? An argument can be made that conversion of the generic op into an affine parallel is the right place for the linalg yield op to be converted.

pmlc/conversion/linalg_to_pxa/linalg_to_pxa.cc
```diff
@@ -338,10 +339,13 @@ struct GenericOpConversion : public OpConversionPattern<GenericOp> {
       op->emitError("No linalg.yield in generic op.");
     }
 
-    for (unsigned i = 0; i < op.getNumResults(); ++i) {
-      op.getResult(i).replaceAllUsesWith(forOp.getResult(i));
+    if (op.getNumResults() == forOp.getNumResults()) {
+      rewriter.replaceOp(op, forOp.getResults());
+    } else {
+      // For some original ops such as conv and copy, they don't return
+      // anything. So we erase the original op directly.
```
flaub, line 346: `linalg.conv` and `linalg.copy` cannot be lowered to PXA because they are already in 'bufferized' form. In other words, we can only lower `linalg` ops that have value-like semantics.

flaub (changes requested):
I was able to come up with a solution for this without requiring the change to YieldOp. Also I don't think the ReturnOpConversion will be guaranteed to be run because there's no rule that forces it. I'll submit my solution when I've got some other parts worked out.

shiyao00:
> I was able to come up with a solution for this without requiring the change to YieldOp. Also I don't think the ReturnOpConversion will be guaranteed to be run because there's no rule that forces it. I'll submit my solution when I've got some other parts worked out.

Please submit any alternative approach. If it works, we could close this MR.

flaub (approved):

flaub:
LGTM

PR #1868, opened 2021-09-16
Verify the path tile->linalg->pxa and fix bugs

shiyao00:


flaub (approved):
I think allowing signless in typeScore may introduce another bug, but I think this is fine for now. I can try and fix it a different way later.

PR #1874, opened 2021-09-20
use tensor.extract_slice

shiyao00:


shiyao00:
Resolved by another PR.

PR #1876, opened 2021-09-20
Add mlir-hlo to external llvm targets

brianretford:
Rolls llvm forward to the version specified in the  (newly added) mlir-hlo vendor submodule.


environment.yml
```diff
@@ -3,7 +3,7 @@ channels:
 dependencies:
   - ccache=4.1
   - cffi=1.14.4
-  - cmake=3.13.4
+  - cmake=3.14
```
flaub: This breaks openvino and other potential integaration.

environment.yml
```diff
@@ -12,6 +12,7 @@ dependencies:
   - numpy=1.18.5
   - pip=21.0.1
   - pre_commit=2.11.1
+  - pybind11
```
flaub: Is this needed?

flaub (dismissed):

environment.yml
brianretford: reverted, not sure why I bumped it forward anymore, everything works as is

environment.yml
brianretford: not now that I'm not building the mlir/mhlo python bindings, removed

flaub (approved):

PR #1877, opened 2021-09-20
Fix return op connection

shiyao00:


flaub (changes requested):
See my other PR. We need to copy the input buffer to the output buffer.

PR #1879, opened 2021-09-21
Fix memref.get_global in Linalg->PXA

shiyao00:


pmlc/conversion/linalg_to_pxa/linalg_to_pxa.cc
```diff
@@ -431,35 +431,45 @@ struct LowerLinalgToPXAPass
     }
   }
 
+  // Copy the input buffer to the output buffer.
+  AffineParallelOp copyBuffer(OpBuilder &builder, Location loc, Value input,
```
flaub: Let's mark this as `static`

pmlc/conversion/linalg_to_pxa/linalg_to_pxa.cc
```diff
@@ -431,35 +431,45 @@ struct LowerLinalgToPXAPass
...
-          auto loadOp =
-              bodyBuilder.create<pxa::PxaLoadOp>(loc, def, identityMap, idxs);
-          auto reduceOp = bodyBuilder.create<pxa::PxaReduceOp>(
-              loc, AtomicRMWKind::assign, loadOp, outputArg, identityMap, idxs);
-          bodyBuilder.create<AffineYieldOp>(loc, reduceOp.result());
+          copyBuffer(builder, loc, def, outputArg, context);
           operand.set(outputArg);
+        } else if (isa<memref::GetGlobalOp>(def.getDefiningOp())) {
+          auto forOp = copyBuffer(builder, loc, def, outputArg, context);
+          operand.set(forOp.getResult(0));
```
flaub, line 474: We should set the operand to `forOp.getResult(0)` in both cases (not just for `memref::GetGlobalOp`).

pmlc/conversion/linalg_to_pxa/linalg_to_pxa.cc
```diff
@@ -431,35 +431,45 @@ struct LowerLinalgToPXAPass
...
     unsigned argNumber =
         funcOp.getType().getNumInputs() - returnOp.getNumOperands();
     for (OpOperand &operand : returnOp->getOpOperands()) {
       // Find very initial allocation of memref
       Value def = pxa::getIndirectDef(operand.get());
       BlockArgument outputArg = funcOp.getBody().getArgument(argNumber++);
       if (def != outputArg) {
+        MLIRContext *context = &getContext();
+        Location loc = returnOp->getLoc();
+        ImplicitLocOpBuilder builder(loc, returnOp);
```
flaub, line 470: These were hoisted because we don't need to redo this work for each operand of the return.

flaub: Also, it's nice if the order of the copies are in the same order as the operands.

pmlc/conversion/linalg_to_pxa/test/shape.mlir
```diff
@@ -0,0 +1,15 @@
...
+}
+
+// CHECK-LABEL: func @main
+// CHECK-SAME: (%[[arg0:.*]]: memref<3x3x3x3x4xf32> {stdx.const}, %[[arg1:.*]]: memref<5xi32>) -> memref<5xi32>
+// CHECK: %[[t0:.*]] = memref.get_global @cst_memref_0 : memref<5xi32>
+// CHECK: %[[t1:.*]] = affine.parallel (%[[arg2:.*]]) = (0) to (5) reduce ("assign") -> (memref<5xi32>)
+// CHECK:   %[[t2:.*]] = pxa.load %[[t0]][%[[arg2]]] : memref<5xi32>
+// CHECK:   %[[t3:.*]] = pxa.reduce assign %[[t2]], %[[arg1]][%arg2] : memref<5xi32>
+// CHECK:   affine.yield %[[t3]] : memref<5xi32>
+// CHECK: return %[[t1]] : memref<5xi32>
```
flaub: Can you format this so that the code lines up?

```
// CHECK-LABEL: func @main
//  CHECK-SAME: (%[[arg0:.*]]: memref<3x3x3x3x4xf32> {stdx.const}, %[[arg1:.*]]: memref<5xi32>) -> memref<5xi32>
//       CHECK:   %[[t0:.*]] = memref.get_global @cst_memref_0 : memref<5xi32>
//       CHECK:   %[[t1:.*]] = affine.parallel (%[[arg2:.*]]) = (0) to (5) reduce ("assign") -> (memref<5xi32>)
//       CHECK:     %[[t2:.*]] = pxa.load %[[t0]][%[[arg2]]] : memref<5xi32>
//       CHECK:     %[[t3:.*]] = pxa.reduce assign %[[t2]], %[[arg1]][%arg2] : memref<5xi32>
//       CHECK:     affine.yield %[[t3]] : memref<5xi32>
//       CHECK:   return %[[t1]] : memref<5xi32>
```

flaub (approved):

PR #1882, opened 2021-09-28
Fix issues with reorder layouts

flaub:


pmlc/target/x86/reorder_layouts.cc
```diff
@@ -172,7 +170,7 @@ struct PropagateReorderThruEltwiseOpPattern
     if (!primary)
       return failure();
 
-    // IVLOG(1, "candidate: " << debugString(op));
+    IVLOG(1, "candidate: " << debugString(op));
```
jbruestle, line 173: Do we want this at such a high log level?

jbruestle (approved):

PR #1886, opened 2021-09-30
Add constraints and bypass the bound check for linalg.generic

shiyao00:


pmlc/conversion/tile_to_linalg/utils.cc
```diff
@@ -139,4 +141,118 @@ llvm::SmallSet<int64_t, 4> getUsedDims(AffineExpr expr) {
...
+           "The input maps have different numbers of dimensions.");
+    for (AffineExpr expr : map.getResults()) {
+      if (auto dimExpr = expr.dyn_cast<AffineDimExpr>()) {
+        dims.insert(dimExpr.getPosition());
+      }
+    }
+  }
+  return dims.size() != numDims;
+}
+
```
flaub: Isn't this the same as:

https://github.com/llvm/llvm-project/blob/main/mlir/lib/IR/AffineMap.cpp#L514

shiyao00: We can determine it using inversePermutation().

flaub (dismissed):

flaub (dismissed):

flaub (approved):

PR #1891, opened 2021-10-13
Generalize reorder pass but check that channels-last is used

flaub:


flaub:
@shiyao00 This should fix the remaining eDSL tests under `PLAIDML_USE_LINALG=1`.

shiyao00 (approved):

PR #1896, opened 2021-10-22
WIP Program Load

tzerrell:


tzerrell:
Added compile to run the pipeline and changed the way the Program was being constructed so that `module` would be initialized from the provided code -- PLAIDML_DUMP is now producing code with actual content (that appears correct), but the test overall is segfaulting.

PR #1901, opened 2021-11-16
Generate memory allocation for each init_tensor use

shiyao00:


pmlc/conversion/linalg_to_pxa/linalg_to_pxa.cc
```diff
@@ -380,10 +379,14 @@ struct InitTensorOpConversion
...
+    auto value = op.result();
+    auto type = value.getType().cast<RankedTensorType>();
     if (llvm::none_of(type.getShape(), ShapedType::isDynamic)) {
-      BufferAllocator allocResult(rewriter, op, type);
-      op.replaceAllUsesWith(allocResult.resultMemRef);
+      while (value.use_begin() != value.use_end()) {
+        // For each use of InitTensorOp, we need an independent allocation
+        BufferAllocator allocResult(rewriter, op, type);
+        value.use_begin()->set(allocResult.resultMemRef);
+      }
```
flaub, line 389: I don't think this solution works when I have multiple uses of a `linalg.fill` op. I think the proper fix is to have the lowering of `linalg.generic` allocate its own output buffer and then copy the buffer that's being passed into for the init/out.

flaub (changes requested):

pmlc/conversion/linalg_to_pxa/linalg_to_pxa.cc
flaub: OK, so it looks like `linalg.fill` is not `NoSideEffect`, even though I would argue that this should be the case when linalg is operating on tensors.

I guess I'm worried about other tensor values that might have multiple uses that may require multiple buffers for each use. If we modify the lowering of `linalg.generic`, then we don't have to have a special case for each value lowering.

shiyao00: I initially thought about implementing alloc for generic only. However, I think the current implementation is better.
Suppose we have two uses A and B of an InitTensorOp. As InitTensorOp doesn't assign any particular value and it's only a declaration, two memory allocations don't cause any problem.
If the result of InitTensorOp is assigned by some other op C and D/E uses C's result, D/E are not the direct uses of the InitTensorOp. So there is still one allocation for op C.
If we implement the allocation for generic only, what about other operations using the result of InitTensorOp? I think we would have to implement the allocation for every op. It's too trouble for the current ops and the future ops.

shiyao00: Here's a more clear version. As InitTensorOp doesn't provide any meaningful value, it's illegal or meaningless to LOAD its result. So its uses must be all STOREs. For each store use, we need an independent memory allocation.

flaub: When lowering from value-like to buffer-like, I think the idea is that we need to have each operation that produces a new value allocate a buffer for that value. And for each value being used as an input operand, we have to copy the value into the output buffer. This approach should work with any op that has input and output operands.

flaub: Later fusion passes should be able to notice that a copy is happening from an uninitialized buffer and so should be able to drop the extra copy.

PR #1902, opened 2021-11-16
Insert memory allocation before linalg.generic

shiyao00:


jbruestle (dismissed):
Looks good!

flaub (dismissed):

pmlc/conversion/linalg_to_pxa/linalg_to_pxa.cc
```diff
@@ -350,12 +389,12 @@ struct GenericOpConversion : public OpConversionPattern<linalg::GenericOp> {
     }
 
     if (op.getNumResults() == forOp.getNumResults()) {
-      rewriter.replaceOp(op, forOp.getResults());
```
flaub, line 353: What don't we just use `replaceOp`?

shiyao00: Consider the program:

  %a = linalg.generic (...)
  %b = linalg.generic (out %a)

After transforming the first generic if we use replaceOp, it's
 %a0 = alloc()
 %a1 = affine.parallel // %a0 <- xxx
  %a = linalg.generic (...)
  %b = linalg.generic (out %a)

MLIR will replace %a in the second generic later after all transformation. However, when we process the second generic, we need to add a copy loop. 

 %a0 = alloc()
 %a1 = affine.parallel // %a0 <- xxx
  %a = linalg.generic (...)
  %a2 = alloc()
  %a3 = affine.parallel  // %a2 <- %a, where %a2 is memref and %a is tensor 
  %b = linalg.generic (out %a)

The copy loop's input is a tensor, but the output is a memref. It triggered the debugging assertion.

PR #1908, opened 2021-12-07
Enable multi-output fusion and fix related issues

shiyao00:
1. Enable multi-output fusion
2. Block the fusion on the reduction indexes
3. Let TileAccumulate process the multi-output layers


jbruestle:
Looks good to me, curious if there is any perf impact.  I would imagine this might fix the missing RELU fusions since singleOutput is no longer required?

flaub (dismissed):

flaub (approved):

PR #1909, opened 2021-12-12
Add Stencil for TPP binary ops

taknevski:
Support for stencil tpp binary and unary ops. Modified stencil implementation to support generic binary/unary ops.
Added test case for binary add.
Support for floating point, relu, exp, tanh, add, sub, mul and div.


pmlc/target/x86/stencil_tpp_binary.cc
```diff
@@ -0,0 +1,219 @@
...
+  }
+  return false;
+}
+
+class StencilImpl : public pxa::StencilBase {
+private:
+  std::string opName;
+  template <typename OpTy>
+  void maybeCaptureGeneric(Optional<pxa::StencilCapture> &capture,
+                           const std::string &inName) {
```
flaub: Use `llvm::StringRef` (which I think is re-exported in `mlir::StringRef`).

pmlc/target/x86/stencil_tpp_binary.cc
```diff
@@ -0,0 +1,219 @@
...
+      pxa::convertToValueMap(op.getContext(), rap->outer);
+  mapOperands.append(outerValueMap.getOperands().begin(),
+                     outerValueMap.getOperands().end());
+
+  AffineMap tileMap = pxa::makeTileMap(op.getContext(), op.getAffineMap(),
+                                       op.getMapOperands(), idxs);
+
+  return TppOperand{op.getMemRef(), outerValueMap.getAffineMap(), tileMap};
+}
+bool isLocallyDefined(AffineParallelOp &op, Value &source) {
```
flaub: `Value` and `Op` types in MLIR have value-like semantics, so you should pass these by value instead of by ref.

pmlc/target/x86/stencil_tpp_binary.cc
taknevski: Thanks, corrected

pmlc/target/x86/stencil_tpp_binary.cc
taknevski: Thanks, done,

pmlc/target/x86/stencil_tpp_binary.cc
```diff
@@ -0,0 +1,217 @@
...
+  }
+  return false;
+}
+
+class StencilImpl : public pxa::StencilBase {
+private:
+  StringRef opName;
+  template <typename OpTy>
+  void maybeCaptureGeneric(Optional<pxa::StencilCapture> &capture,
+                           const StringRef &inName) {
```
flaub: Replace `const StringRef&` with `StringRef`

flaub (approved):
LGTM

PR #1914, opened 2022-01-07
added check for dimension compatibility for inputs and outputs prior to binary op stenciling

taknevski:
-This check will only stencil if the output tensor dimensions and both input tensor dimensions (shapes) are identical.
-This check will prevent stenciling of binary ops that require broadcasting of either of or both the input tensors to match the dimensionality of the output tensor.

### DIFF
--- a/pmlc/target/x86/stencil_tpp_binary.cc
+++ b/pmlc/target/x86/stencil_tpp_binary.cc
@@ -50,6 +50,27 @@ Optional<TppOperand> getTppOperand(TOp op, Block *block,
 
   return TppOperand{op.getMemRef(), outerValueMap.getAffineMap(), tileMap};
 }
+
+bool shapeCompatible(MemRefType output, MemRefType input1, MemRefType input2) {
+  if (!output.hasStaticShape() || !input1.hasStaticShape() ||
+      !input2.hasStaticShape())
+    return false;
+  if (output.getRank() != input1.getRank() ||
+      output.getRank() != input2.getRank())
+    return false;
+
+  ArrayRef<int64_t> outShape = output.getShape();
+  ArrayRef<int64_t> inputShape1 = input1.getShape();
+  ArrayRef<int64_t> inputShape2 = input2.getShape();
+
+  for (int64_t i = 0; i < outShape.size(); i++) {
+    if (outShape[i] != inputShape1[i] || outShape[i] != inputShape2[i])
+      return false;
+  }
+
+  return true;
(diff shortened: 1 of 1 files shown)


pmlc/target/x86/stencil_tpp_binary.cc
```diff
@@ -50,6 +50,27 @@ Optional<TppOperand> getTppOperand(TOp op, Block *block,
...
+    return false;
+  if (output.getRank() != input1.getRank() ||
+      output.getRank() != input2.getRank())
+    return false;
+
+  ArrayRef<int64_t> outShape = output.getShape();
+  ArrayRef<int64_t> inputShape1 = input1.getShape();
+  ArrayRef<int64_t> inputShape2 = input2.getShape();
+
+  for (int64_t i = 0; i < outShape.size(); i++) {
```
jbruestle: ArrayRef has a != operator, so you can just say outShape != inputShape1.  This also automatically checks the length, so it would make the .size() comparisons above redundant as well.

jbruestle (commented):
Seems like a fine thing to work around the broadcast case for now, minor comment regarding use of operator !=

pmlc/target/x86/stencil_tpp_binary.cc
taknevski: Ok sounds good, I can make this change, thanks

pmlc/target/x86/stencil_tpp_binary.cc
```diff
@@ -92,9 +108,15 @@ class StencilImpl : public pxa::StencilBase {
...
-
+    auto outputSource = cast<pxa::PxaReduceOp>(reduce.getDefiningOp()).memref();
     if (isLocallyDefined(op, source1) || isLocallyDefined(op, source2))
       return;
+    auto outputType = outputSource.getType().cast<MemRefType>();
+    auto input1Type = source1.getType().cast<MemRefType>();
+    auto input2Type = source2.getType().cast<MemRefType>();
+
+    if (!shapeCompatible(outputType, input1Type, input2Type))
+      return;
```
flaub: It seems like this could just be:

```
if (outputType.getShape() != input1Type.getShape() || outputType.getShape() != input2Type.getShape())
  return;
```

taknevski: Ok, I'll make this change, thanks

shiyao00 (approved):

PR #1916, opened 2022-01-18
Support for broadcasting in tpp unary stencils

taknevski:
1. Added support for broadcasting 1d array(input) to 2d/4d array output.
2. Added support for broadcasting constants(input) to 2D array output.
3. Modified xsmm dispatch op to include type of broadcast(col, row, scalar, none) as a parameter. 
4. Modified pxa generic op to have input as type AnyFloatOrMemRef instead of strict MemRef.
5.  Included a separate pass for scalar unary tpp stenciling. This pass specifies the stencil index requriements only on the output but no requirements on the input
6. Modified existing stencil idx requirement on the scalar tpp unary op stenciling to have the outer dimension be zero (int hte event of 1D array being broadcasted to 2D) or greater than 1(2D array to 2D array).


pmlc/rt/xsmm.cc
```diff
@@ -1,9 +1,11 @@
 // Copyright 2020 Intel Corporation
 
-#include "mlir/ExecutionEngine/RunnerUtils.h"
-
+#define NO_BCAST 0
+#define ROW_BCAST 1
+#define COL_BCAST 2
+#define SCALAR_BCAST 3
```
flaub, line 6: Use static constexpr in C++ rather than #define

taknevski:
Discarding modifications. Resolving constant broadcasting in upstream conversion pass from linalg to pxa.

PR #1921, opened 2022-02-03
Enable reorder of activations

flaub:


pmlc/target/x86/pipeline.cc
```diff
@@ -262,7 +262,10 @@ void pipelineBuilderStage1(OpPassManager &pm) {
   }
 
   pm.addPass(pmlc::conversion::tile_to_linalg::createLowerTileToLinalgPass());
-  pm.addNestedPass<FuncOp>(createReorderLayoutsPass());
+  if (!util::getEnvVar("PLAIDML_REORDER_WEIGHTS").empty())
```
alheinecke: Do I get it correctly after review the changes, that the default is the re-order weights and activation (and according to the td file try to do with a channel blocking of 32) and we set a env var to just reorder weights?

alheinecke (commented):
Just a small question on the differences between reorder and reorderweights.

I like the composable design which should allow pretty easily to reorder NCHW -> NBcHWc or even NC -> NnNcnc (for regular GEMMs).

pmlc/target/x86/pipeline.cc
flaub: Yes, so that I can more easily compare perf. Once we confirm perf is better for activations + weights we can drop the reorder weights-only pass.

flaub:
I verified that the inception_v3 failure must be related to the reorder activations pass and not any of the other linalg changes because it appears to work with reorder weights only.

flaub:
Because there are remaining issues with the reorder activations+weights pass, I've flipped the default to continue to use reorder weights. I wanted to get this landed however because it's been pending for awhile and also contains lots of fixes that are needed for activations reordering.

alheinecke (approved):

alheinecke:
Is there any indication what the problem with inception_v3 is? Conv, AvgPool, MaxPool are in RN50. Only the concat between the modules might be problematic as the channel dimension is now split into two parts (is this the issue?)

PR #1925, opened 2022-02-09
Fix OpenVINO build

flaub:
Note that I had to use GCC 9.3.0 to get everything to build properly.

### DIFF
--- a/ci/openvino.py
+++ b/ci/openvino.py
@@ -1,6 +1,7 @@
 import argparse
 import atexit
 import os
+import pathlib
 import subprocess
 import sys
 import tarfile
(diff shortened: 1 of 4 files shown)


alheinecke (commented):
so we don't need OpenCV for the OpenVino frontend?

flaub:
I think what's happening is that it's already installed via conda and it must be picking this up during the build. I was able to run all of `ninja check-smoke` which includes all of the OpenVINO unit tests. However it doesn't necessarily test `benchmark_app`, but we'll see how that does during this CI build.

alheinecke:
Maybe it's best to use brute-force:
a) wget opencv ("wget -O opencv.zip https://github.com/opencv/opencv/archive/4.x.zip")
b) unpack and build

PR #1926, opened 2022-02-16
attempt to fix openvino compilation

alheinecke:


ronghongbo (commented):
I saw 3 errors related with openvino runs in CI: https://github.com/plaidml/plaidml/actions/runs/1851002035 . Are they expected at this moment?

alheinecke:
no, it's bug/incompatible of the plaidml plugin ?!

shiyao00:
Why do you build opencv instead of importing from conda env?

alheinecke:
I tried that and that segfaults... so then it's normally best to build everything from source with the same C++ compiler.

alheinecke:
even with version 4.4.0 the PlaidML plugin SegFaults at closure... it seems to be a bug in the PlaidML plugin

alheinecke:
On CentOS 7/ Fedora 35 with latest miniconda I still see seg faults. The only difference now is that we copy the library from conda in the testkit instead of setting LD_LIBRARY_PATH the conda env. So this seems to be an artifact of the unit test organization.  

The seg fault on is observed when using the benchmark_app right after compilation in the same cenv and ldd points to the library which is copied.

PR #1927, opened 2022-02-24
WIP: JSON config for multithreading

KavithaTipturMadhu:
This patch is the first of many to support Tiling and parallelization in PlaidML from an external description. This patch adds support for multithreading by specifying which parallel loop can run on separate OpenMP threads.

1. The patch identifies parallel loops uniquely by associating hierarchical indices to them. For example, if there are two affine parallel loops in the IR, they are assigned indices 0 and 1. Further, if the first loop has a nested affine parallel loop, it gets the id "0,0".

2. A JSON file containing loop ids and their schedules which need to be parallelized is committed.

Depends on https://github.com/plaidml/llvm-project/pull/47

### DIFF
--- a/pmlc/dialect/pxa/transforms/cpu_thread.cc
+++ b/pmlc/dialect/pxa/transforms/cpu_thread.cc
@@ -16,6 +16,9 @@
 
 using namespace mlir; // NOLINT
 
+#include "pmlc/util/env.h"
+#include "pmlc/util/logging.h"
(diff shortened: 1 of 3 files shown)


alheinecke (changes requested):
Please add some lit test for unit testing this functionality in this directory https://github.com/plaidml/plaidml/tree/plaidml-v1/pmlc/dialect/pxa/tests and add them to the auto/unittests

PR #1939, opened 2022-04-30
Reduce kernel stencil

ZhangMZh:
1D stenciling for TPP reduce ops


alheinecke:
unit test were not running - fixed that -> failing
we need to wait for latest libxsmm being available as we don't want have patch files (where are still missing here anyway)

PR #1940, opened 2022-05-05
Fix: batchnorm folding with depthwise conv

KavithaTipturMadhu:
Batchnorm folding with depthwise convolution was disabled earlier and weight(0) was not being set.


alheinecke (commented):
I have merge latest plaidml-v1 into this PR. I will make re-order optimal and controlled by env variable,

alheinecke:
This one is still failing, but I thought after @tzerrell  fix it should go through. It's now going through but gives the wrong answer, it seems.

PR #1942, opened 2022-05-05
Regulate depthwise convolution

shiyao00:


pmlc/dialect/linalgx/analysis/convolution.cc
```diff
@@ -0,0 +1,38 @@
...
+  Block *block = op.getBody();
+  Block::BlockArgListType args = block->getArguments();
+  if (args.size() != 3)
+    return None;
+
+  ValueRange values = op.getOperands();
+  SmallVector<AffineMap> idxMaps = op.getIndexingMaps();
+
+  Operation *yieldOp = block->getTerminator();
+  if (matchPattern(
```
chelini, line 25: Hi @shiyao00  long time ago I developed something to match structurally (both control flow + access patterns) IR - mainly Affine. See for example, https://github.com/LoopTactics/mlir/blob/fd8a51a3e912c0bb3d53df46e27c9a8596450574/mlir/test/lib/IR/TestMatchers.cpp#L227 

Do you think they could be useful to you?

shiyao00: As it's not yet in MLIR official repo, I think we could use it later. Or you could place this matcher in our repo and then modify other parts using it.

alheinecke:
@shiyao00 : can you please merge #1940 to see if the code still works, right now #1940 breaks because of a bug in the relayout pass. #1940 significantly improves performance as batch norm is completely eliminated as folded also for separable convolutions. 
@KavithaTipturMadhu : FYI

PR #1945, opened 2022-05-10
Improve layout reorder for depthwise convolution

shiyao00:


alheinecke (changes requested):
can you please also merge #1940 into this PR and see if it still breaks.

PR #1950, opened 2022-05-23
Disable unused OpenCV in OV

tzerrell:


chelini (approved):
Nice thanks!

tzerrell:
> Nice thanks!

Unfortunately this seems to have stopped building FuncTests, which I didn't think were a dependency... I'm still investigating

PR #1955, opened 2022-05-26
Retire Tile to PXA conversion

chelini:


chelini:
Currently, the test 'contract.mlir' is disabled.  

Current ->
```
memref.global "private" constant @cst_scalar_memref_1 : memref<f32> = dense<0.000000e+00>
  func @global_sum(%arg0: memref<5x10xf32>, %arg1: memref<f32>) {
    %cst = arith.constant 0.000000e+00 : f32
    %0 = memref.alloc() : memref<f32>
    affine.store %cst, %0[] : memref<f32>
    %1 = affine.load %0[] : memref<f32>
    affine.store %1, %arg1[] : memref<f32>
    affine.for %arg2 = 0 to 5 {
      affine.for %arg3 = 0 to 10 {
        %2 = affine.load %arg0[%arg2, %arg3] : memref<5x10xf32>
        %3 = affine.load %arg1[] : memref<f32>
        %4 = arith.addf %3, %2 : f32
        affine.store %4, %arg1[] : memref<f32>
      }
    }
    return
  }
```

Orig ->

```
func @global_sum(%arg0: memref<5x10xf32>, %arg1: memref<f32>) {
    %cst = arith.constant 0.000000e+00 : f32
    affine.store %cst, %arg1[] : memref<f32>
    affine.for %arg2 = 0 to 5 {
      affine.for %arg3 = 0 to 10 {
        %0 = affine.load %arg0[%arg2, %arg3] : memref<5x10xf32>
        %1 = affine.load %arg1[] : memref<f32>
        %2 = arith.addf %1, %0 : f32
        affine.store %2, %arg1[] : memref<f32>
      }
    }
    return
  }
```

tzerrell (commented):
Is the idea with the tile_to_pxa_special_patterns that we will continue to use the tile->pxa conversion for gather/scatter for a while before eventually deprecating that path in favor of other handling of gather/scatter? (I want to confirm the intent here, because this seems like a reasonable temporary approach, but I don't want to move toward permanently having a "fake" tile->pxa conversion sitting inside the linalg->pxa conversion.)

tzerrell:
Oh, and as for the skipped test, unless I'm missing something, it looks essentially the same with an extra spurious copy -- which isn't ideal but should be subject to DCE later so should be fine. Am I missing a problem?

chelini:
> Is the idea with the tile_to_pxa_special_patterns that we will continue to use the tile->pxa conversion for gather/scatter for a while before eventually deprecating that path in favor of other handling of gather/scatter? (I want to confirm the intent here, because this seems like a reasonable temporary approach, but I don't want to move toward permanently having a "fake" tile->pxa conversion sitting inside the linalg->pxa conversion.)

Yes, long term they will disappear.

chelini:
> Oh, and as for the skipped test, unless I'm missing something, it looks essentially the same with an extra spurious copy -- which isn't ideal but should be subject to DCE later so should be fine. Am I missing a problem?

You are right. Just missing a canonicalization.

tzerrell (dismissed):
LGTM

KavithaTipturMadhu (approved):
LGTM

tzerrell (approved):

PR #1958, opened 2022-05-30
Drop OpenVino deps

chelini:


configure
```diff
@@ -116,41 +116,6 @@ class Configure:
 
     def configure_cmake(self, args):
         build_dir = pathlib.Path(f'build-{args.target}').resolve() / args.type
-
-        if not args.no_openvino:
-            openvino_dir = build_dir / '_deps/openvino-build'
-            if args.temp:
-                temp_dir = args.temp / args.target
-                fc_dir = temp_dir / 'fc'
```
tzerrell, line 124: This needs to be retained to maintain correctness of the lines
```
        if args.temp:
            cmd.append(f'-DFETCHCONTENT_BASE_DIR={fc_dir}')
```
later. I'm pretty sure it's actually a bug that this is in the `if not args.no_openvino` block at present.

tzerrell (commented):
One nitpick, otherwise looks good!

alheinecke (approved):

PR #1959, opened 2022-05-31
Shape Analysis pass

KavithaTipturMadhu:
This patch adds a shape analysis pass that dumps TPP shapes
and surrounding affineparallel op shapes in a file which
can be chosen by setting the environment as follows:
PLAIDML_SHAPE_ANALYSIS_OUTPUT=\<path\>.
This currently dumps tpp tile shape only and needs to be
enhanced to dump the surrounding affine parallel op's bounds.

### DIFF
--- a/pmlc/rt/CMakeLists.txt
+++ b/pmlc/rt/CMakeLists.txt
@@ -48,6 +48,12 @@ set_property(SOURCE xsmm.cc
     LIBXSMM_DEFAULT_CONFIG
 )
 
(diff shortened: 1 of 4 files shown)


tzerrell:
May need #1961 to get tests to pass

chelini:
Is this ready for review? I saw a bit of activity here.

KavithaTipturMadhu:
> Is this ready for review? I saw a bit of activity here.

A few more changes are due from my end.

PR #1963, opened 2022-06-09
Reduce kernel stencil

ZhangMZh:
1D stenciling of MAX/ADD for RN50


pmlc/target/x86/stencil_tpp_unary.cc
```diff
@@ -61,6 +61,13 @@ class StencilImpl : public pxa::StencilBase {
                            StringRef inName) {
     if (capture)
       return;
+    AffineValueMap ranges = util::getRangesValueMap(op);
```
chelini, line 64: There is code duplication here. These few lines of code seems to repeated multiple times, maybe move them to an utility function?

pmlc/target/x86/stencil_tpp_unary.cc
```diff
@@ -129,12 +143,60 @@ class StencilImpl : public pxa::StencilBase {
...
   }
 
+  void maybeCaptureReduceOp(Optional<pxa::StencilCapture> &capture,
+                            StringRef inName, arith::AtomicRMWKind agg) {
+    if (capture)
+      return;
+    AffineValueMap ranges = util::getRangesValueMap(op);
+    auto constExpr =
+        ranges.getResult(op.getIVs().size() - 2).dyn_cast<AffineConstantExpr>();
+    if (constExpr && constExpr.getValue() > 1) {
```
chelini: Be consistent with braces (see line 148). For single stmt I prefer not to have them.

chelini: Be consistent with braces (see line 148). For single stmt I prefer not to have them.

pmlc/target/x86/stencil_tpp_unary.cc
```diff
@@ -209,34 +271,82 @@ class StencilImpl : public pxa::StencilBase {
...
+      int64_t step = steps[i];
+      newLowerBounds.push_back(op.lowerBoundsMap().getResult(i));
+      newUpperBounds.push_back(op.upperBoundsMap().getResult(i));
+      newSteps.push_back(step);
+      groups.push_back(1);
+    }
+    auto newArgs = body->getArguments();
+    newArgs[0].replaceAllUsesWith(newArgs[1]);
+
+    // Update attributes
```
chelini: update maps? why attributes?

chelini: update maps? why attributes?

pmlc/target/x86/stencil_tpp_unary.cc
```diff
@@ -209,34 +271,82 @@ class StencilImpl : public pxa::StencilBase {
...
+                                }},
+                        }) {}
 };
 
 } // namespace
 
+// For 1D stenciling, we still map to 2D TPPs. We can make 2D TPPs do 1D
+// by having the outer dim being 1.
+void addSingleIteration(AffineParallelOp op) {
+  if (op.getIVs().size() == 1) {
```
chelini: Please prefer an early exit instead of wrapping everything in an 'if'. Can you please add documentation for the method?

chelini: Can you have an early exit instead of wrapping everything in a 1d loop? Can you please add documentation for the method?

pmlc/util/strides.cc
```diff
@@ -53,6 +53,36 @@ llvm::Optional<StrideArray> computeStrideArray(mlir::AffineMap map) {
   return ret;
 }
 
+llvm::Optional<StrideArray> computeStrideArray(mlir::AffineMap map,
+                                               mlir::MemRefType memRefType) {
+  assert(map.getNumResults() == 1);
```
chelini, line 58: please add a message on the assert.

chelini, line 58: please add a message on the assert.

chelini (commented):
Thanks for the PR. Can you please describe in more details what are you trying to achieve? Thanks again!

pmlc/target/x86/stencil_tpp_unary.cc
ZhangMZh: Thanks for the advice. It is very helpful. Now in resnet50, some operations are not stenciled (e.g, reduce max/add). This PR is focus on "1D" case:
```
%359 = affine.parallel (%arg110) = (0) to (1000) reduce ("assign") -> (memref<1x1xf32>) {
    %360 = pxa.load %354[%c0, %arg110] : memref<1x1000xf32> "softmax"
    %361 = pxa.reduce addf %360, %356[%c0, 0] : memref<1x1xf32> "softmax"
    affine.yield %361 : memref<1x1xf32> "softmax"
} "softmax"
```
As you can see, the loop above has only one inducution variable (%arg110). However, our current StencilImpl requires at least two variables.
```
getOperation().walk([](AffineParallelOp op) {
  if (op.getIVs().size() >= 2) {
    StencilImpl stencil(op);
    stencil.performStenciling();
  }
});
```
My solution is to add a (0, 1) dim (addSingleIteration) so that it can still be mapped to a 2D TPP without major code changes. I will refactor my code and add more comments. Thanks.

PR #1967, opened 2022-06-13
Add some toy examples

tzerrell:

PR #1968, opened 2022-06-16
Backout weight layout reorders when the LoopsToShapeMap can't be inferred for the resulting op

tzerrell:
When we reorder weight layouts we sometimes produce ops for which the LoopsToShapeMap can't be inferred. This appears to be expected behavior for some otherwise-legal ops, at least based on my reading of this comment:
https://github.com/llvm/llvm-project/blob/50f846d63443781a5c00ec7022c381c4f09f543d/mlir/include/mlir/Dialect/Linalg/IR/LinalgInterfaces.td#L946

With this patch, we just skip the reorder for such cases by calling getLoopsToShapeMap and backing out the transformed op if it is null.

### DIFF
--- a/pmlc/target/x86/reorder_layouts.cc
+++ b/pmlc/target/x86/reorder_layouts.cc
@@ -729,15 +729,12 @@ struct ReorderWeightLayoutsPass
     if (!newConv.getShapesToLoopsMap()) {
       IVLOG(1, "Cannot reorder: LinAlg unable to infer ShapesToLoopsMap. op: "
                    << debugString(op));
-      IVLOG(1, "    Attempted reordered op: " << debugString(newConv));
+      IVLOG(2, "    note: attempted reordered op: " << debugString(newConv));
 
       // Back out the reordered op
       newConv.erase();
       reorderFilter.erase();
       return;
-    } else {
-      IVLOG(1, "TEMP ShapesToLoopsMap OK for op " << debugString(op));
-      IVLOG(1, "    TEMP Resulting reordered op: " << debugString(newConv));
     }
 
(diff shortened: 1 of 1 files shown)


chelini (approved):
Looks good to me. Can we also add a test case that fails? Thanks

tzerrell:
Yes, good call on correctness tests. The short answer is that this allows the `toy_xcp` model introduced in #1967 to run; without this patch that model fails to run. However, simplifying this down to a lit test has proved more complex than I expected -- I'm continuing work on that, but I'll give you an update with the details if that fails to produce a simple test.

tzerrell:
> However, simplifying this down to a lit test has proved more complex than I expected

To expand on this a little -- when I try to use the functions from that example in a lit test (and turn logging on), I get log messages about `Unrecognized PadOp` that do not appear in the logs if I just run the model in PlaidBench. I do not understand how these are running via different paths. Then after these messages, in the lit test I continue to see a shape-to-loops map error. However, I am not seeing log messages I would expect along with it if it were happening in the way this patch is intending to correct for, so I'm wondering if some other part of the code is triggering the error (and I'm also wondering if it's related to the Unrecognized PadOp message).

Which leaves me wondering if we should land this (it does fix the PlaidBench test case, which is closer to "real world" usage), or if we should hold off until an appropriate lit test is figured out (and any additional fixes are made, if necessary).

alheinecke:
I suggest we land this, or do you expect some more changes / progress.

tzerrell:
I'll go ahead and land this as-is, as it works standalone, and if I make further progress / tests then I can make a separate PR

PR #1975, opened 2022-06-29
Batch parallelization and allocs to alloca changes

KavithaTipturMadhu:
This wip patch modifies scoped allocs to allocas using PromoteBuffersToStackPass as well as pxa localization pass. As of now, the first pass does not seem to be scoping allocs other than weights. On the other hand, pxa localization pass throws a segfault at runtime for threads>1. This patch also parallelizes layers along batch dimension barring those which don't have batch dimension as the outer loop's induction variable.

### DIFF
--- a/pmlc/target/x86/pipeline.cc
+++ b/pmlc/target/x86/pipeline.cc
@@ -337,6 +337,7 @@ void pipelineBuilderStage2(OpPassManager &pm, const Options &options) {
 
   pm.addNestedPass<func::FuncOp>(pxa::createLocalizePass());
   pm.addNestedPass<func::FuncOp>(pxa::createResizeTmpsPass());
+  // pm.addNestedPass<func::FuncOp>(pxa::createAllocaConversionPass());
   pm.addPass(pxa::createDeallocPlacementPass());
   pm.addNestedPass<func::FuncOp>(
       pxa::createAffineNormalizePass(/*promote=*/true,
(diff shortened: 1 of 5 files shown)


pmlc/dialect/pxa/transforms/localize.cc
```diff
@@ -59,6 +59,15 @@ struct LocalizePass : public LocalizeBase<LocalizePass> {
...
     for (auto &kvp : toMove) {
       kvp.first->moveBefore(&kvp.second->front());
+      Operation *allocOperation = kvp.first;
+      auto allocOp = cast<memref::AllocOp>(allocOperation);
+
+      OpBuilder builder(allocOp);
+      auto allocaOp = builder.create<mlir::memref::AllocaOp>(
+          allocOp.getLoc(), allocOp.getType().cast<MemRefType>(),
+          allocOp->getOperands());
+      allocOp.replaceAllUsesWith(&*allocaOp);
```
chelini: I never saw this pattern. Maybe just `allocaOp.getResult(0)` ?

pmlc/target/x86/pipeline.cc
```diff
@@ -335,6 +336,8 @@ void pipelineBuilderStage2(OpPassManager &pm, const Options &options) {
   pm.addNestedPass<func::FuncOp>(pxa::createMemRefDataFlowOptPass());
   pm.addPass(createCanonicalizerPass());
 
+  pm.addNestedPass<func::FuncOp>(
```
chelini: why this is not enough? you mention something in the PR summary but it is not clear to me.

PR #1981, opened 2022-07-29
Fix: reorder layout fix to keep Bk as outer loop

KavithaTipturMadhu:


alheinecke (changes requested):
can you please also adjust the comments?

alheinecke (approved):

PR #1986, opened 2022-08-12
Large op from identity->gemm->relu pattern

KavithaTipturMadhu:


pmlc/rt/xsmm.cc
```diff
@@ -277,17 +277,87 @@ extern "C" void plaidml_rt_xsmm_binary_invoke(int64_t addr, void *input1,
   kernel(&param);
 }
 
+extern "C" void plaidml_rt_xsmm_brgemm_relu_invoke_f32(int64_t funcAddr,
```
KavithaTipturMadhu, line 457: @egeor, @alheinecke  I see this method segfaulting at runtime and am unable to debug it using gdb as it points to some other unrelated error. Can you point out to any obvious mistakes I may have made in supporting ext op with this invoke and the following dispatch methods?

pmlc/target/x86/tests/tpp_combine_relu.mlir
```diff
@@ -0,0 +1,39 @@
...
+#map10 = affine_map<(d0, d1, d2) -> (0, 0, 0, d0, d1)>
+#map11 = affine_map<(d0, d1, d2, d3) -> (0, d3, 0, d0, d2)>
+#map12 = affine_map<(d0, d1, d2, d3) -> (0, d3, 0, 0, d2, d1)>
+#map13 = affine_map<(d0, d1) -> (0, 0, 1, d0 + 1, d1)>
+
+memref.global "private" constant @cst_scalar_memref_4 : memref<f32> = dense<0.000000e+00>
+// CHECK-LABEL: func @tpp_gemm_relu
+//       CHECK:   affine.parallel
+//	 CHECK:   affine.parallel
+//       CHECK:       pxa.generic ({{.*}}) <assign> @tpp_gemm_relu
```
alheinecke: could we check somehow that the relu affine map is used by the fused op?

alheinecke (commented):
minor nit on the test to check if the right maps was taken into the fused op and we could also check if the temp is gone... e.g. for malloc being not existent.

pmlc/target/x86/tests/tpp_combine_relu.mlir
KavithaTipturMadhu: I have added a pattern for relu affine map. and a check to ensure that the temp is removed from its place. I can't add checks for if temp allocation has moved from its original position though. As for the maps, the affine maps get renamed, so I can't add a check for map names, nor can the maps be inlined.

alheinecke (approved):

